PR #5, opened 2020-04-29
Fix parsing of path-ignore field

aibaars:

- Run test builds as necessary. Can be on this repository or elsewhere as needed in order to test the change - please include links to tests in otehr repos!
  - [x] CodeQL using init/finish actions

### DIFF
--- a/lib/config-utils.js
+++ b/lib/config-utils.js
@@ -84,7 +84,7 @@ function initConfig() {
             });
         }
         const pathsIgnore = parsedYAML['paths-ignore'];
-        if (pathsIgnore && queries instanceof Array) {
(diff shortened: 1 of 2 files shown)


lib/config-utils.js
```diff
@@ -84,7 +84,7 @@ function initConfig() {
             });
         }
         const pathsIgnore = parsedYAML['paths-ignore'];
-        if (pathsIgnore && queries instanceof Array) {
+        if (pathsIgnore && pathsIgnore instanceof Array) {
```
XhmikosR, line 87: @aibaars Maybe `Array.isArray` is better here?

PR #7, opened 2020-04-29
Disable default queries

Daverlo:
This PR adds an option for disabling the queries we run default.

A run without the `ignore-default-queries` set, it runs all the queries: https://github.com/github/codeql-action/runs/629305743?check_suite_focus=true
A run with `ignore-default-queries: false`, it runs all the queries: https://github.com/github/codeql-action/runs/629307092?check_suite_focus=true
A run with `ignore-default-queries: true` it only runs the custom queries: https://github.com/github/codeql-action/runs/629310667?check_suite_focus=true

- Run test builds as necessary. Can be on this repository or elsewhere as needed in order to test the change - please include links to tests in otehr repos!
  - [x] CodeQL using init/finish actions

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ The CodeQL action should be run on `push` events, and on a `schedule`. `Push` ev
 ### Configuration 
 You may optionally specify additional queries for CodeQL to execute by using a config file. The queries must belong to a [QL pack](https://help.semmle.com/codeql/codeql-cli/reference/qlpack-overview.html) and can be in your repository or any public repository. You can choose a single .ql file, a folder containing multiple .ql files, a .qls [query suite](https://help.semmle.com/codeql/codeql-cli/procedures/query-suites.html) file, or any combination of the above. To use queries from other repositories use the same syntax as when [using an action](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses).
 
+You can disable the default queries using `ignore-default-queries: true`.
+
 You can choose to ignore some files or folders from the analysis, or include additional files/folders for analysis. This *only* works for Javascript and Python analysis.
 Identifying potential files for extraction:
 - Scans each folder that's defined as `paths` in turn, traversing subfolders and looking for relevant files.
@@ -97,6 +99,8 @@ A config file looks like this:
 
 ```yaml
 name: "My CodeQL config"
+
+ignore-default-queries: true
 
 queries:
   - name: In-repo queries (Runs the queries located in the my-queries folder of the repo)
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ A config file looks like this:
 ```yaml
 name: "My CodeQL config"
 
-ignore-default-queries: true
+disable-default-queries: true
 
 queries:
   - name: In-repo queries (Runs the queries located in the my-queries folder of the repo)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -79,6 +79,8 @@ The CodeQL action should be run on `push` events, and on a `schedule`. `Push` ev
 ### Configuration 
 You may optionally specify additional queries for CodeQL to execute by using a config file. The queries must belong to a [QL pack](https://help.semmle.com/codeql/codeql-cli/reference/qlpack-overview.html) and can be in your repository or any public repository. You can choose a single .ql file, a folder containing multiple .ql files, a .qls [query suite](https://help.semmle.com/codeql/codeql-cli/procedures/query-suites.html) file, or any combination of the above. To use queries from other repositories use the same syntax as when [using an action](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses).
 
+You can disable the default queries using `ignore-default-queries: true`.
```
robertbrignull, line 82: Any reasons why "ignore" rather than other alternatives? I see it as a non-typical name. Given that the description "You can disable ..." then it makes me feel it should be called `disable-default-queries`. I think `exclude-default-queries` would also be more clear what it does than `ignore-default-queries`. We won't be able to change this so we should be happy with the name we pick.

mveytsman: I agree, i like disable.

src/config-utils.ts
```diff
@@ -81,6 +82,10 @@ function initConfig(): Config {
             config.name = parsedYAML.name;
         }
 
+        if (parsedYAML['ignore-default-queries'] && typeof parsedYAML['ignore-default-queries'] === "boolean") {
```
mveytsman, line 85: curious about this, what happens if they put a string in there?

jhutchings1:
Thanks for taking the initiative on doing this! 

cc: @github/product-docs-dsp because this change will affect the configuration docs.

hubwriter:
Thanks for the ping Justin. We'll add this to the "Configuring code scanning" topic once this is merged.

src/config-utils.ts
Daverlo: It would ignore the field https://github.com/github/codeql-action/runs/632416621?check_suite_focus=true

README.md
```diff
@@ -100,6 +102,8 @@ A config file looks like this:
 ```yaml
 name: "My CodeQL config"
 
+ignore-default-queries: true
```
robertbrignull, line 105: This needs to be updated to "disable"

src/finalize-db.ts
```diff
@@ -102,7 +102,12 @@ async function runQueries(codeqlCmd: string, databaseFolder: string, sarifFolder
   for (let database of fs.readdirSync(databaseFolder)) {
     core.startGroup('Analyzing ' + database);
 
-    const additionalQueries = queriesPerLanguage[database] || [];
+    const queries: string[] = [];
+    if (!config.disableDefaultQueries) {
+      queries.push(database + '-code-scanning.qls');
+    }
+    queries.push(...queriesPerLanguage[database]);
```
robertbrignull, line 109: The behavious here has changed to remove the `|| []` default value. Some manual testing shows that this will generate an error trying to do `push(...undefined)`.

Please change to `queries.push(...(queriesPerLanguage[database] | | []));` or bring back the `additionalQueries` as it was before.

robertbrignull: As I suspected, here's what happens if a repo doesn't define any custom queries: https://github.com/Anthophila/test-electron/runs/632587165?check_suite_focus=true

Daverlo: Good catch! I've fixed it and triggered a build without custom queries for testing it https://github.com/github/codeql-action/runs/632605173?check_suite_focus=true ✅

krukow:
Looks good to me - I have no concerns. Thanks for the ping and making completing this work.

PR #10, opened 2020-04-29
Improve errors & warnings in autobuild

mveytsman:
This adds better messaging to tell the user what languages we aren't autobuilding and prompts them to replace this action with custom build steps if we do fail.

  - [ ] CodeQL using init/finish actions

### DIFF
--- a/src/autobuild.ts
+++ b/src/autobuild.ts
@@ -49,7 +49,7 @@ async function run() {
     core.endGroup();
 
   } catch (error) {
-    core.setFailed(error.message);
+    core.setFailed("We were unable to automatically build your code. Please replace the call to the autobuild action with your custom build steps.  " + error.message);
     await util.reportActionFailed('autobuild', error.message, error.stack);
     return;
   }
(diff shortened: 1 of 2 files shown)


src/autobuild.ts
```diff
@@ -53,6 +58,6 @@ async function run() {
 }
 
 run().catch(e => {
-  core.setFailed("autobuild action failed: " + e);
+  core.setFailed("We were unable to automatically build your code. Please replace the call to the autobuild action with your custom build steps. codeql/autobuild action failed.  " + e);
```
robertbrignull, line 61: Sorry the comment has got lost now. What I was saying earlier was that this catch block won't catch what you want. You want to leave this catch as it is and add this error message to the catch on [line 52](https://github.com/github/codeql-action/pull/10/files#diff-7c2e956e58b7f4cd9020332c66c7ea06R52). This catch block will only get errors from our code like `reportActionSucceeded`.

mveytsman: ah thanks!

PR #13, opened 2020-05-01
Add retrying to the upload

robertbrignull:
This should make the upload retry a few times and with a small amount of delay between attempts. Hopefully this will help get over very minor issues, though if there's a larger outage of github lasting more than a few seconds then it won't be able to help.

  - I created branches to simulate various failure methods.
    - `upload_retry` where everything is normal
      - [x] CodeQL using init/analyze actions - https://github.com/Anthophila/test-electron/actions/runs/92974577
      - [x] 3rd party tool using upload action - https://github.com/Anthophila/test-electron/actions/runs/92974579
    - `upload_retry_error` where all attempts fail
      - In the logs you can see the final error message and that it retries with the expected delays.
      - [x] CodeQL using init/analyze actions - https://github.com/Anthophila/test-electron/actions/runs/92961566
      - [x] 3rd party tool using upload action - https://github.com/Anthophila/test-electron/actions/runs/92961572
    - `upload_retry_partial_error` where the first attempt fails
      - [x] CodeQL using init/analyze actions - https://github.com/Anthophila/test-electron/actions/runs/92971009
      - [x] 3rd party tool using upload action - https://github.com/Anthophila/test-electron/actions/runs/92971012

### DIFF
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -94,7 +94,7 @@ async function uploadPayload(payload) {
 
         } else {
             // If the upload fails with 5xx then we assume it is a temporary problem
-            // with turbo-scan and not an error that the user has caused or can fix.
+            // and not an error that the user has caused or can fix.
             // We avoid marking the job as failed to avoid breaking CI workflows.
             core.error('Upload failed (' + requestID + '): (' + statusCode + ') ' + await res.readBody());
             return;
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -48,7 +48,7 @@ export function combineSarifFiles(sarifFiles: string[]): string {
 }
 
 // Upload the given payload.
-// If the request fails then this will be retry a small number of times.
+// If the request fails then this will retry a small number of times.
 async function uploadPayload(payload) {
     core.info('Uploading results');
 
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -78,7 +78,7 @@ async function uploadPayload(payload) {
 
         // On any other status code that's not 5xx mark the upload as failed
         if (!statusCode || statusCode < 500 || statusCode >= 600) {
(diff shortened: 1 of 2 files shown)


robertbrignull:
> I see we're re-trying not only on 5xx responses but on all. Additionally, the behaviour we had of not setting the job as failed when we get a 500 is changed too. Can you provide more of an insight as to why you've decided to go this way?

Yes, thanks for spotting that. I completely forgot to mention that I made that change as part of this. I can easily back that change out if you'd prefer not to make it at this time.

I believe that the will of @jhutchings1 and @joshhale was to now fail the action if there's a 500 when uploading.

robertbrignull:
In fact, I will back this change out and we can make it at another time.

robertbrignull:
Changed to only retry on 5xx errors, and I've changed the error message to include the status code.

I've run some tests again so I'm happy with the testing of this.

src/upload-lib.ts
```diff
@@ -47,6 +47,61 @@ export function combineSarifFiles(sarifFiles: string[]): string {
...
+
+        // On a 5xx status code we may retry the request
+        if (attempt < backoffPeriods.length) {
+            // Log the failure as a warning but don't mark the action as failed yet
+            core.warning('Upload attempt (' + (attempt + 1) + ' of ' + (backoffPeriods.length + 1) +
+              ') failed (' + requestID + '). Retrying in ' + backoffPeriods[attempt] +
+              ' seconds: (' + statusCode + ') ' + await res.readBody());
+            // Sleep for the backoff period
+            await new Promise(r => setTimeout(r, backoffPeriods[attempt] * 1000));
+            continue;
```
anaarmas, line 93: why do we need this `continue` here?

robertbrignull: To get to the next iteration of the loop. Technically it's not necessary as this is the last statement of the loop in this branch, but it doesn't hurt as far as I can see.

PR #14, opened 2020-05-01
Improve feedback about what files are being uploaded in the upload-sarif Action.

chrisgavin:
This pull request makes two changes designed to assist debugging the upload action. Firstly it increases the log level of the message listing which SARIF files are being uploaded, which is a very useful piece of information.

Secondly it fails the action if an attempt is made to upload a folder with no SARIF files in, since this is very likely to be a mistake.

  - [x] CodeQL using init/analyze actions (N/A)
    - Successful upload here: https://github.com/Anthophila/chrisgavin/runs/636104360?check_suite_focus=true
    - Failed upload here: https://github.com/Anthophila/chrisgavin/actions/runs/92977473
- [x] Confirm this change is backwards compatible with existing workflows. (Yes, though workflows that try to upload an empty directory will now fail.)
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary. (N/A)

### DIFF
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -56,6 +56,7 @@ export async function upload(input: string) {
             .map(f => path.resolve(input, f));
         if (sarifFiles.length === 0) {
             core.setFailed("No SARIF files found to upload in \"" + input + "\".");
+            return;
         }
         await uploadFiles(sarifFiles);
     } else {
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -54,6 +54,9 @@ export async function upload(input: string) {
         const sarifFiles = fs.readdirSync(input)
             .filter(f => f.endsWith(".sarif"))
             .map(f => path.resolve(input, f));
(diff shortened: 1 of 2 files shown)


src/upload-lib.ts
```diff
@@ -54,6 +54,9 @@ export async function upload(input: string) {
         const sarifFiles = fs.readdirSync(input)
             .filter(f => f.endsWith(".sarif"))
             .map(f => path.resolve(input, f));
+        if (sarifFiles.length === 0) {
+            core.setFailed("No SARIF files found to upload in \"" + input + "\".");
```
robertbrignull, line 58: `core.setFailed` doesn't abort, so we'd still try to upload the emptyset of files.
I think it would be better to return here, but it's perhaps not essential.

robertbrignull: But to clarify, this PR is a strict improvement as it is, so I have no opposition to merging as it is.

PR #16, opened 2020-05-04
Create undeclared-action-input.ql

robertbrignull:
Adds a new CodeQL query for detecting when we are referencing action input parameters (by `core.getInput(...)`) from actions where those inputs are not defined. This can easily happen when code is reused between actions, and affects us more than most people because we have multiple actions in the same repo.

The alerts are visible at https://github.com/github/codeql-action/security/code-scanning?query=ref%3Arefs%2Fheads%2Fundeclared-action-input

All the alerts appear to be true positives to me and it's found all the alerts I expected it to find.

(Here's a screenshot for posterity)
results2

### DIFF
--- a/queries/undeclared-action-input.ql
+++ b/queries/undeclared-action-input.ql
@@ -36,7 +36,7 @@ class ActionDeclaration extends File {
 }
 
 Expr getAFunctionChildExpr(Function f) {
-  result = f.getBody().getAChildStmt*().getAChildExpr*()
+  result.getContainer() = f
 }
 
 /*
@@ -44,6 +44,8 @@ Expr getAFunctionChildExpr(Function f) {
  */
 Function calledBy(Function f) {
   result = getAFunctionChildExpr(f).(InvokeExpr).getResolvedCallee()
+  or
+  result.getEnclosingContainer() = f // assume outer function causes inner function to be called
 }
 
 class GetInputMethodCallExpr extends MethodCallExpr {
--- a/queries/undeclared-action-input.ql
+++ b/queries/undeclared-action-input.ql
@@ -1,0 +1,63 @@
+/**
+ * @name Undeclared action input
+ * @description Code tries to use an input parameter that is not defined for this action.
+   Perhaps this code is shared by multiple actions.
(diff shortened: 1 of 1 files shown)


queries/undeclared-action-input.ql
```diff
@@ -0,0 +1,63 @@
...
+    result.getFile().getRelativePath() = getRootNode().
+      (YAMLMapping).lookup("runs").
+      (YAMLMapping).lookup("main").
+      (YAMLString).getValue().regexpReplaceAll("\\.\\./lib/(.*)\\.js", "src/$1.ts") and
+    result.getName() = "run"
+  }
+}
+
+Expr getAFunctionChildExpr(Function f) {
+  result = f.getBody().getAChildStmt*().getAChildExpr*()
```
robertbrignull, line 39: I'm not very knowlegeable about the javascript QL libraries so might need some help from the team. Perhaps @asgerf if you have a chance. Do the `getAFunctionChildExpr` and `calledBy` predicates look correct to you, or are they even necessary and maybe there's something in the library that can do this more easily?

joshhale:
CodeQL is failing because of the alerts detected by this PR, but I think they can/should be addressed separately.

queries/undeclared-action-input.ql
asgerf: `getContainer` is probably what you're looking for:
```suggestion
  result.getContainer() = f
```

There's a subtle difference there in how nested functions are handled, however (yours would include the body of arrow functions).

To handle nested functions, I'd also add this to `calledBy`:
```codeql
Function calledBy(Function f) {
  ...
  or
  result.getEnclosingContainer() = f // assume outer function causes inner function to be called
}
```

robertbrignull: Thanks. Didn't know about that predicate but that makes sense.

robertbrignull:
Queries still seem to work. It finds all the alerts that I expected it to.

The "Code scanning / CodeQL" check is marked as red because this is "introducing" alerts, but this is fine and we've decided we want to fix them in separate PRs.

PR #19, opened 2020-05-05
Use tmp dir for external queries test

robertbrignull:
The external queries tests run in a fixed directory. This is arguably fine when run as part of CI, but if the tests are run multiple times locally then they start to interfere with each other.

To see it's working, you can see in the [test logs](https://github.com/github/codeql-action/runs/645786593)
```
Cloning into '/tmp/codeql-action-1XUxpg/github/codeql-go'
```

This PR creates a random tmp dir for each test run and deletes it when the test is done. I created a shared function for this because I intend to start using it for other tests as well.

Also by populating the `RUNNER_WORKSPACE` env var we can make it required in `checkoutExternalQueries` and this avoids the hardcoded alternative path.

@Daverlo

  - [x] CodeQL using init/analyze actions - https://github.com/Anthophila/test-electron/actions/runs/96083423
  - [x] 3rd party tool using upload action - https://github.com/Anthophila/test-electron/actions/runs/96083425


robertbrignull:
Tested with a repo that uses external queries.

PR #27, opened 2020-05-11
Reduce duplication between help and readme

robertbrignull:
This slims down the readme so we don't duplicate information that's already in the help docs so it'll be easier to keep it up to date. It does look a bit small now, but I'm not sure what else to put in it.
@jhutchings1 what are your thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ The underlying CodeQL CLI, used in this action, is licensed under the [GitHub Co
 
 ## Usage
 
+This is a short walkthrough, but for more information read [configuring code scanning](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning).
+
 To get code scanning results from CodeQL analysis on your repo you can use the following workflow as a template:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


robertbrignull:
I'm now thinking we should include descriptions of the inputs parameters to each action and how to use them. I say this after looking at at other actions (e.g. https://github.com/actions/checkout and https://github.com/actions/upload-artifact) and noting that the help docs don't mention these inputs. What do people think?

jhutchings1:
> I'm now thinking we should include descriptions of the inputs parameters to each action and how to use them. I say this after looking at at other actions (e.g. https://github.com/actions/checkout and https://github.com/actions/upload-artifact) and noting that the help docs don't mention these inputs. What do people think?

I think that's a good minbar for this action. We want to give people a basic understanding of the action here, but can point them at the help documentation for more context.

robertbrignull:
Sorry it took a long time, but I've updated this PR with what I think is a better balance between providing information and reducing duplication. The readme now contains information about the action inputs which should be enough for most people, and mentions the existance of the config file but points to the docs for all technical details.

README.md
```diff
@@ -90,97 +92,18 @@ If you prefer to integrate this within an existing CI workflow, it should end up
...
-    uses: ./codeql-querypacks/complex-python-querypack/rootAndBar.qls
-
-paths:
-  - src/util.ts
-
-paths-ignore:
-  - src
-  - lib
-```
+The configuration file must be located within the local repository. For information on how to write a configuration file, see "[Using a custom configuration](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#using-a-custom-configuration)."
```
jhutchings1, line 105: I didn't know that was a limitation we had. Would be interesting to see how much work it would take to allow a reference in another repository. This would allow us to handle shared configurations across an enterprise, for example.

robertbrignull: Good point. I've opened an issue for this.

PR #29, opened 2020-05-12
Add PR check to remove npm absolute paths

robertbrignull:
This PR aims to reduce the chance of us accidentally checking in a bad node_modules directory.

We add another PR check (and renames the workflow file as having four things in the name is a bit unwieldy) that runs `npm ci` and also `removeNPMAbsolutePaths`. Hopefully this will catch places where we modify a npm dependency but fail to update the `node_modules` directory, and will catch absolute paths being checked in.

The danger is that running npm commands isn't very stable and we could end up with false positives because of differing behaviour between different npm versions or on different operating systems. The actions environment seems consistent with my local environment so I'd propose that we go for it and if this causes problems then we revert it.

The reason why I used `npm ci` instead of `npm install` is that I believe it's intended to be more stable. However the behaviour has changed a few times since it's introduction so it's all very unclear. npm is just not very good if you want reproducible results.

The reason I added the `--force` flag to `removeNPMAbsolutePaths` is so that it rewrites all files to a canonical form. I was seeing some differences in ordering of properties of json objects when there were otherwise no changes. I hope by letting `removeNPMAbsolutePaths` decide this aspect we'll avoid any inconsistencies.


.github/workflows/pr-checks.yml
```diff
@@ -30,10 +30,38 @@ jobs:
...
+        if [ ! -z "$(git status --porcelain)" ]; then
+          # If we get a fail here then this workflow needs attention...
+          >&2 echo "Failed: Repo should be clean before testing!"
+          exit 1
+        fi
+
+        # Reinstall modules and then clean to remove absolute paths
+        # Use 'npm ci' instead of 'npm install' as this is intended to be reproducible
+        npm ci
+        npm run removeNPMAbsolutePaths
```
Daverlo, line 55: Should this have the `--force` flag?

> The reason I added the --force flag to removeNPMAbsolutePaths is so that it rewrites all files to a canonical form. I was seeing some differences in ordering of properties of json objects when there were otherwise no changes. I hope by letting removeNPMAbsolutePaths decide this aspect we'll avoid any inconsistencies.

robertbrignull: No, that's a reference to the script defined in package.json
https://github.com/github/codeql-action/blob/51b42fcf78236ea4f6a453acb7871a84b03e1848/package.json#L10

I did it that way so then we can keep things consistent by only specifying the arguments there.

It is perhaps a bit confusing that the name is the same, unless you know what is going on. What do you think?

Daverlo: I missed that. If it works I'm happy with it

robertbrignull:
Let's go with it. If we find it causes issues for some people then we'll have to rethink this and we can easily revert this PR.

PR #30, opened 2020-05-12
don't fail-fast on integration tests

robertbrignull:
Aborting the other jobs is unhelpful for determining if the issue is limited to only one operating system. Since the jobs are done mostly in parallel it shouldn't take significantly longer. This will make it use more actions minutes but at our level I think that is acceptable.


robertbrignull:
> Could you also remove the pull_request trigger in this PR? As you pointed out we're stopping the action on PRs

Done :+1:

PR #22, opened 2020-05-06
Always report exceptions

joshhale:
We currently catch exceptions in a couple of places where it would be better to let them reach the top level where they would be included in the status report from the action.

* Don't catch exceptions during upload lower than the top level
* Don't catch exceptions during config initialization lower than the top level
* Correct the action name passed from the `analyze` action

### Merge / deployment checklist

- Run test builds as necessary. Can be on this repository or elsewhere as needed in order to test the change - please include links to tests in other repos!
  - [x] CodeQL using init/analyze actions [action run](https://github.com/Anthophila/test-electron/actions/runs/97180916) :heavy_check_mark: 
  - [x] 3rd party tool using upload action [action run](https://github.com/Anthophila/test-electron/runs/649610588) :heavy_check_mark: 
- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary.

PR #36, opened 2020-05-14
Add tests for config-utils

robertbrignull:
Adds tests for the `config-utils` file. This does change the behaviour in a couple of minor cases, but largely this PR is trying to just assert the current behaviour.

I'm not a huge fan of the fact that we silently ignore invalid yaml fields in certain cases. I'd prefer it if we always threw an error, but I didn't want to make that change here. If we decide to do that in the future then I'd feel much happier having tests of this functionality.

A side note, I had to add the `--serial` flag when running the tests to stop running them in parallel. My understanding (according to the docs) is that tests from separate files are run in separate nodejs processes, but tests within a file are run in one process. Therefore since we are messing with process environment variables in order to control our code we need to run tests serially and avoid parallelism.


robertbrignull:
The code scanning check failure is because there's 1 new alert and 1 fixed alert because I changed the line where the alert appears. Nothing to worry about.

robertbrignull:
Also note this PR includes some but not all of the changes from https://github.com/github/codeql-action/pull/22

PR #37, opened 2020-05-14
Use the full amount of memory  when running queries.

chrisgavin:
Currently we call CodeQL without specifying `--ram`. This means the amount of memory it uses is whatever Java happens to decide. In the case of the Actions workers that seems to be 1732MB. The actions workers have a ton more memory than this though, and since Actions steps are generally run serially we can use almost all of it if we want to.

This change adds automatic detection of how much memory to use based on the total system memory minus 256MB for other processes that might be running. I've also added an input that allows overriding this, but I don't think most people will need it so I've left it out of the documentation for now.

  - [x] 3rd party tool using upload action (N/A)
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary. (N/A)


src/finalize-db.ts
```diff
@@ -10,6 +11,23 @@ import * as sharedEnv from './shared-environment';
 import * as upload_lib from './upload-lib';
 import * as util from './util';
 
+function getMemoryFlag(): string {
+  let memoryToUseMegaBytes: number;
+  const memoryToUseString = core.getInput("ram");
+  if (memoryToUseString) {
+    memoryToUseMegaBytes = Number(memoryToUseString);
+    if (Number.isNaN(memoryToUseMegaBytes)) {
```
robertbrignull, line 19: Probably worth adding a check that it's greater than 0 too.
```suggestion
    if (Number.isNaN(memoryToUseMegaBytes) || memoryToUseMegaBytes <= 0) {
```

chrisgavin:
The way I tested it was just to run `ps aux` while the queries are running and see how much memory was being passed to the evaluator. It should be 50% of the amount you actually want as the other 50% is used for off-heap allocations.

Before this change you get:
```
/opt/hostedtoolcache/CodeQL/1.0.0/x64/codeql/tools/linux64/java/bin/java -Xmx866M -Djava.library.path=/usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib -cp /opt/hostedtoolcache/CodeQL/1.0.0/x64/codeql/tools/codeql.jar com.semmle.cli2.CodeQL execute queries -J-Xmx866M --off-heap-ram=866 --verbosity=progress --logdir=/home/runner/work/******/codeql_databases/javascript/log --ml-model-path=/home/runner/work/******/****** --native-library-path=/home/runner/work/******/****** --warnings=show --no-rerun --output=/home/runner/work/******/codeql_databases/javascript/results -- /home/runner/work/******/codeql_databases/javascript/db-javascript javascript-code-scanning.qls
```

And afterwards it is:
```
/opt/hostedtoolcache/CodeQL/1.0.0/x64/codeql/tools/linux64/java/bin/java -Xmx3336M -Djava.library.path=/usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib -cp /opt/hostedtoolcache/CodeQL/1.0.0/x64/codeql/tools/codeql.jar com.semmle.cli2.CodeQL execute queries -J-Xmx3336M --off-heap-ram=3335 --verbosity=progress --logdir=/home/runner/work/******/codeql_databases/javascript/log --ml-model-path=/home/runner/work/******/****** --native-library-path=/home/runner/work/******/****** --warnings=show --no-rerun --output=/home/runner/work/******/codeql_databases/javascript/results -- /home/runner/work/******/codeql_databases/javascript/db-javascript javascript-code-scanning.qls
```

PR #38, opened 2020-05-15
Log some simple stats about the upload

robertbrignull:
Log a few simple stats about the upload we're making, as requested by @marcogario. The logging is done at debug level so most users won't see it, but it could be useful for us.

This makes the assumption that the sarif file is valid and will fail if it isn't, but I intend to introduce validation of the sarif files imminently in another PR.


robertbrignull:
Here are some demos:

- https://github.com/Anthophila/test-electron/actions/runs/107960104
```
##[debug]Raw upload size: 130372 bytes
##[debug]Base64 zipped upload size: 15640 bytes
##[debug]Number of results in upload: 25
```

- https://github.com/Anthophila/test-electron/actions/runs/107960108
```
##[debug]Raw upload size: 319626 bytes
##[debug]Base64 zipped upload size: 19856 bytes
##[debug]Number of results in upload: 752
```

PR #42, opened 2020-05-17
Add more info about dot syntax to the readme

Daverlo:


README.md
```diff
@@ -84,7 +84,7 @@ The CodeQL action should be run on `push` events, and on a `schedule`. `Push` ev
 
 ### Configuration
 
-You may optionally specify additional queries for CodeQL to execute by using a config file. The queries must belong to a [QL pack](https://help.semmle.com/codeql/codeql-cli/reference/qlpack-overview.html) and can be in your repository or any public repository. You can choose a single .ql file, a folder containing multiple .ql files, a .qls [query suite](https://help.semmle.com/codeql/codeql-cli/procedures/query-suites.html) file, or any combination of the above. To use queries from other repositories use the same syntax as when [using an action](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses).
```
robertbrignull, line 87: Typo in "locatio". I personally would also rewrite the last sentence as "Note that when using local queries starting with `./`, the path is relative to the root of the repository and not to the location of the config file."

I think it flows a bit better. I know what you mean currently but constructions like "Note that for using" are a little bit odd, and I think flipping the order of the last two cases makes it more clear.

PR #43, opened 2020-05-18
Check integration test results

robertbrignull:
This PR adds a check to the `multi-language-repo_test-autodetect-languages` test that the expected number of databases were created. Currently we don't check anything except that it didn't crash. We could go even further and check

Here's an example of it failing: https://github.com/github/codeql-action/runs/685732844?check_suite_focus=true

Also only runs that test on one OS as I don't think running it on windows too is testing anything useful. We have other tests that run on all supported OSes. This should make the integration tests a tiny bit faster.


robertbrignull:
@Daverlo, do you have time to review this?

PR #45, opened 2020-05-26
introduce new syntax for built-in query suites

robertbrignull:
Tightens up the rules on relative paths in the config file, and introduces new options to reference suites from the codeql bundle.

This PR is pending the decision on the exact names of the suites, and actually introducing those suites into the codeql bundle.


src/config-utils.ts
```diff
@@ -17,11 +17,17 @@ export class ExternalQuery {
     }
 }
 
+// The set of acceptable values for built-in suites from the codeql bundle
+const builtinSuites = ['security-experimental', 'security-and-quality'] as const;
```
jhutchings1, line 21:
```suggestion
const builtinSuites = ['security-extended', 'security-and-quality'] as const;
```

Let's go with security-extended instead, since it's not strictly experimental, but just more queries of varied quality. Thanks!

robertbrignull: I should have said the suite names in this PR were mostly just a placeholder as I thought that decision was still being made. I'll update this once we know what they'll be.

robertbrignull:
Running all the new suites in all languages at https://github.com/Anthophila/multi-language-test/actions/runs/121386459

robertbrignull:
Note this change is technically not backwards compitable for anyone using the unadvised `./lgtm-full.qls` hack. Anyone using that will get an error, however the error is hopefully fairly understandable. They will need to switch to `- uses: security-and-quality`.

robertbrignull:
```
The command line is too long.
##[error]The process 'C:\hostedtoolcache\windows\CodeQL\1.0.0\x64\codeql\codeql.cmd' failed with exit code 1
```

robertbrignull:
In order to avoid the command line length issue, I've changed it to create a temporary query suite file that just lists the queries. I can't find the comment or email anymore now but I think @Daverlo and @hmakholm had a discussion recently suggesting this plan for another reason, so I hope this is not a bad move to make.

Also note, the issue with the command line length on windows was already there if somebody tried to run a lot of queries. Adding this new syntax just makes it easier to trigger.

jhutchings1:
Thanks for putting this one together @robertbrignull ! I've tagged our friends in docs to get this on their radar too.

hmakholm:
> In order to avoid the command line length issue, I've changed it to create a temporary query suite file that just lists the queries.

Looks good to me.

robertbrignull:
I think this has been tested enough. I manually ran all query suites for all languages, and the integration tests pass. Right now there's not much more we can do.

PR #46, opened 2020-05-26
Make config file parsing more strict

robertbrignull:
This PR aims to make the parsing of the config file stricter, and safer from a code maintenance point of view.

The changes are:
- Improve error messages to be more descriptive.
- Throw an error if any field is of a type that it shouldn't be, instead of silently ignoring it.
- Define constants for all the field names and always use those constants when accessing that field.

This is obviously going to conflict a bit with https://github.com/github/codeql-action/pull/45 but it shouldn't be too bad and it'll be easy to sort out whichever one gets merged first.


robertbrignull:
@sampart would you be ok taking on review of this PR? However @Daverlo, @anaarmas, and @joshhale you should probably keep an eye on this too though for the reasons I explain below.

The main goal here is making the config parsing give more helpful error messages in the case of incorrect input. A side product is throwing an error on any invalid input instead of silently ignoring it in some cases. This is therefore technically not backwards compatible, however only in that it enforces the schema we've already published. I personally think it can still go on the `v1` branch and this change is not dramatic enough or likely enough to cause problems that we should change it or hold it back or start a new branch.

src/config-utils.ts
```diff
@@ -102,24 +109,47 @@ export class Config {
...
+    return getConfigFilePropertyError(
+        configFile,
+        QUERIES_PROPERTY + '.' + QUERIES_USES_PROPERTY,
+        'must be a built-in suite (' + builtinSuites.join(' or ') +
+            '), a relative path, or be of the form "owner/repo[/path]@ref"' +
+            (queryUses !== undefined ? '\n Found: ' + queryUses : ''));
+}
+
+export function getPathsIgnoreInvalid(configFile: string): string {
+    return getConfigFilePropertyError(configFile, PATHS_IGNORE_PROPERTY, 'must be an array of non-empty string');
```
sampart, line 134:
```suggestion
    return getConfigFilePropertyError(configFile, PATHS_IGNORE_PROPERTY, 'must be an array of non-empty strings');
```

src/config-utils.ts
```diff
@@ -102,24 +109,47 @@ export class Config {
...
 }
 
-export function getQueryUsesIncorrect(queryUses: string): string {
-    return '"uses" value for queries must be a built-in suite (' + builtinSuites.join(' or ') +
-        '), a relative path, or of the form owner/repo@ref\n' +
-        'Found: ' + queryUses;
+export function getPathsInvalid(configFile: string): string {
+    return getConfigFilePropertyError(configFile, PATHS_PROPERTY, 'must be an array of non-empty string');
```
sampart, line 138:
```suggestion
    return getConfigFilePropertyError(configFile, PATHS_PROPERTY, 'must be an array of non-empty strings');
```

src/config-utils.ts
```diff
@@ -102,24 +109,47 @@ export class Config {
...
 }
 
-export function getLocalPathOutsideOfRepository(localPath: string): string {
-    return 'Unable to use queries from local path "' + localPath +
-        '" as it is outside of the repository';
+export function getLocalPathOutsideOfRepository(configFile: string, localPath: string): string {
+    return getConfigFilePropertyError(
+        configFile,
+        QUERIES_PROPERTY + '.' + QUERIES_USES_PROPERTY,
+        'is invalid as the local path "' + localPath + '" is output of the repository');
```
sampart, line 145:
```suggestion
        'is invalid as the local path "' + localPath + '" is outside of the repository');
```

robertbrignull:
Thanks for spotting all those :+1:

PR #48, opened 2020-05-28
Allow pull requests, and report correct commit oid and ref

robertbrignull:
Demo branch for allowing code scanning to run on `pull_request` events, running on the correct commit, and reporting the correct data.

@marcogario, @Daverlo 


.github/workflows/integration-testing.yml
```diff
@@ -16,9 +16,8 @@ jobs:
       shell: bash
       run: |
         mkdir ../action
-        shopt -s dotglob
```
robertbrignull, line 19: This change was necessary because the action now relies on being run in a valid git repository, as it accesses the name of the currently checked out commit. Therefore we don't want to move the `.git` directory.

marcogario:
:+1: to merge this

PR #54, opened 2020-06-04
Fix action when using self-hosted runnners

robertbrignull:
This makes the action work on a self-hosted runner for me.

First, I was able to reproduce the action failing on the second build because of directories not being cleared. Switching to this branch fixed it and I could run the action repeatedly.

This PR moves all files that we create to the `RUNNER_TEMP` directory instead of placing them relative to `RUNNER_WORKSPACE`. This means that files get cleaned between action runs.


aibaars:
These environment variables are not mentioned in https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables

Could you point at the documentation of those environment variables? Have you considered using `GITHUB_WORKSPACE` instead? I would consider it a bug in Actions if that folder is not cleared between runs. It would make little sense for unrelated jobs to leave "surprises" for the next job in the workspace folder.

Do you know if `RUNNER_TEMP` is "private" to each runner process. If not then there might be problems if a single machine runs several self-hosted runner processes.

robertbrignull:
For info, here are all the relevant environment variables available on my self-hosted runner
```
PWD=/actions-runner/_work/robertbrignull-self-hosted/robertbrignull-self-hosted
RUNNER_TOOL_CACHE=/actions-runner/_work/_tool
RUNNER_TEMP=/actions-runner/_work/_temp
RUNNER_WORKSPACE=/actions-runner/_work/robertbrignull-self-hosted
GITHUB_EVENT_PATH=/actions-runner/_work/_temp/_github_workflow/event.json
GITHUB_WORKSPACE=/actions-runner/_work/robertbrignull-self-hosted/robertbrignull-self-hosted
```
You're right `GITHUB_WORKSPACE` is cleared between runs because otherwise checkouts would overlap and it would be chaos. The `RUNNER_WORKSPACE` dir isn't and this is the issue we're running into. It is not documented but we have confirmation from the actions team that the `RUNNER_TEMP` dir is also cleared.

As you can see from the directory paths, the temp dir is private to the runner process as you have to give each runner a `_work` directory.

I think the main downside with putting stuff in the `GITHUB_WORKSPACE` directory is you have to worry about clashing with files from the repository. Obviously this can be worked around by generating random names or appending numbers as necessary. I think the question is whether this is better or worse than placing the files in a "hidden" directory like this where the user won't otherwise find them.

aibaars:
The documentation of `GITHUB_WORKSPACE` is quite confusing. The documentation suggests the repository is checked out in a subdirectory of `GITHUB_WORKSPACE` while in fact  `GITHUB_WORKSPACE` is the repository folder. 

From the docs:
> The GitHub workspace directory path. The workspace directory contains a subdirectory with a copy of your repository ...

`RUNNER_TEMP` looks like good place indeed.

It might be worth asking whether `RUNNER_WORKSPACE` is supposed to be cleared by the worker. It sounds like a bug if it isn't.

robertbrignull:
@rneatherway, does it make sense for you to review this as you've done the research into this area?

rneatherway:
Happy to :hand:

rneatherway:
> The documentation of `GITHUB_WORKSPACE` is quite confusing. The documentation suggests the repository is checked out in a subdirectory of `GITHUB_WORKSPACE` while in fact `GITHUB_WORKSPACE` is the repository folder.

I agree, this seems simply wrong. @robertbrignull could you raise an issue with the relevant doc team? I think that you could find the right place using #product-documentation

rneatherway:
> Could you point at the documentation of those environment variables? Have you considered using `GITHUB_WORKSPACE` instead? I would consider it a bug in Actions if that folder is not cleared between runs. It would make little sense for unrelated jobs to leave "surprises" for the next job in the workspace folder.

@aibaars I raised the question again on Slack (https://github.slack.com/archives/CMV6PDVJT/p1591644772301800). Feel free to chip in there if you like.

rneatherway:
I'm not clear on how this will make it to production (which AIUI is currently the `v1` branch). Is that documented or will it be covered in the Deep Dive?

PR #55, opened 2020-06-08
Update CodeQL link in README

Marcono1234:
QL was moved from `semmle/ql` to `github/codeql`.


Marcono1234:
Sure, have changed it now. Also thanks for linking to the style-guide and explaining the rationale, I wasn't aware that this is the US style.

PR #59, opened 2020-06-12
Add script to open PR for updating release branch

robertbrignull:
Adds a script to create a PR once per week to update the `v1` release branch.

See https://github.com/Anthophila/robertbrignull-update-automation/pull/19 for an example of what it looks like.

We can also try running it on this repo before merging it. I've added a `repository_dispatch` trigger so it can be run on demand as well as on a schedule.

### DIFF
--- a/.github/update-release-branch.py
+++ b/.github/update-release-branch.py
@@ -1,0 +1,178 @@
+import datetime
+from github import Github
+import random
+import requests
+import subprocess
+import sys
+
+# The branch being merged from.
+# This is the one that contains day-to-day development work.
+MASTER_BRANCH = 'master'
+# The branch being merged into.
+# This is the release branch that users reference.
+LATEST_RELEASE_BRANCH = 'v1'
+# Name of the remote
+ORIGIN = 'origin'
+
+# Runs git with the given args and returns the stdout.
+# Raises an error if git does not exit successfully.
+def run_git(*args):
+  cmd = ['git', *args]
+  p = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  if (p.returncode != 0):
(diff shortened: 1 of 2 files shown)


robertbrignull:
I thought we could test this out on this repository before merging it, but turns out repository dispatch events only run on the master branch. You can see the runs on the demo repository so hopefully that and a manual review are enough.

If there are any thoughts on improvements to the message of the PR feel free to bring those up too. Though things like that we can easily change after this is merged and in use.

Another thing is there are a couple of places where the error checking is perhaps not the greatest. For instance when getting the commit author login it might not be a user with a github account and I don't know what happens in this case. Since this is an internal tool I'd personally be happy to just leave it as it is and if it breaks in the future on an edge case then we'll fix it then. :shrug:

robertbrignull:
The `PyGithub` library was just one I found. I have no idea if it's the best one to use, but it seems to be pretty up to date. To understand what stuff in that library is doing I found it easiest to just look directly at the code, e.g. https://github.com/PyGithub/PyGithub/blob/master/github/PullRequest.py

robertbrignull:
Those sound like good improvements to me. Hopefully we won't have to trigger it ondemand too often but being able to do so without having to make the API request manually would be nice.

I picked python for the script only out of personal preference as it's quite lightweight and I wasn't really thinking about how it integrates with the existing code in this action. I wouldn't oppose rewriting it in something else if that would be beneficial, however I also think it's still worth merging this as it is.

Also since this is a script that is only run from an internal dev action and has no chance of breaking the main action I see no problem with merging it now. Let's just merge this and wait for it to run on Monday morning. It's been tested on another repo so I have confidence it'll work on this repo just fine. If it does break then fixing on Monday will be fine.

PR #65, opened 2020-06-15
Add matrix input to init action

robertbrignull:
I think this input is missing. It is used to construct the status reports.

Not sure how this is missing, especially as we have a query for it. Though to be fair we haven't gone through to eliminate all the cases of that as they are actually pretty benign in practice.


robertbrignull:
Sorry this PR may not make much sense. To review I think it's best to know that these inputs are used by the code to send status reports, and this is called from all actions. Then look at the `action.yml` files and note that all of them include both the `matrix` and `token` inputs, except for this one which is missing the `matrix` input. The effect this'll have is that the matrix isn't populated in the status reports, which isn't fatal and explains why we haven't noticed it until now.

robertbrignull:
1 fixed alert. Excellent. That's what I expected
https://github.com/github/codeql-action/pull/65/checks?check_run_id=776340119

PR #69, opened 2020-06-17
Use bundle URL version as the cache version

alexkappa:


lib/setup-tools.js
```diff
@@ -55,4 +56,21 @@ async function setupCodeQL() {
     }
 }
 exports.setupCodeQL = setupCodeQL;
+function getCodeQLURLVersion(url) {
+    const match = url.match(/codeql-bundle-([\d+(\.\d+)]+)/);
```
tibbes, line 60: The Code Scanning result on this line is a genuine problem. I think the square brackets on this line should be removed, if I understand the intended pattern correctly.

This is intended to match `codeql-bundle-6` and `codeql-bundle-8.2` and `codeql-bundle-8.3.4`, etc. right?

src/setup-tools.ts
```diff
@@ -48,3 +49,25 @@ export async function setupCodeQL(): Promise<CodeQLSetup> {
         throw new Error("Unable to download and extract CodeQL CLI");
     }
 }
+
+export function getCodeQLURLVersion(url: string): string {
+
+    const match = url.match(/codeql-bundle-([\d+(\.\d+)]+)/);
```
chrisgavin, line 55: I don't think this regex is quite right. The brackets and plus are part of the character class so they are being matched literally.

src/setup-tools.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    for (const version in tests) {
+        if (tests.hasOwnProperty(version)) {
+
+            const expectedVersion = tests[version];
+            const url = `https://github.com/.../codeql-bundle-${version}/...`;
+
+            try {
+                const parsedVersion = setupTools.getCodeQLURLVersion(url);
+                t.assert(parsedVersion, expectedVersion);
```
chrisgavin, line 60: I'm a bit surprised these tests pass. I don't think this `assert` is actually asserting equality.

alexkappa: Ooh! 😞 I guess the `parsedVersion` was truthy making me think my test was good. Harakiri.

src/setup-tools.ts
alexkappa: They would have to be escaped (e.g. `\[`) in order to be treated literally, no?

The alert would also make sense, being more liberal with matching since we are searching within the bounds of `/ ... /`.

lib/setup-tools.js
alexkappa: Indeed thats the intention. I would be inclined to oblige the alert and opting for something greedier: `/codeql-bundle-(.*)/`. We can let semver decide if the `.*` match is valid semver.

src/setup-tools.ts
chrisgavin: Sorry, by brackets I meant the inner `(` and `)` brackets. Because they are in a character class they are literal by default.

I think essentially this character class breaks down to matching:
* any digit
* `.`
* `+`
* `(`
* `)`

This means all the following match the regex, which I'm not sure is what you intended:
* `codeql-bundle-...`
* `codeql-bundle-.(.)().`
* `codeql-bundle-1(2)(3)`
* `codeql-bundle-1+2+3`

chrisgavin: If it is intended, I think the character class can just be simplified as `[\d()+.]` which would mean the same thing.

alexkappa: Ah, I see. No that was not the intention. Thanks!

src/setup-tools.test.ts
```diff
@@ -0,0 +1,66 @@
...
+            const url = `https://github.com/.../codeql-bundle-${version}/...`;
+
+            try {
+                const parsedVersion = setupTools.getCodeQLURLVersion(url);
+                t.deepEqual(parsedVersion, expectedVersion);
+            } catch (e) {
+                t.fail(e.message);
+            }
+        }
+    }
```
robertbrignull, line 65: You can avoid the `hasOwnProperty` check using `Object.entries(tests)`, e.g.
```suggestion
    for (const [version, expectedVersion] of Object.entries(tests)) {
        const url = `https://github.com/.../codeql-bundle-${version}/...`;

        try {
            const parsedVersion = setupTools.getCodeQLURLVersion(url);
            t.deepEqual(parsedVersion, expectedVersion);
        } catch (e) {
            t.fail(e.message);
        }
    }
```

robertbrignull:
The unit tests look really good to me and look like they validate the behaviour pretty well. I'm running a quick test locally too just to make sure, but I don't expect it to fail.

PR #73, opened 2020-06-22
Add threads flag to analyze action

alexkappa:
### Description

Adds an `inputs.threads` flag to the analyze command. Usage is as follows:

```yaml
- name: Perform CodeQL Analysis
  uses: github/codeql-action/analyze@v1
  with:
    threads: 4
```

The value of `threads` defaults to `1` if is not specified. If specified, will be capped to the vale of `os.cpus().length`.


src/finalize-db.ts
```diff
@@ -116,11 +98,11 @@ async function runResolveQueries(codeqlCmd: string, queries: string[]): Promise<
 
   await exec.exec(
     codeqlCmd, [
-      'resolve',
-      'queries',
-      ...queries,
-      '--format=bylanguage'
-    ],
+    'resolve',
```
chrisgavin, line 101: The formatting here also seems a big broken.

src/util.ts
```diff
@@ -393,3 +393,48 @@ export async function withTmpDir<T>(body: (tmpDir: string) => Promise<T>): Promi
     fs.rmdirSync(tmpDir, { recursive: true });
     return result;
 }
+
+/**
+ * Get the value specified for the `ram` input. If no value was specified, the
```
chrisgavin, line 398: The doc comment is not quite accurate. It doesn't return the input, it returns a flag to be passed to CodeQL.

(The same applies to the `getThreadsFlag()` method.)

src/util.ts
```diff
@@ -393,3 +393,48 @@ export async function withTmpDir<T>(body: (tmpDir: string) => Promise<T>): Promi
...
+   * The value will be capped to the number of available CPUs.
+   *
+   * @returns string
+   */
+export function getThreadsFlag(): string {
+    let numThreads = 1;
+    const numThreadsString = core.getInput("threads");
+    if (numThreadsString) {
+      numThreads = Number(numThreadsString);
+      if (Number.isNaN(numThreads) || numThreads < 0) {
```
chrisgavin, line 431: It might be worth allowing negative thread counts. [CodeQL treats it as leaving that number of cores free](https://help.semmle.com/codeql/codeql-cli/commands/query-run.html#cmdoption-codeql-query-run-j).

src/util.ts
```diff
@@ -393,3 +393,48 @@ export async function withTmpDir<T>(body: (tmpDir: string) => Promise<T>): Promi
     fs.rmdirSync(tmpDir, { recursive: true });
     return result;
 }
+
```
chrisgavin, line 396: It seems a bit odd to move these to `util` as they are specific to the CodeQL analysis step.

alexkappa: Indeed, my first instinct was to create a `finalize-db.test.ts` file and add all relevant test cases in there. That gave me some trouble unfortunately which CodeQL was the only thing that pointed to the problem: the `run()` function at the end of `finalize-db.ts`.

Now as to why `config.ts` in particular, I just saw other similar functions and thought it might be alright.

Ideally I would have liked to have this as follows:

`<action-name>/action.yml` -> `lib/<action-name>.js`
`lib/<action-name>.js` -> `lib/<library-or-package>.js`

But it would be quite a bigger change and one that merits discussion first. What are your thoughts about this?

src/util.ts
```diff
@@ -393,3 +393,48 @@ export async function withTmpDir<T>(body: (tmpDir: string) => Promise<T>): Promi
...
+        throw new Error("Invalid RAM setting \"" + memoryToUseString + "\", specified.");
+      }
+    } else {
+      const totalMemoryBytes = os.totalmem();
+      const totalMemoryMegaBytes = totalMemoryBytes / (1024 * 1024);
+      const systemReservedMemoryMegaBytes = 256;
+      memoryToUseMegaBytes = totalMemoryMegaBytes - systemReservedMemoryMegaBytes;
+    }
+    return "--ram=" + Math.floor(memoryToUseMegaBytes);
+  }
```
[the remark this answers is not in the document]
alexkappa: Nice catch, I will fix this. Sorry.

src/util.ts
alexkappa: Good one. Will relax the check to allow negative values. 

What about capping the negative values in that case? Should I cap the value to `-maxThreads`?

src/finalize-db.ts
alexkappa: For some reason vscode formatter did this. I'll avoid running it on the entire file. Will revert this changes.

src/util.ts
alexkappa: Changed it to read: "Get the codeql `--ram` flag as configured ...". Let me know if its ok.

src/util.ts
chrisgavin: That probably makes sense, or more specifically `1-maxThreads`.

src/util.test.ts
```diff
@@ -8,3 +9,55 @@ test('getToolNames', t => {
...
+  const totalMem = Math.floor(os.totalmem() / (1024 * 1024));
+
+  const tests = {
+    "": `--ram=${totalMem - 256}`,
+    "512": "--ram=512",
+  };
+
+  for (const [input, expectedFlag] of Object.entries(tests)) {
+
+    process.env['INPUT_RAM'] = input;
```
robertbrignull, line 24: Glad you worked out how to do this. I also added a function to do this at https://github.com/github/codeql-action/blob/main/src/config-utils.test.ts#L8

Would be good to consolidate these into one place. Perhaps we need a new file for utilities for use during tests. Then this function, and `withTmpDir` could move there.

robertbrignull: Ah, seems @sampart actually introduced a file like this in https://github.com/github/codeql-action/pull/75/files#diff-2bc785dec81a33ef3bbc6dbe25f2f8c5

For this PR perhaps just ignore this comment then and we'll come back to address stuff in a later PR once all the current ones are merged.

alexkappa: Ah great, I'll be sure to use that to be more consistent. It would surely give more context.

PR #74, opened 2020-06-22
Reduce debug output in tests

sampart:
The full test output is long, so here's an example of a single test before and after.

Before:

```
$ npm test

> codeql@0.0.0 test /Users/sam/code/codeql-action
> ava src/fingerprints.test.ts --serial


⠸ hash
⠼ hash

  1 passed::debug::Ignoring location URI "/src/foo/bar.js' as it is outside of the src root
::debug::Ignoring location URI "https:///Users/sam/code/codeql-action/lib/fingerprints.test.js' as the scheme is not recognised
::debug::Ignoring location URI "ftp:///Users/sam/code/codeql-action/lib/fingerprints.test.js' as the scheme is not recognised
::debug::Ignoring location as uri "1" is invalid
::debug::Ignoring location as uri "undefined" is invalid
::debug::Unable to compute fingerprint for non-existent file: /Users/sam/code/codeql-action/lib/fingerprints.test.js2

  4 tests passed
```

After:

```
$ npm test

> codeql@0.0.0 test /Users/sam/code/codeql-action
> ava src/fingerprints.test.ts --serial



  4 tests passed

```

There's still a fair bit of noise that isn't caught by this (e.g. output from system calls), and the output could do with more cleaning generally, but this is a start, and provides a framework for further improvements.

Thanks to @alexrford for pairing with me on much of this.


src/testing-utils.ts
```diff
@@ -0,0 +1,27 @@
...
+  const typedTest = test as TestInterface<{write: any}>;
+
+  typedTest.beforeEach(t => {
+      const processStdoutWrite = process.stdout.write.bind(process.stdout);
+      t.context.write = processStdoutWrite;
+      process.stdout.write = (str: Uint8Array | string, encoding?: any, cb?: (err?: Error) => void) => {
+          // Core library will directly call process.stdout.write for commands
+          // We don't want debug output to be included in tests
+          if (typeof str === "string") {
+            str = str.replace(/::(info|debug|warning).*/, '');
```
sampart, line 13: We can't assume that this is at the start of the line, because it gets interleaved with output from Ava, so you might see lines like:

```
6 passed::debug::{...}
```

sampart:
Replaced by https://github.com/github/codeql-action/pull/75

PR #77, opened 2020-06-23
Call codeql.exe rather than codeql.cmd wrapper on Windows

nickrolfe:
Ever since CodeQL 2.0.2 (~6 months ago IIRC), `codeql.cmd` has been a trivial wrapper around `codeql.exe` on Windows - it just forwards all arguments.

Since we may want to remove it in the future, I think it makes sense for the action to stop using the wrapper. I assume we are using a relatively recent version of the CodeQL tools on Actions, that includes `codeql.exe`.

Sorry, I'm not sure how to test this, and I may not have made the change correctly - I just edited the two files where I found the `.cmd` extension being added. One of them appears to be generated from the other, but I edited it manually.


robertbrignull:
Thanks for this. I'd heard codeql was switching to `exe` but was confused because the current call still works, but that explains it now if it's a trivial wrapper.

You're right the error is due to generated files. Run `npm run build` to generate them and check in the result.

robertbrignull:
Sorry I think this now conflicts because we've normalised all the indentation. You'll have to merge in `main` again. Once that's done and the CI passes then you can just merge.

nickrolfe:
Thanks, I've merged `main` and fixed the trivial conflict.

nickrolfe:
@robertbrignull thanks for your help. Can you please merge? I don't have write access to this repository.

PR #79, opened 2020-06-23
Only output anything from tests when there's a failure

robertbrignull:
This is an alternative / builds upon https://github.com/github/codeql-action/pull/75

The extra bits are:
- Run ava in verbose mode. This disables the fancy spinner and means that printing doesn't mess it up. I think ava is also doing some kind of output suppression but I haven't been able to fully verify, but anyway this turns it off.
- Instead of only suppressing some output, forward all of it to a string in memory, and then only print it all if the test fails. I think this gives a nice user experience by being very minimal when everything passes but giving you all the information when something fails.
- Uses `typedTest.afterEach.always` instead of `typedTest.afterEach` because the latter only runs on passing tests.


src/testing-utils.ts
```diff
@@ -0,0 +1,37 @@
+import {TestInterface} from 'ava';
+
+type TestContext = {stdoutWrite: any, stderrWrite: any, testOutput: string};
+
+function wrapOutput(context: TestContext) {
+  return (str: any): boolean => {
+    if (typeof str === 'string') {
```
sampart, line 7: Technically, `write` could also be passed a `Uint8Array`.  I think it's probably okay to just discard those, but it would be interesting to see what sort of things that's used for.  My testing suggests that it is used a couple of times (I had something fail when I had code that assumed it was always a string), but I don't know what for.

robertbrignull: I'm not really sure. I didn't notice anything missing but I agree it would be nice not to throw them away.

I couldn't find any good documentation on what to do with the `Uint8Array` or the `encoding` parameter so I've just ignored them. I'm not sure what the correct way to write them to a string would be.

In one experiment I was trying to use a `Stream` to store the data because that has the same interface, but I couldn't get it to work nicely as I also can't find any helpful info on streams. All the documentation is unintelligible to me.

sampart: Could you temporarily switch this so that it provides output on success only, and update `wrapOutput` to only store int arrays?  (They'll be of type `"object"`). Then you could see what output you get and whether it's anything worth worrying about.

If that doesn't throw up anything of interest, I'm happy that we just ignore everything except strings here.

robertbrignull: The output that comes as a byte array is apparently the output from git when we run it as a subprocess. So I think we do want to keep that, but I've found a way to do it.

sampart: I might have known it was git being special ;-). Nice one.

src/testing-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Satisfy contract by calling callback when done
+    if (cb !== undefined && typeof cb === 'function') {
+      cb();
+    }
+
+    return true;
+  };
+}
+
+export function silenceDebugOutput(test: TestInterface<any>) {
```
robertbrignull, line 33: In hindsight, should perhaps rename this to something else. I'll open a new PR for that.

PR #81, opened 2020-06-23
Report that an action has been aborted on configuration failures

alexkappa:
### Description

This PR adds support for reporting with `status: aborted` if a configuration failure has occurred.

### Dependencies

We shouldn't merge this PR until the `/repositories/{repository_id}/code-scanning/analysis/status` [schema](https://github.com/github/github/blob/71ebdf179f407d6ebee328d7f3740ee409f375ff/app/api/schemas/v3/schemas/code-scanning-analysis.json#L59-L61) has been updated to allow for the new status.


src/util.ts
```diff
@@ -365,6 +365,16 @@ export async function reportActionSucceeded(action: string) {
...
 }
 
+/**
+ * Report that an action has been aborted.
+ *
+ * Note that the started_at date is always that of the `init` action, since
+ * this is likely to give a more useful duration when inspecting events.
+ */
+export async function reportActionAborted(action: string, cause?: string) {
+  await sendStatusReport(await createStatusReport(action, 'abort', cause));
```
robertbrignull, line 375: I think the status accepted by the dotcom endpoint is `aborted` instead of `abort`

src/setup-tracer.ts
```diff
@@ -139,30 +139,38 @@ function concatTracerConfigs(configs: { [lang: string]: TracerConfig }): TracerC
...
     core.startGroup('Load language configuration');
 
-    const languages = await util.getLanguages();
+    const config = await configUtils.loadConfig();
+
+    languages = await util.getLanguages();
     // If the languages parameter was not given and no languages were
     // detected then fail here as this is a workflow configuration error.
     if (languages.length === 0) {
       core.setFailed("Did not detect any languages to analyze. Please update input in workflow.");
```
robertbrignull, line 159: Would it be reasonable for this case to also report `aborted`?

src/util.ts
alexkappa: Good eye 👀 🕶️

src/setup-tracer.ts
alexkappa: Hmm, yes I suppose you're right. What would happen so far? Would the `starting` report never be paired with a `success` or `failure`?

I am thinking about throwing an error here and letting the `catch` statement catch it.

robertbrignull: Yes, I think we'd just never get another report to pair with the `starting` report.

Throwing an error so there's only one place that does the reporting sounds good to me.

alexkappa:
> Could you run a test?

Looking good in Splunk: https://git.io/JfpSM

PR #82, opened 2020-06-23
Use a single Octokit client for everything rather than a bunch of Octokits and an HTTP client.

chrisgavin:
Currently our requests to the GitHub API are pretty inconsistent. In a few places we use an Octokit, but we construct them at the point of use potentially leading to them being initialized inconsistently. We also sometimes just use an HTTP client.

Perhaps most importantly, none of these methods actually respect the `GITHUB_API_URL` environment variable that tells you where to find the GitHub API.

Because of this I've replaced these custom Octokits and the HTTP client with a single Octokit to ensure that all requests are made consistently and to the correct host.


src/api-client.ts
```diff
@@ -0,0 +1,11 @@
...
+import * as octokit from "@octokit/rest";
+import consoleLogLevel from "console-log-level";
+
+const githubAPIURL = process.env["GITHUB_API_URL"] || "https://api.github.com";
+export const client = new octokit.Octokit({
+  auth: core.getInput("token"),
+  baseUrl: githubAPIURL,
+  userAgent: "CodeQL Action",
+  log: consoleLogLevel({ level: "debug" })
+});
```
alexkappa, line 11: Eventually could we make a class that wraps `octokit.Octokit` and supplies higher level functions [a-la octokit](https://octokit.github.io/rest.js/v18#repos-create-release)? 

For example

```js
octokit.repos.codeScanning.uploadAnalysis({
  owner,
  repo,
  analysis
});
```

robertbrignull: Or hopefully once we're fully out of beta then we can have our methods in the proper Octokit.

robertbrignull:
I think a couple of places that making using custom requests could use built-in octokit methods, but making that change at a later date sounds good to me.

src/api-client.ts
chrisgavin: Yep. Octokit actually does have methods to add custom endpoints which I did experiment with, but it unfortunately adds typing-related complications. Ideally these methods are eventually part of the real API.

chrisgavin:
> I think a couple of places that making using custom requests could use built-in octokit methods, but making that change at a later date sounds good to me.

100%. I was actually pretty surprised we didn't already use built-in methods for things like getting the languages of a repository.

src/api-client.ts
```diff
@@ -0,0 +1,11 @@
+import * as core from "@actions/core";
+import * as octokit from "@octokit/rest";
+import consoleLogLevel from "console-log-level";
+
+const githubAPIURL = process.env["GITHUB_API_URL"] || "https://api.github.com";
```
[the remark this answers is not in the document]
chrisgavin: Hmm, in a test job I did it existed on both. e.g. See https://github.com/chrisgavin/test-actions/runs/803393860?check_suite_focus=true.

I'm happy to require it exists, though I experienced some very strange issues when trying to use `util.getRequiredEnvParam(...)` here. Specifically it seems like the Action fails at runtime with the error `getRequiredEnvParam` is not a function. I wondered if it could be due to an import loop or something but I never managed to figure it out.

PR #85, opened 2020-06-26
getRequiredEnvParams must be non-empty

rneatherway:


src/util.test.ts
```diff
@@ -64,3 +64,10 @@ test('getThreadsFlag() throws if the ram input is < 0 or NaN', t => {
     t.throws(util.getThreadsFlag);
   }
 });
+
+test('getRef() throws on the empty string', t => {
+  for (const input of [""]) {
```
robertbrignull, line 69: Is the use of a loop to make it easier to add new cases? In that case the test name may be a bit too specific. Not a big issue though if you want to keep it this way

PR #87, opened 2020-06-26
Allow config files to be from a remote repository

sampart:
You can see an example of this in use in [a branch that builds off this one](https://github.com/github/codeql-action/compare/support-remote-config...remote-config-example).  Here's a deep link to the Actions run, showing the HTTP request: https://github.com/github/codeql-action/runs/811641896?check_suite_focus=true#step:4:27


src/config-utils.ts
```diff
@@ -250,6 +262,57 @@ function initConfig(): Config {
...
+async function getRemoteConfig(configFile: string): Promise<any> {
+  // retrieve the various parts of the config location, and ensure they're present
+  const format = new RegExp('(?<owner>[^/]+)/(?<repo>[^/]+)/(?<path>[^@]+)@(?<ref>.*)');
+  const pieces = format.exec(configFile);
+  // 5 = 4 groups + the whole expression
+  if (pieces === null || pieces.groups === undefined || pieces.length < 5) {
+    throw new Error(getConfigFileRepoFormatInvalidMessage(configFile));
+  }
+
+  const response = await api.client.repos.getContents({
```
sampart, line 297: Two things to note here.

1. This API method only supports files up to 1mb.  I presume that limit is okay?  If not, we could use `
octokit.git.getBlob` ([docs](https://octokit.github.io/rest.js/v18#git-get-blob)), retrieving the `file_sha` via `octokit.repos.getContent` ([docs](https://octokit.github.io/rest.js/v18#repos-get-content)). ([Slack discussion](https://github.slack.com/archives/C7KL0J81Y/p1593096158214000)).
2. This method is renamed to `getContent` in v4 of `@octokit/plugin-rest-endpoint-methods` ([source](https://unpkg.com/browse/@octokit/plugin-rest-endpoint-methods@4.0.0/dist-types/generated/method-types.d.ts#L4601), [discussion](https://github.com/octokit/plugin-rest-endpoint-methods.js/issues/151)) and marked as deprecated in newer 3.x versions - we're on 3.3.1 at present.

robertbrignull: I think both of these points are fine.

1) I wouldn't expect one of these config files to ever go over 1mb and if it does then there's something very odd.

2) This is fine. When we upgrade to v4 we'll have to fix this, but it'll be a compilation error that we'll notice when upgrading so it'll be fine.

sampart:
Thanks @robertbrignull.  I've had a go at adding an integration test in 4b37db72e44f5b068fbdddd0ca0ff7d4a9d11b69; I'd appreciate your eyes on that one, thanks!  In particular, can we simplify the checkout?

robertbrignull:
I'd copy what the existing integration tests do, instead of using the format from the public template. There's no need to do the `fetch-depth` and custom checkout step for the integration tests. I'd just copy `multi-language-repo_test-custom-queries` but change the reference to the config file to be a remote repository reference.

robertbrignull:
Or you could even change that existing test so it tests both aspects at once. We don't have a good guideline for this. I'm wary of introducing too many extra tests, especially if they all test on multiple OS's, because the tests will start taking up too many action minutes and take too long. It's unclear if we've reached that point or not yet. And since we can always change the test setup at any point, whatever you do we can always adjust in the future.

sampart:
Sounds like a plan, thanks.  Done in 1bb294af6b763c1da9d0d010aa587eeb3b4cdb5c.

PR #88, opened 2020-06-29
Remove `strategy: fail-fast`

robertbrignull:
Setting `strategy: fail-fast` has no effect unless the workflow is run as part of a matrix build. It controls whether one matrix build failing causes other builds in that same matrix to be aborted. It has no effect within an individual run.

See https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast for more info.

I'm removing this from the suggested usage example because it is unnecessary and misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,6 @@ on:
 
 jobs:
   CodeQL-Build:
-
-    strategy:
-      fail-fast: false
-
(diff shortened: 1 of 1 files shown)


robertbrignull:
I'm not sure. I also remember removing it.

PR #90, opened 2020-06-29
add analysis_key to status reports

robertbrignull:
Adds the `analysis_key` field to the status reports. This shouldn't be merged until some work has been done on the server side to allow this field, but I'm opening the PR now so it doesn't get forgotten about.


robertbrignull:
This should work now. Let's find out

PR #94, opened 2020-07-01
Add development instructions to CONTRIBUTING.md

sampart:


CONTRIBUTING.md
```diff
@@ -10,6 +10,24 @@ Contributions to this project are [released](https://help.github.com/articles/gi
 
 Please note that this project is released with a [Contributor Code of Conduct][code-of-conduct]. By participating in this project you agree to abide by its terms.
 
+## Development and Testing
+
+Before you start, you may want to ensure that you have a recent version of node installed. You can see which version of node is installed on the actions runners here: https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners
```
rneatherway, line 15: It'd be nice to include the node version (12) here, unless you are worried about it going out of date. I doubt this will change very quickly.

robertbrignull: You can also point to https://github.com/github/codeql-action/blob/main/init/action.yml which is the actual source of truth

sampart: Good plan, thanks!  Done in dfed1f7eea4b6665f383bc25f4f408ab2a48df75

PR #102, opened 2020-07-07
Exclude lib directory from analysis

robertbrignull:
I don't think we want to be analyzing the `lib` directory as that's all generated code and we're currently getting some duplicate alerts showing up.

It'll theoretically also show up alerts from dependencies, although I can't actually see any such alerts right now. Although this could be informative, it's not the way that codeql is meant to be used. It'll be more useful to rely on things like CVEs and dependabot to find out about vulnerabilities in dependencies.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -11,3 +11,4 @@ queries:
   - uses: security-and-quality
 paths-ignore:
   - tests
+  - lib


robertbrignull:
Good, this reports 1 fix as I would expect
https://github.com/github/codeql-action/pull/102/checks?check_run_id=845489749

PR #109, opened 2020-07-13
Move config parsing earlier + add to codeql search path

robertbrignull:
This change started out just trying to add `--search-path` to the CodeQL args when resolving queries, however in doing that I wasn't able to get the data where I needed it nicely. A simpler but cruder answer could have been to stuff some extra fields in the objects already being passed around, but that didn't seem ideal to me. Instead this PR restructures how we parse and process the config considerably to move everything sooner and to one place.

The main changes are:
- All config parsing and subsequent processing happens in the `init` action. Previously it was mostly in the `init` action, but there was a large chunk in the `analyze` action where we checked out external repositories and converted the user-supplied list of queries and query suites to run into an explicit list of query files.
- The data stored in the internal config file that persists between actions has changed. Some extra data has been added like the list of languages. Some data has changed format, such as the set of queries which is now an explicit array of filepaths to existing ql files.

This is quite a big change, but I think is has a few benefits:
- The user gets notified of problems sooner, because we validate everything in the `init` action. Previously if you specified queries from a remote repository but the repository or file path didn't exist, then you wouldn't find out until after the code has been built.
- All config parsing is done in one place. I hope this makes it easier to us to understand and develop. It also makes it easier to do features like adding the CodeQL `--search-path` because all of the data is available.
- By storing more data and more explicit data in the internal config file, and treating the data as immutable, it makes the other actions simpler.


src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
...
-      let absoluteQueryPath = path.join(workspacePath, localQueryPath);
-
-      // Check the file exists
-      if (!fs.existsSync(absoluteQueryPath)) {
-        throw new Error(getLocalPathDoesNotExist(configFile, localQueryPath));
-      }
+/**
+ * Parse the suitename to a set of queries and update resultMap.
+ */
+async function parseBuiltinSuite(
```
sampart, line 113: If you're changing the comment to avoid the word "parse" you should also rename this function.  How about `addSuiteQueriesToResultMap`?

src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
...
+  const suites = languages.map(l => l + '-' + suiteName + '.qls');
+  await runResolveQueries(resultMap, suites, undefined, false);
+}
 
-      // Check the local path doesn't jump outside the repo using '..' or symlinks
-      if (!(absoluteQueryPath + path.sep).startsWith(workspacePath + path.sep)) {
-        throw new Error(getLocalPathOutsideOfRepository(configFile, localQueryPath));
-      }
+/**
+ * Parse the local path to a set of queries and update resultMap.
```
sampart, line 129: Similarly to the above - you're not parsing the path, you're just using it to find files.  I don't know enough about how the resolve queries function works to be certain about wording here, but how about:

```suggestion
 * Retrieve the set of queries at localQueryPath and add them to resultMap.
```

src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
...
+ */
+async function parseLocalQueryPath(
```
sampart, line 131: This will need renaming if you change the comment.  How about `addLocalQueriesToResultMap`?

src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
...
-    }
+  await runResolveQueries(resultMap, [absoluteQueryPath], rootOfRepo, true);
+}
 
-    // Check none of the parts of the repository name are empty
-    if (tok[0].trim() === '' || tok[1].trim() === '') {
-      throw new Error(getQueryUsesInvalid(configFile, queryUses));
-    }
+/**
+ * Parse the remote repo reference to a set of queries and update resultMap.
```
sampart, line 161:
```suggestion
 * Retrieve the set of queries at the referenced remote repo and add them to resultMap.
```

src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
...
+ */
+async function parseRemoteQuery(configFile: string, resultMap: { [language: string]: string[] }, queryUses: string) {
```
sampart, line 163: How about `addRemoteQueriesToResultMap`?

src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
...
+ * Parse a query 'uses' field to a discrete set of query files and update resultMap.
+ */
+async function parseQueryUses(
+  configFile: string,
+  languages: string[],
+  resultMap: { [language: string]: string[] },
+  queryUses: string) {
+
+  // The logic for parsing the string is based on what actions does for
+  // parsing the 'uses' actions in the workflow file
```
sampart, line 204: Does this comment relate just to the 4 lines that follow?  If not, I'd put it in the function-level comment.

src/config-utils.ts
```diff
@@ -5,112 +5,222 @@ import * as yaml from 'js-yaml';
 import * as path from 'path';
 
 import * as api from './api-client';
+import { getCodeQL } from './codeql';
+import * as externalQueries from "./external-queries";
 import * as util from './util';
 
+// Property names from the user-supplied config file.
 const NAME_PROPERTY = 'name';
 const DISPLAY_DEFAULT_QUERIES_PROPERTY = 'disable-default-queries';
```
sampart, line 14: I know this wasn't added in this PR, but you're making use of it in your changed code, and that code would be much easier to read if it was called `DISABLE_DEFAULT_QUERIES_PROPERTY`.

Consider this bit, for example:

```
if (!parsedYAML[DISPLAY_DEFAULT_QUERIES_PROPERTY]) {
  await addDefaultQueries(languages, queries);
}
```

src/config-utils.ts
```diff
@@ -326,10 +524,33 @@ async function initConfig(): Promise<Config> {
...
+}
+
+/**
+ * Load and return the config.
+ *
+ * This will parse the config from the user input if present, or generate
+ * a default config. The parsed config is then stored to a known location.
+ */
+export async function initConfig(): Promise<Config> {
+  let configFile = core.getInput('config-file');
```
sampart, line 541:
```suggestion
  const configFile = core.getInput('config-file');
```

src/config-utils.ts
```diff
@@ -384,35 +605,46 @@ async function getRemoteConfig(configFile: string): Promise<any> {
...
+function getParsedConfigFolder(): string {
   return util.getRequiredEnvParam('RUNNER_TEMP');
 }
 
-export function getConfigFile(): string {
-  return path.join(getConfigFolder(), 'config');
+/**
+ * Get the file path where the parsed config will be stored.
+ */
+export function getParsedConfigFile(): string {
```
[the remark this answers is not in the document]
sampart: If you do this, I guess the previous method should be renamed to something like `getPathToParsedConfigFileFolder`, although that does sound a bit clunky.

src/external-queries.test.ts
```diff
@@ -2,22 +2,16 @@ import test from 'ava';
...
-  config.externalQueries = [
-    new configUtils.ExternalQuery("github/codeql-go", "df4c6869212341b601005567381944ed90906b6b"),
-  ];
-
   await util.withTmpDir(async tmpDir => {
     process.env["RUNNER_TEMP"] = tmpDir;
-    await externalQueries.checkoutExternalQueries(config);
+    await externalQueries.checkoutExternalRepository("github/codeql-go", "df4c6869212341b601005567381944ed90906b6b");
 
     // COPYRIGHT file existed in df4c6869212341b601005567381944ed90906b6b but not in master
```
sampart, line 16:
```suggestion
    // COPYRIGHT file existed in df4c6869212341b601005567381944ed90906b6b but not in main
```

src/external-queries.ts
```diff
@@ -3,26 +3,26 @@ import * as exec from '@actions/exec';
 import * as fs from 'fs';
 import * as path from 'path';
 
-import * as configUtils from './config-utils';
 import * as util from './util';
 
-export async function checkoutExternalQueries(config: configUtils.Config) {
+/**
+ * Checkout a repository at the given ref, and return the directory of the checkout.
```
sampart, line 9:
```suggestion
 * Check out repository at the given ref, and return the directory of the checkout.
```

Removing the "a" makes it clearer how you're using the `repository` parameter.

src/config-utils.ts
robertbrignull: Thanks for spotting that. I think that must just be a slip of the keyboard from when that property was originally added.

src/external-queries.test.ts
robertbrignull: The default branch on https://github.com/github/codeql-go is still called `master`. I'll make the comment more clear that it's talking about the default branch without naming it.

PR #111, opened 2020-07-14
Add support for debugging from VSCode

sampart:
https://github.com/avajs/ava/blob/master/docs/recipes/debugging-with-vscode.md recommends passing the `debug` option to Ava.  However, in my experience Ava waits indefinitely when this option is used, without actually running any tests.

The increased timeout setting is so that tests won't fail prematurely because you're slowly stepping through code.

- **N/A** Confirm this change is backwards compatible with existing workflows.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,25 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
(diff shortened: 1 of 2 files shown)


sampart:
The `--debug` option doesn't cause hanging for me either (although the extended timeout is still needed), but passing `debug` without dashes (as recommended in the article I linked to) to invoke ava like `ava debug` does.  But, as you say, this works, so let's go with it for now and we can always revise it. Thanks

PR #117, opened 2020-07-20
Allow the codeql-action to be run locally

aeisenberg:
This change allows the codeql-action to be run locally through
[act](https://github.com/nektos/act).

In order to run the action locally, you need to do two things:

1. Add the `CODEQL_LOCAL_RUN: true` environment variable. The only way
   I could figure out how to do this was to add it directly in the
   workflow file in an `env` block. It _should_ be possible to add it
   through a `.env` file and pass it to `act`, but I couldn't get it
   working.
2. Run this command `act -j codeql -s GITHUB_TOKEN=<MY_PAT>`

Setting the `CODEQL_LOCAL_RUN` env var will fill in missing env vars
that the action needs, but isn't set by `act`. It will also avoid
making api calls to github that would fail locally.


src/util.ts
```diff
@@ -45,6 +45,29 @@ export function getRequiredEnvParam(paramName: string): string {
...
+/**
+ * Ensures all required environment variables are set in the context of a local run.
+ */
+export function prepareEnvironment() {
+  if (!isLocalRun()) {
+    return;
+  }
+
+  core.debug('Action is running locally.');
+  if (!process.env.RUNNER_TEMP) {
```
aibaars, line 63: I'm a bit surprise that this line is needed. See also https://github.com/nektos/act/blob/64b8d2afa47ff6938d7617e00f1260a95c35268e/pkg/runner/run_context.go#L82

aibaars:
Setting variables in a file named `.env` seems to work for me. Note that the flag is just a filename, which is resolved against the current working directory.

src/util.ts
aibaars: When I run a workflow containing the following with `act`
```
    - run: |
       printenv
```
I see that `RUNNER_TEMP=/tmp`.

aibaars:
> Hmmm...The .env did not work for me. I wonder if I'm using an older version than you. Also, are you on Linux?

Yes,  on linux.

src/util.ts
aeisenberg: Hmmmm. Again, this was not the case for me. I must be running things differently somehow or a different version of act. I'll have to explore.

robertbrignull:
When generating the status reports, we access the github API to get the workflow ID. This would also need to be dealt with somehow. Probably best to audit all uses of the `getApiClient` method.

aeisenberg:
@robertbrignull I believe that is already handled here: https://github.com/github/codeql-action/pull/117/files#diff-8cd4968b81985f0efc2053eabc37db6dR263

Under local mode, no status messages are sent.

robertbrignull:
> I believe that is already handled here: https://github.com/github/codeql-action/pull/117/files#diff-8cd4968b81985f0efc2053eabc37db6dR263

That stops the reports from being sent, but the call I was thinking of is when we are constructing the data to send, which happens before that. However now I look closely it is already covered by https://github.com/github/codeql-action/pull/117/files#diff-8cd4968b81985f0efc2053eabc37db6dR102 as this is the method I was thinking of.

> There is at least one call to apiClient that we need https://github.com/github/codeql-action/blob/main/src/config-utils.ts#L404. Though, I could probably still explicitly fail other requests if they are invoked from local mode.

I haven't used `act` before. Presumably it still runs in the context of some repository? In which case I expect that call should work as normal. I do wonder if it would make for a safer situation if we took the stance of denying all HTTP requests in local mode, and allowing only the ones we know to be safe, instead of the current situation which is the opposite.

aeisenberg:
Yes, that is what I was proposing indirectly above. I could do something like this:

```ts
export const getApiClient = function(allowLocalRun = false) {
  if (isLocalRun() && !allowLocalRun) {
    throw new Error('Invalid API call in local run');
  }
  ...
```

And then for calls we know to be safe, we pass the `true` flag. And for calls that aren't, we will fail-fast and then need to handle them differently.

src/util.ts
aeisenberg: Ha! Upgraded from 0.2.7 -> 0.2.10 and that environment variable is now being set. Knowing this yesterday would have made things go a lot more smoothly.

aeisenberg: Thanks for catching that.

aeisenberg:
I'm confused why I'm getting errors on the _Code scanning_ action. This is coming from code that I don't think I have changed. It looks like this check was not run for #116, though it was run for #115 (and it failed, too, but for different checks).

Is it possible the `codeql` action is not being run for local pull requests?

aeisenberg:
Even more confusing is that this failing check is not showing up in the actions tab, but previous calls to the codeql actions were invoked on this branch and were successful. Eg- https://github.com/github/codeql-action/actions/runs/176329168

robertbrignull:
The Code Scanning check is the one that reports changes in alerts. So when it is marked as failed it's because that PR introduces alerts. Things might be a little confused right now because we had to revert some changes to the `main` branch so this reintroduced some alerts to that branch that had been fixed.

aeisenberg:
Thanks for the clarification.

aeisenberg:
@robertbrignull is this a worthwhile thing to add to the action?

src/finalize-db.ts
```diff
@@ -80,7 +82,7 @@ async function run() {
     core.info('Analyzing database');
     await runQueries(databaseFolder, sarifFolder, config);
 
-    if ('true' === core.getInput('upload')) {
+    if ('true' === core.getInput('upload') && !util.isLocalRun()) {
```
robertbrignull, line 85: Could you move this check to inside the `upload_lib.upload` method? That way it'll cover the `uploadSarif` action too.

Also from the point of view of testing more code in the action, it might be better to move this check to immediately before we make the action upload, in order to skip less code. The only major thing this would affect is the fingerprinting code, which might be nice to test but is not essential.

src/api-client.ts
```diff
@@ -2,7 +2,12 @@ import * as core from "@actions/core";
 import * as github from "@actions/github";
 import consoleLogLevel from "console-log-level";
 
-export const getApiClient = function() {
+import { isLocalRun } from "./util";
+
+export const getApiClient = function(allowLocalRun = false) {
```
robertbrignull, line 7: Thanks for adding this. I like that API calls are now safe by default in local mode.

src/finalize-db.ts
aeisenberg: Make sense.

src/util.ts
```diff
@@ -45,6 +45,26 @@ export function getRequiredEnvParam(paramName: string): string {
...
+export function isLocalRun(): boolean {
+  return !!process.env.CODEQL_LOCAL_RUN
+    && process.env.CODEQL_LOCAL_RUN !== 'false'
+    && process.env.CODEQL_LOCAL_RUN !== '0';
+}
+
+/**
+ * Ensures all required environment variables are set in the context of a local run.
+ */
+export function prepareEnvironment() {
```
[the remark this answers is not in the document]
aeisenberg: Sure.

aeisenberg:
@sampart I added a section to the readme. Should I move that to contributing?

sampart:
Actually, moving it sounds like a great idea, thanks.  Please don't fully replace what's there, as I think it's helpful that we also tell people about the option of pushing a branch (that's simpler than using act), but please do add the info in.  I don't think we need it in the README if it's in the Contributing guide.  Thanks!

aeisenberg:
Right...this is a little different from the _Running the action_ section. The instructions are for running from the main repository, but locally through act. So, to get this working, you need to run this from a repository different from this one. It's a subtle difference that I didn't call out in the README. I'll make that explicit. The main use case is to debug why a specific repository's codeql action is failing.

sampart:
Oh, I totally hadn't realised that distinction, thanks for mentioning.  Would an end-user of the action use this to troubleshoot their own workflows that use the action?  If so, README would be the right place for this, after all.

CONTRIBUTING.md
```diff
@@ -26,6 +26,25 @@ This project also includes configuration to run tests from VSCode (with support
...
+  ```bash
+  CODEQL_LOCAL_RUN=true
+
+  # Optional, for better logging
+  GITHUB_JOB=<ANY_JOB_NAME>
+  ```
+
+1. Run `act -j codeql -s GITHUB_TOKEN=<PAT>`
+
+Running locally will generate the CodeQL database and run all the queries, but it will avoid uploading and reporting results to GitHub. Note that this must be done a repository that _consumes_ this action, not this repository. The use case is to debug failures of this action on specific repositories.
```
joshhale, line 46:
```suggestion
Running locally will generate the CodeQL database and run all the queries, but it will avoid uploading and reporting results to GitHub. Note that this must be done on a repository that _consumes_ this action, not this repository. The use case is to debug failures of this action on specific repositories.
```

aeisenberg:
@sampart Probably not useful for end users. So, I think contributing is the right place for it. When I discussed this earlier, the use case would be for maintainers or contributors to determine why a particular repository is failing by reproducing locally.

robertbrignull:
@aeisenberg a bunch of js files have got into the `src` directory somehow. They should be in `lib`, so I'm not sure how that happened.

aeisenberg:
Darn. Let me fix that. Not sure what I did.

aeisenberg:
I'm out tomorrow and Monday. Please merge if you feel like this is good. Otherwise, I'll look at it when I get back.

PR #119, opened 2020-07-21
Revert "Merge main into v1"

robertbrignull:
Reverts github/codeql-action#114 as we unfortunately discovered a bug in this week's update. We'll work on fixing it and do another push to `main` or wait until next week.

The bug relates to when a config file is specified but does not include the `disable-default-queries` option and we erroneously don't add the default queries.

@marcogario


robertbrignull:
The "failing check" is because the PR we're reverting fixed trivial alerts, so now we're reintroducing them. But they're harmless.

PR #120, opened 2020-07-21
Default queries fix + reset env vars in tests

robertbrignull:
This primarily fixes a bug from https://github.com/github/codeql-action/pull/109 where we were not running the default queries in certain cases and this was causing workflows to fail. If a config file was provided but `disable-default-queries` was not in that config file then we were not running the default queries.

I've added a unit test to catch this case. I considered adding an integration test too but wasn't sure if that was overkill and a waste of resources given how much longer the integration tests take to run.

While writing the unit test, I discovered another issue that I have also included in this PR. Many of the unit tests set environment variables because this is how action input values are controlled. These values were persisting between tests and causing interference. In this we I add some code to reset environment variables between tests, and add in all the necessary declarations to get the tests to pass again. I could pull this part out to a separate PR if that would be preferable, though the two PRs would conflict.


src/testing-utils.ts
```diff
@@ -49,6 +49,12 @@ export function setupTests(test: TestInterface<any>) {
     const processStderrWrite = process.stderr.write.bind(process.stderr);
     t.context.stderrWrite = processStderrWrite;
     process.stderr.write = wrapOutput(t.context) as any;
+
+    // Many tests modify environment variables. Take a copy now so that
+    // We reset them after the test to keep tests independent of each other.
+    // process.env only has strings fields, so a shallow copy is fine.
+    t.context.env = {};
+    Object.assign(process.env, t.context.env);
```
sampart, line 57:
```suggestion
    Object.assign(t.context.env, process.env);
```

Is this the wrong way round?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign says:

> The Object.assign() method copies all enumerable own properties from one or more source objects to a target object. It returns the target object.
> `const returnedTarget = Object.assign(target, source);`

src/config-utils.test.ts
```diff
@@ -199,6 +200,51 @@ test("load non-empty input", async t => {
   });
 });
 
+test("default queries are used", async t => {
+  return await util.withTmpDir(async tmpDir => {
+    process.env['RUNNER_TEMP'] = tmpDir;
+    process.env['GITHUB_WORKSPACE'] = tmpDir;
+
+    // Check that the default behaviour is to add the default queries.
+    // In this case if a config file is specified by does not include
```
sampart, line 209:
```suggestion
    // In this case if a config file is specified but does not include
```

src/config-utils.test.ts
```diff
@@ -199,6 +200,51 @@ test("load non-empty input", async t => {
...
+          byLanguage: {
+            'javascript': {},
+          },
+          noDeclaredLanguage: {},
+          multipleDeclaredLanguages: {},
+        };
+      },
+    });
+
+    // Just create a generic config object with non-default values for all fields
```
sampart, line 228:
```suggestion
    // Just create a generic config object with a non-default value for one of the fields
```

In the [other place](https://github.com/github/codeql-action/blob/main/src/config-utils.test.ts#L170-L180) where you use this comment, you specify way more fields, hence my suggested change.

robertbrignull: Thanks. That comment was copy-pasted and I didn't spot it was now incorrect. I've updated to say the actual important points of this config.

robertbrignull:
Note I've also run some tests in a private repo with a config that triggers the bug. I've checked that I can reproduce it, and that it's fixed both by reverting the previous PR, and by merging this PR.

src/config-utils.test.ts
```diff
@@ -225,7 +225,9 @@ test("default queries are used", async t => {
       },
     });
 
-    // Just create a generic config object with non-default values for all fields
+    // The important point of this config is that is doesn't specify
```
sampart, line 228:
```suggestion
    // The important point of this config is that it doesn't specify
```

PR #121, opened 2020-07-22
Use templated strings for better readability

aeisenberg:


src/fingerprints.ts
```diff
@@ -167,15 +167,15 @@ export function resolveUriToFile(location: any, artifacts: any[]): string | unde
       location.index < 0 ||
       location.index >= artifacts.length ||
       typeof artifacts[location.index].location !== 'object') {
-      core.debug('Ignoring location as index "' + location.index + '" is invalid');
+      core.debug(`Ignoring location URI "${location.index}" is invalid`);
```
robertbrignull, line 170: The text changed here too which looks like a mistake

src/fingerprints.ts
```diff
@@ -167,15 +167,15 @@ export function resolveUriToFile(location: any, artifacts: any[]): string | unde
...
       return undefined;
     }
     location = artifacts[location.index].location;
   }
 
   // Get the URI and decode
   if (typeof location.uri !== 'string') {
-    core.debug('Ignoring location as uri "' + location.uri + '" is invalid');
+    core.debug(`Ignoring location URI "${location.uri}" is invalid`);
```
robertbrignull, line 178: The removal of the word "as" makes this harder to read, though I admit it is an error message so perfect grammar is not of the upmost importance. Still I think the removal as "as" is negative change.

PR #123, opened 2020-07-24
Search for the CodeQL bundle in multiple places.

chrisgavin:
The aim of this change is to resolve a number of issues with where the CodeQL bundle is fetched from. In many cases it is either not possible or not desired to fetch the CodeQL bundle from the canonical location at `https://github.com/github/codeql-action/releases/download/codeql-bundle-20200630/codeql-bundle.tar.gz`. This might be because you are using the Action on a runner that does not have internet access. It may also be because you've copied the CodeQL Action repository to your own organization so that you can control what code is running on your runner machines.

The general idea of this change was inspired by the way the Actions runner attempts to find an Action. We can apply a similar principle to finding the CodeQL bundle. This process works by looking for the CodeQL bundle in a place that's as local as possible and then working outwards. It checks if the bundle exists:

1. On $our_github_instance at the repository name for the current Action.
2. On $our_github_instance at github/codeql-action.
3. On GitHub.com at github/codeql-action.

It downloads the bundle from the first place it finds it. For symmetry I initially considered also checking GitHub.com at the repository name for the current Action, however I decided against that due to the accidental security risk someone might expose themselves to if they incorrectly mirror the Action without copying the bundle.

To check whether the bundle exists in these locations we use the GitHub API. This is required because on GHES you must authenticate to access the API, even for accessing public projects. The one exception to this is when we're finally getting the bundle from the canonical location. In this case we do not use the API. This is because if we access the API from GHES we cannot authenticate with our token, meaning we have to make the request anonymously. You could easily run out of API rate limit making these anonymous requests. Because we know for sure the bundle exists in the canonical location we can safely download it directly rather than using the API.

This change currently relies on two hacks to work around limitations in other parts of Actions. I've opened pull requests to solve these limitations but it's unclear when they might be reviewed. The two pull requests are:
* actions/runner#585 to avoid having to extract the current Action name from the file path at which it happens to be running.
* actions/toolkit#530 to allow providing headers directly to the `downloadTool(...)` method of the toolcache.

Sorry for the excessive description. This change ended up being a bit more complicated than I expected and I wanted to make it clear why I made the design decision I have.


src/codeql.ts
```diff
@@ -74,16 +82,107 @@ let cachedCodeQL: CodeQL | undefined = undefined;
...
+  // Duplicates will happen either because the GitHub instance is GitHub.com, or because the Action is not a fork.
+  const uniqueDownloadSources = potentialDownloadSources.filter((url, index, self) => index === self.indexOf(url));
+  for (let downloadSource of uniqueDownloadSources) {
+    let [apiURL, repository] = downloadSource;
+    // If we've reached the final case, short-circuit the API check since we know the bundle exists and is public.
+    if (apiURL === GITHUB_DOTCOM_API_URL && repository === CODEQL_DEFAULT_ACTION_REPOSITORY) {
+      return `https://github.com/${CODEQL_DEFAULT_ACTION_REPOSITORY}/releases/download/${CODEQL_DEFAULT_BUNDLE_VERSION}/${CODEQL_DEFAULT_BUNDLE_NAME}`;
+    }
+    let [repositoryOwner, repositoryName] = repository.split("/");
+    let client = apiURL === INSTANCE_API_URL ? api.getApiClient() : new Octokit();
```
robertbrignull, line 125: This check will never be false, because we return early if it's the github.com and the canonical action.

src/codeql.ts
```diff
@@ -74,16 +82,107 @@ let cachedCodeQL: CodeQL | undefined = undefined;
...
+    // GitHub.com, and the canonical Action.
+    [GITHUB_DOTCOM_API_URL, CODEQL_DEFAULT_ACTION_REPOSITORY],
+  ];
+  // We now filter out any duplicates.
+  // Duplicates will happen either because the GitHub instance is GitHub.com, or because the Action is not a fork.
+  const uniqueDownloadSources = potentialDownloadSources.filter((url, index, self) => index === self.indexOf(url));
+  for (let downloadSource of uniqueDownloadSources) {
+    let [apiURL, repository] = downloadSource;
+    // If we've reached the final case, short-circuit the API check since we know the bundle exists and is public.
+    if (apiURL === GITHUB_DOTCOM_API_URL && repository === CODEQL_DEFAULT_ACTION_REPOSITORY) {
```
robertbrignull, line 121: I'd argue you could simplify the logic by removing this default case from the list of potential download sources, and moving this return statement to after the loop.
This would have a number of knock-on advantages:
- We get rid of the worrying "could not find an accessible CodeQL bundle" error case too which is actually dead code at the moment.
- It would also let you simplify the list of potential sources to just the repository name, as the instance URL is no longer needed there as it's always the current instance.
- We can avoid adding `octokit` as a explicit dependency and thus keep the size of this PR down.

src/codeql.ts
```diff
@@ -74,16 +82,107 @@ let cachedCodeQL: CodeQL | undefined = undefined;
  */
 const CODEQL_ACTION_CMD = "CODEQL_ACTION_CMD";
 
+const CODEQL_DEFAULT_BUNDLE_VERSION = "codeql-bundle-20200630";
+const CODEQL_DEFAULT_BUNDLE_NAME = "codeql-bundle.tar.gz";
+const GITHUB_DOTCOM_API_URL = "https://api.github.com";
+const INSTANCE_API_URL = process.env["GITHUB_API_URL"] || GITHUB_DOTCOM_API_URL;
```
robertbrignull, line 88: I'd recommend avoiding constants definitions that do any computation at all even if it's just looking up something in `process.env`. We may want to change that in tests and we can't if it gets evaluated right at the start.

src/codeql.ts
```diff
@@ -74,16 +82,107 @@ let cachedCodeQL: CodeQL | undefined = undefined;
...
+const CODEQL_DEFAULT_ACTION_REPOSITORY = "github/codeql-action";
+
+function getCodeQLActionRepository(): string {
+  // Actions do not know their own repository name,
+  // so we currently use this hack to find the name based on where our files are.
+  // This can be removed once the change to the runner in https://github.com/actions/runner/pull/585 is deployed.
+  const runnerTemp = util.getRequiredEnvParam("RUNNER_TEMP");
+  const actionsDirectory = path.join(path.dirname(runnerTemp), "_actions");
+  const relativeScriptPath = path.relative(actionsDirectory, __filename);
+  if (relativeScriptPath.startsWith("..") || path.isAbsolute(relativeScriptPath)) {
```
robertbrignull, line 98: Can you add a comment that this is accounting for the case when we run CI on the codeql-action repository itself. Assuming I've understood that's what this case is.

src/codeql.ts
```diff
@@ -74,16 +82,107 @@ let cachedCodeQL: CodeQL | undefined = undefined;
...
     if (codeqlFolder) {
       core.debug(`CodeQL found in cache ${codeqlFolder}`);
     } else {
-      const codeqlPath = await toolcache.downloadTool(codeqlURL);
+      if (!codeqlURL) {
+        codeqlURL = await getCodeQLBundleDownloadURL();
+      }
+
+      // We have to download CodeQL manually because the toolcache doesn't support Accept headers.
+      // This can be changed to a one-liner once https://github.com/actions/toolkit/pull/530 is merged and released.
```
robertbrignull, line 159: Could you extract all the bits that can be converted to a one-liner to a function on their own, so that when we do convert them it's easy to tell which code this comment refers to.

src/codeql.ts
```diff
@@ -74,16 +82,107 @@ let cachedCodeQL: CodeQL | undefined = undefined;
  */
 const CODEQL_ACTION_CMD = "CODEQL_ACTION_CMD";
 
+const CODEQL_DEFAULT_BUNDLE_VERSION = "codeql-bundle-20200630";
```
robertbrignull, line 85: I don't think the word "default" is quite accurate in `CODEQL_DEFAULT_BUNDLE_VERSION` and `CODEQL_DEFAULT_BUNDLE_NAME`. It seems these are the only acceptable values and the only values it will try searching for.

robertbrignull:
> For symmetry I initially considered also checking GitHub.com at the repository name for the current Action, however I decided against that due to the accidental security risk someone might expose themselves to if they incorrectly mirror the Action without copying the bundle.

I agree, this seem does seem like a security risk if they pick a non-default repository name. Also I think it's unlikely it would ever usefully trigger.

> Sorry for the excessive description. This change ended up being a bit more complicated than I expected and I wanted to make it clear why I made the design decision I have.

Thank you for the description. Also for raising two improvements to the actions core code.

src/codeql.ts
chrisgavin: Good point. There are a few bits of dead code left after I decided to not use the API in the final case.

src/codeql.ts
chrisgavin: I used "DEFAULT" because you can technically still specify a different version with the `tools` input if you want. I agree it's probably more confusing than descriptive though.

src/codeql.ts
chrisgavin: The code can definitely be simplified, but I don't think we can easily remove the default case from the list, as we rely on filtering out the non-unique sources to handle the case where `INSTANCE_API_URL` is the same as `GITHUB_DOTCOM_API_URL` so we can break out early.

I've restructured the code slightly to achieve some of the advantages you've mentioned while preserving the deduplication , but let me know if I've missed a neater way of doing it.

src/codeql.ts
```diff
@@ -139,13 +142,32 @@ async function getCodeQLBundleDownloadURL(): Promise<string> {
       core.info(`Looked for CodeQL bundle in ${downloadSource[1]} on ${downloadSource[0]} but got error ${e}.`);
     }
   }
-  throw new Error("Could not find an accessible CodeQL bundle.");
+  return `https://github.com/${CODEQL_DEFAULT_ACTION_REPOSITORY}/releases/download/${CODEQL_BUNDLE_VERSION}/${CODEQL_BUNDLE_NAME}`;
```
aeisenberg, line 145: It looks like it's possible to run into an error on line [142 above](https://github.com/github/codeql-action/pull/123/files/813cb0479fbbedd3ededa41c1c28209aa22353d8..5587e128ff75c5ab3df15d9f04817e4f84228f68#diff-fabdc04b7538bbb9cb582f726acf408cR142) and then still return the default URL for the bundle. 

This seems dangerous to me. If a user specified a bundle that can't be found, should we end the action with an error so this can be fixed?

chrisgavin: Hmm, to me the alternative seems more dangerous. People expect to be able to fork the CodeQL Action and have it still work, but if we require that CodeQL exists on the forked repository that won't necessarily be the case.

src/codeql.ts
```diff
@@ -74,16 +81,116 @@ let cachedCodeQL: CodeQL | undefined = undefined;
  */
 const CODEQL_ACTION_CMD = "CODEQL_ACTION_CMD";
 
+const CODEQL_BUNDLE_VERSION = "codeql-bundle-20200630";
+const CODEQL_BUNDLE_NAME = "codeql-bundle.tar.gz";
+const GITHUB_DOTCOM_API_URL = "https://api.github.com";
+const CODEQL_DEFAULT_ACTION_REPOSITORY = "github/codeql-action";
+
+function getInstanceAPIURL(): string {
+  return process.env["GITHUB_API_URL"] || GITHUB_DOTCOM_API_URL;
```
robertbrignull, line 90: My understanding was that this environment variable should always be populated now, so the default is unnecessary. We could use `util.getRequiredEnvParam` to ensure it exists too. Do you agree?

src/codeql.ts
robertbrignull: I'll leave it up to you to decide if it's simpler but I think this function is equivalent to the following which doesn't use a look at all. You'd better check my working though as I found it hard getting my head around the logic to work out if the two implementations are indeed equivalent or not.
```typescript
async function checkRepoForCodeQLBundle(instanceApiUrl: string, repository: string): Promise<string | undefined> {
  let [repositoryOwner, repositoryName] = repository.split("/");
  try {
    const release = await api.getApiClient().repos.getReleaseByTag({
      owner: repositoryOwner,
      repo: repositoryName,
      tag: CODEQL_BUNDLE_VERSION
    });
    for (let asset of release.data.assets) {
      if (asset.name === CODEQL_BUNDLE_NAME) {
        core.info(`Found CodeQL bundle in ${repository} on ${instanceApiUrl} with URL ${asset.url}.`);
        return asset.url;
      }
    }
  } catch (e) {
    core.info(`Looked for CodeQL bundle in ${repository} on ${instanceApiUrl} but got error ${e}.`);
  }
  return undefined;
}

async function getCodeQLBundleDownloadURL(): Promise<string> {
  const instanceApiUrl = getInstanceAPIURL();
  const codeQLActionRepository = getCodeQLActionRepository();
  if (codeQLActionRepository !== CODEQL_DEFAULT_ACTION_REPOSITORY) {
    const assetUrl = await checkRepoForCodeQLBundle(instanceApiUrl, codeQLActionRepository);
    if (assetUrl !== undefined) {
      return assetUrl;
    }
  }
  if (instanceApiUrl !== GITHUB_DOTCOM_API_URL) {
    const assetUrl = await checkRepoForCodeQLBundle(instanceApiUrl, CODEQL_DEFAULT_ACTION_REPOSITORY);
    if (assetUrl !== undefined) {
      return assetUrl;
    }
  }
  return `https://github.com/${CODEQL_DEFAULT_ACTION_REPOSITORY}/releases/download/${CODEQL_BUNDLE_VERSION}/${CODEQL_BUNDLE_NAME}`;
}
```

src/codeql.ts
robertbrignull: I think the implementation that's there now is backwards compatible with how it would work now. Currently the download URL is hardcoded to point to the default action, unless a fork explicitly changes that behaviour. So that makes me lean towards keeping it as it is and always checking the default action.

Note that if a user specifies to use a specific bundle by setting the `tools` input then all of this implementation is skipped and we just use the value they provided.

PR #127, opened 2020-07-27
Allow queries to be specified in the workflow configuration, overriding those set in the config file

sampart:
@robertbrignull do you think it's worthwhile adding an integration test for this feature?

In time, this will be extended to allow additive behaviour as well as overriding, but I'm currently planning on doing that in a separate PR.

Furthermore, re overriding - the way that this PR is implemented means that queries in the workflow file (the newly-added feature) take precedence over the configuration file. I _think_ that's what's wanted from the discussion, but do say if I've got this wrong (CC @jhutchings1).

The queries specified in the workflow configuration are comma separated. I'm deliberately not including the ability to escape commas here, as that adds a lot of complexity to support a definite edge case (commas in file names/paths).  This is different than what @robertbrignull and @rneatherway proposed (they wanted escaping), so feel free to push back on this one, Robert and Robin.


robertbrignull:
> Furthermore, re overriding - the way that this PR is implemented means that queries in the workflow file (the newly-added feature) take precedence over the configuration file. I think that's what's wanted

Yes, I agree.

We want that by default if you see an option in the workflow file then you know exactly what it's doing. Once there's syntax for additive behaviour, then when you see that syntax you know you need to also look in the config file too to fully understand it.

Making the default behaviour to override also means there isn't a difference in behaviour between list-based properties like the queries, and single-valued properties like booleans or strings.

robertbrignull:
> The queries specified in the workflow configuration are comma separated. I'm deliberately not including the ability to escape commas here, as that adds a lot of complexity to support a definite edge case (commas in file names/paths)

I'm mostly ambivalent about escaping. I agree that having commas in one of these options is unlikely so it's definitely an edge case. I also think we should ideally still do it if we can, but we could certainly do it later, or wait until we have a better idea of how much of an edge case it is. So I'd prefer to work with that in mind and try to plan to be compatible for the future, and I think we'll be doing that with this plan so I'm happy.

init/action.yml
```diff
@@ -16,6 +16,9 @@ inputs:
   config-file:
     description: Path of the config file to use
     required: false
+  queries:
+    description: comma-separated list of queries to run; overrides default queries and any set in a configuration file
```
robertbrignull, line 20: "overrides default queries" is different behaviour from queries set in the configuration file.

If you want to not run the default queries you set the `disable-default-queries` property, but otherwise anything specified here is run in addition to the default queries.

src/config-utils.ts
```diff
@@ -613,6 +620,23 @@ function isLocal(configPath: string): boolean {
...
+  if (isLocal(configPath)) {
+    // Treat the config file as relative to the workspace
+    const workspacePath = util.getRequiredEnvParam('GITHUB_WORKSPACE');
+    configPath = path.resolve(workspacePath, configPath);
+  }
+
+  const languages = await getLanguages();
+  const queries = {};
+  queryUses.split(',').forEach(async query => {
+    await parseQueryUses(configPath, languages, queries, query);
```
robertbrignull, line 633: This config filepath parameter is only used when constructing error messages. It's also not correct here because the error isn't in the config file, but is in a workflow input. We'll need to instead change how the errors are generated so it can accomodate that.

src/config-utils.ts
```diff
@@ -613,6 +620,23 @@ function isLocal(configPath: string): boolean {
...
 }
 
+async function updateConfigWithQueries(config: Config, queryUses: string, configPath: string): Promise<Config> {
+  if (isLocal(configPath)) {
+    // Treat the config file as relative to the workspace
+    const workspacePath = util.getRequiredEnvParam('GITHUB_WORKSPACE');
+    configPath = path.resolve(workspacePath, configPath);
+  }
+
+  const languages = await getLanguages();
```
robertbrignull, line 630: Avoid caling `getLanguuages` here because this will try to compute them again, making API requests and in the worst case getting a different answer than we got before. Instead you can get the languages from the config as that part has been populated.

src/config-utils.ts
```diff
@@ -599,6 +599,13 @@ export async function initConfig(): Promise<Config> {
     config = await loadConfig(configFile);
   }
 
+  // If queries were provided using `with` in the action configuration,
```
robertbrignull, line 602: I'm not sure this approach of adding this on to the end of the process will work, because a lot of the data has been munged together and lost. I think instead you'd need to check this input at the place where we check the queries value from the config file, and override it there but otherwise go through the same processing. Does that make sense?

So in very rough pseudocode, instead of
```
loadFromConfigFile();
loadFromWorkflowInputs();
```
as two sequential steps, we do something like
```
queries = workflowInputDefined() ? getWorkflowInputValue() : getConfigFileValue();
loadQueries(queries);
```

rneatherway:
> I'm mostly ambivalent about escaping. I agree that having commas in one of these options is unlikely so it's definitely an edge case. I also think we should ideally still do it if we can, but we could certainly do it later, or wait until we have a better idea of how much of an edge case it is. So I'd prefer to work with that in mind and try to plan to be compatible for the future, and I think we'll be doing that with this plan so I'm happy.

I _think_ I agree with this. However, let's work it through to be sure. Let's imagine that the escaping would work something like:

| Input text | Interpreted result |
| --- | --- |
| `,` | `<list separator>` |
| `\,` | `,` |
| `\\` | `\` |
| `\(?!,)` | `\` |

This means that we are at risk if people use backslashes in their configurations, which is presumably quite possible for Windows users. Will they work in paths at the moment? However, the final rule, which is somewhat debatable, basically allows an invalid escape sequence to be treated just as the literal text and means that the only currently "valid" configuration that we would change the behaviour of is a Windows-style paths where one element of the "list" has a trailing backslash. I don't believe that a directory is a valid input as a query specification is it? If not then we're in the clear.

As an example, I think the current implementation would do:

`"hello.ql,path\to\,query.ql" => ["hello.ql", "path\to\", "query.ql" ]`

and with the above escaping we would instead get:

`"hello.ql,path\to\,query.ql" => ["hello.ql", "path\to,query.ql" ]`

sampart:
> I don't believe that a directory is a valid input as a query specification is it? If not then we're in the clear.

It sounds like you're saying we would be in the clear if a directory wasn't valid input and we also adopted the escaping rules you specify, rather than saying we're in the clear now.  But I think we are in the clear now (at least in terms of your example).  I'll explain why in a mo, but first to your point about whether directories are allowed:

https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#running-additional-queries says:

> You can specify a single .ql file, a directory containing multiple .ql files, a .qls query suite definition file, or any combination.

And here's why I think we're doing the right thing.  Taking your example of current behaviour:

```
"hello.ql,path\to\,query.ql" => ["hello.ql", "path\to\", "query.ql" ]
```

I think that this is the "right" behaviour.  This output assumes that the input was `a file, a directory with a trailing slash, and a file`.  I believe that this is a far more likely scenario than `"hello.ql,path\to\,query.ql"` being intended to mean `a file, and a filepath which contains a comma`.

sampart:
> Code looks mostly fine but I think there's been a misunderstanding as to what the intended semantics are meant to be. Feel free to disagree if I've got the wrong end of things, but I'll explain how I think it should work.

Your points make sense, @robertbrignull; I think this does need changing.  I'll do that.  Thanks for the detail about a better approach; much appreciated.

rneatherway:
> It sounds like you're saying we would be in the clear if a directory wasn't valid input and we also adopted the escaping rules you specify, rather than saying we're in the clear now. But I think we are in the clear now (at least in terms of your example). I'll explain why in a mo, but first to your point about whether directories are allowed:

My comment was actually about whether it would be safe to ship this PR as it is and then _later_ add the escaping rules _if_ it turns out that people need them. It was a thought experiment about whether any configurations would have their meaning changed. Thanks for the quote from the docs:

> You can specify a single .ql file, a directory containing multiple .ql files, a .qls query suite definition file, or any combination.

This means that as it stands `"hello.ql,path\to\,query.ql" => ["hello.ql", "path\to\", "query.ql" ]` would be a valid configuration that we would break if we later added escaping, so this PR is not entirely safe in that respect.

To be clear I think the chance of "specifying a directory using windows path separators and a trailing separator" is pretty remote. I was just being thorough.

> I believe that this is a far more likely scenario than `"hello.ql,path\to\,query.ql"` being intended to mean `a file, and a filepath which contains a comma`.

I agree with this point, however we have to be a bit careful when we make things impossible to express.

init/action.yml
hubwriter, line 20:
```suggestion
    description: Comma-separated list of additional queries to run. By default, this overrides the same setting in a configuration file
```

sampart: Thanks Alistair!

sampart:
Thanks @rneatherway.  I understand what you're getting at now.  I like your checking-for-future-incompatibilities approach 🚀 

I think we should solve this by bumping the major version number if we introduce escaping - it's effectively a backwards-incompatible change, and that's what major version numbers are for.  That way, we get to keep the code simple, avoid doing too much in one PR, but still leave the door open for the future.

init/action.yml
sampart: Alistair's change in 517d9fa fixes the wording here; I'll change the logic accordingly, considering your points in the review comment and elsewhere.  Thank you for explaining it so clearly.

src/config-utils.ts
sampart: Good point.  I've addressed this in https://github.com/github/codeql-action/pull/127/commits/129713f1a0030aba663b339ae4e8cdfb4ffa79cc

src/config-utils.ts
sampart: This is no longer an issue now I've overhauled where this logic happens.

src/config-utils.ts
```diff
@@ -518,13 +525,37 @@ async function getLanguages(): Promise<Language[]> {
...
+/**
+ * Returns true if queries were provided in the workflow file
+ * (and thus added), otherwise false
+ */
+async function addQueriesFromWorkflowIfRequired(
+  codeQL: CodeQL,
+  languages: string[],
+  resultMap: { [language: string]: string[] },
+  tempDir: string,
+  configFile?: string
```
robertbrignull, line 537: What does the `configFile` parameter mean here and when would it be non-null? Aren't we always considering input from the workflow here?

src/config-utils.test.ts
```diff
@@ -273,6 +273,195 @@ test("default queries are used", async t => {
...
+    // Check resolveQueries was called correctly
+    // It'll be called once for the default queries and once for `./override`,
+    // but won't be called for './foo' from the config file.
+    t.deepEqual(resolveQueriesArgs.length, 2);
+    t.deepEqual(resolveQueriesArgs[1].queries.length, 1);
+    t.regex(resolveQueriesArgs[1].queries[0], /.*\/override$/);
+
+    // Now check that the end result contains only the default queries and the override query
+    t.deepEqual(config.queries['javascript'].length, 2);
+    t.regex(config.queries['javascript'][0], /javascript-code-scanning.qls$/);
```
robertbrignull, line 379: Probably worth noting that this won't be the correct contents when running properly as it would resolve the query suite and directory to a set of ql files. This data is here because it's what was returned from `resolveQueries`. That's not a huge isue for this test but probably good to be aware of.

src/config-utils.ts
sampart: Oh yes, good point.  I made the mistake of thinking that the value being available meant that we wanted to use it.  I've fixed this in ab4e7216d3e94457716c4ac9a99c3d8449952426.

src/config-utils.test.ts
sampart: That's helpful, thanks.

robertbrignull (approved):

PR #128, opened 2020-07-30
Avoid sending status reports on enterprise

robertbrignull:
This avoids us sending status reports when running on enterprise. The endpoint currently doesn't do anything that would be useful on enterprise.

This will probably conflict a little bit with https://github.com/github/codeql-action/pull/117 but I'll fix that up when it happens.

The check and the point where we revert is currently just before we actually send the status report, but notably this is after we've constructed the report, so there is a little bit of wasted effort. I think this is acceptable for the first attempt, though there's clearly room for improvement to avoid constructing the report while also keeping the centralised check and avoiding having to add checks to every call.

Things that need to be checked and tested before merging:
- [x] Status reports are still sent as normal on github.com
- [x] Status reports are not sent on enterprise


robertbrignull:
I've tested this now and confirmed the reports are not sent and it doesn't fail.

PR #131, opened 2020-08-04
Improve error messages when no languages or no queries are specified

robertbrignull:
This should help address some unhelpful error messages being seen, for example at https://github.com/octokit/auth-unauthenticated.js/runs/943539485?check_suite_focus=true

The error there is that we didn't detect any languages to analyze, because ther weren't any specified in the workflow file, and linguist didn't detect any because I believe there wasn't any typescript code on the default branch at that time. The branch where the analysis ran was introducing the first piece of analyzable code.

So although the above is perhaps the true cause of the failure, the error message made this very unclear. In the case where there isn't a config file we weren't checking that the set of languages was non-empty. This is a bug and so I've copied the check from the config file case over into a shared method so it also runs when there's no config file.

Another potential configuration setup that would cause the same bad error message is if no queries are specified for a particular language. That didn't happen here but I think it's likely it'll happen at some point. I also add a check for this case so we'll report a better error message and report during config parsing instead of after extraction.


src/config-utils.ts
```diff
@@ -468,6 +471,13 @@ async function getLanguages(): Promise<string[]> {
     core.info("Automatically detected languages: " + JSON.stringify(languages));
   }
 
+  // If the languages parameter was not given and no languages were
+  // detected then fail here as this is a workflow configuration error.
+  if (languages.length === 0) {
+    throw new Error("Did not detect any languages to analyze. " +
+        "Please update input in workflow or check that GitHub detects the correct languages in your repository.");
+  }
+
```
anaarmas, line 480: Sorry if this is a silly question @robertbrignull but will the error appear more prominently (in the Check failure summary, rather than only in the detailed logs) if thrown from here rather than in the calling method as it was before these changes? Or do we also need to change _how_ we throw the error?

robertbrignull: Where the error is thrown from doesn't make a difference to how it is presented.

I'm going to do a test of this to make sure it works and looks presentable before this is merged.

anaarmas: :+1: I'd love a link to see what it looks like too!

src/config-utils.ts
```diff
@@ -572,6 +577,15 @@ async function loadConfig(configFile: string): Promise<Config> {
     });
   }
 
+  // The list of queries should not be empty for any language. If it is then
+  // it is a user configuration error.
+  for (const language of languages) {
+    if (queries[language].length === 0) {
+      throw new Error(`Did not detect any queries to analyze for ${language}. ` +
```
chrisgavin, line 584:
```suggestion
      throw new Error(`Did not detect any queries to run for ${language}. ` +
```

robertbrignull:
I've done some test runs:
- Here's what it looks like when no languages are detected: https://github.com/dsp-testing/robertbrignull_no_language/actions/runs/195948939
- Here's what it looks like when there are no queries to run: https://github.com/dsp-testing/robertbrignull_no_language/actions/runs/195952794

During testing I discovered a bug with checking the set of queries so I've fixed that.

I also ran into a problem of not being able to see the error stacktrace. I've opened https://github.com/github/codeql-action/pull/136 to hopefully address this.

PR #137, opened 2020-08-06
Provide a better error message when languages are not recognised

robertbrignull:
Improves the error message when the user specifies languages that doesn't exist. Currently it does something ugly like saying `A fatal error occurred: c-code-scanning.qls is not a well-known query suite.` but this will make it much nicer. Also adds support for some well known aliases for languages that people might try. This feature is perhaps more debatable, but it might be nice to have since people are more likely to type "c" than to know that CodeQL uses "cpp" to refer to both.

Also includes unit tests of the check for no languages, because I forgot to add that when doing that.

Here are some tests / demos:
- An invalid language: https://github.com/github/codeql-action/actions/runs/197994941
- Using an alias: https://github.com/github/codeql-action/actions/runs/197995109


robertbrignull:
Ah good, that fixes the tests. They passed for me locally the whole time which made it hard to debug.

PR #139, opened 2020-08-07
Support checkout of multiple refs for a single repository

esbena:
This change ensures that two different checkouts are made of `foo/bar` for the config below. The old implementation would only checkout the `@dev` version, and reuse that for the `@legacy` version.

```yaml
      name: my config
      queries:
        - uses: foo/bar@dev
        - uses: foo/bar@legacy
```

  - this is technically a breaking change, but I think the breakage is desirable


src/external-queries.ts
```diff
@@ -13,7 +13,8 @@ export async function checkoutExternalRepository(repository: string, ref: string
 
   core.info('Checking out ' + repository);
 
-  const checkoutLocation = path.join(folder, repository);
+  const suffix = ref ? "@" + ref.replace(/\//g, '__') : "";
```
robertbrignull, line 16: Why have the ref as a suffix instead of having another path segment?

I had to look up to make sure `@` wasn't forbidden on windows or something like that. It appears to be fine though.

Do we need to sanitise against other characters too that might be illegal in a path? I'm unsure. On the one hand it might be nice, but then the git checkout would likely fail anyway, so I don't think we actually gain much.

src/external-queries.test.ts
```diff
@@ -11,9 +11,10 @@ setupTests(test);
 test("checkoutExternalQueries", async t => {
   await util.withTmpDir(async tmpDir => {
     process.env["RUNNER_TEMP"] = tmpDir;
-    await externalQueries.checkoutExternalRepository("github/codeql-go", "df4c6869212341b601005567381944ed90906b6b");
+    const ref = "df4c6869212341b601005567381944ed90906b6b";
+    await externalQueries.checkoutExternalRepository("github/codeql-go", ref);
```
robertbrignull, line 15: It would be nice to have this test exercise the new feature by checking out this repo at another commit as well. However I'm worried as I've seen this test sometimes time out already, so doing another checkout may be too much. I'd appreciate if you could try it out, and we'll see if it's worth it.

Ultimately it would be better if we changed this to checkout a local tiny repository instead of a medium sized repo from github.com and that would make it much faster, but that's a bigger change that I don't expect you to make here in this PR.

esbena: The time cost is twice as high. 

Besides, we now assert that a directory named `df4c6869212341b601005567381944ed90906b6b` exists. It would be an extraordinarily weird behaviour if we reused that for another checkout.

esbena:
Drive-by thought: The following forbids valid paths like `org/repo/path1@path2@ref`.

```js
  let tok = queryUses.split('@');
  if (tok.length !== 2) {
    throw new Error(getQueryUsesInvalid(configFile, queryUses));
  }
```

src/external-queries.ts
esbena: > Why have the ref as a suffix instead of having another path segment?

I thought we allowed the empty ref (but we don't), which would then introduce an actual problem if you mixed `foo/bar` and `foo/bar@dev`. I'll make an additional path segment instead.

> Do we need to sanitise against other characters too that might be illegal in a path? 

git refs are surprisingly permissive, so I think that is an uphill battle: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html. I have changed the check to avoid the classic path traversal outside of `folder`.

robertbrignull:
> Drive-by thought: The following forbids valid paths like `org/repo/path1@path2@ref`.

Good point, that should probably change to only consider the first occurrence of `@` and then anything that comes after it.

esbena:
Merging on green.

PR #140, opened 2020-08-07
Remove some shared env

robertbrignull:
This PR aims to remove a few elements from `shared-environment.ts`. Most of this is pretty easy because the variable is only used from one place, or the data is readily available in the config already.

This doesn't remove `ODASA_TRACER_CONFIGURATION` or `CODEQL_WORKFLOW_STARTED_AT` because those are unfortunately going to be slightly more tricky.

I think the only bit that needs explaining is the removal of `should_abort`. I claim this is safe to do because it's functionality of checking that the `init` action has already run is covered by the call to `getConfig` in the `autobuild` and `analyze` actions. In the `init` and `upload-sarif` actions, all it was doing was validating that the github ref env var was populated, which doesn't seem worth it to keep the function around.


robertbrignull:
https://github.com/github/codeql-action/pull/137 also introduced a list of the known languages to the `config-utils` file, and this PR introduces that list to the `codeql` file. I'd like to keep the data in one place, and based on the dependency direction I think that place has to be the `codeql` file, so I'll move the existing code to there.

robertbrignull:
> #137 also introduced a list of the known languages to the config-utils file, and this PR introduces that list to the codeql file. I'd like to keep the data in one place, and based on the dependency direction I think that place has to be the codeql file, so I'll move the existing code to there.

After looking at this, it's going to involve a fair bit or refactoring and ideally I'd like to even introduce a new `languages.ts` file. So I propose leaving this PR as it is and doing that change separately.

PR #147, opened 2020-08-10
Add a CLI interface to the upload-sarif action

robertbrignull:
This PR is a first attempt at adding a command lint interface to the codeql-action. This PR only deals with the `upload-sarif` action, and largely ignores the others except the minimum changes required to keep them working.

The main things done here are:
- Introduce `src/cli.ts` which is the entrypoint to the CLI. I'm using the [commander](https://www.npmjs.com/package/commander) library here to do the CLI part, and it generates `-h` options and stuff like that. I'm open to other libraries instead if someone has a favourite that is better.
- Introduce a new `build-cli` npm script, that uses webpack to bundle the `cli.ts` file and all its dependencies into a single file, that is output to `cli/cli.js`. This file is not checked into the repository as it isn't needed to run the action normally. I felt it therefore didn't belong in the `lib` directory, but I don't think the `cli` directory name is great either. I'd be very open to suggestions here.
- Introduce `logging.ts` which mostly just wraps the `@actions/core` package, and replaces it with stuff like `console.log` when not on actions.
- Adds lots of extra function parameters. I think the way we're going to have to do things, is pass parameters from the entrypoint all the way into the functions that need them. That way the shared code doesn't have to have different cases for actions vs the CLI. At the entrypoints we pass the inputs, either from command line args, or `core.getInput`, or the environment, and pass it onwards to where it is needed. Hopefully this won't get too ugly.

The diff looks huge because of `node_modules` changes, but if you ignore those and other generated files, the actual churn is only about 300 lines.

This code is largely untested in practice. I wanted to get the PR up soon so people could see what it looks like.


joshhale:
I can't jump to file to see the significant files because there are too many in the PR. Would it help to open a dummy PR with only the significant files?

robertbrignull:
> I can't jump to file to see the significant files because there are too many in the PR, and I can't see, e.g. src/cli.ts by searching in the files page. Would it help to open a dummy PR with only the significant files?

I'll reorganise so the node_modules changes are in a separate commit. That way you should be able to view it in the UI.

robertbrignull:
I'm doing some testing and have found some bits I've missed, so I'm fixing those up and will push a new version shortly.

robertbrignull:
I've now tested this manually, which showed up a few bugs and odditites. I managed to upload to one of my testing repos with
```
node cli/cli.js upload --sarif-file path/to/example.sarif --repository dsp-testing/robertbrignull_test --commit 6548b467960719daec4cf3cca3e59b85f69fc0e1 --ref refs/heads/master --github-url https://api.github.com --github-auth robertbrignull:<PAT> --analysis-key foo
```

Unfortunately everything has become a bit more annoying since there are different upload endpoints for actions vs non-actions, so there's a bit of ugly code there to generate the right payload and upload it to the right place. I'm not sure what the best approach is to avoid duplicating too much code or keep it safe so we won't accidentally do the wrong thing in the wrong environment.

robertbrignull:
@sampart would you be able to review this? You'll have to look at individual commits and ignore the one that's got all the node_modules and generated file changes in it.

src/cli.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+program
+  .command('upload')
+  .description('Uploads a SARIF file, or all SARIF files from a directory, to code scanning')
+  .requiredOption('--sarif-file <file>', 'SARIF file to upload')
+  .requiredOption('--repository <repository>', 'Repository name')
+  .requiredOption('--commit <commit>', 'SHA of commit that was analyzed')
+  .requiredOption('--ref <ref>', 'Name of ref that was analyzed')
+  .requiredOption('--analysis-key <key>', 'Identifies the analysis, for use matching up equivalent analyses on different commits')
```
robertbrignull: A bunch of these options are not used. That's a bug from changing definitions over time. I'll trim these down.

robertbrignull: I've pushed a commit to remove all of these that aren't used in CLI mode. It makes the `upload_lib.upload` interface a bit silly though as lots of the parameters are nullable so the type system isn't giving us much.

Maybe we could do something better by passing all the options as a blob along the with mod actions/cli mode variable.

sampart: I like the idea of passing as an object, so you effectively get named parameters 👍

src/logging.ts
```diff
@@ -0,0 +1,26 @@
+import * as core from '@actions/core';
+
+export interface Logger {
```
sampart: I really like your approach here 🏅

src/upload-lib.ts
```diff
@@ -109,18 +121,48 @@ export interface UploadStatusReport {
 // Uploads a single sarif file or a directory of sarif files
 // depending on what the path happens to refer to.
 // Returns true iff the upload occurred and succeeded
-export async function upload(input: string): Promise<UploadStatusReport> {
-  if (fs.lstatSync(input).isDirectory()) {
-    const sarifFiles = fs.readdirSync(input)
+export async function upload(
+  sarifFile: string,
```
sampart: How about calling this `sarifPath`, since it could be a file or a directory?  I'm not particularly fussed about this, though, so happy to leave if you'd rather.

webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+    extensions: [ '.ts', '.js' ],
+  },
+  output: {
+    filename: 'cli.js',
+    path: path.resolve(__dirname, 'cli'),
+  },
+  optimization: {
+		// We no not want to minimize our code.
+		minimize: false
+	},
```
sampart:
```suggestion
    // We do not want to minimize our code.
    minimize: false
  },
```

sampart: You probably don't even need the comment:

```suggestion
    minimize: false
  },
```

src/upload-lib.ts
```diff
@@ -156,75 +198,69 @@ export function validateSarifFileSchema(sarifFilePath: string) {
 
 // Uploads the given set of sarif files.
 // Returns true iff the upload occurred and succeeded
-async function uploadFiles(sarifFiles: string[]): Promise<UploadStatusReport> {
-  core.startGroup("Uploading results");
-  core.info("Uploading sarif files: " + JSON.stringify(sarifFiles));
-
-  const sentinelEnvVar = "CODEQL_UPLOAD_SARIF";
-  if (process.env[sentinelEnvVar]) {
-    throw new Error("Aborting upload: only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job");
```
sampart: You're now allowing multiple uploads in a job. Is that okay?  Another option might be to clear this env var every time the CLI launches?

src/upload-lib.ts
```diff
@@ -156,75 +198,69 @@ export function validateSarifFileSchema(sarifFilePath: string) {
...
+  if (mode === 'actions') {
+    payload = JSON.stringify({
+      "commit_oid": commitOid,
+      "ref": ref,
+      "analysis_key": analysisKey,
+      "analysis_name": analysisName,
+      "sarif": zipped_sarif,
+      "workflow_run_id": workflowRunID,
+      "checkout_uri": checkoutURI,
+      "environment": environment,
```
sampart: You don't have `started_at` anymore.  Is that deliberate?

src/upload-lib.ts
```diff
@@ -156,75 +198,69 @@ export function validateSarifFileSchema(sarifFilePath: string) {
...
+      "sarif": zipped_sarif,
+      "checkout_uri": checkoutURI,
+      "tool_name": toolNames[0],
+    });
+  }
 
   // Log some useful debug info about the info
   const rawUploadSizeBytes = sarifPayload.length;
-  core.debug("Raw upload size: " + rawUploadSizeBytes + " bytes");
+  console.debug("Raw upload size: " + rawUploadSizeBytes + " bytes");
```
sampart: I think you want `logger.debug` here and below.

sampart:
Because I added my comments to the commit rather than the PR-level diff, some comments are showing as "outdated".  I don't think they are all genuinely redundant, though (although some are), so mind you don't miss them 😉

package.json
```diff
@@ -23,6 +24,7 @@
     "@actions/github": "^2.2.0",
     "@actions/http-client": "^1.0.8",
     "@actions/tool-cache": "^1.5.5",
+    "commander": "6.0.0",
```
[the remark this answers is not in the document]
robertbrignull: As you noted this doesn't make too much difference to use day-to-day as we're using `package-lock.json` to pin to specific versions anyway and the `node_modules` folder is checked into the repo. So yes I suppose it makes the intention slightly more clear when it comes to upgrading.

src/upload-lib.ts
robertbrignull: Thanks for spotting this. Removing this wasn't intentional.

It is worth pointing out the check won't work when running in CLI mode because we can't set env vars in the same way such that they are seen by subsequent processes. But it should be safe to leave in and it'll just only do something on actions.

src/upload-lib.ts
robertbrignull: Thanks. That also was not intentional.

robertbrignull:
I've added a simple workflow that builds and runs the CLI to upload an empty SARIF file. You can still run the CLI on actions, so I think this should be ok for making sure we don't accidentally break it too much.

robertbrignull:
Ok, fixed the error handling so it actually now sets a non-zero exit code when there's an error, and fixed a typo in a filename. Good thing I made that typo otherwise might not have noticed the broken error handling.

With the exit code I tried a few variations of re-throwing the error, or putting the try-catch around the call to `program.parse` but they didn't work. So far putting the catch in the sub-action and setting `process.exitCode` was the only thing that worked.

src/upload-lib.ts
sampart: That's what I thought 👍

package.json
```diff
@@ -5,6 +5,7 @@
   "description": "CodeQL action",
   "scripts": {
     "build": "tsc",
+    "build-cli": "webpack --mode production",
     "test": "ava src/** --serial --verbose",
     "lint": "tslint -p . -c tslint.json 'src/**/*.ts'",
     "removeNPMAbsolutePaths": "removeNPMAbsolutePaths . --force"
```
swinton: Suggestion: add a `postinstall` step that builds the CLI automatically after a successful `npm install`. This will make it easier for new users to acquire the CLI by reducing the number of installation steps.

E.g.

```suggestion
  "scripts": {
    "build": "tsc",
    "build-cli": "webpack --mode production",
    "test": "ava src/** --serial --verbose",
    "lint": "tslint -p . -c tslint.json 'src/**/*.ts'",
    "removeNPMAbsolutePaths": "removeNPMAbsolutePaths . --force",
    "postinstall": "npm run-script build-cli"
```

package.json
```diff
@@ -5,6 +5,7 @@
   "description": "CodeQL action",
   "scripts": {
```
swinton: Suggestion: add a `bin` property to `package.json`, which resolves to the new CLI being introduced. This will allow new users to run the CLI directly, e.g. via `npx ...`.

```suggestion
  "bin": {
    "codeql-cli": "cli/code-scanning-cli.js"
  },
  "scripts": {
```

webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+    ],
+  },
+  target: 'node',
+  resolve: {
+    extensions: [ '.ts', '.js' ],
+  },
+  output: {
+    filename: 'cli.js',
+    path: path.resolve(__dirname, 'cli'),
+  },
```
swinton: Is is possible to make this output executable, so that it can be invoked directly from the command-line?

robertbrignull:
> I like the idea of passing as an object, so you effectively get named parameters 

Sorry I forgot to do this. I still think it would be good, but perhaps better to delay to a separate PR so we can discuss it properly. It could be a pretty big change in direction.

robertbrignull:
Thanks for the comments @swinton. How to distribute this is still largely an open question and not one we have to solve right now, so I think I'm going to merge this without any changes for now and we'll address it later. I've mentioned you on an issue for this so we can discuss there.

PR #149, opened 2020-08-14
Use all available threads for analysis

robertbrignull:
Changes `getThreadsFlag` to default to using all available cores instead of just 1. This makes the behaviour match `getMemoryFlag` which uses all available memory.

I've tried testing this by analyzing git, as an example of a medium-sized repository. These are the times for the "Perform CodeQL Analysis" step, so it excludes stuff like downloading and unpacking the action. I tried on a github.com cloud worker which have 2 cores and 7GB memory, and with a self-hosted runner with 8 cores and 64GB memory.

- `main` branch, cloud worker = 8m 16s
- `main` branch, self hosted worker = 6m 57s
- `max_threads` branch, cloud worker = 8m 45s (I hope this slight increase is just a fluke)
- `max_threads` branch, self hosted worker = 2m 24s

It's worth noting this PR is still only applying the number of threads to analysis. The other place we might get reasonable improvements is to extraction of scanned languages (i.e. javascript / python / go). I think we can do that separately though and get the various language teams involved to make sure we're doing the right thing.


src/util.ts
```diff
@@ -390,12 +392,17 @@ export function getThreadsFlag(): string {
     }
     const maxThreads = os.cpus().length;
```
chrisgavin, line 393: You could just move this outside the `if` statement since it's needed in both branches.

rneatherway:
LGTM

PR #151, opened 2020-08-17
Add query to detect use of actions libs

robertbrignull:
Since some of the code now needs to run flawlessly outside of actions, I think a CodeQL query will be really helpful for spotting accidental uses.

This borrows a lot of the framework from [undeclared-action-input.ql](https://github.com/github/codeql-action/blob/main/queries/undeclared-action-input.ql) but adds in checks for guards that we are running in actions/cli mode.

This finds a few cases, which I've fixed in this PR. I've pushed a branch without the fixes too so you can see what the alerts were at https://github.com/github/codeql-action/security/code-scanning?query=ref%3Arefs%2Fheads%2Faction_libs_query_alerts


chrisgavin (approved):
I don't think it's essential to resolve now, but just to check my understanding, am I right in thinking that so long as a use of an Actions library is guarded on one call path, it will be considered guarded on all of them?

Might make sense to add a global logger at some point so we don't have to pass it around.

PR #153, opened 2020-08-20
Add tempDir and codeQLCmd fields to the config

robertbrignull:
This PR has two goals, that both involve adding data to the internal config:
- Add `tempDir` and `toolCacheDir` to the config. This doesn't make much difference now but it will make it easier in future for these values to come from the CLI instead of the environment.
- Add `codeQLCmd` to the config. Currently we store the path to the CodeQL executable in the environment, but that won't be possible outside of actions. Instead we can store it in the config.


rneatherway:
The changes are mostly fairly mechanical, adding these extra parameters throughout, and look correct.

It seems a bit strange to me that there are still many calls of `util.getRequiredEnvParam('RUNNER_TEMP')`. I would have expected that this wouldn't be necessary, perhaps by passing the `config` object around after it is first created. However, I am not well versed in how the action works so that may not be possible because, for example, the autobuild action is not called by us directly. It does seem surprising that `finalize-db` cannot have the config passed to it though.

At the end of the day it seems that we will need a single environment variable `RUNNER_TEMP` that tells us where to find the cached config after `init` has dealt with it the first time. Does that sound right?

rneatherway (approved):

PR #155, opened 2020-08-20
Python deps setup

Daverlo:
Here I'm adding the support for the auto-install python dependancies. The PR adds the necessary scripts to the `python-setup` folder. The scripts were developed by @RasmusWL 

This process is done in two steps, in the init action we install the necessary tools and try to get the dependencies, in the analyse step we configure the python extractor so it is able to find them.

The whole init process is not done in the runner as we don't want our cli to be messing around with things in the system. For the self hosted runners we don't try to install the tools for similar reasons, but we do try to get the dependencies, and that should work as long as the tools are present in the system.

There is a possibility of disabling this process by setting the input `setup-python-dependencies` to `false`.

The dependencies can be installed manually, and by providing  the `CODEQL_PYTHON` env var in the analyse step the extractor should be able to pick them up. That en var should contain the path to the python executable that has access to the dependencies.

This comment contains more details about the different usage possibilities: https://github.com/github/codeql-python-team/issues/103#issuecomment-663252898


python-setup/extractor_version.py
```diff
@@ -0,0 +1,53 @@
...
+        with suppress_stdout_stderr():
+            return buildtools.discover.get_version()
+    else:
+        return buildtools.discover.get_version()
+
+
+if __name__ == "__main__":
+    codeql_base_dir = sys.argv[1]
+    version = get_extractor_version(codeql_base_dir)
+    print('{!r}'.format(version))
```
marcogario, line 53:
```suggestion
    print('{!r}'.format(version))

```

python-setup/extractor_version.py
```diff
@@ -0,0 +1,53 @@
...
+
+    zippath = os.path.join(extractor_dir, getzipfilename())
+    sys.path = [zippath] + sys.path
+    import buildtools.discover
+
+    if quiet:
+        with suppress_stdout_stderr():
+            return buildtools.discover.get_version()
+    else:
+        return buildtools.discover.get_version()
```
marcogario, line 47:
```suggestion
    return buildtools.discover.get_version()
```

python-setup/extractor_version.py
```diff
@@ -0,0 +1,53 @@
...
+        sys.stderr = devnull
+        try:
+            yield
+        finally:
+            sys.stdout = old_stdout
+            sys.stderr = old_stderr
+
+
+def get_extractor_version(codeql_base_dir: str, quiet: bool = True) -> int:
+
```
marcogario, line 33:
```suggestion
```

python-setup/auto_install_packages.py
```diff
@@ -0,0 +1,141 @@
...
+    # create temporary directory ... that just lives "forever"
+    venv_path = mkdtemp(prefix='codeql-action-python-autoinstall-')
+
+    # virtualenv is a bit nicer for setting up virtual environment, since it will provide
+    # up-to-date versions of pip/setuptools/wheel which basic `python3 -m venv venv` won't
+
+    if version == 2:
+        _check_call(['python2', '-m', 'virtualenv', venv_path])
+    elif version == 3:
+        _check_call(['python3', '-m', 'virtualenv', venv_path])
```
marcogario, line 63: Can `get_extractor_version` return anything else apart from `2` or `3`? If it can, then we do not install the virtual env, right? Is this expected? If it cannot then I would change this into:
```suggestion
    else:
        assert version == 3
        _check_call(['python3', '-m', 'virtualenv', venv_path])
```

python-setup/install_tools.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# The binaries for packages installed with `pip install --user` are not available on PATH
+# by default, so we fix up PATH to suppress warnings by pip. This also needs to be done by
+# any script that needs to access poetry/pipenv.
+#
+# Using `::add-path::` from the actions toolkit is not enough, since that only affects
+# subsequent actions in the current job, and not the current action.
+export PATH="$HOME/.local/bin:$PATH"
+
+python2 -m pip install --user --upgrade pip setuptools wheel
```
marcogario, line 13: Is it worth checking that you actually have a `python2` executable?

python-setup/install_tools.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# virtualenv is a bit nicer for setting up virtual environment, since it will provide up-to-date versions of
+# pip/setuptools/wheel which basic `python3 -m venv venv` won't
+python2 -m pip install --user virtualenv
+python3 -m pip install --user virtualenv
+
+# venv is required for installation of poetry or pipenv (I forgot which)
+sudo apt-get install -y python3-venv
+
+# We're install poetry with pip instead of the recommended way, since the recommended way
```
marcogario, line 24:
```suggestion
# We install poetry with pip instead of the recommended way, since the recommended way
```

python-setup/install_tools.sh
```diff
@@ -0,0 +1,33 @@
...
+python2 -m pip install --user --upgrade pip setuptools wheel
+python3 -m pip install --user --upgrade pip setuptools wheel
+
+# virtualenv is a bit nicer for setting up virtual environment, since it will provide up-to-date versions of
+# pip/setuptools/wheel which basic `python3 -m venv venv` won't
+python2 -m pip install --user virtualenv
+python3 -m pip install --user virtualenv
+
+# venv is required for installation of poetry or pipenv (I forgot which)
+sudo apt-get install -y python3-venv
```
marcogario, line 22: Ok, I dig up https://github.com/github/codeql-python-team/issues/103#issuecomment-692053819 and it seems we are expecting only the linux cloud runners to use this for now. 

I haven't check yet but it would be good to have a comment saying where this should (and should not) work in the `init`, as that is where you can enable/disable this.

marcogario (commented):
This looks good! I think you should probably get a :+1: from @RasmusWL as well.

I left a few minor comments. I think my main concern is that there are quite a few situations that we want to support (poetry, pip, etc.), and it seems worthwhile to add a test scenario for each of them, and make sure we are indeed handling them correctly.

python-setup/extractor_version.py
RasmusWL: I think `pylint` has a standard rule making the same recommendation as you propose here, but I strongly disagree, since the control flow of the code is much easier to follow with the explicit `else` (in my opinion).

python-setup/install_tools.sh
RasmusWL: For a long term solution it probably would be worth checking if `python2` is not available and install it otherwise. For now I am not too worried -- as long as we have some automated tests that run regularly, that should catch this case right? (what do you think @Daverlo)

python-setup/auto_install_packages.py
RasmusWL: it will only return `2` or `3`.

python-setup/auto_install_packages.py
```diff
@@ -0,0 +1,141 @@
...
+
+        _check_call([venv_pip, 'install', '-e', '.'])
+    except subprocess.CalledProcessError:
+        sys.exit('package installation with `pip install -e .` failed, see error above')
+
+    return venv_python
+
+
+def install_packages() -> str:
+    if os.path.exists('poetry.lock'):
```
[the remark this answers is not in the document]
RasmusWL: yes!

python-setup/auto_install_packages.py
```diff
@@ -0,0 +1,141 @@
...
+        return install_packages_with_poetry()
+
+    if os.path.exists('Pipfile') or os.path.exists('Pipfile.lock'):
+        if os.path.exists('Pipfile.lock'):
+            print('Found Pipfile.lock, will install packages with Pipenv', flush=True)
+        else:
+            print('Found Pipfile, will install packages with Pipenv', flush=True)
+        return install_packages_with_pipenv()
+
+    version = extractor_version.get_extractor_version(sys.argv[1], quiet=False)
```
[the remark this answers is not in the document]
RasmusWL: We should be passing in `codeql_base_dir` for sure :+1:

We _could_ run `get_extractor_version` before calling `install_packages`, I just didn't... if you feel very strongly about this one, let me know.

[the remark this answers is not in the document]
RasmusWL: ehh, naming is apparently not too obvious. `extractor_version.get_extractor_version` returns the Python version the extractor thinks this repo is in.

python-setup/auto_install_packages.py
```diff
@@ -0,0 +1,141 @@
...
+        # > uninstall with pip.
+
+        _check_call([venv_pip, 'install', '-e', '.'])
+    except subprocess.CalledProcessError:
+        sys.exit('package installation with `pip install -e .` failed, see error above')
+
+    return venv_python
+
+
+def install_packages() -> str:
```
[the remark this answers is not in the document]
RasmusWL: Agreed

python-setup/auto_install_packages.py
```diff
@@ -0,0 +1,141 @@
...
+    if os.path.exists('requirements.txt'):
+        print('Found requirements.txt, will install packages with pip', flush=True)
+        return install_requirements_txt_packages(version)
+
+    if os.path.exists('setup.py'):
+        print('Found setup.py, will install package with pip in editable mode', flush=True)
+        return install_with_setup_py(version)
+
+    print("was not able to install packages automatically", flush=True)
+
```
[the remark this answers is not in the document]
RasmusWL: sure :+1:

RasmusWL (commented):
Thanks for the review @marcogario :+1:

The context for the Python scripts (that I wrote) is that it's something I made fairly quickly, and having in mind this should hopefully only be used for a limited time, I didn't invest a lot of effort into making this the highest quality code. I'll fix up some of the most blatant errors and bad coding practices.

RasmusWL:
@Daverlo I comitted directly to the branch with my changes. That means I actually haven't run the script, which is kind of scary. Is it easy to add tests so we know the script is working? (or have they already been added?)

Daverlo:
Tests are not going to be too helpful, if it doesn't work we just show a warning, and those are not easily visible. In this action this process failed on purpose https://github.com/dsp-testing/daverlo-djanjo-test/actions/runs/272089425, but you see all green no warning in the actions page https://github.com/dsp-testing/daverlo-djanjo-test/actions and in the pr page https://github.com/dsp-testing/daverlo-djanjo-test/pull/2

I would propose that we run the different tests on the django-test repo (basically update the workflows that I created for https://github.com/github/codeql-python-team/issues/103#issuecomment-663252898 so they run every day).

Do you have any suggestion on how can we do better testing?

python-setup/auto_install_packages.py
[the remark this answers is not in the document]
marcogario: Either is fine, as long as we do not need to reach out for `sys.argv`.

marcogario:
> Do you have any suggestion on how can we do better testing?

I think the typescript wrapper is not too complex. I would expect we can test the python script directly. Ideally these should all complete outputting the env for CODEQL and you can check that. Maybe create a new workflow file for this.

python-setup/auto_install_packages.py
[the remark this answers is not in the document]
RasmusWL: I wasn't able to accept this suggestion since it's on the old code, but I can make the edit myself :stuck_out_tongue:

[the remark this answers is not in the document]
marcogario: Oh, sorry about that.

Daverlo:
I've added a modified version of the playground workflow for testing the scripts. https://github.com/github/codeql-action/actions/runs/278176450

@RasmusWL That should cover all the cases am I right?

RasmusWL:
> I've added a modified version of the playground workflow for testing the scripts. https://github.com/github/codeql-action/actions/runs/278176450
> 
> @RasmusWL That should cover all the cases am I right?

Nice :+1:

Unfortunate about the "expected failures", but not something I will personally worry too much about. 

This should allow us to catch blatant error in the python script early, which is super good :tada: 

> I would propose that we run the different tests on the django-test repo (basically update the workflows that I created for [github/codeql-python-team#103 (comment)](https://github.com/github/codeql-python-team/issues/103#issuecomment-663252898) so they run every day).

Is this still our best approach for something similar for integration tests?

Daverlo:
> Is this still our best approach for something similar for integration tests?

Yep, but as @marcogario suggested, by testing the scripts we should be quite covered

> I think the typescript wrapper is not too complex. I would expect we can test the python script directly. Ideally these should all complete outputting the env for CODEQL and you can check that. Maybe create a new workflow file for this.

.github/workflows/python-deps.yml
```diff
@@ -0,0 +1,63 @@
...
+      with:
+        languages: python
+
+    - name: Test Auto Package Installation
+      run: |
+        set -x
+        $GITHUB_WORKSPACE/python-setup/install_tools.sh
+        echo -e '\n\n\n\n\n' && sleep 0.5
+        cd $GITHUB_WORKSPACE/${{ matrix.test_dir }}
+        $GITHUB_WORKSPACE/python-setup/auto_install_packages.py /opt/hostedtoolcache/CodeQL/0.0.0-20200826/x64/codeql/
```
marcogario, line 55: This will break when we upgrade the bundle version. Can you do a find before to get the version of the bunde against `/opt/hostedtoolcache/CodeQL/0.0.0-XXXX/x64/codeql` ?

marcogario (approved):
With the testing in place, this is good on my side.

We might want to keep the cases in which we know we will fail in the tests, but that is a minor.

.github/workflows/python-deps.yml
RasmusWL: This used to be set as an environment variable, but last time I had to run my scripts it was removed :disappointed: so that one is probably on me, although I have no idea of how to properly fix it.

marcogario: What about https://github.com/github/codeql-action/pull/239/files ?

PR #158, opened 2020-08-24
Break out tracer-config.ts to its own file

robertbrignull:
Pulls a bunch of code from `setup-tracer.ts` to a new `tracer-config.ts` file. This is the code that generates the environment variables to be set when running later processes so that CodeQL can intercept the compiler commands.

This is in preparation for chaning the behaviour when in CLI mode to output the env vars to a file instead of using `core.exportVariable`, while keeping all the code that generates the environment the same. Also while I'm doing it I've written tests for this code which was previously untested.

Also reverts a bad change from a previous PR of mine. In https://github.com/github/codeql-action/pull/140/files#diff-244be00bfdf0600dd7b228caba9f8232L57 I changed the type of the argument from an object to an array because I thought it wouldn't make a difference, but I missed a call to `Object.keys` which then silently did the wrong thing of returning keys like 0, 1, 2, 3... This basically didn't make a difference thankfully so tracing still worked, and the only functionality that was broken by this was the requirement to put the spec from cpp at the end.

Other than the above noted fix of a regression, this PR shouldn't make any changes to behaviour.

@sampart do you have scope to review this?


src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+          'javascript': {spec: spec, env: {'a': 'a', 'b': 'b'}},
+          'python': {spec: spec, env: {'b': 'c'}},
+        },
+        config));
+    t.deepEqual(e.message, 'Incompatible values in environment parameter b: b and c');
+  });
+});
+
+// If cpp is present then it's spec lines always come at the end
+test('concatTracerConfigs - cpp comes last', async t => {
```
sampart, line 165:
```suggestion
test('concatTracerConfigs - C++ spec lines come last if present', async t => {
```

If you do keep the comment, remove the apostrophe from `it's` 😉

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+      }
+    });
+    t.true(fs.existsSync(result.spec));
+    t.deepEqual(
+      fs.readFileSync(result.spec, 'utf8'),
+      path.join(tmpDir, 'compound-build-tracer.log') + '\n3\nghi\nabc\ndef');
+  });
+});
+
+test('concatTracerConfigs - SEMMLE_COPY_EXECUTABLES_ROOT', async t => {
```
sampart, line 204:
```suggestion
test('concatTracerConfigs - SEMMLE_COPY_EXECUTABLES_ROOT is updated to point to compound spec', async t => {
```

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    t.deepEqual(result.env, {
+      'a': 'a',
+      'b': 'b',
+      'SEMMLE_COPY_EXECUTABLES_ROOT': path.join(tmpDir, 'compound-temp')
+    });
+  });
+});
+
+test('concatTracerConfigs - compound environment file', async t => {
```
sampart, line 226: This test name could do with a verb adding too - what is it about the compound environment file that you're testing?

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+        'foo': 'bar_baz',
+      }
+    };
+
+    const result = concatTracerConfigs({ 'javascript': tc1, 'python': tc2 }, config);
+    const envPath = result.spec + '.environment';
+    t.true(fs.existsSync(envPath));
+
+    const buffer: Buffer = fs.readFileSync(envPath);
+    t.deepEqual(28, buffer.length);
```
sampart, line 253: Your expected values and your actual values are the wrong way round in this set of assertions, which will make the output confusing if one fails.

```
.deepEqual(value, expected, message?)
```

(https://github.com/avajs/ava/blob/master/docs/03-assertions.md#deepequalvalue-expected-message)

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    t.deepEqual(2, buffer.readInt32LE(0));
+    t.deepEqual(4, buffer.readInt32LE(4));
+    t.deepEqual('a=a\0', buffer.toString('utf8', 8, 12));
+    t.deepEqual(12, buffer.readInt32LE(12));
```
sampart, line 257: Could you add some comments explaining why these values are expected?  It's not clear to me from the test setup above.

One option might be a big comment block above these assertions showing what the full environment file should look like and explaining why.

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    t.deepEqual('foo=bar_baz\0', buffer.toString('utf8', 16, 28));
+  });
+});
+
+test('getTracerConfig - no traced languages', async t => {
```
sampart, line 262:
```suggestion
test('getTracerConfig - return value undefined when none of the passed languages are traced languages', async t => {
```

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    const codeQL = setCodeQL({
+      getTracerEnv: async function() {
+        return {
+          'ODASA_TRACER_CONFIGURATION': 'abc',
+          'foo': 'bar'
+        };
+      },
+    });
+
+    t.deepEqual(undefined, await getTracerConfig(config, codeQL));
```
sampart, line 277:
```suggestion
    t.deepEqual(await getTracerConfig(config, codeQL), undefined);
```

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+  });
+});
+
+test('getTracerConfig - full', async t => {
```
sampart, line 281: full what?  And when that thing is full, what behaviour are you expecting?

src/tracer-config.ts
```diff
@@ -0,0 +1,159 @@
...
+import * as configUtils from './config-utils';
+import { isTracedLanguage, Language } from './languages';
+import * as util from './util';
+
+export type TracerConfig = {
+  spec: string;
+  env: { [key: string]: string };
+};
+
+const CRITICAL_TRACER_VARS = new Set(
```
sampart, line 14: Is it worth adding a comment to explain what "critical" means here (or renaming the variable)?

src/tracer-config.ts
```diff
@@ -0,0 +1,159 @@
...
+    buffer = Buffer.concat([buffer, sizeBuffer, lineBuffer]);
+  }
+  // Write the compound environment
+  const envPath = spec + '.environment';
+  fs.writeFileSync(envPath, buffer);
+
+  return { env, spec };
+}
+
+export async function getTracerConfig(
```
sampart, line 132: This function and `tracerConfig` have very similar names.  Is it possible to rename them to make their different purposes more obvious?  If not, please could you add a one-liner explaining what each one does?

src/tracer-config.ts
```diff
@@ -0,0 +1,159 @@
...
+  codeql: CodeQL): Promise<TracerConfig | undefined> {
+
+  // Abort if there are no traced languages as there's nothing to do
+  const tracedLanguages = config.languages.filter(isTracedLanguage);
+  if (tracedLanguages.length === 0) {
+    return undefined;
+  }
+
+  // Get all the tracer configs and combine them together
+  let tracedLanguageConfigs: { [lang: string]: TracerConfig } = {};
```
sampart, line 143: I think this can be a constants as you're not reassigning it, just updating its inner values.

```suggestion
  const tracedLanguageConfigs: { [lang: string]: TracerConfig } = {};
```

sampart (approved):
Thanks for this refactor - this is a good step forward.

Many of my comments are about the tests; I think it's important that tests act a bit like documentation (since by nature they describe the system's behaviour), or at least that they make sense without much context (as someone may have to fix a failing test without knowing the project well), hence my insistence on good names.

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+      spec: 'abc',
+      env: {
+        'baz': 'qux',
+        'SEMMLE_PRELOAD_libtrace': 'SEMMLE_PRELOAD_libtrace',
+        'SEMMLE_RUNNER': 'SEMMLE_RUNNER',
+        'SEMMLE_COPY_EXECUTABLES_ROOT': 'SEMMLE_COPY_EXECUTABLES_ROOT',
+        'SEMMLE_DEPTRACE_SOCKET': 'SEMMLE_DEPTRACE_SOCKET',
+        'SEMMLE_JAVA_TOOL_OPTIONS': 'SEMMLE_JAVA_TOOL_OPTIONS',
+        'CODEQL_VAR': 'CODEQL_VAR',
+      }
```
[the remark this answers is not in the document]
robertbrignull: `foo` is not listed here because it's already in the environment with a different value. That's the point of this test but I agree it's very unclear. I'll add a comment.

src/tracer-config.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    process.env['SEMMLE_PRELOAD_libtrace'] = 'abc';
+    process.env['SEMMLE_RUNNER'] = 'abc';
+    process.env['SEMMLE_COPY_EXECUTABLES_ROOT'] = 'abc';
+    process.env['SEMMLE_DEPTRACE_SOCKET'] = 'abc';
+    process.env['SEMMLE_JAVA_TOOL_OPTIONS'] = 'abc';
+    process.env['SEMMLE_DEPTRACE_SOCKET'] = 'abc';
+    process.env['CODEQL_VAR'] = 'abc';
+
+    const result = await tracerConfig(codeQL, config, Language.javascript);
+    t.deepEqual(result, {
```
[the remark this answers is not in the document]
robertbrignull: The difference is that `foo` is missing, and "overwritten" here refers to the values from the CodeQL call overwriting those that exist in the environment already. I've add some comments and swapped setting `process.env` and setting up the mocked CodeQL response incase that makes it easier to see.

src/tracer-config.test.ts
robertbrignull: Honestly I'm not 100% sure what this file does. I'm just asserting that it's contents are as they are now so we'll detect if they change in the future. I've updated the test name to a generic "compound environment file is created correctly"

src/tracer-config.test.ts
robertbrignull: Thanks. It annoys me that the order is swapped for different testing libraries.

src/tracer-config.test.ts
robertbrignull: I've added some comments, but ultimately this should be documented in the algorithm itself when it was written. I think the format we outputting here is internal to CodeQL and thus it's unfortunately not pretty and not documented well.

src/tracer-config.test.ts
sampart: Yeah, me too.  And they have different (often overly terse) ways of showing you expected and actual in the failure output too.

src/tracer-config.test.ts
robertbrignull: I've renamed to "valid spec file". The name "full" came from the fact that we've tested each function individually and now we're testing them all at once by calling the outer function.

src/tracer-config.ts
robertbrignull: I can't answer this, sorry. I'm just copying this code and don't know enough to give this a better name.

src/tracer-config.ts
robertbrignull: I've changed to `getTracerConfigForLanguage` and `getCombinedTracerConfig` and I think that makes things a bit clearer.

robertbrignull:
There are still a couple of places where the test names could be better, but I'm slightly at a loss of what to do. I'll merge this later on today and maybe we can improve things before then, but if not then the tests are something we can easily change after merging.

PR #160, opened 2020-08-24
Rename the action entrypoint files

robertbrignull:
Renames all the entrypoint files of the actions to have a `-action` suffix. The exact suffix isn't too important, but I wanted them to all be easily identifiable, and for the names to match the action names. It's been too long since we renamed the actions and the src files still have the old names.


robertbrignull:
I think this is a slight net-negative because there was some cruft checked into the `lib` folder that shouldn't have been there anymore. I've opened https://github.com/github/codeql-action/pull/161 which should probably be merged before this one, but it doesn't matter too much.

chrisgavin (approved):

PR #161, opened 2020-08-24
Clean up the lib directory

robertbrignull:
There appear to be some files checked in that shouldn't be. I think they are for old ts files, and when we deleted the ts files the js versions stuck around.


robertbrignull:
You can see the checks failed on the first commit (https://github.com/github/codeql-action/runs/1021997379) but succeeded on the second, so that means this check is working.

chrisgavin (approved):
Thanks. I always assumed the TypeScript compiler would just be doing this for us...

PR #164, opened 2020-08-28
Run `npm test` on linux and macos

robertbrignull:
A change I made in https://github.com/github/codeql-action/pull/162 accidentally broke the unit tests on non-linux systems.

I think running just this on all development operating systems will be fine. We don't need to do the other jobs from this workflow.


robertbrignull:
Oh, didn't realise the bug was already on this branch. Should be fixed now. I'm not sure if this is exactly the right fix as it does feel like we're just duplicating the implementation a bit. Alternative would be just deleting this test, but I think let's keep it for now.

robertbrignull:
Ok. For now I've scaled this back to only running on linux and osx. We should get the tests running on windows too but it seems there are a bunch of failures and it'll be more effort.

chrisgavin (approved):

.github/workflows/pr-checks.yml
```diff
@@ -65,9 +65,12 @@ jobs:
         echo "Success: node_modules are up to date"
 
   npm-test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest,macos-latest]
```
aeisenberg, line 70: Any reason why windows isn't included here?

robertbrignull: The tests don't currently pass on windows and I think it's going to be a bit of effort to get them working. I want to do that but we may as well merge this now and do windows separately later.

aeisenberg: Sure.

PR #165, opened 2020-08-28
Allow "additive" queries in workflow by prefixing with "+"

sampart:
Extends the functionality added in https://github.com/github/codeql-action/pull/127 so that you can use queries from a workflow and a config file together, if you wish.  This is the second part of the work for https://github.com/github/code-scanning/issues/1446

@hubwriter FYI, and could you also check the wording in `init/action.yml` and the README please?


src/config-utils.test.ts
```diff
@@ -431,6 +483,70 @@ test("Multiple queries can be specified in workflow file, no config file require
...
+    // These queries shouldn't override anything, because the value is prefixed with "+"
+    setInput('queries', '+./additional1,./additional2');
+
+    fs.mkdirSync(path.join(tmpDir, 'foo'));
+    fs.mkdirSync(path.join(tmpDir, 'additional1'));
+    fs.mkdirSync(path.join(tmpDir, 'additional2'));
+
+    const resolveQueriesArgs: {queries: string[], extraSearchPath: string | undefined}[] = [];
+    const codeQL = setCodeQL({
+      resolveQueries: async function(queries: string[], extraSearchPath: string | undefined) {
```
sampart, line 508: We have this function in lots of tests.  Could it be extracted out?

sampart: Fixed in 23cf700e38cfbee3a56da840563b8491b677859e

src/config-utils.ts
```diff
@@ -525,25 +525,33 @@ async function getLanguages(): Promise<Language[]> {
...
-    return true;
+  }
+}
+
+// Returns true if either no queries were provided in the workflow.
+// or if the queries in the workflow were provided in "additive" mode,
+// indicating that they shouldn't override the config queries but
+// should instead be added in addition
+function shouldAddConfigFileQueries(): boolean {
+  const queryUses = core.getInput('queries');
```
robertbrignull, line 549: This will need merging/rebasing on `main`. Now this part of the codebase has been converted to the 3rd party runner you can't use `core.getInput` in shared code. This input has to be accessed from `init-action.ts` and passed in to here. That should already have been done on `main` so you'll just need to do some rerouting.

src/config-utils.test.ts
```diff
@@ -24,6 +24,11 @@ function setInput(name: string, value: string | undefined) {
   }
 }
 
+function setConfigFile(inputFileContents: string, tmpDir: string) {
+  fs.writeFileSync(path.join(tmpDir, 'input'), inputFileContents, 'utf8');
+  setInput('config-file', 'input');
```
robertbrignull, line 29: I think this `setInput` function has been removed on `main`. You now pass the value to `initConfig` as an ordinary function parameter. Unfortunately this removes a lot of the usefullness of this helper function, so you may unfortunately be best off reverting it.

robertbrignull (commented):
Generally LGTM, though I'll hold off on approving as there are going to be a lot of changes when you merge/rebase this on `main`.

src/config-utils.ts
```diff
@@ -553,6 +549,10 @@ async function addQueriesFromWorkflow(
   githubUrl: string,
   logger: Logger) {
 
+  queriesInput = queriesInput.trim();
```
robertbrignull, line 552: Could be slightly better to avoid doing the trimming in multiple places. Could do it just one before passing to `addQueriesFromWorkflow` and `shouldAddConfigFileQueries`

robertbrignull (approved):
LGTM

We'll test it in person later on

src/config-utils.ts
sampart: I've just written some code to do this - I did the trimming in `Config.initConfig` since that's the best "common ancestor" of the code flows which end up in these functions, so we only need to trim once.

However, I've not pushed it yet as I'm a bit unsure about it - keeping trimming in these functions themselves means that we won't have any odd bugs if they're ever called from anywhere else. It feels a bit risky to have the necessary trimming so far from the code that requires it to have happened.

What do you think about leaving this as it is?

sampart: Robert and I have discussed and we're going to leave this as-is.

PR #168, opened 2020-09-02
make inputting refs easier

robertbrignull:
It's annoying that you have to specify the ref to the API in it's full explicit format. Users definitely won't do this so I think we should prepend `refs/heads/` if whatever they enter doesn't already start with `resf/`.

I admit there is a slight possibilty for confusion if someone makes a branch legitimately called `refs/heads/refs/foo` but this is quite unlikely.


robertbrignull:
@rneatherway, can you review this? Do you think applying this transformation makes sense so users can input the branch name they are familiar with?

rneatherway:
> I admit there is a slight possibilty for confusion if someone makes a branch legitimately called `refs/heads/refs/foo` but this is quite unlikely.

I agree. It's still possible to specify that branch unambiguously using the full form so I think that's fine.

rneatherway (approved):

PR #169, opened 2020-09-02
Print that we are downloading codeql and it may take a while

robertbrignull:
The CodeQL download can take quite a while on a normal internet connection and GitHub appears to also deliberately throttle the download speed. This means it can easily take 10+ minutes, so we need to output something nice. Before this PR the last thing that gets printed is "checking cache" and "not found" which is not informative enough.


aeisenberg:
10 minutes to download??? I almost wonder if it's better to build the bundle ourselves.

aeisenberg (commented):
What you have looks perfectly reasonable. 

I notice that there is a comment on `toolcacheDownloadTool` to remove once https://github.com/actions/toolkit/pull/530 is merged and released. Looks like it has already, so perhaps this is a good time to change? And perhaps going through the tool-cache package would avoid throttling?

robertbrignull:
> Looks like it has already

https://github.com/actions/toolkit/pull/530 doesn't look merged to me. I think the confusion is that it mentions https://github.com/github/codeql-action/pull/123 which is merged so there's a big "merged" icon, but it's for the other PR.

> And perhaps going through the tool-cache package would avoid throttling?

The code in this file was copied from the toolcache, and then slightly modified with the changes from that PR, so changing to use the tool-cache properly wouldn't change anything. The lack of download speed is downloading the release artifact from github.com and I'm pretty sure is imposed at their end and not caused by the method in which we are doing the downloading.

aeisenberg:
My mistake...you're right. It's not merged.

robertbrignull:
@sampart can you review? Or @aeisenberg please feel free to also review. Currently I think putting this message in is better than nothing and I think we should get this in as a priority.

Sometimes the download can take so long that you'll think it's broken, so outputting progress updates on some kind of regular schedule would be better. But I think we can investigate that separately.

sampart (approved):

PR #172, opened 2020-09-07
Use smaller Plarform-Language packages when analyzing only one language

marcogario:
When the action (or runner) analyzes only a single language, we download a smaller codeql bundle.

The main change here is that we now need to first detect the language and then setup codeQL. The language detection used to be a step of the `initConfig` step. We move this out and perform it before `initCodeQL`.

The changes to `setupCodeQL` are meant to correctly cache the tooling. I.e., if we already have a full bundle in the cache, we should not try to download a platform-language bundle. Similarly, if we did download a platforms-language bundle, then we should not cache it as a full bundle.

We extend this logic to only download the platform specific bundle, so that linux runners do not need to download a package with both windows and osx binaries.

I added this comment in the `setupCodeQL` to make the flow a bit more clear:
```
  // The URL identifies the release version. E.g., codeql-20200901 .
  // The plVersion identifies the platform-language combination of the package
  // within the release. E.g., `linux64-cpp` in `codeql-linux64-cpp.tar.gz`.
  // We expect the codeqlUrl (when given) to always point to the main bundle
  // `codeql-bundle.tar.gz`
  //
  // The logic is as follows:
  //  - Always use the Toolcache if available.
  //    - If we would like a platform-language package, but have the
  //      full bundle in the cache, use that.
  //  - If codeqlURL is specified, use that.
  //  - If a single language is being analyzed, try to download the platform-language package.
  //    - If it is not available in the release assets, fallback to the full bundle
  //  - If multiple languages are being anlyzed, use the full bundle
```

I will like to test/merge this code before merging the code to split the bundles, to ensure that it works if we do not create the platform-language packages (it should, and we have tests, but...).


marcogario:
I ran an experiment comparing zip0 and tar.gz decompression time. There is basically no difference between zip and tar.gz. I would keep the tar.gz for consistency.

src/codeql.ts
```diff
@@ -206,16 +211,70 @@ export async function setupCodeQL(
...
 
-    let codeqlFolder = toolcache.find('CodeQL', codeqlURLVersion);
-    if (codeqlFolder) {
-      logger.debug(`CodeQL found in cache ${codeqlFolder}`);
-    } else {
+    let codeqlFolder;
+
+    logger.debug(`PL Version ${plVersion}`);
+    if (plVersion) {
+      codeqlFolder = toolcache.find('CodeQL', `${codeqlURLVersion}-${plVersion}`);
```
robertbrignull, line 250: The `toolcache` library annoyingly is quite strict that the version has to be valid semver. I can't quite tell from looking at it, but would this satisfy that?

In order to better support running outside of actions, I'm thinking we might be better off reimplementing the behaviour we use from the `toolcache` library ourselves. This could be another reason to do that ourselves if it makes dealing with plVersion easier.

src/config-utils.ts
```diff
@@ -491,7 +491,8 @@ async function getLanguagesInRepo(
  * If no languages could be detected from either the workflow or the repository
  * then throw an error.
  */
-async function getLanguages(
+export async function getLanguages(
+  // Maybe move to init.ts
```
robertbrignull, line 495: If these language-related function could be extracted to another file that would be good as this file is getting too big. Would probably be best in `languages.ts` instead, though I agree it'll mostly be used from `init.ts`.

Or even if not moved then probably best to remove this comment from the code. From what I've seen comments like this don't generally help whatever they are suggesting to come true as people overlook them, and they just get outdated which provides incorrect information to people who do read it in the future.

robertbrignull (commented):
A couple of comments, but I admit I haven't read through all that fully. I'd like to see a demo if possible as that would be best for showing it works. I'm a bit worried about our use of the toolcache as we're really not using it how it intends with regards to the version strings.

Overall though, the approach this is taking looks like it'll be good.

I'm sure you already know this, but probably worth noting we're also trying to get the bundle pre-populated into the actions runners. It's not clear exactly when this will happen but it's looking like it will. So that would mean that splitting up the bundle would be less useful on dotcom, but this is still potentially very useful for GHES or self-hosted runners on dotcom.

marcogario:
Thanks @robertbrignull, I will go through your comments in detail tomorrow and also update the code to match main.

>  I'm a bit worried about our use of the toolcache as we're really not using it how it intends with regards to the version strings.

I think we need to dig deeper here. My thought was this would improve all settings. Clearly self-hosted runners, but also dotCom, as I understood there is a window between we releasing a new bundle and the base image being rebuilt. However, also given your comment in https://github.com/github/code-scanning/issues/1801 , I might be misunderstanding how the version information is used in the toolcache. Do you have more information you can point me to, or shall we try to sync?

src/codeql.ts
marcogario: This follows the same idea as the rest of the code, by appending it with a dash. E.g., `0.0.0-20200909-linux64-cpp` . I thought this would be both 1) correct and 2) enough to keep things separate.

marcogario:
This is an example of how we fallback in the search of the tool:
```
 ##[debug]Bundle version 20200826 is not in SemVer format. Will treat it as pre-release 0.0.0-20200826.
  ##[debug]PL Version linux64-javascript
  ##[debug]isExplicit: 0.0.0-20200826-linux64-javascript
  ##[debug]explicit? true
  ##[debug]checking cache: /opt/hostedtoolcache/CodeQL/0.0.0-20200826-linux64-javascript/x64
  ##[debug]not found
  ##[debug]isExplicit: 0.0.0-20200826
  ##[debug]explicit? true
  ##[debug]checking cache: /opt/hostedtoolcache/CodeQL/0.0.0-20200826/x64
  ##[debug]not found
  ##[debug]CodeQL not found in cache
  ##[debug]RUNNER_TEMP=/home/runner/work/_temp
  ##[debug]Using CodeQL URL: https://github.com/github/codeql-action/releases/download/codeql-bundle-20200826/codeql-bundle.tar.gz
```

marcogario:
I was doing some more testing and I ran into the following issue. 

`initCodeQL` takes a codeqlURL that can be a string or undefined. I thought this was because the `tool` argument could be underspecified, and thus we have logic to detect which bundle to use. This is the case for the runner, but not for the init-action.ts, where this is always set to a string:

https://github.com/github/codeql-action/blob/57ef26cffee3737c4115dbe2ec175bf6c6f33d11/src/init-action.ts#L62

What could be the best way to deal with this? Can we detect that the value was not specified in the configuration?

robertbrignull:
> initCodeQL takes a codeqlURL that can be a string or undefined. I thought this was because the tool argument could be underspecified, and thus we have logic to detect which bundle to use. This is the case for the runner, but not for the init-action.ts, where this is always set to a string:

Yes, I agree this is a pain. Your analysis is right that for the runner it might be undefined, but for the action it's always a string. If it's not defined in the workflow then it gets set to the emptystring. So this means the best way to check it may be to do convert it to a boolean to see if it's truthy, but I agree this is no ideal.

I think this is a problem we should deal with and be , though doesn't have to be in this PR. Maybe we should convert all emptystrings to undefined when we read the inputs, but there are other options. I'll open an issue as I think we should track this and make sure it gets improved.

marcogario:
This is ready for review. I am available to walk through the code. The only change I expect is to rebase on https://github.com/github/codeql-action/pull/211, as I currently included the changes from https://github.com/github/codeql-action/pull/208.

src/codeql.test.ts
```diff
@@ -1,31 +1,135 @@
...
+          toolcache.find("CodeQL", toolcacheVersion),
+          `Looking for ${toolcacheVersion}`
+        );
+      }
+    }
 
-      nock("https://example.com")
-        .get(`/download/codeql-bundle-${version}/codeql-bundle.tar.gz`)
+    const cachedVersions = toolcache.findAllVersions("CodeQL");
+    t.is(cachedVersions.length, 4);
```
robertbrignull, line 72: So there should be `win32-cpp`, `linux64-cpp`, `osx64-cpp`, and then the full bundle? Are those the four versions expected? Could you add a comment to the code explaining this?

robertbrignull (commented):
I'm re-reviewing but I'm having to step out now so I'm posting what I have so far. I'll continue reviewing the rest later today.

src/codeql.ts
```diff
@@ -165,7 +166,8 @@ async function getCodeQLBundleDownloadURL(
     // If we've reached the final case, short-circuit the API check since we know the bundle exists and is public.
     if (
       apiURL === util.GITHUB_DOTCOM_URL &&
-      repository === CODEQL_DEFAULT_ACTION_REPOSITORY
+      repository === CODEQL_DEFAULT_ACTION_REPOSITORY &&
+      bundleNames[0] === CODEQL_BUNDLE_NAME
```
robertbrignull, line 170: I'm not convinced this will ever be true. The only place this is called from non-test code, the 0th element is either the platform bundle or the plVersion bundle. The raw full `CODEQL_BUNDLE_NAME` is never the 0th element. I can see why you're checking this here, because the idea is to not default to using the full bundle unless necessary, but I don't think the logic is quite right.

Since we own the release on github.com/github/codeql-action could you do the same thing of just assuming the release artifacts will be there for all platforms and languages? I realise they aren't there currently so that will delay merging this PR until they are there. The alternative would be to stop treating github.com as special and make the request to discover what assets are there. I don't have a strong favourite between those options as they both have positive and negative aspects.

robertbrignull (commented):
Ok, I've gone through everything now.

Only one comment and I can't see anything else wrong, though I'm still a bit uneasy about this feature as a whole because it's just quite complicated and I feel I'm not giving it the brain power / testing time it requires.

Also, https://github.com/github/codeql-action/pull/211 has been merged now, so this could be rebased onto that to simply things.

.github/workflows/integration-testing.yml
```diff
@@ -124,6 +124,45 @@ jobs:
       env:
         TEST_MODE: true
 
+  single-language-bundles:
```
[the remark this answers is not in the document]
marcogario: The idea was not to require this, although I can see how it would be good to test this in a non-failing way.
I did not want to make the bundle-release process more complex, so these tests should pass with only the main bundle (fallback behavior). I see how outputting a message that we used the smaller bundler would be useful for validation.

marcogario:
Thanks for the review. I think there is a conceptual point to address. I did this work trying to be compatible with the current bundle release process. This has a few visible impact in the implementation. As you noted in the review:

1. Integrations tests are currently designed to pass if we do NOT have the language specific bundle
2. The code checking for available bundles gets more complicated and expensive as we need to query the API and cannot shortcut effectively.

Since I automated the creation of the smaller bundles, it would seem that changing the release process has some benefits, with few drawbacks.

The only "problem" I see is that this would require to postpone merging this and first:

1. Change the release process and merge the workflow to generate the language bundles
2. Update the sync tool
3. Finally adjust this PR

@robertbrignull WDYT?

.github/workflows/integration-testing.yml
```diff
@@ -124,6 +124,45 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        language: ["none", "cpp", "csharp", "go", "java", "javascript", "python"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      # Translate OS
```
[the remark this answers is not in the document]
marcogario: Ah, this is a left-over. I might use it now, if I want to test the toolcache (as you suggest below). I am adding a comment, but in general, it is setting an env variable called `OS` with the correct string value depending on the platform.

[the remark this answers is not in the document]
marcogario: I don't need this anymore.

marcogario:
I went ahead and prepared the changes to the process as discussed above. I think that is the most reasonable solution. 

* We need to merge https://github.com/github/codeql-action/pull/216 first before the tests for this PR can succeed (as we now do not fail gracefully anymore)
* I tested the codeql-action-sync-tool and indeed it does download all assets
* Created a PR for the changes to the bundle release process.

marcogario:
As discussed with @robertbrignull , this will be closed in favor of a simpler version that only accounts for platforms.

PR #174, opened 2020-09-07
add error-catching wrapper around calls to toolrunner

nickfyson:
This PR adds a wrapper around specific `exec.exec` calls that are known to sometimes fail in specific ways, and captures both the exit code and stdout/err in order to match against a pre-defined list of conditions. Matching can be either by an exit code that the `codeql-cli` uses to indicate a specific condition, or by regex that matches against either stdout or stderr. When a match is found the wrapper throws a custom error message, which can be used to give snippets of documentation to users.

In this initial PR there is just a single matcher, intended to catch cases where no source code was seen during the build.

You can see the matchers in action in the extra (deliberately failing) integration tests here...
https://github.com/github/codeql-action/actions/runs/243679881

The testing story for this behaviour isn't entirely complete – ideally we would have integration tests like those above that deliberately trigger the errors in question and check they are caught as expected. Perhaps this could be achieved by some additional wrapper to test the wrapper, but that's all getting a little bit inception. I don't think there's a way to run to 'expect' a job to fail, but obviously that would be ideal.

Closes https://github.com/github/code-scanning/issues/1559


src/error-matcher.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  2020-09-07T17:39:53.9050522Z [2020-09-07 17:39:53] [build] Done extracting /opt/hostedtoolcache/CodeQL/0.0.0-20200630/x64/codeql/javascript/tools/data/externs/web/ie_vml.js (3 ms)
+  2020-09-07T17:39:53.9051849Z [2020-09-07 17:39:53] [build-err] No JavaScript or TypeScript code found.
+  2020-09-07T17:39:53.9052444Z [2020-09-07 17:39:53] [build-err] No JavaScript or TypeScript code found.
+  2020-09-07T17:39:53.9251124Z [2020-09-07 17:39:53] [ERROR] Spawned process exited abnormally (code 255; tried to run: [/opt/hostedtoolcache/CodeQL/0.0.0-20200630/x64/codeql/javascript/tools/autobuild.sh])
+  `));
+});
+
+function testErrorMatcher(matcherName: string, logSample: string): boolean {
+
+  const regex = namedMatchersForTesting[matcherName] ? namedMatchersForTesting[matcherName][1] : null;
```
alexrford, line 21: Maybe factor `namedMatchersForTesting[matcherName]` out to a `const matcher = namedMatchersForTesting[matcherName]`.

src/error-matcher.ts
```diff
@@ -0,0 +1,19 @@
+
+export type ErrorMatcher = [number|null, RegExp|null, string];
+
+// exported only for testing purposes
+export const namedMatchersForTesting: { [key: string]: ErrorMatcher } = {
+  /*
+  In due course it may be possible to remove the regex, if/when javascript also exits with code 32.
+   For context see https://github.com/github/semmle-code/pull/36921
```
robertbrignull, line 8: This is private repo so we shouldn't include the link. I know it's annoying. Ideally do the link the other way around, so from the private PR link to this code.

src/error-matcher.ts
```diff
@@ -0,0 +1,19 @@
+
+export type ErrorMatcher = [number|null, RegExp|null, string];
```
robertbrignull, line 2: What's the reasoning for being an array / tuple instead of an object / interface with named fields?
e.g.
```typescript
export interface ErrorMatcher {
  exitCode?: number;
  stdoutRegex?: RegExp;
  message: string;
}
```
assuming I've understood what the fields are for, which is the main benefit you get by naming them. Would also be good to add comments explaining what each field does.

src/error-matcher.ts
```diff
@@ -0,0 +1,19 @@
...
+// exported only for testing purposes
+export const namedMatchersForTesting: { [key: string]: ErrorMatcher } = {
+  /*
+  In due course it may be possible to remove the regex, if/when javascript also exits with code 32.
+   For context see https://github.com/github/semmle-code/pull/36921
+  */
+  noSourceCodeFound: [
+    32,
+    new RegExp("No JavaScript or TypeScript code found\\."),
+    `No source code was seen during the build. Please see...
```
robertbrignull, line 13: As you've discovered here, if you use the backtick multiline strings then you have to avoid any indentation. It's pretty useful if you want to include the entire contents of a file, but I'm not sure that makes it more readable for a string like this. That's just my opinion though and I'll leave it up to you to decide.

alexrford (approved):
LGTM - I'm wondering now in review if it might be clearer to define `ErrorMatcher` as:
```
interface ErrorMatcher {
  returnCode?: number;
  outputRegex?: RegExp;
  message: string;
}
```
rather than as a tuple, but I don't think it's a big deal in the context of how we're using the matchers.

src/error-matcher.test.ts
robertbrignull, line 21: A minor suggestion to make this more readable would be to move the error handling earlier and break out earlier. For example
```typescript
if (!(matcherName in namedMatchersForTesting)) {
  throw new Error(`Unknown matcher ${matcherName}`);
}

const regex = namedMatchersForTesting[matcherName][1];
if (regex === null) {
  throw new Error(`Cannot test matcher ${matcherName} with null regex`);
}

return regex.test(logSample);
```
What do you think?

src/toolrunner-error-catcher.ts
```diff
@@ -0,0 +1,83 @@
+import * as im from '@actions/exec/lib/interfaces';
+import * as toolrunnner from '@actions/exec/lib/toolrunner';
+
+import {ErrorMatcher} from './error-matcher';
+
+/**
+ * Wrapper for exec.exec which checks for specific return code and/or regex matches in console output.
```
robertbrignull, line 7: s/exec.exec/toolrunner.Toolrunner/

Same applies throughout this file, though some of them you might want to rename differently to make it read better

src/toolrunner-error-catcher.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Output will be streamed to the live console as well as captured for subsequent processing.
+ * Returns promise with return code
+ *
+ * @param     commandLine        command to execute (can include additional args). Must be correctly escaped.
+ * @param     matchers           defines specific codes and/or regexes that should lead to return of a custom error
```
robertbrignull, line 12: Parameter order in doc doesn't match the function signature

src/toolrunner-error-catcher.ts
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  // we capture the original return code or error so that if no match is found we can duplicate the behavior
+  let returnState: Error|number;
+  try {
+    returnState = await new toolrunnner.ToolRunner(
+      commandLine,
+      args,
+      {
+        ...options, // pass original options first in order to override below
```
robertbrignull, line 53: Just checking this comment means you want the below lines to override this one? That's the way it works, so the last definition of a field wins. Maybe worth clarifying the comment slightly.

src/toolrunner-error-catcher.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  if (stdoutContents) command += 'console.log(\\"' + stdoutContents + '\\");';
+  if (stderrContents) command += 'console.error(\\"' + stderrContents + '\\");';
+
+  if (command.length === 0) throw new Error("Must provide contents for either stdout or stderr");
+
+  if (desiredErrorMessage) command += 'throw new Error(\\"' + desiredErrorMessage + '\\");';
+  if (desiredExitCode) command += 'process.exitCode = ' + desiredExitCode + ';';
+
+  return toolrunnner.argStringToArray('-e "' + command + '"');
```
robertbrignull, line 150: What is `argStringToArray` providing here? It looks like if it does what I expect it to do it's equivalent to `['-e', command]`, and I'm worried there's a high chance it'll do something else weird because of unexpected characters or escaping.

src/toolrunner-error-catcher.test.ts
```diff
@@ -0,0 +1,151 @@
...
+import * as toolrunnner from '@actions/exec/lib/toolrunner';
+import test from 'ava';
+
+import { ErrorMatcher } from './error-matcher';
+import {setupTests} from './testing-utils';
+import { toolrunnerErrorCatcher } from './toolrunner-error-catcher';
+
+setupTests(test);
+
+test('matchers are never applied if non-error exit', async t => {
```
robertbrignull, line 11: Thanks for adding these. I very much like that there's testing of all scenarios for this.

robertbrignull (commented):
A fair number of comments but all of them minor, so generally looks good to me. Nothing major that I think will cause problems with this approach.

src/toolrunner-error-catcher.test.ts
nickfyson: That was an 'argh, we're not even using `exec.exec` any more, how can I get these tests to run so I can open my PR' kind of a change. :-D

Fixed properly now (by removing the escaping from the command strings).

robertbrignull: I guess another option would be to write it out to a file and then delete it afterwards. Not sure if that's more or less effort, but it would avoid escaping issues to a certain degree.

nickfyson: Actually it's all very neat now with the Toolrunner! 👍

src/error-matcher.ts
nickfyson: No reasoning, just how I hacked it together first and then never revisited! :-)

robertbrignull (approved):
Had another quick look at the fixes and LGTM

Unfortunately looks like there are some conflicts from the merge of https://github.com/github/codeql-action/pull/181, but you may hopefully be able to just run `npm run lint-fix`, or maybe you'll need to fix conflicts manually first.

PR #176, opened 2020-09-09
Add another fingerprinting test

robertbrignull:
This doesn't add too much but this test file came up as a source of fingerprinting inconsistencies w.r.t. CodeQL. I think we should add it as a test case now so we are asserting our behaviour on it and then if it does turn out there's a bug in our algorithm it'll be obvious when we fix it.


chrisgavin (approved):
I didn't realize this happened in such a simple file. Wow.

PR #177, opened 2020-09-09
Make a local repo for checkoutExternalQueries

robertbrignull:
This makes the `checkoutExternalQueries` test create a local git repo instead of downloading one from github.com. This makes the test faster so it won't risk timing out on slower internet connections, and makes it run at all without internet.

This turned out to be moderately complicated because of having to create the git repo. There are probably libraries that would make this easier but for just this one test I didn't think it was worth it.

As a bonus I also add a test of checking out two refs from the same repo, which is now possible without timing out the test.

I did make one small change to the non-test code which is remove the `.git` suffix when cloning. This made the test easier because the directory name was more predictable and because I'm cheating slightly by giving it a `file://` url. This shouldn't matter for github.com as that suffix is optional when cloning by HTTP. The integration tests should catch it if this now fails as one of those clones remote queries.


rneatherway (approved):
LGTM!

PR #179, opened 2020-09-10
Split up builtin and custom queries

robertbrignull:
Splits up the storage and running of builtin and custom queries. This enables us to time each set separately and populate the fields in the status report.

In `runQueries` in `analyze.ts` I had to get a little bit loose with the typing in order to populate the status report. The alternative would have been a massive switch to convert strings into the well-typed field names. It didn't seem worth it to me, but I'm open to being persuaded. I added a test for this method that hopefully will catch if we accidentally rename something that breaks assumptions.


rneatherway:
If we would like to use the `builtin` results to track the performance of the CodeQL builtin queries over time how will we distinguish between e.g. `security-and-quality`, the defaults, or some other suite with a few disabled?

robertbrignull:
In the `init` action the status report contains the queries being run, so we should be able to determine from there which suites were run and therefore get consistent timing data.

robertbrignull:
Although I noticed we aren't populating the status report quite right. I've opened https://github.com/github/codeql-action/pull/219 to fix this.

rneatherway:
Great, I'm happy then.

rneatherway (approved):

robertbrignull:
Just for the record, I've confirmed that the status reports are getting through correctly and this timing data is arriving.

PR #180, opened 2020-09-10
Fix "Using a custom configuration" link in the README

ericcornelissen:

### Description

Fix the "Using a custom configuration" link the README. The actual ID of the header on the page is `#using-a-custom-configuration-file`, so I updated both the link and the link title to match. (as far as I can tell all other links in the README work as expected.)

---

As an aside (not sure where best to report this), the snippet for [Specifying directories to scan](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#specifying-directories-to-scan) in the docs seems to be outdated. `paths-ignore` and `paths` can no longer be specified in the `with:` statement, according to [this error message](https://github.com/ericcornelissen/svgo-action/runs/1097787080#step:3:1):

```
##[warning]Unexpected input(s) 'paths-ignore', valid inputs are ['tools', 'languages', 'token', 'matrix', 'config-file', 'queries']
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Use the `config-file` parameter of the `init` action to enable the configuration
     config-file: ./.github/codeql/codeql-config.yml
 ```
 
-The configuration file must be located within the local repository. For information on how to write a configuration file, see "[Using a custom configuration](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#using-a-custom-configuration)."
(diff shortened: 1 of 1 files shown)


robertbrignull:
Thanks for spotting that and fixing it :+1: 

Thanks for the point about "Specifying directories to scan" too. I've forwarded this onto the docs team and we'll try to get that page fixed as soon as possible.

robertbrignull (approved):

robertbrignull:
We can ignore that one job failure. There's no way this PR broke it. I think that job just doesn't work on PRs from forks, but we'll look into it.

PR #208, opened 2020-09-15
Util for optional/required inputs

marcogario:
Add two utilities to obtain inputs from the action: `getOptionalInput` and `getRequiredInput`. 

These achieve two things. First, we have a more consistent use of `undefined` when optional inputs are not provided. Second, we provide a better semantic characterization of which inputs are optional or required (ensuring that we do throw if required inputs are not given).

The changes mostly affect the action and not the runner (that already has a similar semantics).


marcogario:
see #211.

PR #211, opened 2020-09-15
Make inputs consistent between action and runner

robertbrignull:
This is partially an alternative PR to https://github.com/github/codeql-action/pull/208 but I thought I'd open it anyway. This PR also goes further by:
- Creates `actions-util.ts` that allows us to place higher standards on `util.ts` to work on actions and the runner and makes it easier to see the division of what code runs where.
- Adding a CodeQL query to detect use of `core.input` and point to the correct custom function.
- An extra CodeQL query that detects inconsistent definitions of inputs between actions.


src/actions-util.ts
```diff
@@ -0,0 +1,332 @@
...
+  if (cause) {
+    statusReport.cause = cause;
+  }
+  if (exception) {
+    statusReport.exception = exception;
+  }
+  if (status === "success" || status === "failure" || status === "aborted") {
+    statusReport.completed_at = new Date().toISOString();
+  }
+  const matrix: string | undefined = core.getInput("matrix");
```
marcogario, line 264: Indeed! Code scanning is correct here!
```suggestion
  const matrix = getOptionalInput("matrix");
```

src/actions-util.ts
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Get an environment parameter, but throw an error if it is not set.
+ */
+export function getRequiredEnvParam(paramName: string): string {
+  const value = process.env[paramName];
+  if (value === undefined || value.length === 0) {
+    throw new Error(`${paramName} environment variable must be set`);
+  }
+  core.debug(`${paramName}=${value}`);
```
marcogario, line 39: I think this is a false positive.

src/codeql.ts
```diff
@@ -125,7 +126,7 @@ function getCodeQLActionRepository(mode: util.Mode): string {
   // Actions do not know their own repository name,
   // so we currently use this hack to find the name based on where our files are.
   // This can be removed once the change to the runner in https://github.com/actions/runner/pull/585 is deployed.
-  const runnerTemp = util.getRequiredEnvParam("RUNNER_TEMP");
+  const runnerTemp = getRequiredEnvParam("RUNNER_TEMP");
```
marcogario, line 129: I think this is what the comment above means, but just in case.

This code is called by the runner but uses an `actions-util`. That seems wrong, but I guess this is a workaround.

marcogario (approved):
LGTM!

The CodeQL queries are :sparkling_heart: , but I would suggest you get someone with more experience to review them.

src/actions-util.ts
robertbrignull: I agree I think this is. It's happening because in `getCodeQLActionRepository` we abort early when not in actions mode and the query doesn't understand this. I will fix that but perhaps would be best in a separate PR. I'll open an issue for it as I don't think there is one now. For this PR I think it's enough to see there's 1 new and 1 fixed alert of this type.

src/actions-util.ts
robertbrignull: I agree, although it looks like I got the message in the query the wrong way around. The alert is saying to use `getRequiredInput` and it should be `getOptionalInput`.

src/codeql.ts
robertbrignull: This code is skipped when not on actions as we return just above. But I agree this code is less than ideal and it's on my to-do list to improve the situation in this file.

src/actions-util.ts
robertbrignull: No, the query is correct. The `matrix` input has a default value, so `getRequiredInput` is correct and the type should just be `string`.

robertbrignull:
I was wondering why this wasn't opening alerts for the uses of `core.getInput` in `getRequiredInput` and `getOptionalInput` themselves. Turns out it's because the query relies on the parameter being a string literal, so it skips these cases. This is unintentional behaviour but it ends up doing the right thing, so I'm tempted to leave it. I'll add a comment into the query explaining this though.

robertbrignull:
The alert changes now look correct. One new and one fixed alert because we've moved some code to a different file.

queries/required-action-input.ql
```diff
@@ -0,0 +1,97 @@
...
+   */
+  predicate inputAlwaysHasValue(string input) {
+    exists(YAMLMapping value |
+      value = getRootNode().(YAMLMapping).lookup("inputs").(YAMLMapping).lookup(input) and
+      (exists(value.lookup("default")) or
+       value.lookup("required").(YAMLBool).getBoolValue() = true))
+  }
+
+  /**
+   * The function that is the entrypoint to this action.
```
joshhale, line 45: Technically, I guess it's the _typescript_ source for the function that is the entrypoint to the action.

...not sure if it's important to make that distinction explicit though.

joshhale (approved):
I'm not familiar with the javascript QL library, but the queries lgtm.

queries/required-action-input.ql
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * The given input always has a value, either because it is required,
+   * or because it has a default value.
+   */
+  predicate inputAlwaysHasValue(string input) {
+    exists(YAMLMapping value |
+      value = getRootNode().(YAMLMapping).lookup("inputs").(YAMLMapping).lookup(input) and
+      (exists(value.lookup("default")) or
+       value.lookup("required").(YAMLBool).getBoolValue() = true))
```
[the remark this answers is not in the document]
robertbrignull: Yes unfortunately. This is not a predicate that exists or doesn't, but rather it "returns" a true or false boolean value. In CodeQL I think you need to explicitly use `= true`.

queries/required-action-input.ql
robertbrignull: I'm not sure either. I think I'll just leave it. I think it's implicit that these queries are on the typescript source code instead of the compiled javascript, assuming that's what you meant as the alternative source for the function.

PR #215, opened 2020-09-16
Mergeback v1 to main

marcogario:
Realign `main` with `v1` after https://github.com/github/codeql-action/pull/210

I think this is now correct. I branched from `v1`, and merged `main` into it.


robertbrignull (approved):
> I think this is now correct. I branched from `v1`, and merged `main` into it.

I think I would normally do it the other way around, but the order doesn't really matter so long as both branches are ancestors. I checked locally and everything looks correct from a git graph point of view.

Since this is just a mergeback I'll merge myself now to make sure it gets in before anything else changes.

PR #216, opened 2020-09-17
Workflow to split the bundle into components

marcogario:
Manual workflow to split a codeql bundle into components.

It requires to know both the CLI version and the bundle tag. I tried to trigger this with a new release event, but that was too fragile and prone to errors. Also, right now, I think we cannot know the CLI version in a bundle without running `./codeql --version`.


.github/workflows/split.yml
```diff
@@ -0,0 +1,73 @@
...
+# Split the CodeQL Bundle into platform bundles
+#
+# Instructions:
+#  1. Upload the new codeql-bundle (codeql-bundle.tar.gz) as an asset of the
+#     release (codeql-bundle-20200826)
+#  2. Take note of the CLI Release used by the bundle (e.g., v2.2.5)
+#  3. Manually launch this workflow file (via the Actions UI) specifying
+#     - The CLI Release (e.g., v2.2.5)
+#     - The release tag (codeql-bundle-20200826)
+#  4. If everything succeeds you should 3 new assets.
```
robertbrignull, line 11:
```suggestion
#  4. If everything succeeds you should see 3 new assets.
```

.github/workflows/split.yml
```diff
@@ -0,0 +1,73 @@
+#
+# Split the CodeQL Bundle into platform bundles
+#
+# Instructions:
+#  1. Upload the new codeql-bundle (codeql-bundle.tar.gz) as an asset of the
+#     release (codeql-bundle-20200826)
+#  2. Take note of the CLI Release used by the bundle (e.g., v2.2.5)
+#  3. Manually launch this workflow file (via the Actions UI) specifying
+#     - The CLI Release (e.g., v2.2.5)
+#     - The release tag (codeql-bundle-20200826)
```
robertbrignull, line 10:
```suggestion
#     - The release tag (e.g., codeql-bundle-20200826)
```
for consistency with other input

robertbrignull (approved):
Workflow looks good to me so long as it's tested and works. The bit about finding the upload URL is unfortunately a bit hacky, but it seems fine.

As noted in person, we should move in the future to having the per-platform bundles be generated alongside the full bundle, but as the release is currently in progress that will have to wait until next time. I see no problem having this workflow here in the meantime until the codeql/bundle release process is updated.

PR #217, opened 2020-09-17
Do not always overwrite the GITHUB_REF for PRs

rneatherway:
As we move towards analysing the merge commit for pull requests by
default, we should stop sending `/refs/pull/n/head` rather than
`refs/pull/n/merge` _unless_ the checked-out SHA has actually changed.
Here we assume that any change (compared to GITHUB_SHA) indicates that
`git checkout HEAD^2` has been run earlier. This may sometimes be
incorrect (e.g. `git checkout mybranch`), but in that case the ref
would be wrong either way.

- [X] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary.


robertbrignull (approved):
LGTM

chrisgavin (approved):
I've tested this with the backend changes required to map merge commit annotations to the head commit and it worked great!

rneatherway:
Thanks!

PR #219, opened 2020-09-17
Include workflow queries in status report

robertbrignull:
I realised that queries from the workflow were not being included in the status report from the init action.


sampart (approved):
👍

robertbrignull:
Thanks for the review. I'm going to add some quick tests of this if I caan work out a good way to do it. Should be all test code though so less of a need to re-review unless you want to. Let me know.

sampart:
I'd like to take a look, but you can merge prior to my re-review. Thanks

robertbrignull:
Unfortunately writing a unit test that creates a status report is quite a pain at the moment because it relies on lots of state. We should work towards making this easier, but right now I'm not sure it's worth it for this change.
I'm running some manual tests now and I'll check that the status reports show up correctly at the other end.

PR #221, opened 2020-09-18
Update to the latest version of `@actions/github`.

chrisgavin:
This will hopefully mean we get a newer version of Octokit and mean we can use Octokit plugins.


robertbrignull (approved):
Changes to our code look behaviour preserving. However I'm confused why not update the use in `api-client.ts` to `github.getOctokit` as well? It would seem to be the intended entrypoint to the library. I think you'd need to move the auth parameter field to a parameter instead of being in the options object, but otherwise it's entirely equivalent.

chrisgavin:
That's a fair point. It feels like a slightly ugly interface but I agree it probably makes sense to use it.

PR #227, opened 2020-09-22
Prioritizing pre-downloaded CodeQL bundle in some circumstances

cbraynor:
This allows CodeQL bundles pre-included in the hosted runners to be used instead of the latest version if:

- There is only one version in the tool cache
- The version has been marked as `pinned` ([related](https://github.com/actions/virtual-environments/pull/1557))
- No explicit version has been provided with the `tools` flag (it will be honored)
- The `tools` flag is not the specific literal `latest` (which will force using the latest version)


.github/workflows/codeql.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest,windows-latest,macos-latest]
+        tools: [~, latest]
```
robertbrignull, line 13: For this workflow I'd recommend we don't run on both versions. My reasoning here is although this workflow is testing the product we're mainly doing it for the analysis results. If we want to test the parsing of the tools input, that should be done in one of the `integration-testing` workflows. If we did run on both codeql-bundle versions but they produced different sets of alerts it could be confusing too.

.github/workflows/integration-testing.yml
```diff
@@ -7,6 +7,9 @@ on:
 
 jobs:
   multi-language-repo_test-autodetect-languages:
+    strategy:
```
robertbrignull, line 10: Do we need to do this for both versions in so many cases? I'm just worried that it doubles the number of jobs being run on every PR. Could we just do this on one job and still get almost the same amount of testing?

robertbrignull (approved):
Sorry I didn't get to reviewing this yesterday like I say I would.

Code changes LGTM. Couple of comments about the test workflows, but nothing major.

.github/workflows/codeql.yml
cbraynor: That's fair - it would be confusing

.github/workflows/integration-testing.yml
cbraynor: I debated whether just one would be fine and erred on the side of completeness, but we do just need one. I decided to revert all but `multi-language-repo_test-custom-queries-and-remote-config` where we get coverage on all the platforms

PR #229, opened 2020-09-23
Resolve violations of no-sparse-arrays lint

miqh:
Hello,

These changes should remove the need to suppress the `no-sparse-arrays` lint in `eslintrc.json`.

---


robertbrignull (approved):
Don't know this array had double commas in it, but I've been through the history (including before it became open source) and it's been like this literally from the beginning.

Looking at where we use `CRITICAL_TRACER_VARS` I think this should have no behavioural change, or if it does then something very weird and confusing is going on.

PR #231, opened 2020-09-25
Enable the CodeQL SARIF multi-cause markdown flag.

chrisgavin:
This enables the CodeQL multi-cause markdown flag to allow for better display of alerts that have multiple related locations.

This can't currently be merged as the multi-cause markdown feature is not yet present in the CodeQL bundle we are using, but after the next CodeQL upgrade we should enable this flag.


chrisgavin:
The CI now passes as the new CodeQL bundle is present on the runners in Ring 0.

I don't think it's necessarily able to be merged yet though as the image might not yet have reached runners outside of Ring 0.

robertbrignull (approved):

PR #233, opened 2020-09-28
Resolve violations of no-useless-escape lint

miqh:
Hello,

These changes should remove the need to suppress the `no-useless-escape` lint in `eslintrc.json`.

I guess worth pointing out is the fact most characters are interpreted literally inside a character class regular expression rule, which is why this lint was complaining.

---


cbraynor (approved):
👍🏼 thanks!

PR #235, opened 2020-09-28
Make URL parsing more robust, and handle subdomain isolation

robertbrignull:
This PR makes the runner's parsing of the URL a bit more robust and should make it handle the case where GHES is using [subdomain isolation](https://docs.github.com/en/enterprise/2.19/admin/configuration/enabling-subdomain-isolation).

I need to test that this actually works and I will do so before merging, but from my prelimanary experimentation it looks like it should do as the `/api/v3` route on the main GHES domain still seems to work. So we don't need to be able to reconstruct the isolated subdomain.

The other changes to the URL parsing make it so if you don't pass a scheme then it'll assume it's `https`. Also makes the output URL more consistent.


chrisgavin (approved):
Looks like a nice improvement. :+1:

robertbrignull:
I've verified manually that the `/api/v3` path still seems to be available when subdomain isolation is turned on, so that means applications like ours can go from the main web URL to the API URL without having to know if subdomain isolation is turned on or not.

With the unit tests added in this PR I'm reasonable confident the URL parsing is working as expected. So this should all work fine.

robertbrignull:
This PR accidentally broke the break-out clause in `getCodeQLBundleDownloadURL`. By using `GITHUB_DOTCOM_URL` from `parseGithubUrl` we hopefully avoid that. Sorry for the false starts on this and force pushes. I've left in the trailing slash normalisation for non-dotcom URLs, but notably `GITHUB_DOTCOM_URL` does not have one as that constant matches the URL given to us by the hosted actions runners. The situation is now a little bit inconsistent but hopefully acceptable.

@chrisgavin assuming the tests now pass, it's still probably best if you check that last commit just to make I've not done anything silly.

chrisgavin (approved):
Thanks! Yes, it looks right to me and the log output indicates it's following the short-circuit path so :+1: from me.

PR #236, opened 2020-09-28
Exclude the temporary directory from scanning.

chrisgavin:
This fixes the case where the CodeQL Runner is downloaded on-the-fly to the source directory of the repository being analyzed. It gets copied to the toolcache which is outside of the working directory, but still remains in the repository directory too. We could delete it from the repository directory after copying it to the toolcache but a more robust solution seems like it would just be to ignore the temporary directory completely.

It's not ideal that people download the runner in this way because even compiled languages could take issue with unexpected files in the repository but it seems like people do so we should handle it as well as we can.


src/analysis-paths.ts
```diff
@@ -48,10 +52,17 @@ export function includeAndExcludeAnalysisPaths(config: configUtils.Config) {
...
-    process.env["LGTM_INDEX_EXCLUDE"] = buildIncludeExcludeEnvVar(
-      config.pathsIgnore
+  // If the temporary directory is in the working directory ignore that too.
+  const tempRelativeToWorking = path.relative(process.cwd(), config.tempDir);
+  let pathsIgnore = config.pathsIgnore;
+  if (!tempRelativeToWorking.startsWith("..")) {
+    logger.warning(
+      "Storing the CodeQL Runner in the directory being analyzed is not recommended."
     );
+    pathsIgnore = pathsIgnore.concat(config.tempDir);
```
robertbrignull, line 62: I think you want to add `tempRelativeToWorking` here. These paths are relative to the cwd.

robertbrignull (commented):
Looks generally good to me with one comment. I want to test it to confirm my understanding of the behaviour. Should get that testing done today or tomorrow.

src/analysis-paths.ts
chrisgavin: Ah okay. At least with the JavaScript extractor I had no trouble with an absolute path, but if we're using relative elsewhere that does seem safer in case other extractors are less forgiving.

robertbrignull: I think the python extractor is more picky and only works with relative paths. That's what I remember. There's an issue on the CodeQL side to make that behaviour more consistent.

src/analysis-paths.ts
```diff
@@ -48,10 +52,17 @@ export function includeAndExcludeAnalysisPaths(config: configUtils.Config) {
   if (config.paths.length !== 0) {
     process.env["LGTM_INDEX_INCLUDE"] = buildIncludeExcludeEnvVar(config.paths);
   }
-  if (config.pathsIgnore.length !== 0) {
-    process.env["LGTM_INDEX_EXCLUDE"] = buildIncludeExcludeEnvVar(
-      config.pathsIgnore
+  // If the temporary directory is in the working directory ignore that too.
```
robertbrignull, line 55: Can you do the same thing but with the toolcache dir as well. You can also get that from the `config` object.

src/analysis-paths.ts
```diff
@@ -48,10 +52,17 @@ export function includeAndExcludeAnalysisPaths(config: configUtils.Config) {
...
+  const tempRelativeToWorking = path.relative(process.cwd(), config.tempDir);
+  let pathsIgnore = config.pathsIgnore;
+  if (!tempRelativeToWorking.startsWith("..")) {
+    logger.warning(
```
robertbrignull, line 59: I don't think this is accurate. See my main review comment but I've reproduced this without the runner executable being in the same directory. It's to with where the temp dir gets created which is by default relative to the cwd.

If we're able to work around it by ignoring the directory, then perhaps this warning isn't necessary.

robertbrignull (commented):
I've reproduced the bug with the `codeql-runner-linux` executable not in the cwd or the directory of the repository. As noted earlier the issue is with the temporary directory being in the current directory. By default the temp directory is `./codeql-runner` relative to the cwd, i.e. where the runner executable was run from but not necessarily where it is located.

Something else I realised when testing this is this bug won't happen every time. The files being detected are those from the downloaded and unpacked bundle. This only happens if we don't find the bundle in the toolcache, and hence we download a new version to the temp dir and copy it to the toolcache. So on the next run the temp dir is wiped, but the bundle is not re-downloaded because we already have one in the toolcache that is no wiped.

You could get a similar problem if the toolcache is located in the cwd. The toolcache by default is created in the home directory, so this could happen if for some reason analysis is being run directly from the home directory.

On the plus side, I've tested using a runner built from this PR and it wokred. I tested it for javascript and python.

src/analysis-paths.ts
chrisgavin: Ah, okay. I agree if if will always happen regardless of where you put the runner then the warning probably doesn't make sense.

robertbrignull (approved):
Sorry for the delay. I've done another quick test of this to make sure. Everything looks good and it's working perfectly.

chrisgavin:
No worries! I'm glad it all works.

PR #237, opened 2020-09-29
Resolve violations of sort-imports lint

cbraynor:
These changes should remove the need to suppress the sort-imports lint in eslintrc.json.


robertbrignull (commented):
I note we also have `"sort-imports": ["error", { "allowSeparatedGroups": true }]`. What was the reason for adding that configuration? Do we want to remove that?

Previously the imports were sorted by those from external sources first and then those from local files, which doesn't appear to be a thing supported by the eslint rule. This is not necessarily a problem but I don't think we should try to emulate that with `allowSeparatedGroups`. If we want that we should look for a rule that can enforce it for us.

Overally I'm not hugely convinced by this rule as I can't see why sorting by the _type_ of import (i.e. importing all members or just one) is helpful, especially when it means that the import sources are no longer alphabetical.

cbraynor:
I switched to `import/order` instead which seems to be more aligned with what we want and has the added benefit of implementing the `--fix` functionality

robertbrignull (approved):

PR #238, opened 2020-09-29
Resolve violations of import/no-extraneous-dependencies lint

cbraynor:
These changes should remove the need to suppress the import/no-extraneous-dependencies lint in eslintrc.json.

We were relying on a dependency happening to also use this, so this makes it explicit and robust. We also switch to using the named export as the deep require has been [deprecated since](https://www.npmjs.com/package/uuid#deep-requires-now-deprecated)

Predicated on #237 to avoid merge conflicts ([see diff just for these changes](https://github.com/github/codeql-action/compare/cbraynor/fix206...cbraynor/fix201))


robertbrignull (commented):
LGTM (by which I mean the changes on top of #327)

robertbrignull (approved):

PR #240, opened 2020-09-30
Platform specific bundle

marcogario:
Download the platform specific bundle. These bundles are generated via the split workflow https://github.com/github/codeql-action/blob/main/.github/workflows/split.yml 

If the tool url is provided, we will still respect that. However, in other cases we download `codeql-bundle-${platform}.tar.gz`


robertbrignull (approved):
I think this is all good. I've checked that it still uses a pinned version from the environment in preference to downloading the platform specific pack. And it still uses a user-supplied download URL if given.

As noted in person earlier, this now relies on all the platform specific versions being present. But this will be true for dotcom, and I believe you checked that the sync tool copies all artefacts. So this requirement should be fine.

marcogario:
> As noted in person earlier, this now relies on all the platform specific versions being present. But this will be true for dotcom, and I believe you checked that the sync tool copies all artefacts. So this requirement should be fine.

Yes, this is worth underlying one more time: the platform bundles are now a requirement.

Indeed, I did verify that the sync tool supports this, and did also update the bundle release process accordingly.

PR #253, opened 2020-10-05
Fix making too many requests and triggering the rate limit.

chrisgavin:
This hopefully fixes the fact we are no longer hitting the short circuiting logic due to `githubUrl` having a trailing slash, whereas `util.GITHUB_DOTCOM_URL` does not.


chrisgavin:
Fixed by 2466f2a6a0ab4398f72fb6fc8bb120c6214b0a75.

PR #254, opened 2020-10-05
Provide API token to downloads via GitHub API

henrymercer:
Previously, the token was not provided for downloads via $GITHUB_API_URL and therefore downloads of releases in private repositories were susceptible to failing with a 404.


robertbrignull:
Can you give an example of the download URL you were providing? I'm confused why it started with the API url. I would have expected it to look something like https://github.com/github/codeql-action/releases/download/codeql-bundle-20200630/codeql-bundle.tar.gz

henrymercer:
An example is `https://github.com/some-org/some-private-repo/releases/download/codeql-bundle-20200630/codeql-bundle.tar.gz`.

Looking at the `getCodeQLBundleDownloadURL` function in `src/codeql.ts`, if we're not downloading from `github.com/github/codeql-action`, then we will try to get the asset URL using the GitHub API.  These asset URLs from the GitHub API are typically prefixed by `$GITHUB_API_URL`.

My understanding is that getting the asset URL using the GitHub API is necessary to download an asset from a release on a private repository, the `https://github.com/some-org/some-private-repo/releases/download/codeql-bundle-20200630/codeql-bundle.tar.gz` URL will 404 even with an API token — it only works in the browser.

PR #255, opened 2020-10-06
Guard python deps install

Daverlo:
With this PR we will only try to install the python deps if a python analysis will be executed


marcogario (approved):
LGTM

PR #257, opened 2020-10-06
Port python deps setup to windows

Daverlo:
Closes https://github.com/github/code-scanning/issues/2108

This PR ports the auto install python deps functionality so it also works on windows workers.

I tested this branch with the django example and it worked as expected, providing two alerts:
https://github.com/dsp-testing/daverlo-djanjo-test/actions/runs/291392369
https://github.com/dsp-testing/daverlo-djanjo-test/security/code-scanning

@RasmusWL would you mind taking a look at the python scripts? I created a copy of `auto-install-packages` and modified it so it works on windows.


RasmusWL (changes requested):
I would rather have seen `auto_install_packages.py` being adapted to work on both Windows and Linux instead of duplicating the script and making adjustments (using [sys.platform](https://docs.python.org/3.8/library/sys.html#sys.platform) or [os.name](https://docs.python.org/3.8/library/os.html#os.name)). I really think that is the right approach.

As an example of why this is important, the version of the linux script you ported is not the most up to date (compare with https://github.com/github/codeql-action/blob/main/python-setup/auto_install_packages.py)

I also don't understand the need to strip out the first 3 characters when using pipenv and poetry, can you explain?

Daverlo:
> I also don't understand the need to strip out the first 3 characters when using pipenv and poetry, can you explain?

They were producing a path like
`/d/a/codeql-python-autobuild-playground/virtualenvs/autoinstall-test-FP1YIZRL-py2.7/Scripts/python`

and windows didn't like the `/d` prefix, it complained about the file not existing. 🤷

python-setup/auto_install_packages.py
```diff
@@ -137,7 +162,10 @@ def install_packages(codeql_base_dir) -> Optional[str]:
 
     # The binaries for packages installed with `pip install --user` are not available on
     # PATH by default, so we need to manually add them.
-    os.environ['PATH'] = os.path.expanduser('~/.local/bin') + os.pathsep + os.environ['PATH']
+    if sys.platform.startswith('win32'):
+        os.environ['PATH'] = os.path.expandvars('%APPDATA%\Python\\Python38\\scripts') + os.pathsep + os.environ['PATH']
```
RasmusWL, line 166: This hardcoded Python version (3.8) looks a bit scary to me. Is there a way to _not_ hardcode the version? :grimacing:

RasmusWL (commented):
Nice, this version looks much better, good job :1st_place_medal: 

There is an issue with hardcoding version of python, but otherwise this looks good to me!

python-setup/auto_install_packages.py
Daverlo: That is the default version of Python in Windows workers and it's always there, no matter if you add the setup-python action with like 2.7 or 3.5 or whatever, in fact, that action is not needed at all. I

 prefer to leave it like this if you don't mind, right now it is working for all the cases and I don't think its the best use of my time at this moment.

https://github.com/dsp-testing/codeql-python-autobuild-playground/runs/1213872774?check_suite_focus=true
https://github.com/dsp-testing/codeql-python-autobuild-playground/runs/1221157769?check_suite_focus=true
https://github.com/dsp-testing/codeql-python-autobuild-playground/runs/1221281906?check_suite_focus=true

python-setup/auto_install_packages.py
```diff
@@ -36,10 +38,14 @@ def install_packages_with_poetry():
     poetry_out = _check_output(['poetry', 'run', 'which', 'python'])
     python_executable_path = poetry_out.decode('utf-8').splitlines()[-1]
 
+    if sys.platform.startswith('win32'):
+        python_executable_path = python_executable_path[2:]
```
aibaars, line 42: Could you explain the purpose of chopping off the start of the `python_executable_path`? If it is to get rid off the drive letter (eg  `C:` ) then this code is likely wrong.

python-setup/auto_install_packages.py
```diff
@@ -36,10 +38,14 @@ def install_packages_with_poetry():
...
     return python_executable_path
 
 
 def install_packages_with_pipenv():
+    if sys.platform.startswith('win32'):
+        os.environ['WORKON_HOME'] = os.environ['RUNNER_WORKSPACE'] + '\\virtualenvs'
```
aibaars, line 48:
```suggestion
        os.environ['WORKON_HOME'] = os.path.join(os.environ['RUNNER_WORKSPACE'], 'virtualenvs')
```

python-setup/auto_install_packages.py
```diff
@@ -48,29 +54,43 @@ def install_packages_with_pipenv():
     pipenv_out = _check_output(['pipenv', 'run', 'which', 'python'])
     python_executable_path = pipenv_out.decode('utf-8').splitlines()[-1]
 
+    if sys.platform.startswith('win32'):
+        python_executable_path = python_executable_path[2:]
```
aibaars, line 58: This looks strange to me.

python-setup/auto_install_packages.py
```diff
@@ -48,29 +54,43 @@ def install_packages_with_pipenv():
...
     return python_executable_path
 
 
 def _create_venv(version: int):
     # create temporary directory ... that just lives "forever"
-    venv_path = mkdtemp(prefix='codeql-action-python-autoinstall-')
+    venv_path = os.environ['RUNNER_WORKSPACE']+'/codeql-action-python-autoinstall'
+    print ("Creating venv in "+venv_path, flush = True)
```
aibaars, line 65:
```suggestion
    venv_path = os.path.join(os.environ['RUNNER_WORKSPACE'], 'codeql-action-python-autoinstall')
    print ("Creating venv in " + venv_path, flush = True)
```

python-setup/auto_install_packages.py
aibaars: 3.8 is "always" there until someone updates the virtual machine image. 3.9 was just released, so it may change sooner than you think. You should be able to get the USER_BASE directory from the `site` module. See also https://docs.python.org/3/library/site.html#site.USER_BASE

src/init.ts
```diff
@@ -186,21 +186,28 @@ export async function injectWindowsTracer(
 export async function installPythonDeps(codeql: CodeQL, logger: Logger) {
   logger.startGroup("Setup Python dependencies");
 
-  if (process.platform !== "linux") {
+  if (process.platform === "darwin") {
     logger.info(
-      "Currently, auto-installing python dependancies is only supported on linux"
+      "Currently, auto-installing python dependancies is not supported on MacOS"
```
aibaars, line 191:
```suggestion
      "Currently, auto-installing python dependencies is not supported on MacOS"
```

aibaars: Does it really not work, or have you simply not tested it yet?

python-setup/auto_install_packages.py
aibaars: If I understand correctly you should also be able to set [PYTHONUSERBASE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUSERBASE) to have full control over the install location. You could consider stuffing things in the RUNNER_TEMP folder to avoid interfering with future other jobs on self_hosted workers.

python-setup/auto_install_packages.py
Daverlo: Poetry and pipenv display the path wrongly. They produce `/d/a/codeql-python-autobuild-playground/virtualenvs/requests-2-z_EHYAbo/Scripts/python` instead of `D:\a\codeql-python-autobuild-playground\codeql-action-python-autoinstall\Scripts\python`. Windows doesn't like that way of specifying the drive letter. I completely removed because it is not needed as everything is in the same drive (We are installing the dependencies in the RUNNER_WORKSPACE)

python-setup/auto_install_packages.py
Daverlo: See the other comment

python-setup/auto_install_packages.py
RasmusWL: since I also asked about this, can we maybe just add that comment to the code? :blush:

python-setup/auto_install_packages.py
RasmusWL: I think it's a good idea @aibaars is pointing out, and I'm probably to blame for not thinking about this in the first place.

python-setup/auto_install_packages.py
aibaars: The `/d` silly paths are due to cygwin or msys. You might want to check whether the paths are bad before chopping off letters. You never know when they'll replace `pipenv` and `poetry` with more windows-like versions.

aibaars: I wonder why `poetry` or `pipenv` would behave this way. These are python programs. Are you accidentally running them with a non-windows (ie. cygwin/msys) python binary?

aibaars: Perhaps you should invoke `poetry` like `py -3 -m poetry` or similar or even use `sys.executable` like `[sys.executable like , '-m', 'poetry', 'run', 'which', 'python']`

python-setup/auto_install_packages.py
RasmusWL: I think @aibaars added [this](https://github.com/github/codeql-action/pull/257#discussion_r501668667) comment to the wrong thread:

> Perhaps you should invoke `poetry` like `py -3 -m poetry` or similar or even use `sys.executable` like `[sys.executable like , '-m', 'poetry', 'run', 'which', 'python']`

For solving this problem, I think using `sys.executable` instead of hardcoding the 3.8 path is a very good trade-off.

We just need to make sure that we only invoke this script with Python 3, since we only install pipenv/poetry for Python 3. (but I guess the JS code already does ensures this, right?)

Daverlo: I wasn't able to make it work with the sys.executable
```
    _check_call([sys.executable like , '-m', 'poetry', 'install', '--no-root'])
                                ^
SyntaxError: invalid syntax
```

But I managed to make it work with the `py -3`!!

> We just need to make sure that we only invoke this script with Python 3, since we only install pipenv/poetry for Python 3. (but I guess the JS code already does ensures this, right?)


Yep, it is always called with `py -3`

python-setup/auto_install_packages.py
Daverlo: I'm now calling it with `py -3 -m poetry` and is still producing the same output, so I still need to chop off the start of the string

Daverlo:
@aibaars @RasmusWL I got this working using the `py -3 -m poetry`/`py -3 -m pipenv` calls, so no more hardcoded python version. Could you re-review the PR please?

python-setup/auto_install_packages.py
RasmusWL: Real code without syntax error would be 

```
_check_call([sys.executable, '-m', 'poetry', 'install', '--no-root'])
```

python-setup/auto_install_packages.py
```diff
@@ -137,7 +164,8 @@ def install_packages(codeql_base_dir) -> Optional[str]:
 
     # The binaries for packages installed with `pip install --user` are not available on
     # PATH by default, so we need to manually add them.
-    os.environ['PATH'] = os.path.expanduser('~/.local/bin') + os.pathsep + os.environ['PATH']
+    if sys.platform.startswith('linux'):
+        os.environ['PATH'] = os.path.expanduser('~/.local/bin') + os.pathsep + os.environ['PATH']
```
RasmusWL, line 168: Don't think we need this anymore than that we're using `[sys.executable, '-m'` for `pipenv` and `poetry`. 

```suggestion
```

python-setup/auto_install_packages.py
```diff
@@ -23,8 +23,11 @@ def _check_output(command):
 
 
 def install_packages_with_poetry():
+    command = [sys.executable, '-m', 'poetry']
+    if sys.platform.startswith('win32'):
+        os.environ['POETRY_VIRTUALENVS_PATH'] = os.path.join(os.environ['RUNNER_WORKSPACE'], 'virtualenvs')
```
RasmusWL, line 28: Could you add a comment (to the code) explaining if this is _required_ for things to work, or if it's just to behave nicely? (if it's only for behaving nicely, shouldn't we do it on linux as well?)

python-setup/auto_install_packages.py
```diff
@@ -33,44 +36,63 @@ def install_packages_with_poetry():
     # virtualenv for the package, which was the case for using poetry for Python 2 when
     # default system interpreter was Python 3 :/
 
-    poetry_out = _check_output(['poetry', 'run', 'which', 'python'])
+    poetry_out = _check_output(command + ['run', 'which', 'python'])
     python_executable_path = poetry_out.decode('utf-8').splitlines()[-1]
 
+    if sys.platform.startswith('win32'):
+        python_executable_path = python_executable_path[2:]
```
RasmusWL, line 43: I would still very much like having a comment inside the code explaining why this is needed. (I know you've explained it here on the PR, but that knowledge might be lost if it's not present in the code).

RasmusWL (changes requested):
A few minor nitpicks, but generally I won't be against merging it now :blush:

RasmusWL (approved):
Nice, looks good to me now :+1:

Thanks for working through all the things :muscle:

PR #260, opened 2020-10-08
Update default CodeQL version to codeql-bundle-20201008

rneatherway:

### DIFF
--- a/lib/defaults.json
+++ b/lib/defaults.json
@@ -1,3 +1,3 @@
 {
-    "bundleVersion": "codeql-bundle-20200826"
+    "bundleVersion": "codeql-bundle-20201008"
 }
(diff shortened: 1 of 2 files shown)


robertbrignull:
Release LGTM. Obviously all the tests pass, I've checked the logs and queries are not being compiled from scratch, and I've downloaded the artefact and given it a look too.

I am surprised that the codeql-runner artefacts have changed size. Down to 45MB from 70MB. I'm not sure why this is and they're still 70MB if I build them on my machine. The ones I download from the release still seem to work though.

robertbrignull:
I added a workflow to build and upload the runner artefacts: https://github.com/github/codeql-action/runs/1246905697?check_suite_focus=true

So they are 45MB when built on actions. Our hypothesis is that it's the node version. I have v13 on my machine here, whereas @rneatherway and actions are on v12, so this could be the reason. Regardless, I don't think this needs to hold up this release.

robertbrignull (approved):

PR #261, opened 2020-10-08
Don't use ::set-env in python-setup

RasmusWL:
Is now deprecated as described in https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

@Daverlo I don't know how this plays together with version compatibility, just spotted the warning on an action run, and thought I could fix it up.


Daverlo (approved):
Thanks for changing this!

PR #266, opened 2020-10-13
Add a workflow to build and upload the runner

robertbrignull:
Instead of building the runner locally we should probably standardise on building it within actions to get a consistent environment. This PR introduces a workflow that just builds the runner and upload the artefacts.

I'm intending this currently state as just a starting point. A couple of open questions I have are:
- Should this run on `push` or just `workflow_dispatch`? I only added `push` so I could test it before merging to the default branch.
- Should we merge this with `split.yml`? The primary use case for this is when doing a release, but it could be useful to be able to run separatetly too.

@rneatherway


rneatherway:
I suggest just using `workflow_dispatch` and adding the ability for it to upload the artefacts directly to the specified release like this: https://github.com/github/codeql-action/blob/ffaa8aa19736375a41c680e6fc725b5cb59df03e/.github/workflows/split.yml#L65-L73

Keep it separate though as `split.yml` will be able to go away when we start preparing those platform-specific bundles ahead of time.

rneatherway:
I've tested this using the `on: push` trigger and a hardcoded `RELEASE_TAG`, which I then removed in the last commit.

* https://github.com/github/codeql-action/actions/runs/317389622
* https://github.com/github/codeql-action/releases/tag/test-build-runner

Once this is merged I will delete the test release.

.github/workflows/build-runner.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+    env:
+      RELEASE_TAG: "${{ github.event.inputs.bundle-tag }}"
+
+    strategy:
+      matrix:
+        extension: ["linux", "macos", "win.exe"]
+
+    steps:
+      - uses: actions/checkout@v2
```
chrisgavin, line 21:
```suggestion
      - uses: actions/checkout@v2
        with:
          ref: refs/tags/${{ github.event.inputs.bundle-tag }}
```

It's potentially confusing to build the code from `main` and upload it as an artifact on a tagged release.

(It would be nice if we could just use the workflow from the release tag too, but workflows can't be dispatched on tags currently.)

chrisgavin (approved):
One small change, but otherwise it looks good to me.

robertbrignull:
Thanks for taking on the rest of this @rneatherway.

.github/workflows/build-runner.yml
rneatherway: I've followed Robert's suggestion to make the bundle-tag optional, which now makes this a bit awkward. Do you see a way apart from having two calls to `actions/checkout@2` each with an `if`?

.github/workflows/build-runner.yml
```diff
@@ -4,31 +4,45 @@ on:
   push:
   workflow_dispatch:
     inputs:
+      bundle-tag:
+        description: 'Tag of the bundle release (e.g., "codeql-bundle-20200826")'
+        required: true
```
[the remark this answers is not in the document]
rneatherway: I've done this, just waiting for actions to come back to finish testing.

.github/workflows/build-runner.yml
rneatherway: In the interests of getting a version of this ready I am going to merge. We can later update this if we see a way to resolve the inconsistency.

PR #271, opened 2020-10-22
Fail the analyze action when some language fails to run the queries

Daverlo:
In [analyze:runQueries](https://github.com/github/codeql-action/blob/6ac5978b443875154367ac20221468a766c4901b/src/analyze.ts#L193) we are warding the codeql call with a try, and if we catch something we set an error message, update the status report with the failed language and returning it. Just after that in runAnalyze we are always doing the upload without considering if the runQueries failed or not. This caused that we were uploading partial results, and that doesn't make sense. Partial results would mean that alerts could disappear without being fixed.

The reason why the action is not marked as failed is because this is only done at the topmost catch, in `analyze-action`, and we are not bubbling up the exception, just catching it and setting the status report.

As we want to keep the status report, I've created an `Error` subclass where we can attach it and submit it as a failed status report.

Here is a run with the old behaviour: https://github.com/dsp-testing/daverlo-test-fail-analyze/actions/runs/321647298
The job is green but it has an error saying the analysis failed.

Here is the same repo but it run with this branch: https://github.com/dsp-testing/daverlo-test-fail-analyze/actions/runs/321654493


robertbrignull (approved):
LGTM

Thanks for including the link to the test run.

PR #272, opened 2020-10-22
Fix alignment of cron comment

sonicdoe:
With https://github.com/github/codeql-action/pull/259 changing the minute field from `0` to `30`, the alignment of the comment above broke.


robertbrignull (approved):
Thanks

PR #281, opened 2020-10-29
Add the --threads config option to finalize db

aeisenberg:
This flag is already being used for `runQueries`, so let's use it for
finalize as well.


aeisenberg:
I have not updated the README yet. I'll handle that soon. Also, any sort of unit tests or integration tests I should be adding or updating?

aeisenberg:
Also, there's a failing `Integration Testing / runner-upload-sarif`. I don't understand why.

robertbrignull:
> Also, there's a failing `Integration Testing / runner-upload-sarif`. I don't understand why.

That's because you opened your PR from a fork and so the actions token doesn't have write scopes. You can just ignore the failure for this PR.

robertbrignull (approved):
I see no problem with this. I trust you that passing this flag to this command makes sense.

No need to update the readme or tests for this change. If the existing integration tests pass then that should be enough.

aeisenberg:
Thanks! Will resolve conflicts and rebase.

PR #289, opened 2020-11-04
Fix invalid API call when running locally

simon-engledew:
The method `getAnalysisKey` would call `getWorkflowPath` and raise an API Error.

This change follows the pattern in `prepareLocalRunEnvironment` to set a dummy value for the required environment variable, therefore shortcutting the API request.

I have tried to mirror the existing mechanism for stubbing out `GITHUB_JOB` as closely as possible.


src/actions-util.test.ts
```diff
@@ -33,30 +33,59 @@ test("getRef() returns head PR ref if GITHUB_SHA not currently checked out", asy
...
+  process.env.CODEQL_ACTION_ANALYSIS_KEY = "";
+  process.env.GITHUB_JOB = "";
+
+  actionsutil.prepareLocalRunEnvironment();
+
+  const actualAnalysisKey = await actionsutil.getAnalysisKey();
+
+  t.deepEqual(actualAnalysisKey, "LOCAL-RUN:UNKNOWN-JOB");
+
+  process.env.CODEQL_LOCAL_RUN = origLocalRun;
```
robertbrignull, line 49: The env is reset between tests, so setting this back shouldn't be necessary. It's a bit hacky but look in `testing-utils.ts` for the implementation.

Although I realise the test below this also has this pattern. It should be safe to remove `origLocalRun` from both of these tests.

robertbrignull (commented):
Looks reasonable. So just to confirm, was it previously failing with a "Invalid API call in local run" error?

simon-engledew:
> Looks reasonable. So just to confirm, was it previously failing with a "Invalid API call in local run" error?

Yes, exactly! 👍 Any attempts to access the API when local run is turned on automatically raise an Error.

src/actions-util.test.ts
simon-engledew: Ah nice! 🧹  I have removed the unnecessary lines.

robertbrignull (approved):
LGTM

PR #291, opened 2020-11-04
Add a codeql-path output to the init action

simon-engledew:
I've added a bare bones implemention of the codeql-path output variable. Opening this PR to get feedback and check it is aligned with expectations. :+1:

- [x] Expand readme documentation
- [x] See if there is a better way of testing other than implicitly via the python-deps workflow (– added a second use of the output. Happy that the build will break if there is a regression.)


.github/workflows/python-deps.yml
```diff
@@ -62,8 +63,7 @@ jobs:
         esac
         echo ${basePath}
 
-        codeql_version="0.0.0-$(cat "$GITHUB_WORKSPACE/src/defaults.json" | jq -r .bundleVersion | rev | cut -d - -f 1 | rev)"
-        $GITHUB_WORKSPACE/python-setup/auto_install_packages.py "${basePath}/hostedtoolcache/CodeQL/$codeql_version/x64/codeql"
+        $GITHUB_WORKSPACE/python-setup/auto_install_packages.py "$(dirname ${{steps.init.outputs.codeql-path}})"
```
chrisgavin, line 66: :+1: Good idea to use it here. It's a nice built-in test for it.

init/action.yml
```diff
@@ -23,6 +23,9 @@ inputs:
     description: Try to auto-install your python dependencies
     required: true
     default: 'true'
+outputs:
+  codeql-path:
+      description: The path of the CodeQL binary used for analysis
```
chrisgavin, line 28:
```suggestion
    description: The path of the CodeQL binary used for analysis
```

Just to make the indentation consistent with the rest of the file.

src/init-action.ts
```diff
@@ -191,6 +191,8 @@ async function run() {
     return;
   }
   await sendSuccessStatusReport(startedAt, config);
+
+  core.setOutput("codeql-path", config.codeQLCmd);
```
chrisgavin, line 195: It might be worth moving this in to the final `try` block, just so it's still reported as a failure in the really unlikely event it ever crashes.

chrisgavin (approved):
:+1: Awesome. It ended up being an even smaller change than I hoped and it seems like people will find it useful.

init/action.yml
simon-engledew: Oh, nice catch! 👍 Need to set up vscode better 🤔

.github/workflows/python-deps.yml
simon-engledew: All credit goes to @marcogario who pointed out this usage! 🙇

src/init-action.ts
simon-engledew: Ah, good idea! :+1: Under the hood this appears to be implemented as an [object instantiation and a stdout write](https://github.com/actions/toolkit/blob/af821474235d3c5e1f49cee7c6cf636abb0874c4/packages/core/src/command.ts#L23) – it is possible it could go awry.

simon-engledew:
I can't see an obvious way to test this feature other than the pair of integration tests I've added. The build will break if this feature stops working for any reason - is that good enough coverage?

I'm not super happy with the [README changes](https://github.com/github/codeql-action/blob/c87f3021d45bb310d6736f884852a1a3c8093ad7/README.md). It feels like something is missing or the tone doesn't quite match the copy above. Happy to take suggestions for better wording. ✍️

.github/workflows/codeql.yml
```diff
@@ -15,7 +15,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: ./init
+      id: init
       with:
         languages: javascript
         config-file: ./.github/codeql/codeql-config.yml
+    # example from README.md
+    - name: Print CodeQL Version
+      run: ${{steps.init.outputs.codeql-path}} version --format=json
```
marcogario, line 24: This should error if `${{steps.init.outputs.codeql-path}}` is not defined, or returns the empty string, right? If so, this is good enough as coverage.

README.md
```diff
@@ -125,6 +125,24 @@ By default, this will override any queries specified in a config file. If you wi
     queries: +<local-or-remote-query>,<another-query>
 ```
 
+### Calling CodeQL Directly
```
marcogario, line 128: I am not sure how we are keeping the balance between the docs and the README. I would expect this to be an "advanced" use of the tooling, and thus prefer to put it only in the docs.

simon-engledew: Sounds sensible to me! We always have the examples in two of our workflows - if someone is likely to need to use this they would probably notice these and be able to infer how it works. ⚙️

.github/workflows/codeql.yml
simon-engledew: Ok, great! I think I'll consider this tested then and check that off as done - thanks! ⭐

chrisgavin (approved):
Yep, I can see this is probably hard to test directly without it just being an assertion that `codeQLCmd` is correct, which is already tested in a bunch of other places.

Having the test in the workflow is pretty nice, even if it's not very explicit about what it's testing.

.github/workflows/codeql.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       with:
         languages: javascript
         config-file: ./.github/codeql/codeql-config.yml
-    # example from README.md
+    # confirm steps.init.outputs.codeql-path points to the codeql binary
```
simon-engledew, line 22: I have left a comment of intent to hopefully make it more clear why this step is in the workflow :+1:

PR #298, opened 2020-11-10
Make use of the `GITHUB_ACTION_REPOSITORY` environment variable if it is set.

chrisgavin:
Now that the change in actions/runner#585 is merged it should be possible to make use of the new `GITHUB_ACTION_REPOSITORY` environment variable, rather than relying on the more brittle guess at the name of the Action being used.

Unfortunately we can't fully remove the logic where we guess the Action repository name by the path at which it is checked out because users of GHES 2.22 might still need this, but it does mean we can put a final sunset date on the removal of this code once 2.22 reaches end-of-support.

I've also added some info-level logging in the hopefully rare case that `GITHUB_ACTION_REPOSITORY` is not set in order to help debug any issues caused by the guessing logic.


robertbrignull (approved):
LGTM

I agree we have to keep the compatibility code for now. Thanks for adding instructions into the code about when it can go.

PR #301, opened 2020-11-11
Send action ref and tool version in status reports

robertbrignull:
Requires some changes on the server side before this will work / can be merged.

This sends the action ref, and the tools input and resolved version as part of the status reports. Getting the resolved tools version in a nice format proved slightly awkward, but hopefully the result in this PR isn't too bad.


robertbrignull:
The failed checkrun is just some alert churn because I changed `src/codeql.ts` too close to an alert.

I checked the data that's coming through and the `action_ref` was set to `v2` which I found odd. Especially as in the workflows on this repo we run the action from a local copy, so it's unclear what the ref would be. Turns out the ref value is coming from `actions/checkout@v2` which was the previous step to run in the workflow. I've added a check to work around this in this PR and I'll make sure it's working. I've also opened https://github.com/actions/runner/issues/803 because I think this behaviour is a bug in the actions runner.

robertbrignull:
Can confirm the `action_ref` is no longer being sent when running as a local action (there's probably a better name for this than "local")

chrisgavin (approved):
LGTM. I was a bit worried it might not be compatible with 2.22, but we don't send status reports on Enterprise so it seems like it should be fine. :+1:

PR #302, opened 2020-11-13
Don't use the word "trace" in log messages

robertbrignull:
@shati-patel, does this wording look ok to you? The basic suggestion was to use "monitoring" instead, and I can't think of anything better.

Happy for @shati-patel to approve, or alternatively @sampart if you could give official approval.


init/action.yml
```diff
@@ -1,5 +1,5 @@
 name: 'CodeQL: Init'
-description: 'Setup the CodeQL tracer'
+description: 'Setup CodeQL'
```
sampart, line 2:
```suggestion
description: 'Set up CodeQL'
```

"setup" is a noun, and "set up" is a verb, although I think I'm in the minority in recognising this distinction...

sampart (approved):

init/action.yml
robertbrignull: I agree

lib/runner.js
```diff
@@ -133,7 +133,7 @@ program
                 .join("\n");
             fs.writeFileSync(powershellEnvFile, powershellEnvFileContents);
             logger.info(`\nCodeQL environment output to "${jsonEnvFile}", "${batEnvFile}" and "${powershellEnvFile}". ` +
-                `Please export these variables to future processes so the build can be traced. ` +
+                `Please export these variables to future processes so the build can be monitored by CodeQL. ` +
```
shati-patel, line 136: Very optional suggestion:

```suggestion
                `Please export these variables to future processes so that CodeQL can monitor the build. ` +
```


Using active voice (i.e. "CodeQL can monitor" rather than "monitored by CodeQL") might be a bit clearer?

init/action.yml
shati-patel: @sampart - your comment makes me very happy 😄

shati-patel (approved):
Thanks, this looks good! I added a small comment, but feel free to ignore 😉

lib/runner.js
robertbrignull: Done

PR #308, opened 2020-11-17
Abort CodeQL action if the status cannot be reported

simon-engledew:
Save action minutes by only running the CodeQL check if code scanning is enabled for the repository.

This reverts some work done in https://github.com/github/codeql-action/pull/128

This is _nearly_ a no-op against Enterprise but there is value in not maintaining two separate behaviours.

~Enterprise will log a warning and not fail on a 422. This may occur if the bundled version of CodeQL was updated on an old version of Enterprise.~

- [x] I have only ever seen the Octokit request raise HTTPErrors rather than return a response with non-200 status codes. I need to check with `@robertbrignull` if this behaviour is new, or if there is an edge case that I have not managed to reproduce in my local setup.

~I think this is likely a change in octokit. As both the response and the HTTP error have a status field it seems reasonable to check it on both success and failure using the same code. If it changes in future this code should continue to work the same.~


simon-engledew:
@robertbrignull I had a chat with Marco and he is keen to keep using the status endpoint to determine if code scanning is enabled.

To do so, I have had to undo your previous change and start treating 404 and 403 as stop conditions.

I have attempted to keep to the spirit of the original code by treating other errors in a similar way:

422 - ignore it in Enterprise, it's probably just a mismatched codeql-action<->api due to an upgrade of the action (i.e: broken forward compatibility)
non-200 - ignore it for initial reporting as it may be transient. do not ignore the error for later reporting.

Unstitching something makes me nervous. I would be very grateful if you could give this a once over and see if it makes sense! ❤️

src/actions-util.ts
```diff
@@ -319,37 +323,42 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
       );
       return false;
-    }
+    case 422:
+      // schema incompatibility when reporting status
+      // on enterprise this may be down to an upgraded action sending new data and we should
+      // not stop the scanning process
+      return errorMessage(
+        "Invalid status report sent to code scanning.",
+        getRequiredEnvParam("GITHUB_SERVER_URL") !== GITHUB_DOTCOM_URL
```
robertbrignull, line 354: Should this be `ignoreFailures || (getRequiredEnvParam("GITHUB_SERVER_URL") !== GITHUB_DOTCOM_URL)`?

src/actions-util.ts
```diff
@@ -319,37 +323,42 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
       return false;
-    }
-    if (statusResponse.status === 404) {
+    case 404:
       core.setFailed(
         "Not authorized to used the CodeQL code scanning feature on this repo."
       );
       return false;
-    }
+    case 422:
```
robertbrignull, line 348: While this will achieve the goal of not crashing, I don't think "ignore it after the fact" is the way we should be handling version mismatches. From a philosophical point of view it seems a bit ugly, and practically we could get GHES admins complaining about the errors. However I don't want to put extra burden onto you or onto this issue, so I would be happy for us to merge this now and then address differently in the future.

My personal vote for how we should handle this better is that we should check the GHES at the start of the action run and abort if the GHES version is not supported. That would then guard future calls like this so there's no worry of it failing unexpectedly with a forwards compatibility failure. Note I'm not expecting you to do that in this PR, but I think it would be good to have a think / discussion about whether this PR on its own is safe and whether the above is a good future goal that makes sense to aim for.

robertbrignull (commented):
Generally looks good and looks like it should work, though I had a few comments.

src/actions-util.ts
simon-engledew: So, this would only happen if dotcom sent an invalid initial status report.

I figure we always want to surface this error, otherwise the initial status reporting could be broken for a long time and we might not notice.

src/actions-util.ts
```diff
@@ -319,37 +323,42 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
+
+  const status = await client
+    .request("PUT /repos/:owner/:repo/code-scanning/analysis/status", {
       owner,
       repo,
       data: statusReportJSON,
-    }
-  );
+    })
+    .then(getStatus, getStatus);
```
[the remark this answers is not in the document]
simon-engledew: I can do this explictly with a try catch, yes. I originally did implement it that way, but I was worried that there was some behaviour that I could not reproduce where octokit would return 403/4's rather than throwing.

It sounds like that is no longer the case, in which case I shall simplfiy this. :+1:

src/actions-util.ts
```diff
@@ -290,6 +290,15 @@ export async function createStatusReportBase(
   return statusReport;
 }
 
+function getStatus({ status }: { status?: number } = {}) {
```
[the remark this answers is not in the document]
simon-engledew: This function can go now I know that octokit will never return a non 200 response.

src/actions-util.ts
simon-engledew: I totally agree that this should be handled by robust version checking. 😁 This is definitely a band-aid. :+1:

My understanding is that this PR doesn't make the situation any worse than it currently is.

simon-engledew: Ok I have changed the action to fail on a 422, regardless of environment.

It's an implicit API version check as the status endpoint should not have changed, although obviously it does not guarentee that the analysis endpoint is still compatible.

Additional work could be done to do a version check at the beginning of the run as you have described, but I think it's not necessary for this PR.

robertbrignull: > My understanding is that this PR doesn't make the situation any worse than it currently is.

I wouldn't classify it as better or worse, but this PR definitely makes a difference because we're now calling the status endpoint on enterprise. I think it should be fine though and the error handling should stop the action from failing. I would like to test this before merging just to make sure, but we should be able to do that. I'll give it a go.

src/actions-util.ts
```diff
@@ -324,37 +327,62 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
+          core.setFailed(
+            "Not authorized to used the CodeQL code scanning feature on this repo."
+          );
+          return false;
+        case 422:
+          // schema incompatibility when reporting status
+          // this means that this action version is no longer compatible with the API
+          // and it is possible that other endpoints have changed too
+          if (getRequiredEnvParam("GITHUB_SERVER_URL") !== GITHUB_DOTCOM_URL) {
+            core.setFailed(
```
robertbrignull, line 360: Should this one (or maybe both of them) be `core.error` instead? Calling `core.setFailed` is equivalent to calling `core.error` and failing the action.

robertbrignull (approved):
Thanks. The changes LGTM now, and I've tried it out on a GHES instance and it works as expected.

simon-engledew:
I have also run a scan on dotcom:

https://github.com/dsp-testing/simon-engledew-lite/runs/1430257987?check_suite_focus=true

PR #309, opened 2020-11-18
Ensure unqualified program names are present on `PATH` before executing them.

chrisgavin:
This avoids some potential issues when executing unqualified program names by using a custom library (which you can review [here](https://github.com/chrisgavin/safe-which/)) to resolve them against the `PATH` variable first.


robertbrignull (approved):
The new dependency looks fine (the code looks good, and we got confirmation that using your library is fine). The rest of the code changes are pretty minimal. I've also reproduced the vulnerability and then that using your branch fixes it.

PR #310, opened 2020-11-19
Update code so "no-shadow" passes

### LINKED ISSUE #202: ESLint transition: update code so "no-shadow" passes
This issue relates to the [no-shadow](https://eslint.org/docs/rules/no-shadow) rule.

During the transition from TSLint (now deprecated) to ESLint for typescript linting we introduced some new rules that don't yet pass and added a temporary override in the `.eslintrc.json` file to temporarily disable them.

This issue covers removing that temporary override and any changes to the typescript code required to get the rule passing. Use `npm run lint` to run the linter and once you've done changing the typescript don't forget to also run `npm run build` to reflect the changes in the built javascript (which also needs to be checked in).

ericcornelissen:

* * *

Closes #202

This updates the configuration so that `@typescript-eslint/no-shadow` is enabled (instead of just `no-shadow`, following [the `@typescript-eslint/no-shadow` documentation](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-shadow.md#how-to-use); otherwise some false positives will be outputted). I tried to split my commits per updated file so that I can motivate my changes. Below is an overview of this as well. Suggestions are certainly welcome, and the maintainers can feel free to change things as they see fit 🙂 

<details>
<summary>Changes</summary>

In `api-client.ts`:

- Duplicate use of "_" placeholder argument name.

In `codeql.ts`:

- Two simple variable renames from "path" to "paths" since the types are arrays of strings (not just one string).
- One function definition inside a function moved outside that function to avoid shadowing the "options" argument.

In `config-utils.test.ts`:

- Rename "queries" variable in test cases to "testQueries" to avoid shadowing it in a subsequent helper function call (4 times).
- Rename "path" twice in a hlper function to "validPath" and "invalidPath" to avoid shadowing "path". The new names are more explicit.

In `config-utils.ts`:

- Rename throwaway variable "suite" to "found" when assigned from "find".
- Rename local variable "path" to "newPath" as it is a modification of the "originalPath" provided to `validateAndSanitisePath`.
- Rename instances of "path" to more explicit varients "ignorePath" and "includePath". Maybe "ignoredPath" and "includedPath" are better names?

In `fingerprints.test.ts`:

- Rename shadowing "uri" argument to the more explicit "artifactURI".

In `fingerprints.ts`:

- Rename various instances of "hash", shadowing the function with that name.

In `upload-lib.ts`:

- Rename one instance of "path"  to avoid shadowing.
</details>


chrisgavin (approved):
Thanks! This looks good to me.

PR #311, opened 2020-11-20
Fix typos in source code, logging, comments, and config files

ericcornelissen:

* * *

This fixes typos in both the source code (variable names, logging, and comments) and a couple of config files (`.yml`). Notable changes:

- Rename "toolrunnner" (three 'n') to "toolrunner" in various files
- Rename "relativeFilepaht" to "relativeFilepath" in ` src/fingerprints.test.ts`
- In the config files, only `description` values were changed


src/autobuild-action.ts
```diff
@@ -7,7 +7,7 @@ import { Language } from "./languages";
 import { getActionsLogger } from "./logging";
 
 interface AutobuildStatusReport extends actionsUtil.StatusReportBase {
-  // Comma-separated set of languages being autobuilt
+  // Comma-separated set of languages being autobuild
```
robertbrignull, line 10: This one I'm not so sure about. It's debatable because "autobuild" or "autobuilt" are not really real words. I'd argue that "autobuilt" is more correct here since it would be correct if you removed the "auto" prefix. Maybe it would be more clear with a hyphen as "auto-built".

robertbrignull (commented):
Wow, we're apparently really bad at spelling. Thanks for finding all of these.
One comment, and also you'll have to merge `main` and rebuild the javascript since another PR has been merged.

src/autobuild-action.ts
ericcornelissen: Actually, this one is my mistake I would say. I think autobuilt would be "correct", but I like auto-built better :smiley:

robertbrignull:
Sorry I think you've conflicted with your own PR again (https://github.com/github/codeql-action/pull/310)

robertbrignull:
Ah, snap! You just pushed.

robertbrignull (approved):

PR #316, opened 2020-11-20
Make sure a Python 2 pip is installed

The-Compiler:
In the README of this repo, it's suggested to use the `ubuntu-latest` environment.
That environment will soon be upgraded to Ubuntu 20.04: https://github.com/actions/virtual-environments/issues/1816

As pointed out in that issue, the updated image comes without a Python 2 pip preinstalled:

    Setup Python dependencies
    /home/runner/work/_actions/github/codeql-action/v1/python-setup/install_tools.sh
    [...]
      + python2 -m pip install --user --upgrade pip setuptools wheel
      /usr/bin/python2: No module named pip
    Warning: Unable to download and extract the tools needed for installing the python dependecies. You can call this action with 'setup-python-dependencies: false' to disable this process.

### DIFF
--- a/python-setup/install_tools.sh
+++ b/python-setup/install_tools.sh
@@ -10,6 +10,10 @@ set -e
 # subsequent actions in the current job, and not the current action.
 export PATH="$HOME/.local/bin:$PATH"
 
+# The Ubuntu 20.04 GHA environment does not come with a Python 2 pip
+curl https://bootstrap.pypa.io/get-pip.py --output get-pip.py
+python2 get-pip.py
+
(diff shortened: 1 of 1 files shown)


Daverlo (approved):
Thanks for this ❤️ 

It will prevent us from failing when `ubuntu-latest` its updated.

The-Compiler:
That was quick, thank you! Out of curiosity, when will I see this via `github/codeql-action/init@v1` in my GitHub Actions workflow? (Just asking so I don't forget to revert my commit pinning Ubuntu 18.04 there)

Daverlo:
It got there today! We create the merge PR every Monday.

PR #318, opened 2020-11-23
Introduce parameter object for API params that travel together

sampart:
I've been thinking about how to slim down our long parameter lists, and noticed that we almost always pass `githubAuth` and `githubUrl` as a pair. This PR introduces an interface for a data object which holds those two, so we can slim down parameter lists and introduce a little more type safety.

This change may not be worth the overhead it introduces, but I think it probably is.


robertbrignull (approved):
All changes look correct to me. Variables are going to the right places and there's no spurious changes.

I think this change is an improvement overall. It's not an improvement in terms of linecount but it makes each method definition and call-site more clear, and especially when methods delegate to each other and just pass the parameters on. It also removes a very real possibility of getting the auth/url parameters the wrong way around when calling a method.

PR #319, opened 2020-11-23
Validate the codeql-action workspace for common errors

simon-engledew:
This PR will automatically debug common issues with CodeQL workspaces:

- [x] `run: git checkout HEAD^2`
- [x] push with paths (the baseline wont update for every commit and we will end up with missing baselines)
- [x] no pull_request hook (this will stop working when we phase out guessing about intent via feature branches)
- [x] no push hook (we can't generate a baseline)
- [x] pull_request on branches that are not in push (you can open a PR against a branch with no baseline)

Todo:

- [x] Validate the current workflow looking for common CodeQL gotchas
- [x] Find a way to feed back the messages to the user
- [x] Test with a range of broken workflows

~In an ideal world, I would love if I could get the actions scheduler to pass in the current workflow file as GITHUB_WORKFLOW_FILE. This would make introspection easier, running the action locally easier and save two API requests~

I explored this idea and opened a feature request here: https://github.com/actions/runner/issues/825


simon-engledew:
This gives you output like:

image

krukow:
Definitely a `win` :)

simon-engledew:
Chris G and I did some spelunking through Launch, Actions Service and Runner to see how hard it would be to add the workflow file feature.

To get the converstation started, I opened a PR here: https://github.com/github/launch/pull/3540

It contains a write up of everything that I learned.

The TLDR is, the launch team don't think anyone else is asking for this. They suggested I open an issue against runner to see if anyone else in the community has similar requirements.

src/actions-util.ts
```diff
@@ -100,10 +102,210 @@ export const getCommitOid = async function (): Promise<string> {
...
+  }, {} as ReturnType<typeof toCodedErrors>);
+}
+
+export const WorkflowErrors = toCodedErrors({
+  MismatchedBranches: `Please make sure that every branch in on.pull_request is also in on.push so that CodeQL can establish a baseline.`,
+  MissingHooks: `Please specify on.push and on.pull_request hooks.`,
+  MissingPullRequestHook: `Please specify an on.pull_request hook so CodeQL is run against new pull requests.`,
+  MissingPushHook: `Please specify an on.push hook so CodeQL can establish a baseline.`,
+  PathsSpecified: `Please do not specify paths at on.pull.`,
+  CheckoutWrongHead: `Git checkout HEAD^2 is no longer necessary. Please remove this line.`,
```
rneatherway, line 158: I'd like to make these messages give a little more explanation, possibly with a link to the docs for the current suggested workflow. Is it interpreted as markdown so we can neatly include links?

* `baseline` is a bit internal/technical. I would rephrase it to something like 'so that Code Scanning can compare your pull requests against the state of the base branch'
* Why not specify paths for on:push? (because you might then not always have an analysis to compare your PR against)
* wrong head: lower-case `git` I think. You can get rid of the whole step, not just the line (so that includes the `if: pr`). I would also say something like 'Code Scanning recommends analyzing the merge commit for best results'

simon-engledew: > I'd like to make these messages give a little more explanation, possibly with a link to the docs for the current suggested workflow. Is it interpreted as markdown so we can neatly include links?

Great idea! I'll check. :+1:

simon-engledew: Sadly it does not look like you can output markdown here.

simon-engledew: Ok, I tried outputing a help URL and it ends up looking a bit messy. 🍝 

If you have a matrix you also end up seeing the warnings multiple times. 😭 

I think there is still value in emitting these warnings as-is, but at some point it would be worth investigating alternate ways of surfacing this information.

rneatherway (approved):

src/actions-util.ts
rneatherway: Turns out it is a pretty narrow pipe to shove information through! It's still going to help people though, not least our own support people.

src/actions-util.ts
```diff
@@ -100,10 +102,213 @@ export const getCommitOid = async function (): Promise<string> {
...
+  push?: WorkflowTrigger | null;
+  pull_request?: WorkflowTrigger | null;
+}
+
+interface Workflow {
+  jobs?: { [key: string]: WorkflowJob };
+  on?: string | string[] | WorkflowTriggers;
+}
+
+function isObject(o): o is object {
```
robertbrignull, line 128: Could this be `isObject(o: unknown)`? That'll make it easier when we eventually enable some more of the eslint checks.

src/actions-util.ts
```diff
@@ -100,10 +102,213 @@ export const getCommitOid = async function (): Promise<string> {
...
+  steps?: WorkflowJobStep[];
+}
+
+interface WorkflowTrigger {
+  branches?: string[];
+  paths?: string[];
+}
+
+interface WorkflowTriggers {
+  push?: WorkflowTrigger | null;
```
robertbrignull, line 119: The question mark on the field name means "or undefined", so this type here is really `WorkflowTrigger | null | undefined`. Is that what was intended? I think you can just remove the `| null` here and elsewhere in these definitions.

src/actions-util.ts
```diff
@@ -100,10 +102,213 @@ export const getCommitOid = async function (): Promise<string> {
...
+  pull_request?: WorkflowTrigger | null;
+}
+
+interface Workflow {
+  jobs?: { [key: string]: WorkflowJob };
+  on?: string | string[] | WorkflowTriggers;
+}
+
+function isObject(o): o is object {
+  return o !== null && typeof o === "object";
```
robertbrignull, line 129: Probably need to check for undefined too. Or you could just do `return o && typeof o === "object"` as I think objects are always truthy.

src/actions-util.ts
```diff
@@ -100,10 +102,213 @@ export const getCommitOid = async function (): Promise<string> {
...
+        errors.push(WorkflowErrors.PathsIgnoreSpecified);
+      }
+    }
+
+    if (doc.on.push) {
+      const push = doc.on.push.branches || [];
+
+      if (doc.on.pull_request) {
+        const pull_request = doc.on.pull_request.branches || [];
+        const intersects = pull_request.filter(
```
robertbrignull, line 227: "intersects" is not quite the right word here. I think this is the difference between the two sets, not the intersection.

src/actions-util.ts
```diff
@@ -100,10 +102,213 @@ export const getCommitOid = async function (): Promise<string> {
...
+      break;
+    case MissingTriggers.Push:
+      errors.push(WorkflowErrors.MissingPushHook);
+      break;
+  }
+
+  return errors;
+}
+
+export async function getWorkflowErrors(): Promise<CodedError[] | undefined> {
```
robertbrignull, line 258: What does it mean if it returns the empty list? Currently the code would output a warning saying "0 errors detected" which is a bit odd. Is it worth removing this ambiguity by changing the return type to `Promise<CodedError[]>` and make the empty list the "success" case?

src/actions-util.ts
```diff
@@ -100,10 +102,213 @@ export const getCommitOid = async function (): Promise<string> {
...
+    return undefined;
+  }
+
+  return workflowErrors;
+}
+
+export function formatWorkflowErrors(errors: CodedError[]): string {
+  const issuesWere = errors.length === 1 ? "issue was" : "issues were";
+
+  const errorsList = errors.map((e) => e.message).join(", ");
```
robertbrignull, line 277: The error messages are full sentences ending with fullstops, so joining them with a comma would be odd. Have I misunderstood? Some of these warning annotations can handle multiple lines so that could be an option, but you'd have to test how it looks in practice.

robertbrignull (commented):
A few comments but generally looks good and pretty clean and effective for notifying people of changes to how the workflow should look.

src/actions-util.ts
simon-engledew: Whoops! Good catch. 🤦 

I did try doing multiple lines and it just removed them 😭

src/actions-util.ts
simon-engledew: Yes, I think that is better - thanks!

src/actions-util.ts
simon-engledew: :+1:

src/actions-util.ts
simon-engledew: I took this from the recommendation in the type header for util.isObject. Does this look right?

```
❯ node
Welcome to Node.js v14.15.0.
> function isObject(o) {
...   return o !== null && typeof o === "object";
... }
undefined
> isObject(undefined);
false
> isObject(null);
false
> isObject("");
false
> isObject({});
true
```

src/actions-util.ts
simon-engledew: They can be null and undefined, which I found out the hard way! 😁

src/actions-util.ts
simon-engledew: Good idea! Thanks 👍

src/actions-util.ts
robertbrignull: Good point it works fine as it is. No need to do anything then. Ignore my comment here.

robertbrignull (approved):
I think this is fine now and all LGTM

src/actions-util.ts
simon-engledew: I've added a comment to document this information :+1:

PR #321, opened 2020-11-25
Update code so "github/no-then" passes

### LINKED ISSUE #200: ESLint transition: update code so "github/no-then" passes
This issue relates to the [github/no-then](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/no-then.md) rule.

During the transition from TSLint (now deprecated) to ESLint for typescript linting we introduced some new rules that don't yet pass and added a temporary override in the `.eslintrc.json` file to temporarily disable them.

This issue covers removing that temporary override and any changes to the typescript code required to get the rule passing. Use `npm run lint` to run the linter and once you've done changing the typescript don't forget to also run `npm run build` to reflect the changes in the built javascript (which also needs to be checked in).

ericcornelissen:

* * *

Closes #200 by wrapping the calls to `run()`-like functions that had `.catch` appended (possibly to avoid warnings by [the `@typescript/no-floating-promises` rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md)?) in functions that catch potential errors using a try-catch statement.

The `void` in front of the `runWrapper` function is necessary to avoid warnings by [the `@typescript/no-floating-promises` rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md).


robertbrignull:
This area is surprisingly tricky. The problem is that we're going from non-async code into async code and we need to catch any errors that happen within. Async stuff is just syntactic sugar around promises so that's why we call the async `run` function non-async and use `catch` explicitly. You can't use `async` at the top level in a file, or at least if you do it doesn't behave as you'd want it to.

Your solution here looks like it should work, but I'd like to test it to make sure. If errors are still reported correctly then LGTM.

robertbrignull:
Ok, I've tried it out (didn't take long) and the error handling is working exactly as it was before, so I'm happy with this and it gets us one eslint rule closer.

robertbrignull (approved):

PR #322, opened 2020-11-25
Suggestion: merge main and update lib/ in PR on dispatch

ericcornelissen:

* * *

This is a suggestion to change the way contributing to this project works. Currently, when a PR is submitted the submitter has to run `npm run build` and include the changes to the `lib/` directory in the PR. With the changes in this PR, including the changes to `lib/` is no longer necessary (*). This should avoid a bunch of merge conflicts with regards to the `lib/` directory - even though they're simple to fix it is annoying you have to.

To me, it seems these changes don't require any changes to [the Contributing Guidelines](https://github.com/github/codeql-action/blob/378f1f95d77928be059077e8fb39f1fd2529f45d/CONTRIBUTING.md), but if desired I wouldn't mind adding a note about this new workflow to [the "Submitting a pull request" section](https://github.com/github/codeql-action/blob/378f1f95d77928be059077e8fb39f1fd2529f45d/CONTRIBUTING.md#submitting-a-pull-request).

<sup>(*): It is still possible though, and if a PR is merged with updates to lib/ included, the changes in this PR won't have any effect.</sup>

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -19,8 +19,26 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Check that building works
-      run: npm run-script build
+    - name: Check generated JavaScript
+      run: |
+        # Sanity check that repo is clean to start with
+        if [ ! -z "$(git status --porcelain)" ]; then
+          # If we get a fail here then this workflow needs attention...
+          >&2 echo "Failed: Repo should be clean before testing!"
+          exit 1
+        fi
+        # Wipe the lib directory incase there are extra unnecessary files in there
+        rm -rf lib
+        # Generate the JavaScript files
+        npm run-script build
+        # Check that repo is still clean
+        if [ ! -z "$(git status --porcelain)" ]; then
+          # If we get a fail here then the PR needs attention
+          >&2 echo "Failed: JavaScript files are not up to date. Run 'npm run-script build' to update"
+          git status
+          exit 1
+        fi
+        echo "Success: JavaScript files are up to date"
 
   check-node-modules:
     runs-on: ubuntu-latest
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -19,26 +19,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Check generated JavaScript
-      run: |
-        # Sanity check that repo is clean to start with
-        if [ ! -z "$(git status --porcelain)" ]; then
-          # If we get a fail here then this workflow needs attention...
-          >&2 echo "Failed: Repo should be clean before testing!"
-          exit 1
-        fi
-        # Wipe the lib directory incase there are extra unnecessary files in there
-        rm -rf lib
-        # Generate the JavaScript files
-        npm run-script build
-        # Check that repo is still clean
-        if [ ! -z "$(git status --porcelain)" ]; then
-          # If we get a fail here then the PR needs attention
-          >&2 echo "Failed: JavaScript files are not up to date. Run 'npm run-script build' to update"
-          git status
-          exit 1
-        fi
-        echo "Success: JavaScript files are up to date"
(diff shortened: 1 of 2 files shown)


robertbrignull:
You raise a good point that it's a bit annoying when you have to checkout the branch locally and fix the generated files. For people working at GitHub we merge our own PRs after they're approved, and we're working full time during the day, so it's not a huge effort to update it locally and merge once the CI is green. I agree that for PRs raised from outside persons it's annoying that the PR owner has to fix some conflicts manually and then they can't merge it themselves so it has to wait for someone with access to see it and merge it and there's the very real chance of more conflicts happening in the meantime.

I'm just not sure that having a workflow that builds the generated files and pushes them is the way to go. There are other reasons that you want to build the generated files yourself and it can cause confusion if they aren't up to date. The unit tests run on the built javascript so you won't get accurate unit test results. Also the action itself obviously runs the built javascript so the action runs on the PR will be using an outdated copy of the code. Of course the workflow to build the JS will then run and update the PR and all the other workflows will then run again on correctly built JS, but you still have this weird commit where everything ran on the wrong version and the results are meaningless.

I'm also not a huge personal fan of automated tools pushing to PRs. It's a minor point but it creates a messy history. Creating infinite loops of automation is a worry, but actions has some builtin protection for that so I'm not worried in this case.

I wonder if it would make sense to instead have a workflow for solving the specific case of merging `main` into a branch and fixing conflicts in the `lib` directory. We could do it using a `workflow_dispatch` trigger. Those are only triggerable by repo owners but that's fine as we could trigger it and then merge the PR without needing the PR writer's involvement to fix trivial conflicts. That would improve the experience for yourself just as well, and avoids the potential downsides (though I admit those downsides are subjective).

ericcornelissen:
I wasn't quite aware of how the other workflows depend on the `lib/` directory being up-to-date (though, of course, those workflows can build it themselves - which of course has its own drawbacks). GitHub Actions does indeed protect against infinite loops of automation - the commit created by my change won't trigger any new workflows - if the maintainers of this project don't like it so be it.

I do like your suggestion of a manual trigger to automatically update the `lib/` directory. This still has the advantage of minimal interaction from the maintainer, and also doesn't stop the maintainer from updating the `lib/` directory if the contributor disabled changes from maintainers. I'm not sure if I will be able to put together a working workflow file for this, but I will try something 🙂

ericcornelissen:
I just pushed a draft of your suggestion @robertbrignull, I haven't tested it yet (I will start doing that now) but let me know what you think.

For the record, I wanted to make explicit that I'm using two community-made Actions. One to checkout a PR: [dawidd6/action-checkout-pr](https://github.com/marketplace/actions/checkout-pr) and one to push back the changes [stefanzweifel/git-auto-commit-action ](https://github.com/marketplace/actions/git-auto-commit).

robertbrignull:
I had a go at testing this in another test repository and unfortunately I ran into a few problems. The main ones were that `git merge` returns non-zero if there are merge conflicts, so you have to add some bash logic around that to handle it and only fail in the right circumstances. We don't want it to always succeed as there could be other errors, such as the git username/email address not being configured. I also had some trouble with https://github.com/stefanzweifel/git-auto-commit-action and the commits it created were a bit weird. For example it does a `git checkout` at the start which caused a bit of havoc. It also wasn't able to commit merge conflicts that weren't explicitly resolved via `git add` so I had to add that in manually. At that point the 3rd party actions weren't adding too much so I felt it was simpler to remove them.

I've opened a new PR at https://github.com/github/codeql-action/pull/340 as I thought that was the easiest way to show the state I got to. What's in that PR was working for me and could fix conflicts in the `lib` directory, or just merge main when there weren't conflicts.

ericcornelissen:
Thank you very much for taking the effort to improve upon my draft, your PR looks totally fine to me and is certainly more robust than what I cobbled together here. I'd say #340 can be reviewed and merged instead and this can be closed 🙂

robertbrignull:
Thanks for having a look @ericcornelissen. Let's close this then and continue with #340

PR #324, opened 2020-11-27
Test Go autobuild

robertbrignull:
We should be testing using autobuild with Go as well as a manual build command, just in case.

### DIFF
--- a/.github/workflows/integration-testing.yml
+++ b/.github/workflows/integration-testing.yml
@@ -137,10 +137,6 @@ jobs:
       CODEQL_EXTRACTOR_GO_BUILD_TRACING: "on"
 
     steps:
-    - uses: actions/setup-go@v2
-      if: ${{ matrix.os ==  'macos-latest' }}
-      with:
-        go-version: '^1.13.1'
     - uses: actions/checkout@v2
     - name: Move codeql-action
       shell: bash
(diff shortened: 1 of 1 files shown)


robertbrignull:
I wanted to make sure the Go tracing was working if you used the autobuild action and thought we may as well add an integration test for it. Turns out it does work.

.github/workflows/integration-testing.yml
```diff
@@ -129,6 +129,33 @@ jobs:
...
+    # we're testing Go custom tracing with a manual build on all OSes.
+    runs-on: ubuntu-latest
+    env:
+      CODEQL_EXTRACTOR_GO_BUILD_TRACING: "on"
+
+    steps:
+    - uses: actions/setup-go@v2
+      if: ${{ matrix.os ==  'macos-latest' }}
+      with:
+        go-version: '^1.13.1'
```
chrisgavin, line 143:
```suggestion
```

Does this do anything? It looks like the job always runs on Ubuntu so this step will always be skipped.

chrisgavin (approved):
Looks good with the exception of one question.

.github/workflows/integration-testing.yml
robertbrignull: Thanks. I accidentally copied that from the other job.

PR #326, opened 2020-11-27
Add an explicit deprecation date

rneatherway:
So that we can remove support for the beta configuration of
`on: push`-only we should communicate a date after which this
will not work.


src/actions-util.ts
```diff
@@ -157,7 +157,7 @@ function toCodedErrors(errors: {
 export const WorkflowErrors = toCodedErrors({
   MismatchedBranches: `Please make sure that every branch in on.pull_request is also in on.push so that Code Scanning can compare pull requests against the state of the base branch.`,
   MissingHooks: `Please specify on.push and on.pull_request hooks so that Code Scanning can compare pull requests against the state of the base branch.`,
-  MissingPullRequestHook: `Please specify an on.pull_request hook so that Code Scanning is run against pull requests.`,
+  MissingPullRequestHook: `Please specify an on.pull_request hook so that Code Scanning is explicitly run against pull requests. This will be required to see results on pull requests from 2021-01-31.`,
```
robertbrignull, line 160: Would it be more or less inclusive to spell out the date in words? Though people should realise the format when they try to work out what the 31st month is...

rneatherway: It crossed my mind. I think that's a good suggestion. How about `31 January 2021`?

robertbrignull: Thanks. That looks much less ambiguous

rneatherway: Updated.

robertbrignull (approved):
LGTM

PR #330, opened 2020-12-01
Add a check to ensure expected CodeQL release files exist.

chrisgavin:
This adds a check to make sure all the expected files on the CodeQL bundle release exist, including the runner and the CodeQL complete bundle, neither of which are covered by other checks.

(Adding this because I initially forgot to rename the `codeql-bundle-all.zip` file to `codeql-bundle.zip`.)


robertbrignull (approved):
Thanks. This is pretty neat

PR #332, opened 2020-12-01
Handle the case where branches may be strings, wildcards

simon-engledew:
This PR:

* fixes an oversight where branches could also be set to a string
* handles wildcard branches, testing if they overlap or not


robertbrignull (commented):
I agree this fixes the case we know about that breaks it. What I'm wondering is if there are more cases like this and how we can find them / make them safe.

For example there are a few places it could fail if fields are truthy values of unexpected types, like if `on.push.branches` was a number or something like that. Of course that particular isn't a valid workflow, so maybe we don't care about it. There is the danger that workflow syntax could change in the future to become more permissive.

How much extra work would it be to explicitly check types, instead of using shorthand like checking truthy-ness?

robertbrignull:
> I am worried that if we alter the linter to ignore structure it does not expect then it will silently fail in the background.

I agree, though having it fail is also unfortunate. Perhaps the best way to address this, whichever direction we go with failures, is with excessive amounts of unit tests.

> That way I can add a dashboard metric and keep an eye on it.

Unfortunately getting metrics out of the action is not too easy as it's untrusted code. It's not a great situation. The only way to do this is via the status report, and changing that is a pain. We could add it but I'm not sure it's a perfect fit.

simon-engledew:
> Unfortunately getting metrics out of the action is not too easy as it's untrusted code. It's not a great situation. The only way to do this is via the status report, and changing that is a pain. We could add it but I'm not sure it's a perfect fit.

I've made it report the same way as a workflow lint :)

src/actions-util.test.ts
```diff
@@ -230,3 +268,29 @@ test("formatWorkflowCause()", (t) => {
...
+  t.true(actionsutil.patternsOverlap("*", "*"));
+  t.true(actionsutil.patternsOverlap("*", "main-*"));
+  t.false(actionsutil.patternsOverlap("main-*", "*"));
+  t.false(actionsutil.patternsOverlap("main-*", "main"));
+  t.true(actionsutil.patternsOverlap("main", "main"));
+  t.false(actionsutil.patternsOverlap("*", "feature/*"));
+  t.true(actionsutil.patternsOverlap("**", "feature/*"));
+  t.false(actionsutil.patternsOverlap("feature-*", "**"));
+  t.false(actionsutil.patternsOverlap("a/**/c", "a/**/d"));
+  t.false(actionsutil.patternsOverlap("a/**/c", "a/**"));
```
robertbrignull, line 283: Are double stars only allowed as complete path segments? i.e. is `foo-**` not allowed?

src/init-action.ts
```diff
@@ -98,7 +98,7 @@ async function run() {
 
     const workflowErrors = await actionsUtil.getWorkflowErrors();
 
-    if (workflowErrors.length > 0) {
+    if (workflowErrors.filter(o => o.code !== 'LintFailed').length > 0) {
```
robertbrignull, line 101: This will mean we hide the warning if we fail to link the workflow. I thought we wanted to show it?

src/actions-util.ts
```diff
@@ -134,6 +134,81 @@ function isObject(o: unknown): o is object {
...
+      indexB += 1;
+    }
+    if (advance) {
+      indexA += 1;
+    }
+  }
+  return false;
+}
+
+export function patternsOverlap(patternA: string, patternB: string): boolean {
```
robertbrignull, line 195: What does "overlap" mean? That the patterns are equivalent, or they have non-empty intersection? Could you add comments / docstring here and to any other non-obvious methods?

robertbrignull (commented):
My mind's not working well enough to say if your algorithm is correct or not, but the tests are helpful for that. I'm also worried about going further down the rabbit hole and I think we should probably stop here.

So long as we err on the side of not warning people I think that's fine. I think it
s better that someone runs a slightly dodgy workflow than that we erroneously pop up warnings. What do you think?

src/actions-util.test.ts
simon-engledew: > My mind's not working well enough to say if your algorithm is correct or not, but the tests are helpful for that. I'm also worried about going further down the rabbit hole and I think we should probably stop here.

Totally agree. :+1:

> Are double stars only allowed as complete path segments? i.e. is foo-** not allowed?

Unclear in the documentation. I've added a test to show how it would work.

src/actions-util.ts
simon-engledew: Yes, bad name. 👍 I think `patternIsSuperset` is marginally better?

It should return true if the pattern b is represented by the pattern a.

src/init-action.ts
simon-engledew: I've added a comment. I don't think we want to warn people visually but we do want to report the error via status.

simon-engledew:
> My mind's not working well enough to say if your algorithm is correct or not, but the tests are helpful for that. I'm also worried about going further down the rabbit hole and I think we should probably stop here.

Ugh. I should have taken a break sooner! Ate my dinner and realised that I should not have discounted the idea to turn patterns into a RegExp so quickly. Turns out that's a way easier way of doing it.

robertbrignull (approved):
Unfortunately still fails if you do
```typescript
actionsutil.validateWorkflow({
  on: {
    push: {
      branches: 1
    },
    pull_request: {
      branches: 1
    },
  },
} as any),
[]
);
```
however this is a minor case and it shouldn't be possible to set this as it's an invalid workflow.

And of course in practice it catches any errors and returns `LintFailed`, so it won't fail in practice.

So, this looks good to go to me. If you want to fix the above case that's up to you. You can go like this or fix that and then merge as you like. So long as we get this on `main` today so the release on Monday can go ahead.

PR #339, opened 2020-12-07
Install Python 3 tools before installing Python 2 tools.

chrisgavin:
This change is designed to reduce the fallout of issues with Python 2 dependency installation, since Python 2 is both less popular than Python 3 and also end-of-life meaning there are more likely to be issues with it as tools like Pip drop support for it.

Firstly it moves all the Python 2 steps to the end. This ensures that if they fail the script will have already done the Python 3 dependency installation, which is the one we likely care most about. It also wraps the Python 2 dependency installation in a check that Python 2 is actually present to avoid spurious warnings when you're only using Python 3 which is going to become more and more common.

I note that [there is a [PowerShell version of this script](https://github.com/github/codeql-action/blob/main/python-setup/install_tools.ps1), which I have not updated. This is because the PowerShell version already totally ignores all exit codes, so is already fairly safe when it comes to the Python 2 steps failing (although ignoring all exit codes obviously has its own problems).


RasmusWL (approved):
Nice, thanks :+1:

I was slightly worried what would happen if we try to analyze some project as Python 2, and this step failed because `python2` command wasn't available. But I guess we will figure out quick enough when trying to invoke the extractor with `python2` :smile: 

As a side question, do you know if `python2` is available by default in environment the `codeql-action` runs on when it's based off Ubuntu 20.04 instead of Ubuntu 16.04?

chrisgavin:
> I was slightly worried what would happen if we try to analyze some project as Python 2, and this step failed because `python2` command wasn't available. But I guess we will figure out quick enough when trying to invoke the extractor with `python2` :smile:

Haha, yep my thoughts too.

> As a side question, do you know if `python2` is available by default in environment the `codeql-action` runs on when it's based off Ubuntu 20.04 instead of Ubuntu 16.04?

Yes, `python2` is still present, it's just Pip for Python 2 that's not installed by default anymore.

PR #341, opened 2020-12-08
Gate execution of expensive tests on JS being up to date

### LINKED ISSUE #182: CI: Reduce unnecessary jobs
We currently have 3 workflows that run as part of the CI, and they run on both `push` and `pull-request`.

- codeql.yml: 
  - Basic sanity check of the CodeQL behavior. 
  - Splits into 3 jobs of up to 9 minutes. 
  - Hopefully this will improve soon with https://github.com/github/codeql-action/pull/172 and similar changes

- pr-checks.yml:
  - Code quality checks
  - Splits into 5 jobs of less than 30 seconds

- integration-testing.yml:
  - Extensive coverage of cross-feature behavior.
  - Splits into 19 jobs of between 5 and 15 minutes.
  - https://github.com/github/codeql-action/pull/172 is planning to add 21 additional jobs

All those parallel jobs compete for the same resources. Running them twice means that each push we are running 27x2 -> 54 jobs, filling up the queue.

**Proposal**
1. Change `pr-checks.yml` to be only on `push` (maybe rename to `linting`?)
2. Change `codeql.yml` to be on `push` only for `main` and `v1`, and `pull-request` for all.
3. Change `integration-testing` to be on  `pull-request` and only on push for `main` and `v1`
4. Ideally sort the jobs in `integration-testing` from the more likely to fail to the less likely to fail, and make the latter conditional on the former.
    * This is the strategy I am hoping to adopt for https://github.com/github/codeql-action/pull/172

This would avoid duplication of jobs without reducing the coverage we get.

rneatherway:

Fixes #182.

I actually thought while doing this that we could just merge these two workflow files, I am not sure that we are getting any benefit from them being separate are we?

### DIFF
--- a/.github/workflows/integration-testing.yml
+++ b/.github/workflows/integration-testing.yml
@@ -6,7 +6,27 @@ on:
   pull_request:
 
 jobs:
+  # This job and check-node-modules below dupliacte checks from `pr-checks.yml`.
+  # We run them here as well to gate the more expensive checks which wouldn't
+  # even be executing the correct code if these fail.
+  check-js:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Check generated JavaScript
+      run: .github/workflows/script/check-js.sh
+
+  check-node-modules:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Check node modules up to date
+      run: .github/workflows/script/check-node-modules.sh
+
   multi-language-repo_test-autodetect-languages:
+    needs: [check-js, check-node-modules]
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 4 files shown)


robertbrignull (approved):
Seems reasonable and correct to me.

Do you know how often it actually happens that the integration tests are executed unnecessarily in this way? Are we saving much? What was it that prompted you to do this?

I don't have any strong opinions about merging the workflow files. By merging them we'd save a tiny bit of performance by not duplicating jobs, and we wouldn't need to have the implementation in separate files elsewhere. The downside of merging them is that it makes browsing the UI slightly harder (arguably) and changes which set of checks you can retrigger as a group as you have to retrigger all jobs from a workflow at once.

robertbrignull:
> What was it that prompted you to do this?

I should have followed the link to #182 before commenting so that's my bad. I see we all agreed to do this before and I just forgot about it. I'd still be interested to know what magnitude of effect it'll have if that's possible to calculate, but it might be quite hard to work out.

rneatherway:
It was mainly a desire to close out #182 I'll be honest. Personally I often forget to regenerate the javascript, and it would save running them in that case, so it seems worthwhile.

rneatherway:
I think I would vote for merging to avoid the duplication, but that could be a follow-up.

PR #348, opened 2020-12-21
Only report the first CheckoutWrongHead lint error

simon-engledew:
Some people are running `git checkout HEAD^2` many times during their workflow.

To look a bit tidier, this will only warn them once instead of one warning for each instance.


robertbrignull (approved):
The actual code change looks fine. I don't think we use break with named labels anywhere else, but eslint isn't complaining so it's apparently within our style guide. I don't have a problem with it.

I do wonder why people would be running it multiple times. That's weird and definitely not something we've recommended or included in our default workflows. Do you have an example repo where you've seen this?

Worth nothing that `git checkout HEAD^2` is not idempotent, so it's definitely different from just including it once. So from that point of view I wonder if we should warn them at all since they're clearly doing something unusual. But then also I can't think why you would legitimately want to run it multiple times, so perhaps we should warn them and prompt them to change their workflow because it's weird.

simon-engledew:
> I do wonder why people would be running it multiple times. That's weird and definitely not something we've recommended or included in our default workflows. Do you have an example repo where you've seen this?

Thanks for the prompt! I had a double check and it happens when there are multiple jobs all using the same anti-pattern.

All the linting we do is global accept this so it was polluting the lint output of other jobs. I've improved the fix so that it reports the issue to the job which includes it.

PR #349, opened 2020-12-22
Prevent accidental analysis of the `get-pip.py` script.

chrisgavin:
It seems like in some cases this script gets analyzed. We can either remove it after running it, or just not save it at all.


cbraynor (approved):

chrisgavin:
I've verified that this does fix the extraction of `get-pip.py`. The Action previously output these lines which are gone when running it with this branch:

```
  [2020-12-22 14:31:21] [build] [INFO] [2] Extracted file /home/runner/work/py2004/py2004/get-pip.py in 774ms
  [2020-12-22 14:31:21] [build] [INFO] [2] Extracted file /home/runner/work/py2004/py2004/get-pip.py in 774ms
```

Issue #220, opened 2020-09-18
Recursive SARIF file detection from `upload-sarif` sarif_file parameter


tlsa:
I [created a workflow](https://github.com/tlsa/libcyaml/blob/tlsa/codeql/.github/workflows/static-analysis.yaml#L74) that runs Clang scan-build with the `-sarif` parameter, to output SARIF format, and the `-o build/sarif` argument to put the output in the directory `build/sarif`.

When it is run, scan-build puts the actual SARIF files in a sub-directory of the specified output directory.

The output from scan-build is:

```
scan-build: Analysis run complete.
scan-build: Analysis results (sarif files) deposited in '/home/runner/work/libcyaml/libcyaml/build/sarif/2020-09-18-110804-4323-1'
```

And the sarif files are in the randomly named sub-directory:

```
$ find build/sarif
build/sarif
build/sarif/2020-09-18-110804-4323-1
build/sarif/2020-09-18-110804-4323-1/report-B4xj3N.sarif
build/sarif/2020-09-18-110804-4323-1/report-yK7dxl.sarif
build/sarif/2020-09-18-110804-4323-1/report-SrmlCl.sarif
build/sarif/2020-09-18-110804-4323-1/report-QPrVcF.sarif
build/sarif/2020-09-18-110804-4323-1/report-67tbyT.sarif
build/sarif/2020-09-18-110804-4323-1/report-OfHm5E.sarif
build/sarif/2020-09-18-110804-4323-1/report-T8GZn7.sarif
build/sarif/2020-09-18-110804-4323-1/report-NKjLYw.sarif
build/sarif/2020-09-18-110804-4323-1/report-z81jQ7.sarif
build/sarif/2020-09-18-110804-4323-1/report-bmXRUK.sarif
build/sarif/2020-09-18-110804-4323-1/report-6aH1dl.sarif
build/sarif/2020-09-18-110804-4323-1/report-iK7eHK.sarif
build/sarif/2020-09-18-110804-4323-1/report-vnqcCF.sarif
build/sarif/2020-09-18-110804-4323-1/report-aATCA3.sarif
```

You can see the CI run for this here: https://github.com/tlsa/libcyaml/runs/1133489433

At the moment, with

```
    - name: upload scan-build
      uses: github/codeql-action/upload-sarif@v1
      with:
        sarif_file: build/sarif
```

I get the following error:

```
#[error]No SARIF files found to upload in "build/sarif".
Error: No SARIF files found to upload in "build/sarif".
[...]

PR #352, opened 2020-12-23
Support find .sarif files recursively

chao2zhang:
This addresses #220 by recursively calling `fs.readdirSync()`

- [X] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary.


robertbrignull (approved):
Thanks for implementing this. The implementation is nice and simple and looks good to me.

The only thing I was worried about was if we'd get into loops with symbolic links, but looking at the documentation it looks like it'll be fine, and I even gave it a try locally and it was fine.

I think we could add a unit test of this but rather than delaying this PR it'll probably be easiest to merge now and I'll add a quick test.

jwgmeligmeyling:
The workings of this feature is not documented in the action.yml nor readme, despite the checkmark for it.

Some examples would be helpful. For example, this recursive implementation does only work on folder names and not file names.

aeisenberg:
> Some examples would be helpful. For example, this recursive implementation does only work on folder names and not file names.

Can you be more specific? You can specify a directory (which gets searched recursively for sarif files or a single file. Is there a feature you are looking for, but is not implemented here? Or are you just seeking clarity on the behaviour?

PR #357, opened 2021-01-04
Introduce external repository token

robertbrignull:
Potential alternative to https://github.com/github/codeql-action/pull/273 to demonstrate an idea.

Introduces `GitHubApiExternalRepoDetails` and `GitHubApiCombinedDetails` (names very open to changing) to get us type safety on the different types of api token.

Closes #273.


chrisgavin (approved):
This does seem like a fairly neat option to maintain the encapsulation. I'm happy with this method instead of #273.

robertbrignull:
I forgot to include a couple of bits of https://github.com/github/codeql-action/pull/273 like the changes to the readme and to `init/action.yml`. I think I've got it all now and this should work but I'm going to give it a manual test before merging. @chrisgavin, could you have another quick look and see if you spot anything else I've missed?

robertbrignull:
I've tested this out locally with a private repository that could only be downloaded by passing the external repository token. It all seems good.

PR #360, opened 2021-01-11
Output path to preload_tracer in env file

robertbrignull:
When building code on macos with SIP enabled it's necessary to prefix your build command with `preload_tracer`, so to make that easy we output a path to that executable in the env that you need to source. This means you should be able to run `$CODEQL_PRELOAD_TRACER ./my-build-command.sh`.


robertbrignull:
@p0, thanks for your suggestion to use `${CODEQL_DIST}/tools/${CODEQL_ARCH}/preload_tracer`.
I've used `CODEQL_DIST` in this but I'm not sure `CODEQL_ARCH` is available so I've hardcoded the path as `${CODEQL_DIST}/tools/osx64/preload_tracer`. Is this ok? The code path is only executed on osx.

p0:
Yes, that would be safe enough.

adityasharad (approved):

aibaars:
> @p0, thanks for your suggestion to use `${CODEQL_DIST}/tools/${CODEQL_ARCH}/preload_tracer`.
> I've used `CODEQL_DIST` in this but I'm not sure `CODEQL_ARCH` is available so I've hardcoded the path as `${CODEQL_DIST}/tools/osx64/preload_tracer`. Is this ok? The code path is only executed on osx.

The variable is called `CODEQL_PLATFORM` and should be available. 

Why do we need to prefix things with the `preload_tracer`; wouldn't the `codeql/tools/osx64/runner` be enough?

robertbrignull:
> The variable is called `CODEQL_PLATFORM` and should be available. 

Thanks. You're right. I've changed it to use that value.

src/runner.ts
```diff
@@ -203,6 +203,19 @@ program
...
+      // On macos include the path to preload_tracer in the env as that
+      // executable is needed to trace when System Integrity Protection is enabled.
+      if (process.platform === "darwin") {
+        const codeqlDist = tracerConfig.env["CODEQL_DIST"];
+        const codeqlPlatform = tracerConfig.env["CODEQL_PLATFORM"];
+        tracerConfig.env["CODEQL_PRELOAD_TRACER"] = path.join(
+          codeqlDist,
+          "tools",
+          codeqlPlatform,
+          "preload_tracer"
```
aibaars, line 215: @p0 Wouldn't it be better to use the `runner` instead. It is available for all platforms including windows so we simply populate an env variable like `CODEQL_RUNNER` for all platforms. The advantage is that things keep on working in case a user tries to run the CI job on various platforms (or copy pastes the `"$CODEQL_PRELOAD_TRACER" ./build-command.sh` command line into a linux job). This could lead to unexpected problems due to undefined or empty environment variables.

robertbrignull:
I've switched over to using the `runner` and outputting the env var on all platforms. I'm going to write some tests for this.

src/runner.ts
```diff
@@ -203,6 +203,17 @@ program
...
 
+      // On macos it's necessary to prefix the build command with the runner exectuable
+      // on order to trace when System Integrity Protection is enabled.
+      // The exectuable also exists and works for other platforms so we output this env
+      // var with a path to the runner regardless so it's always available.
+      tracerConfig.env["CODEQL_RUNNER"] = path.join(
+        tracerConfig.env["CODEQL_DIST"],
+        "tools",
+        tracerConfig.env["CODEQL_PLATFORM"],
+        "runner"
```
aibaars, line 214: I think it is fine to leave off the `.exe` extension on Windows, but you might want to add it just in case.

robertbrignull: Thanks. I was just looking at this. May as well add on the extension.

I also note there are `runner-linux`, `runner-win.exe`, and `runner-osx`. They appear to maybe be duplicates of the non-suffixed versions. Should these be ignored?

aibaars: I think these are left-overs of the legacy tools.

robertbrignull:
I've added the `CODEQL_RUNNER` var to the unit tests, and added usages of it to some of the integration tests. Unfortunately the unit tests don't run on windows (I only just discovered this) and they don't pass either so enabling them will be a bit of work. So for windows we're relying on the integration tests to show that things are working, but that's probably fine.

robertbrignull:
@aibaars or @adityasharad do you mind having another quick look over this and make sure I haven't done anything stupid since it was last reviewed?

aibaars (approved):
Looks good to me.

PR #363, opened 2021-01-13
Fix overzealous warnings when PR scanning is not required

simon-engledew:
This should relax the linter to allow the case where users do not want PR scanning.

See https://github.com/github/codeql/issues/4850

I've used a helper when making deepEqual calls on error messages to improve the test output. Before it simply said `{ object }`, which made it quite hard to see what was wrong. Now it pretty prints the message codes.


src/actions-util.ts
```diff
@@ -236,11 +236,11 @@ export function validateWorkflow(doc: Workflow): CodedError[] {
   let missing = MissingTriggers.None;
 
   if (doc.on === undefined) {
-    missing = MissingTriggers.Push | MissingTriggers.PullRequest;
+    // codeql will scan the default branch
```
robertbrignull, line 239: Having a missing `on` section doesn't appear to be a valid workflow at all, so it probably doesn't matter what we do in this case. So not erroring is perfectly fine. Not suggesting changing anything here but just trying to confirm my understanding.

src/actions-util.ts
```diff
@@ -250,19 +250,22 @@ export function validateWorkflow(doc: Workflow): CodedError[] {
...
-      missing = missing | MissingTriggers.PullRequest;
-    }
-    if (!Object.prototype.hasOwnProperty.call(doc.on, "push")) {
+    const hasPush = Object.prototype.hasOwnProperty.call(doc.on, "push");
+    const hasPullRequest = Object.prototype.hasOwnProperty.call(
+      doc.on,
+      "pull_request"
+    );
+
+    if (!hasPush) {
```
robertbrignull, line 265: Is this definitely something we want to warn about, or should it `if (!hasPush && hasPullRequest)`?

I'm thinking about a workflow like
```
on: [workflow_dispatch]
```
so it's entirely manual. That would seem fine to me.

If you add this then adding the above as a test case would be good.

src/actions-util.ts
```diff
@@ -276,24 +279,28 @@ export function validateWorkflow(doc: Workflow): CodedError[] {
       }
     }
 
-    const push = branchesToArray(doc.on.push?.branches);
+    // check the user is scanning PRs right now
+    // if not the warning does not apply
+    if (doc.on.pull_request !== undefined) {
```
robertbrignull, line 284: Why `doc.on.pull_request` instead of `hasPullRequest`?

It probably helps the type checker be convinced you're accessing something that exists, but in that case why not use `doc.on.pull_request !== undefined` to define `hasPullRequest` and then you might get that for free here too.

Also, the type of `doc.on.pull_request` is `WorkflowTrigger | null | undefined` so you're missing another bad case potentially.

src/actions-util.test.ts
```diff
@@ -5,6 +5,13 @@ import sinon from "sinon";
 import * as actionsutil from "./actions-util";
 import { setupTests } from "./testing-utils";
 
+function errorCodes(
```
robertbrignull, line 8: Since `errorCodes` is always used in the form `t.deepEqual(...errorCodes(...))` would it make things simpler to change this to
```typescript
import { ExecutionContext } from "ava";

function errorCodesDeepEqual(
  t: ExecutionContext,
  actual: actionsutil.CodedError[],
  expected: actionsutil.CodedError[]
){
  t.deepEqual(actual.map(({ code }) => code), expected.map(({ code }) => code));
}
```

robertbrignull (commented):
Looks pretty good though one comment where I think we're still being too strict and will warm people about acceptable edge cases.

src/actions-util.ts
simon-engledew: Ah, I should probably add a comment! 🤔 🤦  Null actually means 'all branches' where undefined means 'not configured'. I thought I added a test case to show this but even if I did, it's not very clear.

Checking against undefined is intentional here as we want to skip the mismatched branches check if pull_requests are not enabled.

src/actions-util.test.ts
simon-engledew: I originally wrote it that way but then the error message points into that helper function instead of the test, which forces you to find the line number by hand and then look it up in the test file.

I looked into extending the test library but it is minimialist to the point of brutalism. I couldn't extend it and keep the typechecker happy, and it involved a prototype overload which seemed dirty.

src/actions-util.ts
simon-engledew: Ah, nice catch! Thanks ❤️

src/actions-util.ts
simon-engledew: Ah ok! I thought it might be valid - some strange stuff seems to be. I'll add a comment based on this feedback.

src/actions-util.test.ts
robertbrignull: Ah fair. Making the test failures more readable is pretty helpful. Feel free to leave this as it is. My suggestion here was only because I thought it might be cleaner and not because there's anything bad about what's there currently.

robertbrignull (approved):

PR #367, opened 2021-01-19
Remove date specific pull request warnings

simon-engledew:
After January 2021 CodeQL will no longer display pull request results if `on.pull_request` is not set.

Now we only need to warn if on.pull_request is configured without a corresponding `on.push`.


src/actions-util.ts
```diff
@@ -177,30 +177,20 @@ function branchesToArray(branches?: string | null | string[]): string[] | "**" {
...
-
 export interface CodedError {
   message: string;
   code: string;
 }
-function toCodedErrors(errors: {
-  [key: string]: string;
-}): { [key: string]: CodedError } {
+
+function toCodedErrors<T>(errors: T): Record<keyof T, CodedError> {
```
simon-engledew, line 185: I updated this so that the keys are typechecked and you get an error if you try and use one that is not defined.

chrisgavin (approved):
LGTM.

Do we want to have a period where we continue to post a re-worded version of the warning even after making the change in case people go looking for reasons why their Actions are no longer working as expected? I guess it should be very rare people are depending on the old behavior now, so I think just removing the warning is also fine. :+1:

simon-engledew:
>  I guess it should be very rare people are depending on the old behavior now, so I think just removing the warning is also fine.

:+1: Cool. We can always put it back in if people run into it a lot.

PR #368, opened 2021-01-19
Only insert external repos token if supplied

robertbrignull:
Addresses a bug reported by @lcartey that we accidentally introduced in https://github.com/github/codeql-action/pull/357 which was breaking the workflow for people using a workaround we suggested at https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#accessing-private-repositories and also a version using a deploy key and rewriting to a SSH URL.

The PR changes it so we only insert the external repositories token into the URL if it is explicitly supplied and don't fall back to using the main github token. Also avoids adding in a double slash to the URL sometimes.

@chrisgavin do you think this is ok? Why was the falling back to the other token there? It's unlikely to have access to other repos. Was it just in case it did to avoid having to add the token argument twice?


chrisgavin (approved):
Falling back to the other token is useful on GitHub Enterprise Server, where it will have access to all public repositories.

That said, I'm happy with the change. I think the implicit credentials should technically still work, so the fallback is still there; it's just in a different place.

robertbrignull:
> I think the implicit credentials should technically still work, so the fallback is still there; it's just in a different place.

Could you explain what you mean here?

Also I now realise that I haven't actually tried testing that this fixes the bug. If the github token is being picked up through some other means there's a chance this won't fix anything.

chrisgavin:
The `GITHUB_TOKEN` is stored in the Git configuration of the runner automatically when you do a `checkout` so it will also be used for Git operations against external repositories even if we don't pass it explicitly.

robertbrignull:
I've tested this out in all possible configurations that I could think of and it seems to be working as expected.

- If you use the workaround from https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#accessing-private-repositories with an old version of the codeql-action (before this week's update) then it works
- If you change the branch to `main` then it fails
- If you change the branch to this one then it works again (still using the workaround)
- If you change to using `external-repository-token` then it still works

PR #369, opened 2021-01-19
Update description to include limit on number of results

felicitymay:
👋🏻 I'm proposing this change alongside some similar changes to the documentation. We want to make sure that people are aware of the limit on the number of results that will be processed per SARIF file. This is rarely a problem for results generated by CodeQL but a few people have encountered it when running 3rd party tools that may be "noisy".

### DIFF
--- a/upload-sarif/action.yml
+++ b/upload-sarif/action.yml
@@ -3,7 +3,7 @@ description: 'Upload the analysis results'
 author: 'GitHub'
 inputs:
   sarif_file:
-    description: The SARIF file or directory of SARIF files to be uploaded.
+    description: The SARIF file or directory of SARIF files to be uploaded. Each SARIF file should contain a maximum of 1000 results, any additional results are ignored.
     required: false
     default: '../results'
   checkout_path:
--- a/upload-sarif/action.yml
+++ b/upload-sarif/action.yml
@@ -3,7 +3,7 @@ description: 'Upload the analysis results'
 author: 'GitHub'
 inputs:
   sarif_file:
-    description: The SARIF file or directory of SARIF files to be uploaded. Each SARIF file should contain a maximum of 1000 results, any additional results are ignored.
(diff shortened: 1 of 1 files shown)


upload-sarif/action.yml
```diff
@@ -3,7 +3,7 @@ description: 'Upload the analysis results'
 author: 'GitHub'
 inputs:
   sarif_file:
-    description: The SARIF file or directory of SARIF files to be uploaded.
+    description: The SARIF file or directory of SARIF files to be uploaded. Each SARIF file should contain a maximum of 1000 results, any additional results are ignored.
```
robertbrignull, line 6: I worry saying "each SARIF file" is misleading. If you upload a directory of files then it doesn't upload them separately but instead first concatenated them into one big SARIF file and uploads that. It would be more accurate to say "A maximum of 1000 results will be processed and any additional results will be ignored."

Does that seem ok?

hubwriter, line 6:
```suggestion
    description: The SARIF file or directory of SARIF files to be uploaded. Each upload should contain a maximum of 1000 results, any additional results are ignored.
```

hubwriter (approved):

upload-sarif/action.yml
felicitymay: Ah. Thanks. I hadn't realized that you could have more than one file before looking at the action, so wondered if this was a way to work around the 1000 result limit.

felicitymay: I've updated the description with Alistair's suggestion.

robertbrignull (approved):

felicitymay:
Many thanks for the rapid turnaround on reviews @hubwriter and @robertbrignull

felicitymay:
I don't have permission to merge this. Can I leave it in your hands @robertbrignull?

PR #370, opened 2021-01-21
Do not warn users if a workflow cannot be read

simon-engledew:
Hide these warnings:

image

This will still send a LintFailed message status report.


robertbrignull:
> This will still send a LintFailed message status report.

I don't think this will send a `LintFailed` message status report. If `getWorkflow` returns undefined then it does nothing. You want to change `getWorkflowErrors` to do
```typescript
if (workflow === undefined) {
  return [WorkflowErrors.LintFailed];
}
```
Ideally we'd also have different kinds of `LintFailed` so we can tell between errors in the workflow and failing to read the workflow.

robertbrignull:
I agree this will silence the error message. Up to you if you want to also upload the failure message in the status report or just leave it.

simon-engledew:
> I don't think this will send a LintFailed message status report. If getWorkflow returns undefined then it does nothing. 

Ugh, sorry - quite right. I should have actually checked this. 🤦 🙇

I've had a bit of a refactor and tweaked the way we report errors to include the exception string if something failed during linting

src/actions-util.ts
```diff
@@ -317,17 +316,23 @@ export function validateWorkflow(doc: Workflow): CodedError[] {
...
+      }
 
-    return validateWorkflow(workflow);
+      return `warning: ${formatWorkflowCause(workflowErrors)}`;
+    } catch (e) {
+      return `error: getWorkflowErrors() failed: ${e.toString()}`;
+    }
   } catch (e) {
-    return [WorkflowErrors.LintFailed];
+    return `error: getWorkflow() failed: ${e.toString()}`;
```
robertbrignull, line 335: To avoid this try-catch covering more than necessary you could do
```typescript
let workflow: Workflow;
try {
  workflow = await getWorkflow();
} catch (e) {
  return `error: getWorkflow() failed: ${e.toString()}`;
}
```
and that way you can assign a variable to a non-nullable type but have the try-catch be more contained so it's easier to read and you won't unexpectedly catch something you didn't mean to.

robertbrignull (approved):
One suggestion about a try-catch but LGTM regardless

src/actions-util.ts
simon-engledew: Oh, it's nice that the type checker is sophisticated enough to realise that the variable is not nullable!

I've used that technique to add more fine grained handling over the whole function, thanks! :+1:

robertbrignull (approved):

PR #371, opened 2021-01-21
Move .github/workflows back where it was

robertbrignull:
We're getting a lot of "Could not read workflow" errors on the pr-checks because they move the repository files out of the way in order to move some test files back in. We can avoid the error by making sure we don't move the workflow file that's executing, or at least we move it back to where it was.


simon-engledew (approved):
Great! LGTM

PR #374, opened 2021-01-25
Upload debug logs automatically when `ACTIONS_STEP_DEBUG` is enabled.

chrisgavin:
This change ensures that CodeQL debug logs are printed when running with `ACTIONS_STEP_DEBUG` enabled. I was initially going to upload them as an artifact but it actually seemed more convenient to include them in the action output. The logs will then be included when using the "download log archive" option and that archive will contain all relevant information for debugging.

I included each log file in an output group, but really this just makes them easier to find by grepping the output file since in practice they tend to be so big that the online log viewer truncates them.


src/analyze-action.ts
```diff
@@ -115,6 +120,27 @@ async function run() {
...
+      core.info("Debug mode is on. Printing CodeQL debug logs...");
+      for (const language of config.languages) {
+        const logsDirectory = util.getCodeQLDatabasePath(
+          config.tempDir,
+          language
+        );
+        const logGlobber = await glob.create(
+          path.join(logsDirectory, "log", "**")
+        );
+        const logFiles = await logGlobber.glob();
```
robertbrignull, line 134: I wonder if we need this library just to walk the file tree. We already have `findSarifFilesInDir` in `upload-lib.ts` which recursively finds SARIF files and it's not that complicated (it ignores symlinks). You could copy that implementation (or make a util method if you like).

robertbrignull (approved):
One comment about whether to use a library to find the logs files or do it with plain `fs` methods. LGTM regardless. I tried it out and the output looks fine.

src/analyze-action.ts
chrisgavin: :+1: That sounds sensible. I've used a manual walk instead of the glob library.

chrisgavin:
I've also had to move the output to `stdout` instead of `stderr`. I'm not a huge fan of logging to `stdout`, but it seems like you have to since the Actions commands go to `stdout`, so if you don't also use it the grouping can be a bit incorrect depending on when each stream gets flushed.

robertbrignull (approved):
👍

PR #377, opened 2021-01-26
Fix the formatting of a warning message.

chrisgavin:
The GitHub version number was converted from a string to an object but the warning messages were not updated. Currently they are displayed as:

> The CodeQL Action version you are using is too old to be compatible with GitHub Enterprise [object Object]. If you experience issues, please upgrade to a more recent version of the CodeQL Action.


robertbrignull (approved):
Thanks for finding this.

robertbrignull:
Would be nice if there was a query for "formatting an object as a string" like this. Maybe there is and we're just not running it.

PR #385, opened 2021-02-01
Allow overriding path of temporary files.

chrisgavin:
This makes it so you can override the temporary directory used by the Action by setting the `CODEQL_ACTION_TEMP` environment variable.

I went with an environment variable instead of an input, both because I think it's advantageous to be able to override this for all jobs on self-hosted runners and because it allows you to set it once for the entire workflow rather than for each step.

This option is a useful escape hatch for Windows hosted runners, which have a very fast, but also very small temporary disk. This is used by default, but if the project is too big to build on this disk it's potentially useful to be able to use the larger, slower disk instead.

- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary. (I've opted not to update the readme; I don't think we necessarily want to advertise this option yet, but I think it's useful to have it ready if someone eventually needs it.)


src/actions-util.ts
```diff
@@ -45,6 +45,11 @@ export function getRequiredEnvParam(paramName: string): string {
   return value;
 }
 
+export function getTemporaryDirectory(): string {
+  const value = process.env["CODEQL_ACTION_TEMP"];
+  return value !== undefined ? value : getRequiredEnvParam("RUNNER_TEMP");
```
aeisenberg, line 50: If someone explicitly sets the environment variable to `''` then we will fail here. Is this your intended behaviour? I've seen patterns where people set environment variables to an empty string instead of unsetting them.

chrisgavin: Hopefully most people know better than to do that, but I've added some handling for it.

aeisenberg (approved):
This looks good to me.

chrisgavin:
I believe the Rubocop CI check failure is caused by a bug in the latest release of Bundler (rubygems/rubygems#4380). We might want to pin to a specific version in future, but for now I think we can just wait for an update.

PR #391, opened 2021-02-12
Add capability to specify auth from env var or stdin

aeisenberg:
This commit adds two new ways of specifying GitHub auth:

1. from the GITHUB_TOKEN environment variable
2. from standard input

This commit does not include any documentation changes and the
descriptions of new command line options will need to be tweaked.
I will work on that later.


src/util.ts
```diff
@@ -301,3 +302,60 @@ export function apiVersionInRange(
...
+  githubAuth: string | undefined,
+  fromStdIn: boolean | undefined,
+  readable = process.stdin as Readable
+): Promise<string> {
+  if (githubAuth && fromStdIn) {
+    throw new Error("Cannot specify both '--github-auth' and '--with-token'.");
+  }
+
+  if (githubAuth) {
+    return githubAuth;
```
adityasharad, line 333: Can we log a deprecation warning here? As a reference, Docker uses `WARNING! Using --password via the CLI is insecure. Use --password-stdin.` (We could also adopt that naming convention: `--github-auth-stdin`).

src/util.ts
```diff
@@ -301,3 +302,60 @@ export function apiVersionInRange(
...
+ *
+ * @return a promise resolving to the auth token.
+ */
+export async function getGitHubAuth(
+  githubAuth: string | undefined,
+  fromStdIn: boolean | undefined,
+  readable = process.stdin as Readable
+): Promise<string> {
+  if (githubAuth && fromStdIn) {
+    throw new Error("Cannot specify both '--github-auth' and '--with-token'.");
```
adityasharad, line 329: Recommend that the user uses the stdin option.

src/util.ts
```diff
@@ -301,3 +302,60 @@ export function apiVersionInRange(
...
+  }
+
+  if (fromStdIn) {
+    return new Promise((resolve, reject) => {
+      let token = "";
+      readable.on("data", (data) => {
+        token += data.toString("utf8");
+      });
+      readable.on("end", () => {
+        token = token.split(/\s+/)[0].trim();
```
adityasharad, line 343: Make sure you don't crash here if it's empty.

src/util.ts
```diff
@@ -301,3 +302,60 @@ export function apiVersionInRange(
...
+        reject(err);
+      });
+    });
+  }
+
+  if (process.env.GITHUB_TOKEN) {
+    return process.env.GITHUB_TOKEN;
+  }
+
+  throw new Error("No GitHub authentication token is specified.");
```
adityasharad, line 360:
```suggestion
  throw new Error("No GitHub authentication token was specified. Please provide a token via the GITHUB_TOKEN environment variable, or by adding the `--<flagname>` flag and passing the token on standard input.");
```

src/runner.ts
```diff
@@ -442,9 +451,13 @@ program
   )
   .requiredOption("--ref <ref>", "Name of ref that was analyzed. (Required)")
   .requiredOption("--github-url <url>", "URL of GitHub instance. (Required)")
-  .requiredOption(
+  .option(
     "--github-auth <auth>",
-    "GitHub Apps token or personal access token. (Required)"
+    "GitHub Apps token or personal access token. This option is insecure...(WIP come up with a better description)"
```
adityasharad, line 456: Suggest something like "Using this option is insecure, and will be deprecated. Please use --<safer option>."

adityasharad (commented):
Thanks! Generally looks sensible, suggestions are mainly about user-facing wording.

aeisenberg:
All your suggestions make sense. Let me work on that.

src/util.ts
aeisenberg: I don't see how there is a chance of crashingl

adityasharad: Ah if `token` is empty, split gives a singleton list. Okay nothing to see here :)

adityasharad: Ah if `token` is empty, split gives a singleton list. Okay nothing to see here :)

aeisenberg:
I missed an instance of `--github-auth`. It's added now.

robertbrignull (commented):
Haven't formally approved yet. I'll let @adityasharad do that once the requirements are nailed down. But just saying that the code here LGTM from a low-level code point of view.

aeisenberg:
Is there any documentation change that needs to occur? Or are the new cli options sufficient?

adityasharad (approved):
Looks good. Docs team suggested we file an issue internally to update the Code Scanning help.

JLLeitschuh:
Is there a plan to issue a CVE number for this?

PR #392, opened 2021-02-13
Add detection for GitHub AE.

chrisgavin:
This adds detection for GitHub AE, mostly to avoid showing warnings about compatibility. There could in fact be compatibility issues, but since GitHub AE is unversioned we have to assume there is not, and the GitHub AE instance is as up-to-date as GitHub.com is.


src/util.ts
```diff
@@ -221,6 +221,7 @@ let hasBeenWarnedAboutVersion = false;
 
 export type GitHubVersion =
   | { type: "dotcom" }
+  | { type: "ghae" }
   | { type: "ghes"; version: string };
```
aeisenberg, line 225: Have you considered converting `type` into an enum?

chrisgavin: I was just extending the existing variable, so I was going to leave it as-is, but I've pushed another commit to convert it to an enum.

robertbrignull (approved):

PR #393, opened 2021-02-15
Also look for the CodeQL bundle at the custom GitHub AE endpoint.

chrisgavin:
This updates the CodeQL bundle finding logic to work on GitHub AE, where we always have to use a custom endpoint to get the CodeQL bundle rather than going to the repository directly.


src/codeql.ts
```diff
@@ -291,11 +317,17 @@ export async function setupCodeQL(
...
+      const parsedQueryString = queryString.parse(parsedCodeQLURL.search);
       const headers: IHeaders = { accept: "application/octet-stream" };
       // We only want to provide an authorization header if we are downloading
       // from the same GitHub instance the Action is running on.
       // This avoids leaking Enterprise tokens to dotcom.
-      if (codeqlURL.startsWith(`${apiDetails.url}/`)) {
+      // We also don't want to send an authorization header if there's already a token provided in the URL.
+      if (
+        codeqlURL.startsWith(`${apiDetails.url}/`) &&
+        parsedQueryString["token"] === undefined
```
aeisenberg, line 329: Should we be concerned about the security of sending the token as part of the query string? The url will be encrypted, but there is a possibility that the token will be logged and secret scanning may or may not find it.

It's low risk, but would it be better to remove the token from the query string and add it as an `authorization` header?

chrisgavin: I don't think we get control over how authentication to the assets server is done, so unless we can change what the server accepts we're probably stuck with sending the token in the URL.

Given that the token only grants access to the CodeQL bundle, which is already publicly available on GitHub.com I think the risk of someone obtaining it is pretty minimal.

aeisenberg: I'm fine with that.

aeisenberg (approved):
Looks reasonable to me.

PR #396, opened 2021-02-16
Increase the default amount of RAM reserved for the OS

adityasharad:
Mitigation for OOM errors (137/SIGKILL) seen by users when we overcommit the available memory.
For Unix, reserve 1GB.
For Windows, reserve 1.5GB, as the OS needs more memory and estimates inaccurately.

Potential impact: There's a risk of this slightly increasing run times, but I think 1-1.5GB is rarely enough to dramatically change performance, and we've discussed that the improved stability is worth the tradeoff. I'm doing some simple performance testing, but to evaluate the full impact we may need to ship this and monitor overall runtimes.


chrisgavin (approved):
LGTM. I suppose there is some risk that repositories that could previously be analyzed with 6.75GB of RAM might not be able to be analyzed with 6GB, but I agree that improving stability is the most important thing.

PR #397, opened 2021-02-17
Set externalRepoAuth

robertbrignull:
@aeisenberg, I noticed in https://github.com/github/codeql-action/pull/395 I unfortunately don't think it'll have the effect we wanted. I believe we wanted the standard token input to be used for normal auth and the external repository token. Currently that means you need to set it for both fields. I realise this is a bit error prone. If you have ideas for how to improve this situation do please go ahead.


aeisenberg:
I missed that. Thanks for catching. I'll take this over.

aeisenberg:
I think this solution is fine. Another option would be to fall back on the default API key [here](https://github.com/github/codeql-action/blob/cb574a7d600f13c14044c6f670a05ab1ed07d318/src/external-queries.ts#L54-L57) if `externalRepoAuth`, but I don't think that's any better than what you have.

robertbrignull:
Thanks for pushing the updated JS. I could have swore I ran the build but apparently not.

> Another option would be to fall back on the default API key here if externalRepoAuth

Yeah that's another option. The only downside of that is we then have to implement the fallback in each place that uses it.

aeisenberg (approved):

PR #400, opened 2021-02-23
Don't run update release branch workflow on forks.

chrisgavin:

### DIFF
--- a/.github/workflows/update-release-branch.yml
+++ b/.github/workflows/update-release-branch.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   update:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


robertbrignull (approved):
Seems reasonable as forks shouldn't need to use this workflow, but at the same time it's not harmful. Should the same change be applied to some of the other more "operational" workflows, e.g. `check-expected-release-files`, `release-runner`, `split`, `update-supported-enterprise-server-versions`?

chrisgavin:
The main reason that I modified this workflow specifically is that it creates a PR that mentions everyone who's contributed to the repository. In general I don't think we mind if people are running the other workflows, even if it is a bit pointless.

PR #409, opened 2021-03-07
When downloading the CodeQL bundle, only use the GitHub AE endpoint on GitHub AE, and check it first.

chrisgavin:
This makes the requests to find the CodeQL bundle a bit more efficient. Firstly there's no point trying the GitHub AE endpoint when we're not on GitHub AE, so I've threaded through the GitHub variant and skipped that endpoint unless it's GHAE.

Secondly, if we are on GitHub AE we should check that endpoint first since the other endpoints are much less likely to work. 


src/init-action.ts
```diff
@@ -101,9 +101,7 @@ async function run() {
   };
 
   const gitHubVersion = await getGitHubVersion(apiDetails);
-  if (gitHubVersion !== undefined) {
```
chrisgavin, line 104: I think these are safe to remove. `getGitHubVersion` seems to always return an answer.

robertbrignull (approved):
LGTM

PR #424, opened 2021-03-17
Ensure error correct error message on 403 error

aeisenberg:
This ensures that error messages for 403 errors on upload display the raw error to the user. Also, all errors are written to the console log so that they can more easily be debugged.

I have not been able to reproduce the exact problem discussed, but this seems to fix it. Any help with reproducing would be appreciated.


robertbrignull (approved):
I agree this should fix the issue of the 403 message, and doesn't affect any other messages. Not sure whether we should or not, as we could get the same problem again but then also some of the error messages the API returns are a bit unhelpful. I think keeping this PR as it is is fine, and we can always change error message display again in the future.

Not the exact same scenario as the original bug, but I tried it out to demonstrate it showing a different error message.

Before:
Screenshot 2021-03-17 at 10 53 59

After:
Screenshot 2021-03-17 at 10 54 07

aeisenberg:
Thanks for trying this out. I was able to look at this in a similar (but different) situation and it looks fine.

PR #426, opened 2021-03-17
Display better error message on invalid sarif

aeisenberg:
Specifically, some third party tools do not include a `results`
block for runs when there is an error. This change adds a more
explicit error message for this situation.


src/upload-lib.ts
```diff
@@ -177,6 +177,9 @@ function getSarifFilePaths(sarifPath: string) {
 export function countResultsInSarif(sarif: string): number {
   let numResults = 0;
   for (const run of JSON.parse(sarif).runs) {
```
adityasharad, line 179: Shall we also handle the case where `runs` is null/undefined? It's valid SARIF according to https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10540920 but I think we currently crash rather than returning 0.

adityasharad (approved):
Looks sensible, one further suggestion.

src/upload-lib.ts
aeisenberg: Sure. Probably a good idea to be more defensive in general. I'll take a look and see if there are any more simple changes to make that will strengthen our handling of invalid sarif.

chrisgavin (approved):
Yep, this looks like a good change to me.

> Specifically, some third party tools do not include a results block for runs when there is an error.

Some tools unfortunately just don't include a results block at all if there were no results, which this change should catch. Hopefully if there's an error the tool should fail with a non-zero exit code and the SARIF upload shouldn't be attempted at all.

PR #428, opened 2021-03-22
Fix race condition with actions/checkout@v1

simon-engledew:
Work around a race condition in Actions where sometimes `GITHUB_SHA != git rev-parse head` after an `actions/checkout@v1`

image

This leads to the upload incorrectly rewriting the ref, causing analyses to go missing. 😭 

Do not rewrite the ref when we:

* are on GITHUB_SHA
* are on GITHUB_REF


src/actions-util.ts
```diff
@@ -425,19 +425,27 @@ export async function getRef(): Promise<string> {
...
+  // in actions@v1 this may not be true as it checks out the repository
+  // using GITHUB_REF. There is a subtle race condition where
+  // git rev-parse GITHUB_REF != GITHUB_SHA, so we must check
+  // git git-parse GITHUB_REF == git rev-parse HEAD instead.
+  const hasChangedRef = sha !== head && (await getCommitOid(ref)) !== head;
+
+  if (pull_ref_regex.test(ref) && hasChangedRef) {
+    const newRef = ref.replace(pull_ref_regex, "refs/pull/$1/head");
+    core.info(
+      `No longer on merge commit, rewriting ref from ${ref} to ${newRef}.`
```
rneatherway, line 446: Won't this be visible to users of the action? Should it be?

src/actions-util.ts
```diff
@@ -425,19 +425,27 @@ export async function getRef(): Promise<string> {
...
-  const checkoutSha = await getCommitOid();
-
-  if (
-    pull_ref_regex.test(ref) &&
-    checkoutSha !== getRequiredEnvParam("GITHUB_SHA")
-  ) {
-    return ref.replace(pull_ref_regex, "refs/pull/$1/head");
+  const head = await getCommitOid("HEAD");
+  // in actions@v2 we can check if git rev-parse HEAD == GITHUB_SHA
+  // in actions@v1 this may not be true as it checks out the repository
```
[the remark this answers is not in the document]
simon-engledew: Whoops! Thanks

src/actions-util.ts
simon-engledew: I'll drop it to debug.

rneatherway (approved):

rneatherway (approved):

PR #435, opened 2021-03-30
Add special error message case for dependabot

robertbrignull:
This error currently comes through as "Resource not accessible by integration" but it would be helpful if we said something better. I had a hard time writing this error message as I wasn't sure what to say. If other people have other ideas please do let me know.

Here's an example of what it looks like in practice: https://github.com/dsp-testing/robertbrignull-dependabot-test/pull/7/checks?check_run_id=2227856119

I also considered putting some example configuration like `branches-ignore: ['dependabot/**']` in the message but then decided against it. I think that would be a bit hard to read in a single-line message like this and I worry about giving explicit practical advice as it's never going to be right for every workflow.


cannist:
I fear the error message might be slightly confusing if the user does not have enough context already. But I cannot come up with something better in the short space we have for an error message. Although perhaps you could add "Uploading code scanning results requires write access." as second sentence?

I wonder if we should have something like a FAQ entry that explains the situation and exemplifies a solution in more detail  that we could link to in the error message?
Do we have a FAQ page or something similar for code scanning or the CodeQL action? I cannot seem to find anything like that.

src/actions-util.ts
```diff
@@ -628,7 +628,15 @@ export async function sendStatusReport<S extends StatusReportBase>(
     if (isHTTPError(e)) {
       switch (e.status) {
         case 403:
-          core.setFailed(e.message || GENERIC_403_MSG);
+          if (isDependabotActor()) {
+            core.setFailed(
+              'Workflows triggered by Dependabot on the "push" event run with read-only access. ' +
+                'To use Code Scanning with Dependabot please ensure you are using the "pull_request" event for this workflow and avoid triggering on the "push" event for dependabot branches. ' +
+                "See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags for more information on how to configure these events."
```
adityasharad, line 635: Minor tweaks, but in general I think this is good.
Should we link instead to https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning#scanning-pull-requests? (That in turn links to the URL you have here.)

```suggestion
              'Workflows triggered by Dependabot on the "push" event run with read-only access. ' +
                'To use Code Scanning with Dependabot, please ensure you are using the "pull_request" event for this workflow and avoid triggering on the "push" event for Dependabot branches. ' +
                "See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags for more information on how to configure these events."
```

adityasharad:
https://docs.github.com/en/code-security/secure-coding/troubleshooting-the-codeql-workflow is the closest to an FAQ page. We can put out a docs update there too.

src/actions-util.ts
```diff
@@ -628,7 +628,15 @@ export async function sendStatusReport<S extends StatusReportBase>(
     if (isHTTPError(e)) {
       switch (e.status) {
         case 403:
-          core.setFailed(e.message || GENERIC_403_MSG);
+          if (isDependabotActor()) {
+            core.setFailed(
+              'Workflows triggered by Dependabot on the "push" event run with read-only access. ' +
```
aeisenberg, line 633: I notice here that you don't actually check for a `push` event. It's _probably_ safe to assume that it is, but it's also an easy check to make. How confident are you that this can _never_ happen on `pull_request` events?

src/actions-util.ts
robertbrignull: Hmm, I agree https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning#scanning-pull-requests could be good. Would it better to link to the `scanning-on-push` section which is immediately above? That way I expect a user would read that and the one below, and there's a big warning box about scanning pull requests. If we link to the pull requests section I worry stuff above that might be missed.

rneatherway:
Something that concerns me a bit about this error message is that it is most problematic when the branch gets squash-merged into the default branch, and the wording doesn't quite account for that.

robertbrignull:
> Something that concerns me a bit about this error message is that it is most problematic when the branch gets squash-merged into the default branch, and the wording doesn't quite account for that.

I don't know how to explain this aspect in a message like this. This is further evidence that we should opt for a docs troubleshooting page and link there I think.

src/actions-util.ts
```diff
@@ -655,6 +664,16 @@ export async function sendStatusReport<S extends StatusReportBase>(
   }
 }
 
+// Was the workflow run triggered by a `push` event, for example as opposed to a `pull_request` event.
+function workflowIsTriggeredByPushEvent() {
+  return process.env["GITHUB_EVENT_NAME"] === "push";
```
robertbrignull, line 669: I've gone with `process.env` instead of `getRequiredEnvVar` here because if the env var for some reason doesn't exist I'd rather it returned false instead of crashing.

aeisenberg:
I think this is good to go in (almost) as is and then we can add a FAQ entry for troubleshooting. When that's available, we can update the link here to point to the FAQ.

Aditya had two small grammar changes that should probably be fixed and then 🚢 .

adityasharad (approved):

PR #440, opened 2021-04-07
Add external git repositories to search path for custom queries

edoardopirovano:
When running a query from an external repository, we may need some additional dependencies that are specified in this repository. To find these, we should add the the root of the external repository to the search path in our CLI invocations. At the moment, this is done in `resolve queries` but not in `database analyze`. This PR adds the search path in the latter call. As a result of this, we may perform multiple `database analyze` invocations (one for each external repository, with the respective search path in each one) and have to merge the SARIF results into one file in order to not break consumers who are expecting all custom query results to be in a single file.


src/analyze.ts
```diff
@@ -154,48 +156,43 @@ export async function runQueries(
...
+          sarifFolder,
+          undefined
+        );
+        statusReport[`analyze_builtin_queries_${language}_duration_ms`] =
+          new Date().getTime() - startTimeBuliltIn;
+      }
+      const startTimeCustom = new Date().getTime();
+      const temporarySarifDir = actionsUtil.getTemporaryDirectory();
+      const temporarySarifFiles: string[] = [];
+      for (let i = 0; i < queries["custom"].length; ++i) {
```
aeisenberg, line 174: nit: Using a `for..of` loop here might be nicer.

src/config-utils.ts
```diff
@@ -46,11 +46,23 @@ type Queries = {
   [language: string]: {
     /** Queries from one of the builtin suites */
     builtin: string[];
+
     /** Custom queries, from a non-standard location */
-    custom: string[];
+    custom: QueriesWithSearchPath[];
```
aeisenberg, line 51: Does this change the format of the config file? We are going to need to support the old format as well.

aeisenberg (commented):
Sorry, this slipped through the cracks. Reviewed now. I don't think we can be changing the format of the config file.

src/analyze.ts
```diff
@@ -154,48 +156,43 @@ export async function runQueries(
...
+        if (queries["custom"][i].queries.length > 0) {
+          await runQueryGroup(
```
aeisenberg, line 176: Something I just noticed now is that this code is `await`ing in a loop. `runQueryGroup` is a long running operation and it's too bad we are serializing this. We've always had this problem in this part of the code. An ideal solution would run this in a loop and wrap it in `Promise.all()`. This could probably save a lot of time since we're parallelizing all the IO operations.

edoardopirovano: Ah that seems like a good idea, yes. Though is it safe to have concurrent CLI invocations on the same database? Won't they share the same cache and run into issues?

src/analyze.ts
aeisenberg: I missed that. Good point.

src/analyze.ts
aeisenberg: Right...not sure of that. There might be some better parallelization we can do, but best not mess with it right now.

src/config-utils.ts
edoardopirovano: As I understand it, there are two types of config files that we store. The configuration that the user supplies is stored in a `UserConfig` object, which this PR **does not** change. When the `init` task is run, this gets read in and further processed (by carrying out some computation such as resolving queries) into a `Config` object, which this PR **does** change (by augmenting it with extra information about the search path). This `Config` object is stored in a temporary directory on the runner and used by later tasks. From my understand of how Actions work, data in this temporary runner directory is wiped between different runs of the action so we do not need worry about finding an outdated `Config` object.

aeisenberg: Ah....that makes sense Thanks.

aeisenberg (approved):
Looks good now that I understand it all.

edoardopirovano:
Pushed a rebase to fix conflict, will merge once tests pass.

PR #441, opened 2021-04-08
PR checks: Run integration tests against both `tools: null` and `tools: latest`

adityasharad:
Always test against both the default and latest CodeQL bundle.

This improves test coverage shortly after a CodeQL bundle release, where the latest bundle
may not yet be built into the Actions VM image as the default bundle.

It also saves a manual step during bundle release testing,
since we no longer need to temporarily change the PR checks to `tools: latest`.

There is some redundancy when the latest bundle is the same as the default bundle on the VM image,
but this can be considered a test for the `tools: latest` configuration.

- [N/A] Confirm the [readme](https://github.com/github/codeql-action/blob/master/README.md) has been updated if necessary.


aeisenberg (commented):
So, right now it looks like both `latest` and `~` are the same. It would be really nice if you could detect this situation and skip one of the tests. I wouldn't say this is necessary, though.

adityasharad:
That would be good. Outline of how that might work:
- create an initial job that runs before the test jobs
- run the `init` action twice, once with `tools: latest` and once without, and record the `codeql-path` outputs each time
- run `codeql version` from both paths, and output the array of unique values
- matrix the remaining test jobs over this array

adityasharad:
Added a commit that implements the above idea, comparing `tools: null` and `tools: latest`, and only running the matrixed test against the distinct versions. This prevents duplication in the PR tests. I haven't yet modified the `codeql.yml` workflow.

.github/workflows/codeql.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest,windows-latest,macos-latest]
+        tools: [~, latest]
```
aeisenberg, line 14: Do you want to call `check-codeql-versions` here as well?

aeisenberg (commented):
This looks good. Is there a way you can force it to run against two versions? It _looks_ like the job name includes the matrixed parameters.

.github/workflows/codeql.yml
adityasharad: Done, just copied the code over. This means we might see many extra `init` calls from this repo, and might have alerts wobbling if the two bundles give different results, but I think that is both rare and something we can deal with.

aeisenberg (approved):
Nice. Code looks good. Have you tried running this when the versions are actually different?

adityasharad:
Tested in a draft PR that changed the referenced bundle to an older release, and verified that it produces `[null, "latest"]` as the matrix and starts parallel jobs with the right versions: https://github.com/github/codeql-action/pull/444/checks?check_run_id=2310079094

Issue #418, opened 2021-03-12
Unable to upload SARIF file


aetos382:
When I attempted to upload a sarif file using the `upload-sarif@v1` action, I encountered the following error.

> Unable to upload "/home/runner/work/path-to.sarif" as it is not valid SARIF:
> \- instance.runs[0].tool.driver.language does not match pattern "^[a-zA-Z]{2}|^[a-zA-Z]{2}-[a-zA-Z]{2}]?$"

The language property of my sarif file is `ja-JP`, which is a valid form of language tag, but it did not match the regular expression pattern described above.

I think the regular expression pattern for this schema is a mistake.
https://github.com/github/codeql-action/blob/cfec2bbc35f12252507dd0de0e172c08b65e8c13/src/sarif_v2.1.0_schema.json#L2342

I think this schema is coming from https://github.com/oasis-tcs/sarif-spec .
So, I'm submitting an issue to that repository as well.
https://github.com/oasis-tcs/sarif-spec/issues/488

aeisenberg:
Yes, that does seem suspicious. For now, can you change the language property to `ja`? I will investigate further.

aeisenberg:
Apologies for not following up until now. I think the pattern we would want to use is this: `^[a-zA-Z]{2}(-[a-zA-Z]{2})?$`.

@lcartey, does this look right to you?

aetos382:
@aeisenberg 

> For now, can you change the language property to ja?

My sarif file was generated by the C# compiler, so I can't change the language tag.

I have now disabled this action in my workflow, as it was not important for my project to use the compiler generated sarif file.

> I think the pattern we would want to use is this: ^[a-zA-Z]{2}(-[a-zA-Z]{2})?$

Yes. I also think that pattern is correct.

stephenfuqua:
Still getting this error with [release 2.9.4](https://github.com/github/codeql-action/releases/tag/codeql-bundle-20220615). My situation is also using a sarif file generated by the package `Microsoft.CodeAnalysis` in a .NET project. [Example build failure](https://github.com/stephenfuqua/Ed-Fi-Analytics-Middle-Tier/runs/6974054613?check_suite_focus=true). Locally I see that the sarif file has "en-US" for the language. I will continue to investigate as time is available, to see if I can find a workaround on my side or find and report on the root cause.

aeisenberg:
[...]

PR #442, opened 2021-04-09
Fixes a regex for language and locale recognition

aeisenberg:
See https://github.com/oasis-tcs/sarif-spec/pull/490
See #418

Note that this changes the sarif spec file. Unless this
change is actually merged in the sarif spec repo, the
version used by the action will be slightly different.


chrisgavin (approved):
Awesome! Thank you very much! Hopefully the fix gets merged upstream too, but it makes sense to me to fix it here.

PR #443, opened 2021-04-09
Use externalRepoAuth when getting a remote config

aeisenberg:
This allows users to specify a different token for retrieving the
codeql config from a different repository.


adityasharad (approved):
Looks good, thanks for the quick fix.
I see we were already passing around a`GitHubApiExternalRepoDetails` to check out query repos configured in the workflow file, so that case is already correct.

aeisenberg:
> I see we were already passing around a`GitHubApiExternalRepoDetails` to check out query repos configured in the workflow file, so that case is already correct.

That's right. It's handled.

PR #446, opened 2021-04-16
Add automationdetails id to runs

Daverlo:
We are starting the process of deprecating the use of `analysis_key` and `environment` in code scanning, and instead use the [`automationDetails`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Toc16012479) property. This PR starts populating this field, using the `analysis_key` and `environment`.


src/upload-lib.test.ts
```diff
@@ -131,3 +131,36 @@ test("finding SARIF files", async (t) => {
...
+
+  // check that an empty environment produces the right results
+  expectedSarif =
+    '{"runs":[{"automationDetails":{"id":".github/workflows/codeql-analysis.yml:analyze/"}}]}';
+  modifiedSarif = uploadLib.populateRunAutomationDetails(
+    sarif,
+    analysisKey,
+    "{}"
+  );
+  t.deepEqual(modifiedSarif, expectedSarif);
```
starcke, line 165: Should we add  test to assert that if there already is an ID then it is not overwritten?

starcke (commented):
Generally looks good to me. Asked for one more test. It'll probably be good to get a sanity check from someone that have worked on the action.

src/upload-lib.ts
```diff
@@ -40,6 +40,38 @@ export function combineSarifFiles(sarifFiles: string[]): string {
...
+    }
+  }
+
+  const sarif = JSON.parse(sarifContents);
+  for (const run of sarif.runs || []) {
+    if (run.automationDetails === undefined) {
+      run.automationDetails = {
+        id: automationID,
+      };
+    }
```
aeisenberg, line 69: Do you also want to handle the case where `run.automationDetails` is defined, but `id` is missing? Eg-

```
run: {
  automationDetails: { }
}
```

Daverlo: My initial idea was not to handle that case. The are some other fields in the object that serve the same purpose (`guid` and `correlationGuid`), so one option would be to recreate the `id` based on that, but following the [example](https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Toc16012479) is not the case.

```
"id": "Nightly CredScan run for sarif-sdk/master/x86/debug/2018-10-05",
"guid": "11111111-...",
"correlationGuid": "22222222-...."
```

The `correlationGuid` identifies the `Nightly CredScan run for sarif-sdk/master/x86/debug/` part of the `id` but there is no direct way of getting one from the other.

starcke (approved):
:+1:

PR #452, opened 2021-04-20
Ignore non-string values in populateRunAutomationDetails

Daverlo:
This PR is an addition to https://github.com/github/codeql-action/pull/446

We have found that in code scanning we do not process the non-string environment values, so we have to do the same in the action, in order to not cause a baseline mismatch.


starcke (approved):
:+1:

PR #456, opened 2021-04-22
Introduce our own toolcache implementation for use by the runnner

robertbrignull:
This PR aims to help with our use of actions libraries outside of the actions environment. Some of them are safe to use in this way, but some are not. Notably `@actions/toolcache` makes assumptions about being run on actions. Possibly this dependence is limited to just environment variables and hence we've been working around it by setting those environment variables. However this is not an ideal situation and things could break in the future if the toolcache library changes its assumptions, so I think we should aim to avoid using it outside of actions.

In this PR I introduce `src/toolcache.ts` which acts as an interface to the parts of the `@actions/toolcache` library that we use. When running on actions it delegates straight to the library, and when not on actions we include our own implementation. I've just copied the code from the toolcache library javascript, converted it to typescript, and removed some cases we weren't using.

Hopefully this means that the behaviour of the runner will not change because the implementation is effectively not changing. However if the toolcache library does change in the future then we'll keep up with that because we're still using the library when running on actions. If in the future we delete the runner then we can delete the toolcache implementation in this repository and switch to just using the library everywhere and everything should be fine.

Also includes a drive-by fix to the `unguarded-action-lib.ql` query to also catch uses of `actions-util.ts`, which does flag a single case in `src/analyze.ts` where we were using that code in the runner.


package.json
```diff
@@ -23,6 +23,7 @@
     "@actions/exec": "^1.0.1",
     "@actions/github": "^4.0.0",
     "@actions/http-client": "^1.0.8",
+    "@actions/io": "^1.1.0",
```
robertbrignull, line 26: Adds this explicit dependency because the toolcache library was using it to be a recursive copy. From what I can tell there isn't a good way to do that with just builtin node libs, so we'd end up introducing a library dependency anyway. May as well be this one. I've added this library to the list of actions libraries that are ok to use as I believe it's fine.

src/toolcache.ts
```diff
@@ -0,0 +1,315 @@
...
+
+import { Logger } from "./logging";
+import { Mode } from "./util";
+
+/*
+ * This file acts as an interface to the functionality of the actions toolcache.
+ * That library is not safe to use outside of actions as it makes assumptions about
+ * the state of the filesystem and available environment variables.
+ *
+ * On actions we can just delegates to the toolcache library, however outside of
```
aeisenberg, line 18:
```suggestion
 * On actions we can just delegate to the toolcache library, however outside of
```

aeisenberg (commented):
This looks reasonable to me. I just want to be sure that everything in the else blocks of `mode === "actions"` is just copied from tool-cache? There's one code scanning issue, which I'm not sure is relevant.

robertbrignull:
Thanks for reviewing so quickly @aeisenberg.

> I just want to be sure that everything in the else blocks of `mode === "actions"` is just copied from tool-cache?

Yes, it's all copied from https://github.com/github/codeql-action/blob/main/node_modules/%40actions/tool-cache/lib/tool-cache.js and then cleaned up to make it into typescript. If you want to check I didn't make any mistakes your best bet is probably to go through that file and check the functions against the ones I add.

> There's one code scanning issue, which I'm not sure is relevant.

May as well fix it since it's so easy to do. Though worth pointing out the vulnerability, such that there is one, was already there in the toolcache library code. It's only now that the query is detecting it since we copied the code.

robertbrignull:
I also tested this out locally in an environment that definitely isn't actions. Just to make sure the CI tests weren't spuriously succeeding because the CI environment happens to be actions.

aeisenberg (approved):
Looks good.

PR #457, opened 2021-04-22
Restrict Actions token permissions in CodeQL workflow.

chrisgavin:
This both helps improve the security of the workflow and ensures we don't accidentally add a dependency on any new permissions without realizing.


adityasharad (approved):

chrisgavin:
It turns out that while this is probably still useful, it was actually passing with less permissions than it should have. This was for two reasons:

* `actions: read` is needed, but only if the repository being analyzed is private. If it is public then `actions: read` is effectively always granted.
* `security-events: write` is not needed in pull requests. This is handled somewhat magically by GitHub in the pull request case, but for pushes it has to be explicitly specified.

I feel like we can probably still test whether the required permissions have changed, but it will involve having to test changes against a private repository which is a bit of a pain.

I'll merge this anyway because it seems like it's good practice to start being more explicit about what permissions we require.

PR #458, opened 2021-04-22
Add the github-linguist package

aeisenberg:
This commit only adds a single package and all of its transitive
dependencies. The github-linguist package will be used for counting
lines of code as a baseline for databases we are analyzing.

The actual work to calculate the lines of code will come in a future PR.

This PR is surprisingly large and I'm not sure if that is a genuine
problem or not. The package-lock.json file has been upgraded from
v1 to v2 and this introduced some changes, but even when I tried to
install the package using and older version of npm, there were still
a large number of changes.


robertbrignull (approved):
LGTM

As you say, this is just adding the library. So I wouldn't worry about the size of the PR. That's just an unfortunate side effect of having the node modules in the repository.

I don't really know if the upgrading of the package lock file is a problem or not, but I'm going to assume it's not. I still have an old version of npm and I tried checking out your branch and I could still do npm things without errors.

PR #459, opened 2021-04-22
Add baseline metrics for lines of code

aeisenberg:
This commit uses a third party library to estimate the lines of code in
a database that is to be analyzed by codeql.

The estimate uses the same includes and excludes globs for determining
which files should be counted.

The lines of code count is returned by language and injected into the
SARIF as appropriate.

~~Currently, this PR adds the LoC data in the `metricResults` property of the sarif in a blob like this:~~

```js
      {
          metric: `baseline/${language}/lines-of-code`,
          value: lineCounts[language]
        }
```

~~We haven't agreed on what this will look like, so injecting the metric may change.~~

We've decided that the lines of code will be injected into metrics with id like this: `${language}/summary/lines-of-code` and a new `baseline` property is added to the metric. Languages that we have a count for, but no metric, will be ignored.


src/analyze.ts
```diff
@@ -145,6 +146,15 @@ export async function runQueries(
 ): Promise<QueriesStatusReport> {
   const statusReport: QueriesStatusReport = {};
 
+  // count the number of lines in the background
+  const locPromise = countLoc(
```
robertbrignull, line 150: This is clever, so the promise it's only resolved when it's used but is used in potentially multiple places. Worth noting that in practice there will always be some queries to evaluate so we'll always end up using this promise. It's an error for there not to be any queries to analyse and it would have error-ed back in the init step. Up to you if you therefore want to leave this as it is or potentially simplify it.

src/count-loc.ts
```diff
@@ -0,0 +1,64 @@
...
+export async function countLoc(
+  cwd: string,
+  include: string[],
+  exclude: string[],
+  dbLanguages: string[],
+  logger: Logger
+): Promise<Record<string, number>> {
+  const result = await new LocDir({
+    cwd,
+    include,
```
robertbrignull, line 36: Worth noting that we analyze everything in the current directory unless explicitly excluded. Does this also have that behaviour?

src/analyze.ts
aeisenberg: The main reason why I'm doing this is so that the line counting can happen in "parallel". On large projects, counting can take 10-20s and there's lots of disk IO, so it's nice to be able to run this while other things are happening.

src/count-loc.ts
aeisenberg: It's inheriting the include/excludes from the configuration. So, yes, the behaviour should be the same, or at least similar. The globbing rules are slightly different between the cli and the npm package, so there will be edge cases. I tried a few glob patterns from projects I found in the wild and they behave similarly with this library and the cli, but I expect there will be some differences. Hopefully nothing big.

robertbrignull: Sorry I wasn't very clear. What I meant was that the action will always analyze everything in the repo / current directory unless something is explicitly excluded. If you put something from the current directory in the `include` section it's basically a no-op. The only real use therefore is including something from outside the repository.

This is different to how I believe linguist behaves. From what I can tell if you provide nothing for the `includes` option then it'll include everything in the current directory, but if you do pass something then it'll only include that.

Sorry I've still not explained too well. Does the above make sense?

You could potentially make the behaviours more similar by passing `['**'].concat(include)` here.

aeisenberg: Oh...that makes sense. I didn't realize the behaviour. I will make the change. Thanks for pointing it out.

src/count-loc.ts
```diff
@@ -0,0 +1,64 @@
+import { LocDir } from "github-linguist";
+
+import { Logger } from "./logging";
+
+const supportedLanguages = {
```
adityasharad, line 5: This is a mapping from Linguist language names to CodeQL language prefixes in query IDs? Worth a comment.

src/count-loc.ts
```diff
@@ -0,0 +1,64 @@
...
+  c: "cpp",
+  "c++": "cpp",
+  "c#": "csharp",
+  go: "go",
+  java: "java",
+  javascript: "javascript",
+  python: "python",
+  ruby: "ruby",
+  typescript: "jypescript",
```
adityasharad, line 14: Awkwardly, although the CodeQL language identifiers are `javascript` and `python`, the conventional prefixes we appear to have adopted for query IDs targeting these languages is `js/` and `py/`... Easier to handle that here.

```suggestion
  javascript: "js",
  python: "py",
  ruby: "ruby",
  typescript: "js",
```

adityasharad (commented):
Nice and clear. Couple of recommendations based on CodeQL conventions for labelling languages.

src/count-loc.ts
adityasharad: and apparently `csharp` -> `cs`

robertbrignull (approved):
LGTM from my point of view. Probably best let @adityasharad also review the recent changes.

src/analyze.ts
```diff
@@ -269,3 +284,29 @@ export async function runAnalyze(
...
+  const lineCounts = await locPromise;
+  if (language in lineCounts) {
+    const sarif = JSON.parse(fs.readFileSync(sarifFile, "utf8"));
+    if (Array.isArray(sarif.runs)) {
+      for (const run of sarif.runs) {
+        const metricId = `${language}/summary/lines-of-code`;
+        run.properties = run.properties || {};
+        run.properties.metricResults = run.properties.metricResults || [];
+        const metric = run.properties.metricResults.find(
+          (m) => m.metricId === metricId
```
adityasharad, line 302: We deviated slightly from the original design when implementing this in the CLI, so for now I think you should look for `m.metric.id` instead of `m.metricId`. The reporting descriptors are currently inlined rather than being references.

Eventually I think we will go back to making these references with `metricId`. So perhaps it is worth handling both cases here. cc @henrymercer

aeisenberg: so, you're saying something like: 

```js
          (m) => m.metricId === metricId || m.metric?.id === metricId
```

?

src/count-loc.ts
```diff
@@ -0,0 +1,65 @@
...
+// Map from linguist language names to language prefixes used in our metrics
+const supportedLanguages = {
+  c: "cpp",
+  "c++": "cpp",
+  "c#": "cs",
+  go: "go",
+  java: "java",
+  javascript: "js",
+  python: "py",
+  ruby: "ruby",
```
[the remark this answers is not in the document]
aeisenberg: Darn...fixed that, but never pushed.

src/analyze.ts
adityasharad: Exactly!

adityasharad: might have to update your tests too

src/analyze.test.ts
```diff
@@ -103,5 +138,47 @@ test("status report fields and search path setting", async (t) => {
...
+    });
+  }
+
+  function verifyLineCountForFile(
+    lang: string,
+    filePath: string,
+    lineCount: number
+  ) {
+    const sarif = JSON.parse(fs.readFileSync(filePath, "utf8"));
+    t.deepEqual(sarif.runs[0].properties.metricResults, [
```
adityasharad, line 167: You need to update this to check your three different test `runs`: run 0 has the `metricId` form of result, run 1 has the `metric.id` form, and run 2 doesn't have it. That should fix the tests.

adityasharad (approved):

PR #463, opened 2021-04-28
Add category input

Daverlo:
Follow up to https://github.com/github/codeql-action/pull/446

This PR introduces a new action input for both the analyze and upload actions for further configurability. The category is only applied as a substitute for the default `analysis_key/environment`, and it won't be applied if the sarif already contains an automation id.


adityasharad:
Upcoming versions of `codeql database analyze` will support the `--category` flag directly and put it into the SARIF. Should we look for that mechanism when it is supported? cc @aeisenberg

src/upload-lib.test.ts
```diff
@@ -137,10 +137,31 @@ test("populateRunAutomationDetails", (t) => {
   const analysisKey = ".github/workflows/codeql-analysis.yml:analyze";
 
   let expectedSarif =
-    '{"runs":[{"automationDetails":{"id":".github/workflows/codeql-analysis.yml:analyze/language:javascript/os:linux/"}}]}';
+    '{"runs":[{"automationDetails":{"id":"language:javascript/os:linux/"}}]}';
```
aeisenberg, line 140: Is this format specifically required? Currently, the cli will produce: `javascript/linux/` (ie- `<language>/<os>/`) if the category is not specifically overriden.

aeisenberg (commented):
The cli v2.5.3 will automatically include an automationDetails.id value. It will be in the form specified in my comment. The cli will also accept a `--sarif-category` parameter that allows users to override the value with any string (though, the `/` is always appended if it's not there).

Even though v2.5.3 will support the category, sarif produced from other tools may not have a category. Also, it will be necessary to support older versions of the cli (or at least not crash on them).

I think, by default, the action should not be overriding the automationDetails.id value if one already exists, though there probably should be a way to force a custom id in there if requested.

src/upload-lib.test.ts
Daverlo: No specific format is required, that's just a test example. It is based on the default format we decided for code scanning in https://github.com/github/codeql-action/pull/446, but we support any* string.

*As long as it complies with the sarif definition (the `/` at the end if there is no `runId`, etc)

robertbrignull:
Should this be updating the runner too? My understanding of this work was it's to allow parallel builds on 3rd party CI/CD, which means people using the runner and the external `/sarifs` endpoint.

Daverlo:
>Should this be updating the runner too? My understanding of this work was it's to allow parallel builds on 3rd party CI/CD, which means people using the runner and the external /sarifs endpoint.

That makes sense, thanks!

src/upload-lib.ts
```diff
@@ -44,12 +44,47 @@ export function combineSarifFiles(sarifFiles: string[]): string {
...
     return sarifContents;
   }
+  const automationID = getAutomationID(category, analysis_key, environment);
+
+  const sarif = JSON.parse(sarifContents);
+  for (const run of sarif.runs || []) {
+    if (run.automationDetails === undefined) {
+      run.automationDetails = {
+        id: automationID,
+      };
```
aeisenberg, line 61: v2.5.3 of the cli will populate this value automatically. Is this something you want to overwrite here?

Here is a suggestion on how to handle this:

1. If the user supplies a custom automation id, then always overwrite what already exists.
2. If the action is using a generated automation id, then only overwrite if the id does not exist in the sarif yet.
3. Also, I think you need to be careful about setting `run.automationDetails` to a new object. It's not happening yet, but in the future, we may be using other properties of this object.

src/upload-lib.ts
```diff
@@ -44,12 +44,47 @@ export function combineSarifFiles(sarifFiles: string[]): string {
...
+    return automationID;
+  }
+
+  return computeAutomationID(analysis_key, environment);
+}
+
+function computeAutomationID(
+  analysis_key: string | undefined,
+  environment: string | undefined
+): string {
```
aeisenberg, line 87: The format of the cli-generated automationId is something `linux/javascript/`. This is different from what you are generating here. Is this something we should be unifying? Will code scanning work better if we change the cli to using the format you have here?

src/upload-lib.ts
Daverlo: I don't think we should overwrite the id, that's why I went with your second suggestion.

Overwriting the value is quite tricky, and we should be careful about it. As you said, if we replace the full `automationDetails` object, we lose the possibility of using other properties in the future. And we can't just replace the id, as the other fields are correlated with it: https://github.com/github/codeql-action/pull/446#discussion_r615586394

src/upload-lib.ts
Daverlo: Code scanning will accept any valid string, so that is not a problem. Also, I don't think the cli can generate the same id, as the one generated by the action uses stuff specific to it, like the workflow file path and job name.

On the other hand, If we don't make it match we are going to have a mismatch on the comparison. I'll check with @starcke what do we want to do. My suggestion would be that the cli gets an input like the one I'm adding here, and from the action we generate the right string and pass it to the cli.

src/actions-util.test.ts
```diff
@@ -73,6 +73,47 @@ test("getAnalysisKey() when a local run", async (t) => {
...
+    ".github/workflows/codeql-analysis.yml:analyze",
+    "{}"
+  );
+  t.deepEqual(
+    actualAutomationID,
+    ".github/workflows/codeql-analysis.yml:analyze/"
+  );
+
+  // check non string environment values
+  actualAutomationID = actionsutil.computeAutomationID(
```
aeisenberg, line 107: Is this something we are likely to hit in the wild? If so, it would be nice if we could handle this case better.

src/actions-util.ts
```diff
@@ -425,6 +425,36 @@ export async function getAnalysisKey(): Promise<string> {
...
+  const environment = getRequiredInput("matrix");
+
+  return computeAutomationID(analysis_key, environment);
+}
+
+export function computeAutomationID(
+  analysis_key: string | undefined,
+  environment: string | undefined
+): string {
+  let automationID = `${analysis_key}/`;
```
aeisenberg, line 439: `analysis_key` can be undefined (according to its type). I don't think that case is handled here.

src/actions-util.test.ts
```diff
@@ -73,6 +73,47 @@ test("getAnalysisKey() when a local run", async (t) => {
...
+    ".github/workflows/codeql-analysis.yml:analyze",
+    '{"number": 1, "object": {"language": "javascript"}}'
+  );
+  t.deepEqual(
```
aeisenberg, line 111: Can you also add some tests when one or both of `analysis_key` and `environment` are undefined?

src/actions-util.ts
```diff
@@ -425,6 +425,36 @@ export async function getAnalysisKey(): Promise<string> {
...
+
+  // the id has to be deterministic so we sort the fields
+  if (environment !== undefined && environment !== "null") {
```
aeisenberg, line 442: The function parameter type does not permit `null` for the `environment`. Either the type specification is incorrect, or you don't need to check `environment !== "null"`. But maybe a simpler solution would be:

```suggestion
  if (environment) {
```

This would also avoid an error on an empty string.

src/actions-util.ts
```diff
@@ -425,6 +425,36 @@ export async function getAnalysisKey(): Promise<string> {
...
+    const environmentObject = JSON.parse(environment);
+    for (const entry of Object.entries(environmentObject).sort()) {
+      if (typeof entry[1] === "string") {
+        automationID += `${entry[0]}:${entry[1]}/`;
+      } else {
+        // In code scanning we just handle the string values,
+        // the rest get converted to the empty string
+        automationID += `${entry[0]}:/`;
```
aeisenberg, line 450: I mentioned in the tests above in a comment that this could cause ambiguity in the automationID. Is this serious enough to need to handle?

src/codeql.ts
```diff
@@ -689,6 +691,9 @@ function getCodeQLForCmd(cmd: string): CodeQL {
       if (extraSearchPath !== undefined) {
         args.push("--search-path", extraSearchPath);
       }
+      if (automationDetailsId !== undefined) {
+        args.push("--sarif-category", automationDetailsId);
+      }
```
aeisenberg, line 696: This is good as long as we are sure that users will never user the cli < 2.5.3. I think that's the case, but just want to verify.

src/actions-util.test.ts
Daverlo: Yes, that is something we have already hit. For the moment we can't do it better, since we have been discarding this information in code scanning since the beginning.

src/actions-util.ts
Daverlo: `analysis_key` should not be undefined now, I'll change it.

src/actions-util.ts
Daverlo: If we don't handle it all the users that have this kind of data will get some stale tips

src/codeql.ts
Daverlo: I think that should be the case, as the default bundle version is already updated https://github.com/github/codeql-action/pull/472/files.

For enterprise I think it should be safe too, but I'm not completely sure.

@robertbrignull could you please confirm if this change would be safe?

robertbrignull: Since the action knows what version of the bundle it wants to download, it should be fine to rely on the CLI being new enough.

On dotcom you do need to wait for the actions virtual environment to be updated. Looking at a recent actions run on the repo that didn't set `tools: latest`, CodeQL appears to be at 2.5.4 so that's fine.

On enterprise server, the action and bundle should if anything stay more in sync because the action always tries to download/use the right bundle. When the action is updated in an air-gapped instance using the sync tool it syncs the action and the bundle together.

Of course people can always pass the `tools` input either with the action or the runner, but here we have to rely on people using the correct bundle version for the action/runner version. So for instance if they update the runner then they update the bundle at the same time. Hopefully it'll be fine.

robertbrignull (commented):
I've looked through the production code at least and it all looks fine to me. I'll leave it to @aeisenberg to give a final review as you've been following more closely than I have and have already reviewed everything.

src/actions-util.ts
Daverlo: I did change this, but rolled back because it was indeed needed https://github.com/github/codeql-action/pull/463/checks?check_run_id=2508425922. I don't know why, but sometimes we get `environment = "null"`

aeisenberg: Hmmm...that's very odd. OK. I guess we need to keep it.

aeisenberg (approved):
Looks good. Thanks for working through this and discovering the incident.

👍🏼

Daverlo:
I run a test and this worked, it populated the category on the sarif https://github.com/dsp-testing/daverlo-main-flow/actions/runs/816024707

I'm going to hold off this PR a bit more, as I needed to specify `tools: latest` in order to succeed, as the default was still picking up v2.5.2 20210421 https://github.com/dsp-testing/daverlo-main-flow/actions/runs/816011280

Daverlo:
The tools are already updated so I'll proceed to merge this https://github.com/dsp-testing/daverlo-main-flow/actions/runs/828192431

PR #465, opened 2021-04-28
Avoid analyzing excluded language files for line counting

aeisenberg:
This change passes in a list of file types to the line counting
analysis. These are the languages for the databases being analyzed.
Line count analysis is restricted to these files.


src/count-loc.ts
```diff
@@ -15,6 +15,16 @@ const supportedLanguages = {
   typescript: "js",
 };
 
+const supportedLanguagesReversed = {
```
adityasharad, line 18: May want to compute this reverse map rather than hardcoding, but this is fine for now if we're not going to change it often.

adityasharad (approved):

src/count-loc.ts
aeisenberg: It's done.

PR #474, opened 2021-05-04
Change from `metric` to `rule`

aeisenberg:
The SARIF that we are interpreting has moved away from using `metric`
to the more general term, `rule`. We need to adapt our baseline lines of
code counting to use `rule` as well.


src/analyze.ts
```diff
@@ -295,16 +295,16 @@ async function injectLinesOfCode(
...
       for (const run of sarif.runs) {
-        const metricId = `${language}/summary/lines-of-code`;
+        const ruleId = `${language}/summary/lines-of-code`;
         run.properties = run.properties || {};
-        run.properties.metricResults = run.properties.metricResults || [];
-        const metric = run.properties.metricResults.find(
-          // the metric id can be in either of two places
-          (m) => m.metricId === metricId || m.metric?.id === metricId
+        run.properties.results = run.properties.results || [];
+        const rule = run.properties.results.find(
```
henrymercer, line 301: Metric results will still be stored in the `run.properties.metricResults` property, so I think this should still refer to `run.properties.metricResults`.

henrymercer (approved):

PR #475, opened 2021-05-05
Clarify the missing baseline lines of code warning message

aeisenberg:
See https://github.com/github/codeql/issues/5838


src/count-loc.ts
```diff
@@ -72,7 +72,15 @@ export async function countLoc(
       logger.debug(`  ${language}: ${count}`);
     }
   } else {
-    logger.warning("Did not find any lines of code in database.");
+    logger.warning(
+      "Could not determine a lines of code baseline for this repository. " +
+        "Because of this, it will not be possible to compare lines of code" +
```
henrymercer, line 77: (There's a missing space here)

src/count-loc.ts
```diff
@@ -72,7 +72,15 @@ export async function countLoc(
       logger.debug(`  ${language}: ${count}`);
     }
   } else {
-    logger.warning("Did not find any lines of code in database.");
+    logger.warning(
+      "Could not determine a lines of code baseline for this repository. " +
```
henrymercer, line 76: Do you think it would be best to avoid mentioning the term "baseline" to users and instead say things like "Could not determine the number of lines of code in this repository" and "Because of this, it will not be possible to compare the number of lines of code analyzed by code scanning with the total number of lines of code in the repository"?

src/count-loc.ts
```diff
@@ -72,7 +72,15 @@ export async function countLoc(
...
+        "Because of this, it will not be possible to compare lines of code" +
+        "metrics produced by code scanning with the baseline. This will not " +
+        "affect the results of the code scan. If you have any questions, you " +
+        "can raise an issue at https://github.com/github/codeql-action/issues. " +
+        "Please include a link to the repository if public, or else information " +
+        "about the language used."
```
henrymercer, line 82:
```suggestion
        "Please include a link to the repository if public and information " +
        "about the language used."
```

src/count-loc.ts
aeisenberg: Hmmm...I see your point about _baseline_, there's also ambiguity in your suggestion since it's not clear that code scanning produces its own count and which count is missing.

Maybe that's too much information in this context and being ambiguous is fine.

henrymercer: Good point.  I've updated my suggestion to say "Could not determine the total number of lines of code in this repository".  Does that seem clearer?

src/count-loc.ts
```diff
@@ -72,7 +72,16 @@ export async function countLoc(
       logger.debug(`  ${language}: ${count}`);
     }
   } else {
-    logger.warning("Did not find any lines of code in database.");
+    logger.warning(
+      "Could not determine the number of lines of code for this repository. " +
```
henrymercer, line 76:
```suggestion
      "Could not determine the total number of lines of code in this repository. " +
```

src/count-loc.ts
```diff
@@ -72,7 +72,16 @@ export async function countLoc(
...
+        "Because of this, it will not be possible compare the number of lines " +
```
henrymercer, line 77:
```suggestion
        "Because of this, it will not be possible to compare the number of lines " +
```

src/count-loc.ts
```diff
@@ -72,7 +72,16 @@ export async function countLoc(
...
+        "of code analyzed by code scanning with the total number of lines of " +
+        "code in the repository. This will not affect the results of the code " +
+        "scan. If you have any questions, you can raise an issue at " +
```
henrymercer, line 80:
```suggestion
        "code in the repository. This will not affect the results produced by code " +
        "scanning. If you have any questions, you can raise an issue at " +
```

src/count-loc.ts
```diff
@@ -72,7 +72,16 @@ export async function countLoc(
...
+        "https://github.com/github/codeql-action/issues. Please include a link " +
+        "to the repository if public, otherwise information code scanning " +
```
henrymercer, line 82:
```suggestion
        "to the repository if public, or otherwise information about the code scanning " +
```

henrymercer (approved):
Couple more suggestions on the message

aeisenberg:
All good suggestions.

adityasharad:
Do you want to keep this at the warning level (which makes it visible in the Actions summary page), or drop it down to info/debug so it's only present in the logs view?

aeisenberg:
Hmmm...didn't realize that this would be raised to being visible in the summary page. This is a good point. I'll drop down its verbosity.

aeisenberg:
A debug message won't appear at all unless the debug flag is set to true. I'd be leaning towards using an info message.

adityasharad (approved):

PR #477, opened 2021-05-05
fix value escaping in codeql-env.sh

hmakholm:
CodeQL itself complained (correctly) that the escaping here would fail if the value contains backslashes.

Switch to single-quoting, which is less tricky to escape.

Hmmm, how confident are we that there are not consumers that try to parse `codeql-env.sh` themselves and will break if they don't find double quotes? They _should_ be using the JSON output instead, but still ...


src/runner.ts
```diff
@@ -249,7 +249,7 @@ program
         const shEnvFileContents = Object.entries(tracerConfig.env)
           // Some vars contain ${LIB} that we do not want to be expanded when executing this script
           .map(
-            ([key, value]) => `export ${key}="${value.replace(/\$/g, "\\$")}"`
+            ([key, value]) => "export " + key + "='" + value.replace(/'/g, "'\"'\"'") + "'"
```
aeisenberg, line 252:
```suggestion
            ([key, value]) => `export ${key}='${value.replace(/'/g, "'\"'\"'")}'`
```

aeisenberg (approved):
Yes, it's safer to use single quotes instead of escaping the `$`.

alexet:
I know there are consumers converting the windows output to sh (as they use bash on windows) but I don't know about users converting the sh output.

src/runner.ts
```diff
@@ -249,7 +249,7 @@ program
         const shEnvFileContents = Object.entries(tracerConfig.env)
           // Some vars contain ${LIB} that we do not want to be expanded when executing this script
           .map(
-            ([key, value]) => `export ${key}="${value.replace(/\$/g, "\\$")}"`
+            ([key, value]) => `export ${key}='${value.replace(/'/g, "'\"'\"'")}'`
```
aeisenberg, line 252:
```suggestion
            ([key, value]) =>
              `export ${key}='${value.replace(/'/g, "'\"'\"'")}'`
```

hmakholm: Will that work? The linter apparently said it wants a newline _between_ the backticks; you're adding one _in front of_ them.

aeisenberg: You can ignore the open/close `. It is just being used to delineate the start and end of the insertion string. 

```
Insert `⏎·············`
```

It makes sense if you see this in an IDE (well, vscode at least), but less sense if you see it as a cli message.

aeisenberg:
Arrrrgh....also need to update the compile files. Let me just push a change up.

hmakholm:
Do we need a change note warning users of the different quoting?

aeisenberg:
Based on how we are doing things now, the answer is _no_. And as we discussed, this might change, but I don't think we should be holding this PR up until we make any process changes.

aeisenberg:
@hmakholm, merged this for you because you mentioned earlier that you kept having merge conflicts if you waited too long.

PR #483, opened 2021-05-07
Fix lines of code mapping

aeisenberg:
We were inadvertently using codeql language ids instead of the action's
language ids. There is now a 3-way mapping between the ids used by
the lines counter library, the action, and codeql.

Fixes https://github.com/github/codeql/issues/5838


src/count-loc.ts
```diff
@@ -1,29 +1,59 @@
...
+    action: Language.javascript,
+    codeql: "js",
+  },
+  python: {
+    action: Language.python,
+    codeql: "py",
+  },
+  typescript: {
+    action: Language.javascript,
+    codeql: "js",
```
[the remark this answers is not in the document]
adityasharad: Not yet possible, but no harm being ready for it while we're modifying this code :)

aeisenberg:
Added ruby and renamed the constants.

aeisenberg:
Hmmm....doesn't like ruby. https://github.com/github/codeql-action/pull/483/checks?check_run_id=2529409389#step:4:28

adityasharad:
Oh dear, bad call from me. Fine with doing Ruby separately and getting the rest of your changes underway.

adityasharad (approved):

src/count-loc.ts
```diff
@@ -40,30 +70,28 @@ export async function countLoc(
   cwd: string,
   include: string[],
   exclude: string[],
-  dbLanguages: string[],
+  dbLanguages: Language[],
   logger: Logger
-): Promise<Record<string, number>> {
+): Promise<Partial<Record<IdPrefixes, number>>> {
```
RA80533, line 75: I’m not sure if it makes sense to use partial here. Are you trying to describe a partial of the full matrix of supported languages?

src/count-loc.ts
```diff
@@ -1,29 +1,59 @@
...
+  },
+  python: {
+    name: Language.python,
+    prefix: "py",
+  },
+  typescript: {
+    name: Language.javascript,
+    prefix: "js",
+  },
 };
```
RA80533, line 46: Use `as const` at the very end.

src/count-loc.ts
```diff
@@ -1,29 +1,59 @@
...
+  (obj, [key, { name: action }]) => {
+    if (!obj[action]) {
+      obj[action] = [];
     }
-    obj[value].push(key);
+    obj[action].push(key);
     return obj;
   },
-  {}
+  {} as Record<Language, string[]>
```
RA80533, line 56: Should `string[]` be `IdPrefixes[]`?

aeisenberg: No. The `string[]` refers to language ids used in our library that counts LoC, which is different from the `IdPrefixes` used by codeql.

src/count-loc.ts
aeisenberg: There are compile errors in the tests [here](https://github.com/github/codeql-action/blob/b6ae33fbe6bf5391303516675e7e6f2f803e4307/src/count-loc.test.ts#L21-L25) and elsewhere.

Typescript assumes that `Record<IdPrefixes, number>` will have every key in `IdPrefixes`. Adding `Partial` there removes the compile errors. I was surprised by this, too. I'm not super happy about this solution. Do you know a better way of doing this?

src/count-loc.ts
aeisenberg: That's not a language feature we have been using in this project, but we should probably start.

PR #487, opened 2021-05-08
Use `--build`

### LINKED ISSUE #486: Transpile with `--build`

RA80533:

---

### Before

```console
$ time npm run build

> codeql@0.0.0 build
> tsc


real    0m7.986s
user    0m11.848s
sys     0m0.684s
```

### After

```console
$ time npm run build

> codeql@0.0.0 build
> tsc --build


real    0m0.557s
user    0m0.427s
sys     0m0.131s
```

The change results in ~1,300% faster builds.

Closes #486


robertbrignull:
@RA80533, the CI is complaining that the node modules are out of date, but what it really is I think is that trailing newline in `package.json`. Can you run `npm ci && npm run removeNPMAbsolutePaths` to let it format this how it wants. We use this check to make sure the actual `node_modules` directory is kept up to date, but unfortunately it also sometimes fails on trivial things like this.

RA80533:
Your observation is correct and I should’ve been more specific in my metrics. Builds during development should ultimately end up being much faster, with the ~1,300% increase being the ceiling when no changes have been made. Enabling `incremental` may also be of additional use but I can’t recall if `--build` does so implicitly.

I’ll run those commands on the branch when I get home.

RA80533:
It appears that `npm ci` creates a huge .diff. None of the package.json files in node_modules have a final newline. Is this something that should be done separate from this PR?

aeisenberg:
Did you also run `npm run removeNPMAbsolutePaths`? If you just run `npm ci`, then you will see a large diff. The second command will remove all the diffs.

aeisenberg (approved):

aeisenberg:
Thanks for sticking with this change!

PR #490, opened 2021-05-09
dummy PR to get tests to run

hmakholm:


aeisenberg:
@hmakholm Instead of having to do this all the time, might be more efficient to add a `workflow_dispatch` trigger handler on the workflows you want to run.

hmakholm:
I've concluded that I can probably just get away with starting the `PR checks` workflow manually, instead of creating a PR.

PR #495, opened 2021-05-10
Output environment file for Windows

edoardopirovano:
Currently, the environment file for the tracer is only being output in the Unix format. Thus, these variables are not being read by the tracer on Windows machines. This PR addresses this by also outputting the environment file in the Windows format.

I have also modified the `runner-analyze-csharp-windows` test to unset the `CODEQL_EXTRACTOR_CSHARP_ROOT` variable before the build step. This would cause the test to fail if the environment file being produced was not correct.


.github/workflows/pr-checks.yml
```diff
@@ -465,7 +465,12 @@ jobs:
 
     - name: Build code
       shell: powershell
+      # Note this step unsets the CODEQL_EXTRACTOR_CSHARP_ROOT to ensure that the .win32env file is read correctly
```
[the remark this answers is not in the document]
edoardopirovano: Before this PR the file was not being produced at all, so I figured we need to modify an integration test to make sure it actually gets read when it is produced.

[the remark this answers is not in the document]
aeisenberg: Oh. I see, I misunderstood the comment and what you're doing here. By un-setting the `CODEQL_EXTRACTOR_CSHARP_ROOT` variable, you are ensuring that `.win32env` is being read correctly. 

Could you update the comment to something like this?

```suggestion
      # Note we want to make sure that the .win32env file is read correctlyl, so we unset the CODEQL_EXTRACTOR_CSHARP_ROOT from the .sh file.
```

aeisenberg (approved):

.github/workflows/pr-checks.yml
```diff
@@ -465,7 +465,12 @@ jobs:
...
       run: |
+        (Get-Content ./codeql-runner/codeql-env.sh) | Select-String -pattern ".*CODEQL_EXTRACTOR_CSHARP_ROOT.*" -NotMatch | ? {$_.ToString().trim() -ne "" } | Set-Content ./codeql-runner/codeql-env.sh
+        $jsonEnv = Get-Content -Path ./codeql-runner/codeql-env.json -Raw | ConvertFrom-Json
+        $jsonEnv.PSObject.Properties.Remove('CODEQL_EXTRACTOR_CSHARP_ROOT')
+        $jsonEnv | ConvertTo-Json -Depth 1 | Out-File ./codeql-runner/codeql-env.json -Force
```
adityasharad, line 473: Could we make this simpler, by removing `CODEQL_EXTRACTOR_CSHARP_ROOT` from the environment after the `Invoke-Expression` step below? I think `Clear-Variable` is the PowerShell command.
The step below doesn't use the JSON file.

src/tracer-config.ts
```diff
@@ -138,10 +138,30 @@ export function concatTracerConfigs(
     sizeBuffer.writeInt32LE(lineBuffer.length, 0);
     buffer = Buffer.concat([buffer, sizeBuffer, lineBuffer]);
   }
-  // Write the compound environment
+  // Write the compound environment for Unix
   const envPath = `${spec}.environment`;
   fs.writeFileSync(envPath, buffer);
 
+  // Prepare the content of the compound environment file on Windows
```
adityasharad, line 145: What do you think about doing this step only when we're on Windows, vs unconditionally?

.github/workflows/pr-checks.yml
hmakholm: I would have a (weak!) preference to only write the file that corresponds to the current platform. I doubt it makes any _concrete_ difference either way, but it feels most right to try to mimic what the actual `preload_tracer` does.

src/tracer-config.ts
hmakholm: I would have a (weak!) preference to only write the file that corresponds to the current platform. I doubt it makes any _concrete_ difference either way, but it feels most right to try to mimic what the actual `preload_tracer` does.

.github/workflows/pr-checks.yml
edoardopirovano: Aha, done. For reason I thought those files needed patching to stop the runner reading them again.

src/tracer-config.ts
edoardopirovano: Sure, done - except in the unit tests where we still test both regardless of platform.

.github/workflows/pr-checks.yml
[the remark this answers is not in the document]
edoardopirovano: Sure, done ✅

PR #496, opened 2021-05-10
Set the CODEQL_WRAPPER env var

aeisenberg:
The CODEQL_WRAPPER environment variable will be set to "action" if
running in an actions context and "runner" otherwise. This variable
will be used in the cli to determine which context it is in.

Also, use a similar mechanism to provide a different user agent
when communicating with the github api.


src/codeql.ts
```diff
@@ -399,6 +399,9 @@ export async function setupCodeQL(
       throw new Error(`Unsupported platform: ${process.platform}`);
     }
 
+    // set a  environment variable to let CodeQL know where we are running
+    process.env.CODEQL_WRAPPER = util.isAction() ? "true" : "false";
```
adityasharad, line 403: I recall you considered using `action` and `runner` vs just using a Boolean. I think that makes sense and is more flexible.

src/util.ts
```diff
@@ -390,3 +390,11 @@ export async function getGitHubAuth(
     "No GitHub authentication token was specified. Please provide a token via the GITHUB_TOKEN environment variable, or by adding the `--github-auth-stdin` flag and passing the token via standard input."
   );
 }
+
+/**
+ * Returns true if this process is running as part of github actions.
+ * False otherwise.
+ */
+export function isAction() {
+  return "GITHUB_WORKFLOW" in process.env;
```
adityasharad, line 399: Neat. You can also expect `GITHUB_ACTIONS` to be `true`.

src/codeql.ts
aeisenberg: Let me fix that.

hmakholm:
This would be most useful _together_ with a version indication of some sort. I think it would be least messy to add both in a single PR, such that we avoid the risk of needing to recognize three different states (no variable set, kind variable set but version unknown, both kind and version known).

src/util.ts
```diff
@@ -390,3 +390,11 @@ export async function getGitHubAuth(
...
   );
 }
+
+/**
+ * Returns true if this process is running as part of github actions.
+ * False otherwise.
+ */
+export function isAction() {
+  return "GITHUB_ACTIONS" in process.env;
+}
```
hmakholm, line 400: What if someone writes an Actions workflow that explicitly invokes `codeql-runner-linux64`? In that case I think we should announce ourselves as the _runner_ even though `$GITHUB_ACTIONS` also happens to be set.

I feel it would be more robust simply to have a global flag that `runner.ts` sets early on.

aeisenberg: Is there a use case for why someone would do this?

aeisenberg:
We haven't decided yet how to best version the action, so we can hold off on this PR until we make the decision.

src/util.ts
adityasharad: Henning is right. We do it in this repo for integration testing!

aeisenberg:
Going to hold off on this PR for now until we can figure out the proper release strategy for the action.

RA80533:
It appears that `CODEQL_WRAPPER` is sourced directly from `GITHUB_WORKFLOW`. Should the latter be used instead?

src/util.ts
aeisenberg: Implemented this idea here https://github.com/github/codeql-action/pull/516.

aeisenberg:
Not going in this direction any more. Superseded by #516.

PR #497, opened 2021-05-11
Avoid counting lines of code in parallel with running queries

aeisenberg:
This can lead to OOME in larger repositories. Unfortunately, running
the LoC serially will cause the action to take longer, but I don't think
there is any other way around this.

Fixes https://github.com/github/codeql/issues/5873

I am running an analysis in a forked version the repository mentioned in that issue.

This run https://github.com/aeisenberg/Mudlet/runs/2557912173?check_suite_focus=true is using the current version of the action (ie- counting in parallel). I will do another run with the count happening after the queries are completed.


aeisenberg:
This change is no longer necessary.

PR #499, opened 2021-05-12
Update loc count library

### LINKED ISSUE #498: Prevent OOME when counting lines of code
Some users, (eg- https://github.com/github/codeql/issues/5873) are seeing out of memory errors during the lines of code counting phase. A mitigation is to ensure that non-analyzed code is added to the `paths-ignore` property. However, we still should figure out why we are running out of memory in these situations.

aeisenberg:
This version will count lines of code in each file serially. It still
runs all file system operations asynchronously. The only difference now
is that it will only count one file at a time. It is slower, but it
is able to count large repositories without running out of memory.

Fixes #498


aeisenberg:
Before merging, I want to try this on at least one repository that is currently giving us OOME.

aeisenberg:
This run is against a repository that had an OOME in the past, with the `v1` branch of the action:
https://github.com/aeisenberg/Mudlet/actions/runs/837223527

I feel confident that this will fix the problem.

adityasharad (approved):

PR #500, opened 2021-05-13
Change include path for lines of code counting

aeisenberg:
Previously, we were always using `**` in the include path. the
effect of this was to always count lines in the entire
repository unless explicitly added to the paths-ignore. This
was incorrect behaviour. Now we only using `**` if the include
path is otherwise empty.


src/count-loc.test.ts
```diff
@@ -63,7 +63,23 @@ test("ensure lines of code can handle includes", async (t) => {
...
 
   t.deepEqual(results, {
-    js: 15,
+    js: 12,
+  });
+});
+
+test("ensure lines of code can handle empty includes", async (t) => {
+  // note that "**" is always included. The includes are for extra
+  // directories outside the normal structure.
```
adityasharad, line 72: Outdated comment?

src/count-loc.test.ts
```diff
@@ -63,7 +63,23 @@ test("ensure lines of code can handle includes", async (t) => {
   );
 
   t.deepEqual(results, {
-    js: 15,
+    js: 12,
```
adityasharad, line 66: Might conflict with your other PR.

adityasharad (approved):
Looks reasonable, look out for conflicts with the other LOC PR.

PR #502, opened 2021-05-13
Use the prefix id for keying into metrics rules

aeisenberg:
Fixes a bug where we were looking for incorrect keys for metrics rules.
Previously, we were using full language names in the keys. Now, we use
the short language names in the metric ids. This is done through a
simplification of the code.

Also, this change does two smaller things:

1. Prints out the baseline count to the logs
2. Adds the `assertNever` function to ensure we never miss a case in
   a switch statement. This function is borrowed from vscode-codeql.### Merge / deployment checklist


adityasharad (approved):
Good catch.

aeisenberg:
Thanks for the quick review @adityasharad, but I decided to make a change. I think this simplifies things a bit. Could you take another look?

adityasharad (approved):
Makes sense. We use the `Language` entity wherever possible, and convert only at the points where the ID prefix or Linguist name is needed.

PR #503, opened 2021-05-13
Add warnings about CLI hack to codeql.ts

hmakholm:
These warnings refer to a hack that was added to the CLI in https://github.com/github/semmle-code/pull/39335

They can be removed again once the CLI has a more principled way for the CLI to recognize that it's being invoked by the runner/action.


src/codeql.ts
```diff
@@ -530,7 +537,14 @@ function getCodeQLForCmd(cmd: string): CodeQL {
         fs.writeFileSync(process.argv[2], JSON.stringify(env), 'utf-8');`
       );
 
+      // BEWARE: The name and location of this file is recognized by `codeql database
+      // trace-command` in order to enable special support for concatenable tracer
+      // configurations. Consequently the name must not be changed.
+      // (This warning can be removed once a different way to recognize the
+      // action/runner has been implemented in `codeql database trace-command`
+      // _and_ is present in the lastest CLI release.)
```
aeisenberg, line 545:
```suggestion
      // _and_ is present in the lastest support CLI release.)
```

aeisenberg (approved):

PR #505, opened 2021-05-17
Add Action input to set CodeQL DB location

edoardopirovano:
This adds a new option that lets the user control where the CodeQL databases that the Action creates are placed.

### Documentation information
@github/docs-content-codeql Note that this documentation page will need updating: https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning#using-a-custom-configuration-file

The new option has the name `db-location` and should be a string containing a path that the CodeQL Action can write to. If provided, this will result in the Action storing the CodeQL databases it creates at that location instead of the default temporary directory. A typical use case for this option would be that the user wishes for the CodeQL database to be in a consistent location so that a later step in their workflow can upload the DB somewhere.

We should highlight that it is the responsibility of the user that if they set a non-temporary directory in this configuration option, they should ensure that they clear it in a step later in their workflow when they are done using the database.


felicitymay:
@edoardopirovano - thanks for the docs team ping. 🙇🏻‍♀️ 

I think this will be easier to discuss and track in the related issue, so I've added that to our first responder board instead.

src/config-utils.ts
```diff
@@ -933,6 +948,19 @@ async function loadConfig(
...
   }
 
+  let dbLocation;
+  if (DB_LOCATION in parsedYAML) {
+    if (typeof parsedYAML[DB_LOCATION] !== "string") {
+      throw new Error(dbLocationInvalid(configFile));
+    }
+    if (parsedYAML[DB_LOCATION]!.length === 0) {
+      throw new Error(dbLocationInvalid(configFile));
+    }
```
aeisenberg, line 958: Nit: to avoid the type assertion `!.length`.

```suggestion
    if (typeof parsedYAML[DB_LOCATION] !== "string" || parsedYAML[DB_LOCATION] == "") {
      throw new Error(dbLocationInvalid(configFile));
    }
```

src/util.ts
```diff
@@ -171,18 +171,11 @@ export function getThreadsFlag(
...
-  return path.resolve(tempDir, "codeql_databases");
-}
-
 /**
  * Get the path where the CodeQL database for the given language lives.
  */
-export function getCodeQLDatabasePath(tempDir: string, language: Language) {
-  return path.resolve(getCodeQLDatabasesDir(tempDir), language);
+export function getCodeQLDatabasePath(dbDir: string, language: Language) {
+  return path.resolve(dbDir, language);
```
aeisenberg, line 178: Nit: This is a small function. Would it make sense to inline?

aeisenberg (approved):
I think this looks fine. Two small suggestions that are not necessary to address.

src/config-utils.ts
```diff
@@ -933,6 +948,19 @@ async function loadConfig(
     }
   }
 
+  let dbLocation;
+  if (DB_LOCATION in parsedYAML) {
```
adityasharad, line 952: The new field is in the CodeQL YAML config file. I was originally thinking this would be an Actions input to the `init` or `analyze` steps (see their `action.yml` files for the existing inputs), so that the option is directly settable from the Actions workflow YAML, and can reference other Actions context and environment variables.

I'm open to revising that though. What are the advantages of having this in the config file?

src/util.ts
edoardopirovano: I think it's nice to have everywhere we resolve the location of a database flow through this function since it makes future changes to where databases are located easier.

src/config-utils.ts
edoardopirovano: Aha, adding it as an Actions input sounds totally reasonable given the above. I have no strong argument for putting it in the config file. I will change the option to an Actions input.

src/config-utils.ts
edoardopirovano: (No longer applicable given we are changing it to an Actions input per below)

src/config-utils.ts
```diff
@@ -524,6 +529,14 @@ export function getPathsIgnoreInvalid(configFile: string): string {
   );
 }
 
+export function dbLocationInvalid(configFile: string): string {
+  return getConfigFilePropertyError(
```
adityasharad, line 533: This is no longer a config file error, so consider reporting a different error.

src/util.ts
adityasharad: Also nit: Should this instead be taking a `config`, and then internally using `config.dbLocation`?

.github/workflows/pr-checks.yml
```diff
@@ -59,14 +59,16 @@ jobs:
         mv ../action/tests/multi-language-repo/{*,.github} .
         mv ../action/.github/workflows .github
     - uses: ./../action/init
+      with:
+        db-location: "${{ runner.temp }}/customDbLocation"
```
adityasharad, line 63: Have we got a similar test where we use the default?

adityasharad (commented):
Code looks reasonable, though I'll let Andrew have the final say.
Some comments mainly to ensure we have enough test coverage for both cases.

src/config-utils.ts
edoardopirovano: Whoops, this code wasn't actually used any more, removed.

.github/workflows/pr-checks.yml
edoardopirovano: All the other tests were using the default location anyway, but I've changed one of them to explicitly check that DBs are created there.

src/util.ts
```diff
@@ -171,18 +172,11 @@ export function getThreadsFlag(
...
-export function getCodeQLDatabasesDir(tempDir: string) {
-  return path.resolve(tempDir, "codeql_databases");
-}
-
 /**
  * Get the path where the CodeQL database for the given language lives.
  */
-export function getCodeQLDatabasePath(tempDir: string, language: Language) {
-  return path.resolve(getCodeQLDatabasesDir(tempDir), language);
+export function getCodeQLDatabasePath(config: Config, language: Language) {
```
aeisenberg, line 178: Any reason why you made this change?

aeisenberg (approved):
Nice work. My question is just curiosity and doesn't require any changes.

src/util.ts
edoardopirovano: This was in response to @adityasharad's suggestion above 🙂

aeisenberg: Yep...thanks. Missed that. 👍🏼

PR #510, opened 2021-05-19
Add release and mergeback workflows

aeisenberg:
This commit ensures that the changelog is updated before a release with
the correct date and version.

Also, after a release, a mergeback PR is created to ensure that the
changelog update and version bump is available in main.


aeisenberg:
You can try out the new `update-release-branch` behaviour by running that workflow on this branch. You can see that it updates the changelog file with the expected text.

I have not been able to try the `post-release-mergeback` workflow yet. Since it's not merged into main, I can't run it as workflow dispatch. If it looks ok, can we merge, and then I will test it out?

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,75 @@
...
+    inputs:
+      baseBranch:
+        description: 'The base branch to merge into'
+        default: v1
+        required: false
+
+  # Commented for now to ensure we don't run this accidentally after a real merge
+  # push:
+  #   branches:
+  #     - main
```
adityasharad, line 18: I think you want these the other way round? Pushes to v1, and then open a PR against main as the base?

aeisenberg: yes.

aeisenberg:
This is roughly working. It's ready for a good review. You can see the automated PRs that this PR is making. Once it gets reviewed, I can figure out the TODOs.

.github/update-release-branch.py
```diff
@@ -4,6 +4,16 @@
 import requests
 import subprocess
 import sys
+import json
+import datetime
+import os
+
+EMPTY_CHANGELOG = """
+# CodeQL Action: Changelog
```
adityasharad, line 12: Make sure this matches the current header.

.github/update-release-branch.py
```diff
@@ -71,6 +81,13 @@ def open_pr(repo, all_commits, short_main_sha, branch_name):
       body += ' - ' + get_truncated_commit_message(commit)
       body += ' (@' + commit.author.login + ')'
 
+  body += '\n\nPlease review the following:\n'
```
adityasharad, line 84: Minor: consider creating a list of lines, and then calling `'\n'.join(lines)`.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      baseBranch:
+        description: 'The base branch to merge into'
+        default: main
+        required: false
+
+  # Commented for now to ensure we don't run this accidentally after a real merge
```
adityasharad, line 15: Can be removed now?

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+      BASE_BRANCH: "${{ github.event.inputs.baseBranch || 'main' }}"
+      # TODO head_ref won't exist in workflow dispatch, but not sure what to put here until I can start trying this out using workflow_dispatch
+      HEAD_BRANCH: "${{ github.head_ref }}"
+      NEW_BRANCH: "mergeback/${{ github.head_ref }}-to-${{ github.event.inputs.baseBranch || 'main' }}"
+      DEFAULT_REVIEWER: "github/codeql-core"
+
+    steps:
+      - name: Dump GitHub context
+        env:
+          GITHUB_CONTEXT: '${{ toJson(github) }}'
```
adityasharad, line 37: Better to just dump the event context `github.event`.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Dump branches
+        run: |
+          echo "BASE_BRANCH $BASE_BRANCH"
+          echo "HEAD_BRANCH $HEAD_BRANCH"
+          echo "NEW_BRANCH $NEW_BRANCH"
+
+      - name: validate
+        run: |
+          if [ "$BASE_BRANCH" -eq "$HEAD_BRANCH" ]; then
+            echo "error::Base branch and head branch are the same $BASE_BRANCH."
```
adityasharad, line 49:
```suggestion
            echo "::error Base branch and head branch are the same $BASE_BRANCH."
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+            exit -1
+          fi
+
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+
+      - name: Update git config
+        run: |
+          git config --global user.email "codeql-core@github.com"
+          git config --global user.name "CodeQL Actions Bot"
```
adityasharad, line 59: This will make it look like the global Actions bot user.
```suggestion
          git config --global user.email "github-actions@github.com"
          git config --global user.name "github-actions[bot]"
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Create mergeback branch
+        run: |
+          git checkout -b $NEW_BRANCH
```
adityasharad, line 63:
```suggestion
          git checkout -b "$NEW_BRANCH"
```

.github/workflows/update-release-branch.yml
```diff
@@ -22,12 +22,17 @@ jobs:
...
 
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         pip install PyGithub==1.51 requests
 
+    - name: Update git config
+      run: |
+        git config --global user.email "codeql-core@github.com"
+        git config --global user.name "CodeQL Actions Bot"
```
adityasharad, line 35:
```suggestion
        git config --global user.email "github-actions@github.com"
        git config --global user.name "github-actions[bot]"
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+          git ls-remote --tags origin | grep $TAG
+          EXISTS="$?"
+          if [ "$EXISTS" -ne 0 ]; then
+             echo "::set-output name=exists::true"
+             echo "Tag $TAG exists. Not going to re-release."
+          fi
+
+      # we didn't tag the release during the update-release-branch workflow because the
+      # commit that actually makes it to the main branch may be different than the one
+      # created by the workflow. We tag now because we know the correct commit
```
adityasharad, line 80:
```suggestion
      # we didn't tag the release during the update-release-branch workflow because the
      # commit that actually makes it to the release branch is a merge commit,
      # and not yet known during the first workflow. We tag now because we know the correct commit.
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Create mergeback branch
+        run: |
+          git checkout -b $NEW_BRANCH
+
+      - name: Check for tag
+        id: check
+        run: |
+          set +e # don't fail on an errored command
+          VERSION="$(jq '.version' -r 'package.json')"
```
adityasharad, line 69: Minor: You could make the version an output of this step, so you don't need to compute it again below.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
+# This workflow runs after a release of the action.
+# It merges any changes from the release back into the
+# main branch. Typically, this is just a single commit
+# that updates the changelog.
+name: Mergeback after release
```
adityasharad, line 5: Minor: should we name this something like "Tag release and merge back" or "Complete release"? The tagging is a useful step.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
+# This workflow runs after a release of the action.
+# It merges any changes from the release back into the
```
adityasharad, line 2:
```suggestion
# It tags the release and merges any changes from the release back into the
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+  #     - v1
+
+  pull_request:
+    paths:
+      - .github/workflows/post-release-mergeback.yml
+
+jobs:
+  merge-back:
+    runs-on: ubuntu-latest
+    env:
```
adityasharad, line 27: Add an `if` clause to ensure this only runs on the parent repo, identified by name, and not on forks.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+      BASE_BRANCH: "${{ github.event.inputs.baseBranch || 'main' }}"
+      # TODO head_ref won't exist in workflow dispatch, but not sure what to put here until I can start trying this out using workflow_dispatch
+      HEAD_BRANCH: "${{ github.head_ref }}"
```
adityasharad, line 30: I think `${{ github.ref || github.event.pull_request.head.ref }}` will work.
`github.ref` gets the chosen branch for `workflow_dispatch` and `push` events.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          git checkout -b $NEW_BRANCH
+
+      - name: Check for tag
+        id: check
+        run: |
+          set +e # don't fail on an errored command
+          VERSION="$(jq '.version' -r 'package.json')"
+          TAG="v$VERSION"
+          git ls-remote --tags origin | grep $TAG
```
adityasharad, line 71:
```suggestion
          git ls-remote --tags origin | grep "$TAG"
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,104 @@
...
+          git push origin --follow-tags "v$VERSION"
+
+      - name: Create mergeback branch
+        if: steps.check.outputs.exists == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -exu
+          PR_TITLE="Mergeback $HEAD_BRANCH into $BASE_BRANCH"
+          PR_BODY="Updates version and changelog."
```
adityasharad, line 95: Consider having "after version <version>" in the PR body or the commit message.

.github/update-release-branch.py
aeisenberg: Sure. this should be irrelevant now that the changelog is in `main`.

.github/workflows/post-release-mergeback.yml
aeisenberg: `github.ref` looks like this:

```
"ref": "refs/heads/main"
```

But, we can't generate a branch name from that. Maybe, I could just remove the leading `refs/heads` if it exists.

adityasharad: Ah yes. The most common trick I've seen is `REF_SANITIZED=${REF//\//-}`.

aeisenberg: Good idea.

aeisenberg: Actually, made a little change. This is no longer necessary.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,114 @@
...
+
+          # Update the changelog
+          perl -i -pe 's/^/## \[UNRELEASED\]\n\n/ if($.==3)' CHANGELOG.md
+          npm version
+          git add .
+          git status
+          git commit -m "Update changelog and version after $VERSION"
+
+          git push origin "$NEW_BRANCH"
+          gh pr create --head "$NEW_BRANCH" --base "$BASE_BRANCH" --title "$PR_TITLE" --body "$PR_BODY" --reviewer "$DEFAULT_REVIEWER"
```
adityasharad, line 114: Consider making this a draft PR when `"$GITHUB_EVENT_NAME" == "pull_request"`.

aeisenberg: good idea.

aeisenberg:
Here's the latest PR that gets created: https://github.com/github/codeql-action/pull/526

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,127 @@
...
+      - .github/workflows/post-release-mergeback.yml
+
+jobs:
+  merge-back:
+    runs-on: ubuntu-latest
+    if: github.repository == 'github/codeql-action'
+    env:
+      BASE_BRANCH: "${{ github.event.inputs.baseBranch || 'main' }}"
+      HEAD_BRANCH: "${{ github.head_ref || github.ref }}"
+      # Would like to use the github/codeql-core team, but that is not working
```
adityasharad, line 30: Hm that will only work if the user is a member of the `github` org (which the GitHub Actions bot user /`GITHUB_TOKEN` does not have). Names are fine for now, or we don't set a default reviewer since we're all subscribed to the repo anyway.

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,127 @@
...
+      DEFAULT_REVIEWER: "aeisenberg"
+
+    steps:
+      - name: Dump GitHub context
+        env:
+          GITHUB_CONTEXT: "${{ toJson(github) }}"
```
adityasharad, line 36:
```suggestion
          GITHUB_CONTEXT: "${{ toJson(github.event) }}"
```

.github/workflows/post-release-mergeback.yml
```diff
@@ -0,0 +1,127 @@
...
+          fi
+
+          git push origin "$NEW_BRANCH"
+          gh pr create \
+            --head "$NEW_BRANCH" \
+            --base "$BASE_BRANCH" \
+            --title "$PR_TITLE" \
+            --body "$PR_BODY" \
+            --reviewer "$DEFAULT_REVIEWER" \
+            "$DRAFT"
```
adityasharad, line 127: Minor: I believe this will substitute null rather than the empty string if the variable is empty. See whether you need it.
```suggestion
            "${DRAFT:+"$DRAFT"}"
```

adityasharad (approved):
Looks good. A few minor comments, feel free to merge once addressed.

package.json
```diff
@@ -9,7 +9,8 @@
     "test-debug": "ava src/** --serial --verbose --timeout=20m",
     "lint": "eslint --report-unused-disable-directives --max-warnings=0 . --ext .js,.ts",
     "lint-fix": "eslint --report-unused-disable-directives --max-warnings=0 . --ext .js,.ts --fix",
-    "removeNPMAbsolutePaths": "removeNPMAbsolutePaths . --force"
+    "removeNPMAbsolutePaths": "removeNPMAbsolutePaths . --force",
+    "version": "cd runner && npm version patch && git add ."
```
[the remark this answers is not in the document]
aeisenberg: Not exactly. This script ensures that when `npm version ...` is called, the `version` script runs and that also triggers the `npm version` on the runner. So, both will be updated in tandem.

See `npm help version`

>  If preversion, version, or postversion are in the scripts property of the  package.json,  they  will  be executed as part of running npm version.
> ...
>  4. Run the version script. These scripts have access to the new version in  package.json...

.github/workflows/post-release-mergeback.yml
aeisenberg: Is there a reason why you are asking for this? I find it useful to have the entire github context available for debugging (in fact, I used it for this workflow already).

package.json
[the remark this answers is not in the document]
adityasharad: Oh I see. So calling `npm version` at the root will then invoke this script, which bumps the version in the runner. Makes sense.

Consider avoiding the `git add .` here, and keeping that in the Actions workflow instead.

.github/workflows/post-release-mergeback.yml
aeisenberg: That would explain why it was failing. Sure. I will remove a default reviewer.

.github/workflows/post-release-mergeback.yml
adityasharad: It contains tokens. The default one gets masked but it's a safer idea to avoid logging the entire context.

.github/workflows/post-release-mergeback.yml
aeisenberg: It's explicitly set to `""` if not a pull request. So, that's fine, I think.

.github/workflows/post-release-mergeback.yml
aeisenberg: Sure. I'll remove it for security reasons.

package.json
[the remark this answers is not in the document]
aeisenberg: By running `git add .`, this ensures that the version changes to the runner get included in the same commit as the version changes to the containing action. I _could_ run `git commit --amend` in the workflow, but that seems messier if someone is running the `npm version` command manually.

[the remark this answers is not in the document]
adityasharad: Got it. Do you know which file it will touch? Perhaps just add the `package.json` file then?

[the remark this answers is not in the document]
aeisenberg: `package.json` and `package-lock.json`. It's pretty safe running `add .` since the initial `npm version` will fail if the working directory is not clean. I have a slight preference for keeping the `.` since this will be more future proof if `npm version` ever does start touching different files.

Also note that the invocation of `npm version patch` in the `runner` directory doesn't create a commit since commits are only created if the `package.json` is in the root of the git repo.

adityasharad (approved):
Update from main and :shipit:!

PR #513, opened 2021-05-20
Add Ruby as a possible language

aibaars:


src/codeql.ts
```diff
@@ -654,6 +663,19 @@ function getCodeQLForCmd(cmd: string): CodeQL {
...
+      let output = "";
+      await new toolrunner.ToolRunner(cmd, codeqlArgs, {
+        listeners: {
+          stdout: (data: Buffer) => {
+            output += data.toString();
+          },
+        },
+      }).exec();
+
+      return JSON.parse(output);
```
aeisenberg, line 677: Catch malformed json errors and rethrow with a meaningful message.

```suggestion
      try {
        return JSON.parse(output);
      } catch (e) {
        throw new Error("Something meaningful.);
      }
```

aibaars: Wouldn't that suppress the likely interesting error message of `e` ? In Java I'd write `new Error("Unexpected output from codeql resolve languages", e)` . What would be the equivalent in Javascript?

robertbrignull: I think just construct the message as a string. So like `new Error("Unexpected output from codeql resolve languages: " + e.message)` and maybe some type checks to make sure that `e.message` exists.

aibaars: I opted to just write `Unexpected output from codeql resolve languages: ${e}` . This should include the original error's type and side-steps the problem of a potentially empty message string.

aeisenberg (approved):
We're populating the changelog now. Since this is a public-facing feature, could you add an entry. And then it's good to merge. Thanks!

aibaars:
> We're populating the changelog now. Since this is a public-facing feature, could you add an entry. And then it's good to merge. Thanks!

There shouldn't be an entry. This is just for internal tests. CodeQL for Ruby is not available yet.

PR #515, opened 2021-05-20
Analyze: Improve log grouping and log the analysis summary in its own group

adityasharad:
- Fix grouping of the analysis logs, so that custom query logs and the upload steps also get grouped. (Bug spotted by @AlonaHlobina.)
- Capture the stdout of `codeql database analyze`, which contains the analysis summary from summary and diagnostic queries. Log this output in its own group, along with the baseline metrics computed in the Action.


src/codeql.ts
```diff
@@ -712,7 +712,16 @@ function getCodeQLForCmd(cmd: string): CodeQL {
         args.push("--sarif-category", automationDetailsId);
       }
       args.push(querySuite);
-      await new toolrunner.ToolRunner(cmd, args).exec();
+      // capture stdout, which contains analysis summaries
+      let output = "";
+      await new toolrunner.ToolRunner(cmd, args, {
+        listeners: {
+          stdout: (data: Buffer) => {
+            output += data.toString();
```
aeisenberg, line 720:
```suggestion
            output += data.toString("utf8");
```

src/analyze.ts
```diff
@@ -188,7 +190,7 @@ export async function runQueries(
       const temporarySarifFiles: string[] = [];
       for (let i = 0; i < queries["custom"].length; ++i) {
         if (queries["custom"][i].queries.length > 0) {
-          const sarifFile = await runQueryGroup(
+          const { sarifFile } = await runQueryGroup(
```
aeisenberg, line 193: Should you be capturing the stdout here as well?

adityasharad: Good question. My thinking was that custom queries wouldn't produce diagnostic results. Perhaps that's wrong, or inconvenient for those running a specific version of the queries instead of the builtin suites. What do you think?

aeisenberg: Hmmm...this leads me to several more questions. If the stdout is not captured, would it still be printed to the console, or is it just discarded? If it's just discarded, is that really the behaviour we want? What about stderr?

What would the downside of capturing and printing stdout for custom queries? If it's going to print some unrelated messages, then that might also happen in the default queries. I'd think that it's more informative to include stdout from the custom queries, but I'm not entirely sure what is included in stdout.

adityasharad: Stdout from `codeql database analyze` only contains the analysis summary table. The remaining progress logs from CodeQL go to stderr.

To run CodeQL from the Action, we use the `exec` library from `actions/toolkit`, which automatically streams both stdout and stderr from the wrapped process into the Actions logs. Here I've added an extra listener for stdout to record the output. So nothing will be lost from the output we see.

adityasharad: What happens if we show the stdout for custom queries?
- if the custom queries have no diagnostic queries (most likely) it's empty, no downside
- if there are diagnostic queries (according to the formats we've chosen), then you'll see a second summary of results, which may be misleading

aeisenberg: Right. Probably not worth it, then.

The only danger is if a customer is explicitly adding their own diagnostic queries that are not default and expecting their table to be shown. This is not our intended use case, so I think it is safe to ignore.

src/codeql.ts
adityasharad: Done.

aeisenberg (approved):

src/codeql.ts
```diff
@@ -717,7 +717,7 @@ function getCodeQLForCmd(cmd: string): CodeQL {
       await new toolrunner.ToolRunner(cmd, args, {
         listeners: {
           stdout: (data: Buffer) => {
-            output += data.toString();
+            output += data.toString("utf8");
```
aibaars, line 720: @adityasharad Is CodeQL guaranteed to output UTF-8 ? Doesn't it output things in the "system" encoding to make sure things render correctly when running in a non-UTF-8 terminal?

adityasharad: Good point, it uses the Java default, so that's a little harder to anticipate here. Is it better to just avoid passing in a locale?

PR #516, opened 2021-05-20
Send version and mode to code scanning via user agent

aeisenberg:
The initial goal of this change is to send more complete information to code scanning via the user agent. This also opens the possibility of sending the action/runner version to the CLI.

This commit changes the way the action determines if running in action
or runner mode. There is now an environment variable that is set at the
beginning of the process and elsewhere in the process, we can check to
see if the variable is set.


src/api-client.ts
```diff
@@ -37,7 +40,7 @@ export const getApiClient = function (
   return new githubUtils.GitHub(
     githubUtils.getOctokitOptions(auth, {
       baseUrl: getApiUrl(apiDetails.url),
-      userAgent: "CodeQL Action",
+      userAgent: `CodeQL ${getMode()} ${pkg.version}`,
```
adityasharad, line 43: I believe this is the convention and follows the HTTP spec.
```suggestion
      userAgent: `CodeQL-${getMode()}/${pkg.version}`,
```

adityasharad (commented):
Neat! I defer to Robert on whether the mode initialisations are in the right place, but the logic looks sensible to me.

src/actions-util.ts
```diff
@@ -727,3 +732,25 @@ export function getRelativeScriptPath(): string {
...
+export function getMode(): Mode {
+  // Make sure we fail fast if the env var is missing. This should
+  // only happen if there is a bug in our code and we neglected
+  // to set the mode early in the process.
+  const mode = getRequiredEnvParam(CODEQL_RUN_MODE_ENV_VAR) as Mode;
+
+  if (mode !== Mode.actions && mode !== Mode.runner) {
+    throw new Error(`Unknown mode: ${mode}.`);
+  }
+  return mode;
```
RA80533, line 751: Be very careful about using type assertions. When used to mark sources of input as valid, it can potentially lead to untrusted data being allowed in places it shouldn't. While that's not the case here because the input is validated, it warrants a note considering that this snippet is quite close to something CodeQL would flag.

Instead of a type assertion, allow `mode` to settle to whatever type the compiler infers it to be (in this case, it would be inferred to be a `string`). The conditional block underneath its assignment provides enough information to the TypeScript compiler for it to correctly infer `mode` to be of type `Mode` when it's returned on line 751.

RA80533 (changes requested):

src/actions-util.ts
```diff
@@ -727,3 +732,25 @@ export function getRelativeScriptPath(): string {
...
+
+export function setMode(mode: Mode) {
+  core.exportVariable(CODEQL_RUN_MODE_ENV_VAR, mode);
+}
+
+export function getMode(): Mode {
+  // Make sure we fail fast if the env var is missing. This should
+  // only happen if there is a bug in our code and we neglected
+  // to set the mode early in the process.
+  const mode = getRequiredEnvParam(CODEQL_RUN_MODE_ENV_VAR) as Mode;
```
RA80533, line 746:
```suggestion
  const mode = getRequiredEnvParam(CODEQL_RUN_MODE_ENV_VAR);
```

aeisenberg: Yes. This type assertion is not necessary. Thanks.

src/actions-util.ts
aeisenberg: Good point.

RA80533:
The conditional branching within the updated getCodeQLBundleDownloadURL function should be tested based on the new method of determining the mode. Every occurrence in which an explicit mode was passed in now implicitly relies upon the environment in which it is called.

Should that implicit behavior be moved to a higher level of abstraction, e.g., into functions directly consumed by the end-user? There are a lot of lower-level functions reliant upon the results from getCodeQLBundleDownloadURL.

aeisenberg:
Thanks for the comment @RA80533. I added a couple of more unit tests. I feel like the code is reasonably safe now. We are sure to set the mode at the beginning of every entry point. If there is a call to get the mode and it hasn't been set yet, then that will fail.

In some ways, the logic is more complex because some low level functions now rely on a new environment variable, but the trade off is that we no longer need to pass the `mode` parameter around. I had originally tried keeping the `mode` parameter and passing it around, but then almost every function needed access to it. Using the environment variable in a single place seems like a reasonable trade-off. There are other parts of the code that also rely on environment variables, so this situation is not too different.

RA80533:
> In some ways, the logic is more complex because some low level functions now rely on a new environment variable, but the trade off is that we no longer need to pass the `mode` parameter around. I had originally tried keeping the `mode` parameter and passing it around, but then almost every function needed access to it. Using the environment variable in a single place seems like a reasonable trade-off. There are other parts of the code that also rely on environment variables, so this situation is not too different.

It is very much a case of deciding which to trade off. Environment variables become unwieldy in larger projects because they create a mathematical side effect in the flow of logic; for smaller projects, such as this one, the flow of data is easy enough to follow to the point that such dependent behavior is easy to track down.

/lgtm

aeisenberg:
Ooooh...code scanning found a bug. Thank you code scanning.

src/actions-util.ts
```diff
@@ -727,3 +732,30 @@ export function getRelativeScriptPath(): string {
   const actionsDirectory = path.join(path.dirname(runnerTemp), "_actions");
   return path.relative(actionsDirectory, __filename);
 }
+
+const CODEQL_RUN_MODE_ENV_VAR = "CODEQL_RUN_MODE";
+
+export function setMode(mode: Mode) {
+  // avoid accessing actions core when in runner mode
+  if (mode === Mode.actions) {
+    core.exportVariable(CODEQL_RUN_MODE_ENV_VAR, mode);
```
aeisenberg, line 741: Note this is a code scanning error, but a false positive. We cannot have a code path accessible from the runner that will call into the core library. The if clause above prevents this from happening.

aeisenberg:
Updated ql queries that look for runner/actions access. Could I get another look at this?

RA80533:
Without the guard, what would be available to non-Actions runners through the core library?

aeisenberg:
Yes, mostly right...there are _some_ actions libraries that are ok to use for the runner. https://github.com/github/codeql-action/blob/3a474e8ed612f1b0ae2ff3b2586171b7ea29a6fc/queries/unguarded-action-lib.ql#L16-L21

queries/unguarded-action-lib.ql
```diff
@@ -90,17 +90,96 @@ class RunnerEntrypoint extends Function {
...
 
+/**
+ * A generic check to see if we are in actions or runner mode in a particular block of code.
+ */
+abstract class ActionsGuard extends IfStmt {
+
+  /**
+   * Get an expr that is only executed on actions
+   */
+  abstract Expr getAnActionsExpr();
```
adityasharad, line 101: You could instead declare:
```ql
abstract Stmt getActionsBlock();

/**
  * Gets an expr that is only executed on actions
 */
final Expr getAnActionsExpr() { getActionsBlock().getAChildStmt*().getAChildExpr*() = result }
```

This way each subclass needs to describe what they consider an Actions-specific statement, and you don't need to repeat the logic for an Actions-specific expression.

adityasharad (approved):
Nice. Logic makes sense and neat adjustment to the query; a few minor suggestions there.

There's probably room to combine more of the logic in that query, and use some builtin features of the CodeQL JS libraries, but we can do that separately. No need to hold up this PR.

queries/unguarded-action-lib.ql
aeisenberg: Yes. That's better. Thanks.

src/runner.ts
```diff
@@ -149,7 +150,9 @@ program
     "(Advanced, windows-only) Inject a windows tracer of this process into a parent process <number> levels up."
   )
   .action(async (cmd: InitArgs) => {
+    setMode(Mode.runner);
```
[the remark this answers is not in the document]
aeisenberg: Hmmm...it looks like there is an event mechanism for `commander`, but it's not quite what we need and it is internal, so I think we will need to create a query here.

aeisenberg:
Wait...commander v8 is soon to be released with a `hook` for `preAction` and `postAction` in it. I will get this merged, and then do the update for v8.

PR #518, opened 2021-05-21
Ensure queries[language] objects are initialized

aibaars:
This PR improves the CodeQL action in case a query suite is used that contains no queries. This is a bit of a corner case and resulted in an error like `no property 'builtin' for 'undefined'` instead of the more friendly message `Unable to analyse ruby as no queries were selected for this language`


src/config-utils.ts
```diff
@@ -760,6 +760,14 @@ export async function getDefaultConfig(
     logger
   );
   const queries: Queries = {};
+  for (const language of languages) {
+    if (queries[language] === undefined) {
```
robertbrignull, line 764: Won't this always be true? The `queries` variable has just been defined to be `{}`.

robertbrignull:
So in this case, all the requisite codeql suites exist (i.e. `ruby-code-scanning.qls`, `ruby-security-extended.qls`, `ruby-security-and-quality.qls`) but they are empty? Otherwise I would have expected codeql to error trying to resolve the suite, but maybe it doesn't.

Another thing to consider to make the error message better is there's a chunk of code at the end of `loadConfig` that does
```typescript
// The list of queries should not be empty for any language. If it is then
// it is a user configuration error.
for (const language of languages) {
  if (
    queries[language] === undefined ||
    (queries[language].builtin.length === 0 &&
      queries[language].custom.length === 0)
  ) {
    throw new Error(
      `Did not detect any queries to run for ${language}. ` +
        "Please make sure that the default queries are enabled, or you are specifying queries to run."
    );
  }
}
```
If you move that to `initConfig` so it runs when using the default config too then it'll give you a friendly error message quicker before it tries to build any code or run any queries.

aibaars:
> So in this case, all the requisite codeql suites exist (i.e. `ruby-code-scanning.qls`, `ruby-security-extended.qls`, `ruby-security-and-quality.qls`) but they are empty?

@robertbrignull  Exactly. We don't have any queries yet that match the default query selectors.

aibaars:
@robertbrignull Thanks for the quick review. I addressed your comments.

src/config-utils.ts
```diff
@@ -1016,6 +1013,21 @@ export async function initConfig(
...
+  // it is a user configuration error.
+  for (const language of config.languages) {
+    if (
+      config.queries[language] === undefined ||
+      (config.queries[language].builtin.length === 0 &&
+        config.queries[language].custom.length === 0)
+    ) {
+      throw new Error(
+        `Did not detect any queries to run for ${language}. ` +
+          "Please make sure that the default queries are enabled, or you are specifying queries to run."
```
robertbrignull, line 1026: You could change this error message to be more accurate for the case of ruby, but also I assume this situation of having empty suites will be temporary so it doesn't matter too much.

robertbrignull (approved):
LGTM
Should be fine

src/config-utils.ts
aibaars: Yes, indeed. The important thing is that the Action no long crashes with an `undefined property` error.

PR #519, opened 2021-05-21
Delete Ruby test file

aibaars:
Drop Ruby file from the repository. This should fix the test failure on https://github.com/github/codeql-action/pull/513 .


aibaars:
This is not the right way to fix this .

PR #529, opened 2021-05-25
Introduce an option to upload CodeQL DBs created

edoardopirovano:
This introduces an option on the `analyze` step of `db-upload` which results in the CodeQL databases created by the action being uploaded as artifacts. Further, a `db-artifact-prefix` option can be used to change the name of the created artifacts (the language is always appended to disambiguate between the potentially several databases created in multi-language repos).


edoardopirovano:
After some discussion, it seems that it would be preferable to instead clean up DBs and output their location to let users upload them themselves in a later step.

aeisenberg:
Once this work is done and @shati-patel 's work on authentication with github is completed in the extension, then we can create a feature to download databases directly from github.

PR #534, opened 2021-05-31
Fix the mergeback workflow

aeisenberg:
And manually mergeback v1 into main.

Needed to do a few fixes here:

- Fix the logic in `post-release-mergeback.yml` workflow
- Unshallow before pushing a tag
- Fix the `version` script so that newlines are correctly stripped to keep `removeNPMAbsolutePaths` happy
- Manually updated the changelog.

### DIFF
--- a/.github/update-release-branch.py
+++ b/.github/update-release-branch.py
@@ -8,8 +8,7 @@ import json
 import datetime
 import os
 
-EMPTY_CHANGELOG = """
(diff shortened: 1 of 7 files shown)


aeisenberg:
There are 39 commits being mergedback. I think this is ok. These are all the commits that exist in v1, but not main. The actually changed files are correct. Future mergebacks should not have so many commits in them.

adityasharad (approved):

PR #539, opened 2021-06-01
Add environment variables to signal feature and version to the CLI

aeisenberg:
This PR ensures environment variables are set before any invocation of
the CLI.  Here is a list of vars that are set:

https://github.com/github/codeql-coreql-team/issues/1124#issuecomment-852463521

This ensures the CLI knows the features and versions of the containing
actions/runner.

Additionally:

- Fix the user agent so that it more closely aligns with user agent
  spec
- Refactor environment variable initialization so that it all happens in
  one place and call.

Note that `prepareLocalRunEnvironment()` should only be run in actions mode. 
And the method is most likely deprecated and
should be removed. I originally added it because I had a way of working
where I would run the action from my local machine to test out changes,
but this was always a little flaky. So, I no longer use this way of
working. I will probably remove it soon.


src/actions-util.ts
```diff
@@ -733,22 +733,61 @@ export function getRelativeScriptPath(): string {
   return path.relative(actionsDirectory, __filename);
 }
 
-const CODEQL_RUN_MODE_ENV_VAR = "CODEQL_RUN_MODE";
+/**
+ * Environment variables to be set by the action and possibly used by the
+ * CLI.
+ */
+enum EnvVar {
```
robertbrignull, line 740: Are all these options only relevant to the action and not to the runner? Just wondering if they should be in `actions-util.ts` or somewhere shared.

hmakholm: They should be set by both the action and the runner. (We might need a short word to use about "action and runner both").

aeisenberg: Yes, I've confusingly been using `codeql-action` to signify both the action and the runner. And `actions` or `runner` to signify one mode or the other. (And I've probably not been entirely consistent at that.) If you have better terms, please let me know.

aeisenberg: If there are no better solutions, I'll just make sure to clarify the meaning of all variables, etc in comments.

robertbrignull: Comments are fine by me. My original point was more just that this is in `actions-util` which was meant to be for code only relating to actions. I'm surprised that when we call `setMode`/`initializeEnvironment` from in `runner.ts` it isn't generating an alert from our custom query.

I think the query might be wrong in fact. I've opened https://github.com/github/codeql-action/pull/542

aeisenberg: Ahhh...I see what you mean. I was never clear on the distinction between `util.ts` and `actions-utl.ts`. I should probably move the mode-related methods to `util.ts`.

aeisenberg:
Oops. A bit of a PR screw-up here. The dependent PR #540 was merged into this one. 

I also made changes to migrate `getMode`, `setMode`, and the env vars from `actions-util` into `util`. Can you all take another look? Probably best to go commit-by-commit for this now.

src/util.ts
```diff
@@ -396,3 +398,94 @@ class ExhaustivityCheckingError extends Error {
...
+  runner = "Runner",
+}
+
+/**
+ * Environment variables to be set by codeql-action and used by the
+ * CLI. These environment variables are relevant for both the runner
+ * and the action.
+ */
+enum EnvVar {
+  // either 'actions' or 'runner'
```
adityasharad, line 402: Minor: these could be `/** */` doc comments so we see them in IDE hovers.

src/util.ts
```diff
@@ -396,3 +398,94 @@ class ExhaustivityCheckingError extends Error {
...
+  RUN_MODE = "CODEQL_ACTION_RUN_MODE",
+
+  // semver of this action
```
adityasharad, line 405: action -> action/runner in all these comments

src/util.ts
```diff
@@ -396,3 +398,94 @@ class ExhaustivityCheckingError extends Error {
...
+  VERSION = "CODEQL_ACTION_VERSION",
```
adityasharad, line 406: I think we spoke about `DRIVER` to indicate either? Not a big deal either way, as long as the comments are clear.

src/util.ts
```diff
@@ -396,3 +398,94 @@ class ExhaustivityCheckingError extends Error {
...
+  // own sandwiched workflow mechanism
+  FEATURE_SANDWICH = "CODEQL_ACTION_FEATURE_SANDWICH",
+}
+
+export function initializeEnvironment(mode: Mode, version: string) {
+  // avoid accessing actions core when in runner mode
+  if (mode === Mode.actions) {
+    core.exportVariable(EnvVar.RUN_MODE, mode);
+    core.exportVariable(EnvVar.VERSION, version);
+    core.exportVariable(EnvVar.FEATURE_SARIF_COMBINE, "true");
```
adityasharad, line 431: Minor: you may want to have a map from EnvVar to values, or pull each of these values out before using them in each of the branches below. I can totally see myself making the mistake of changing an env var value in the `actions` block here and forgetting to change it in the `runner` block.

adityasharad (approved):
Looks reasonable, though definitely worth another look from Robert.

src/util.ts
robertbrignull: Or define something like
```typescript
const exportVar = (name: string, value: string) => (mode === Mode.actions) ? core.exportVariable(name, value) : process.env[name] = value;
```

src/util.ts
```diff
@@ -466,7 +466,10 @@ export function getRequiredEnvParam(paramName: string): string {
   if (value === undefined || value.length === 0) {
     throw new Error(`${paramName} environment variable must be set`);
   }
-  core.debug(`${paramName}=${value}`);
+
+  if (process.env[EnvVar.RUN_MODE] === Mode.actions) {
```
robertbrignull, line 470: Why do we print this here? It looks like it could be left-over debugging code.

src/util.ts
```diff
@@ -396,3 +398,94 @@ class ExhaustivityCheckingError extends Error {
...
+export function getRequiredEnvParam(paramName: string): string {
+  const value = process.env[paramName];
+  if (value === undefined || value.length === 0) {
+    throw new Error(`${paramName} environment variable must be set`);
+  }
+  core.debug(`${paramName}=${value}`);
+  return value;
+}
+
+export function getTemporaryDirectory(): string {
```
robertbrignull, line 476: Should this stay in the actions-only area? The `CODEQL_ACTION_TEMP` and `RUNNER_TEMP` variables it looks for only exist on actions. Confusingly, that `RUNNER_TEMP` is referring to the process that "runs" the actions, and is not related to our codeql-runner.

robertbrignull:
Generally looks good to me too. I've made a few comments throughout but nothing major. Feel free to merge when you think it's ready.

src/util.ts
aeisenberg: This is just copied from the original method. I can remove in order to simplify things.

src/util.ts
aeisenberg: Good point. I will move back.

src/util.ts
aeisenberg: I'm reluctant to change to `DRIVER` since `CODEQL_ACTION` is used pretty consistently already for env vars. I'll make sure to be clear in the comments.

src/util.ts
aeisenberg: Just copied from earlier. I will remove.

src/util.ts
aeisenberg: Moved back to the `actions-util.ts` module.

aeisenberg:
Thanks for the review!

PR #540, opened 2021-06-01
Remove local environment running

aeisenberg:
This is a functionality that never worked perfectly and hasn't been
used for a while.

This allows developers to run the action on their local machine, but
the run was always flaky and never 100% mirrored what was happening on
the actions runner.


adityasharad (approved):
Sounds reasonable if we're not using this for local testing.

robertbrignull (approved):
LGTM too

PR #542, opened 2021-06-02
Update unguarded-action-lib.ql to catch uses of actions-util.ts

robertbrignull:
I'm not 100% sure this change is correct but the PR checks should tell us as this should introduce at least one alert because it'll now spot accesses of `actions-util.ts`. Because it's using `matches` but then with a string with no wildcards in it I'm assuming this must be a mistake.

Alternatively we could probably just use `getImportedPath().getValue() = "./actions-util")` since we currently have all our source in one directory. Is that better? Using `matches` lets us catch this potential future edge case.

### DIFF
--- a/queries/unguarded-action-lib.ql
+++ b/queries/unguarded-action-lib.ql
@@ -43,7 +43,7 @@ class ActionsLibImport extends ImportDeclaration {
   ActionsLibImport() {
     getImportedPath().getValue().matches("@actions/%") and
     not isSafeActionLib(getImportedPath().getValue()) or
-    getImportedPath().getValue().matches("/actions-util$")
+    getImportedPath().getValue().matches("%/actions-util$")
   }
 
   string getName() {


adityasharad (approved):
Sensible. I was similarly worried that we would break the existing pattern by moving files around.

aeisenberg:
Thanks for the fix. It's not showing any new alerts. Is that expected?

robertbrignull:
> It's not showing any new alerts. Is that expected?

I was hoping this would generate a new alert. I think that it should be catching the import of `setMode` in `runner.ts` should be triggering an alert. I made this PR just off a hunch but probably someone will need to try the query out locally to work it why it's not catching that. Unless of course I'm mistaken and you think that shouldn't be an alert.

aeisenberg:
Actually, I think the reason is that you based this change on `main`, but the offending code was still in a PR. When I rebased #539 on main, I started seeing the warning.

robertbrignull:
Ah thanks for working that out. That makes perfect sense.

PR #543, opened 2021-06-03
Cleanup CodeQL DBs and output their location for later steps

edoardopirovano:
This PR does two things that should facilitate user workflows that want to upload CodeQL databases somewhere:

1. Databases created during the `analyze` step are cleaned up at the end of the step (i.e. items that are not needed to run further queries like evaluation logs and the cache are removed). The level of cleanup performed is configurable via a new parameter of the `analyze` step. I don't think we should expose this in user-facing documentation as most users will just want the default, but it may be useful for internal debugging to have an option there.
2. The `analyze` step now has an output, `db-locations` that contains a JSON object mapping from languages that a CodeQL database was created for to the location of this database on disk. I've modified one of the PR checks to use this to verify that it is populated correctly.


analyze/action.yml
```diff
@@ -13,6 +13,10 @@ inputs:
     description: Upload the SARIF file
     required: false
     default: "true"
+  cleanup-level:
+    description: "Level of cleanup to perform on CodeQL databases at the end of the analyze step. This should either be 'none' to skip cleanup, or be a valid argument for the --mode flag of the CodeQL CLI command 'codeql database cleanup'"
```
aeisenberg, line 17: Would be nice if you could list the valid options here. Or maybe provide a link to the documentation (or both).

analyze/action.yml
```diff
@@ -34,6 +38,9 @@ inputs:
     default: ${{ github.token }}
   matrix:
     default: ${{ toJson(matrix) }}
+outputs:
+  db-locations:
+    description: A map from languages that CodeQL databases were created for to their location.
```
aeisenberg, line 43:
```suggestion
    description: A map from language to absolute path for each database created by CodeQL.
```

aeisenberg (commented):
Looks good. A couple of suggestions around comments.

analyze/action.yml
edoardopirovano: I'm hesitant to list the options because it duplicates where we have to make a documentation change if we modify them later. Have added a link to the CLI docs, though.

aeisenberg (commented):
👍🏼

aeisenberg (approved):

PR #545, opened 2021-06-03
Allow the codeql-action to run packages

aeisenberg:
This commit adds a `packs` option to the codeql-config.yml file. Users
can specify a list of ql packs to include in the analysis.

For a single language analysis, the packs property looks like this:

```yaml
packs:
  - pack-scope/pack-name1@1.2.3
  - pack-scope/pack-name2   # no explicit version means download the latest
```

For multi-language analysis, you must key the packs block by lanaguage:

```yaml
packs:
  cpp:
    - pack-scope/pack-name1@1.2.3
    - pack-scope/pack-name2
  java:
    - pack-scope/pack-name3@1.2.3
    - pack-scope/pack-name4
```

This implementation adds a new analysis run (alongside custom and 
builtin runs). The unit tests indicate that the correct commands are
being run, but I have not actually tried this with a real CLI.

Also, convert `instanceof Array` to `Array.isArray` since that is
sightly better in some situations. See:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs_isarray


src/config-utils.test.ts
```diff
@@ -1177,3 +1310,109 @@ test("path sanitisation", (t) => {
...
+);
+test(
+  "not an array",
+  parsePacksErrorMacro,
+  { [Language.cpp]: "c/d" },
+  [Language.cpp],
+  /The configuration file "\/a\/b" is invalid: property "packs" must be an array of non-empty strings/
+);
+
+test(invalidPackNameMacro, "c");
```
edoardopirovano, line 1414: Maybe add a comment about why this is invalid? It's not immediately obvious to me.

src/config-utils.ts
```diff
@@ -954,6 +1016,78 @@ async function loadConfig(
   };
 }
 
+// Only alpha-numeric characters, with `-` allowed as long as not the first or last char
+const PACK_IDENTIFIER_PATTERN = (function () {
```
edoardopirovano, line 1020: Explain what constitutes a valid pack here?

src/config-utils.ts
```diff
@@ -1019,11 +1153,11 @@ export async function initConfig(
   // The list of queries should not be empty for any language. If it is then
   // it is a user configuration error.
   for (const language of config.languages) {
-    if (
-      config.queries[language] === undefined ||
-      (config.queries[language].builtin.length === 0 &&
-        config.queries[language].custom.length === 0)
-    ) {
+    const hasPacks = config.packs[language]?.length > 0;
+    const hasQueries =
```
edoardopirovano, line 1157: Maybe split this into `hasBuiltinQueries` and `hasCustomQueries`?

edoardopirovano (commented):
Mostly looks good, a few minor comments/questions

src/config-utils.ts
```diff
@@ -31,6 +33,11 @@ export interface UserConfig {
   }>;
   "paths-ignore"?: string[];
   paths?: string[];
+
+  // If this is a multi-language analysis, then the packages must be split by
+  // language. If this is a single language analysis, then no split by
+  // language is necessary.
+  packs?: Record<string, string[]> | string[];
```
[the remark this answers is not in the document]
aeisenberg: My thinking here is that 90% of the time, repositories will work on a single language. It would be a shame if you need to specify the language in these cases even though there's only one.

src/codeql.ts
```diff
@@ -121,6 +128,17 @@ export interface ResolveQueriesOutput {
   };
 }
 
+export interface PackDownloadOutput {
+  packs: PackDownloadItem[];
+}
+
+interface PackDownloadItem {
```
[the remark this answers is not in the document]
aeisenberg: I like this idea, seems like a lot of work, though. Maybe good for a hackathon?

src/codeql.ts
```diff
@@ -747,9 +766,43 @@ function getCodeQLForCmd(cmd: string): CodeQL {
       }).exec();
       return output;
     },
+
+    // Download specified packs into the package cache. If the specified
+    // package and version already exists (e.g., from a previous analysis run),
```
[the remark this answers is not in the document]
aeisenberg: Each run of the CLI will re-request the list of available versions. So, if a new version is uploaded then the CLI will download it on subsequent runs.

I will add a comment.

src/codeql.ts
```diff
@@ -747,9 +766,43 @@ function getCodeQLForCmd(cmd: string): CodeQL {
...
+            output += data.toString("utf8");
+          },
+        },
+      }).exec();
+
+      try {
+        return JSON.parse(output) as PackDownloadOutput;
+      } catch (e) {
+        throw new Error(
+          `Attempted to download specified packs but got error ${e}.`
```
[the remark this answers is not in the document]
aeisenberg: good point.

adityasharad:
Is now a good point to create a feature branch?

aeisenberg:
Sure. I'll create one now.

src/analyze.ts
```diff
@@ -291,6 +302,26 @@ export async function runQueries(
...
+) {
+  return packsWithVersion.map(packWithVersionToQuerySuiteEntry).join("\n");
+}
+
+function packWithVersionToQuerySuiteEntry(
+  pack: configUtils.PackWithVersion
+): string {
+  let text = `- qlpack: ${pack.packName}`;
+  if (pack.version) {
+    text += `${"\n"}  version: ${pack.version}`;
```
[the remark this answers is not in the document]
RA80533: @aeisenberg This is actually not the case. `${"\n"}` is evaluates to `\n` which will display a newline when printed. You might be thinking of String.raw`\n` which would produce "\n" when printed.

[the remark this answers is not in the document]
aeisenberg: You know what...I was just mistaken about this....either way should be fine:

```
$ node
Welcome to Node.js v12.16.3.
> console.log(`\n`)


undefined
> console.log(`${"\n"}`)


undefined
```

I'll change to the simpler version.

src/config-utils.ts
```diff
@@ -954,6 +1016,81 @@ async function loadConfig(
...
+  return new RegExp(`^${component}/${component}$`);
+})();
+
+// Exported for testing
+export function parsePacks(
+  packsByLanguage: string[] | Record<string, string[]> | undefined,
+  languages: Language[],
+  configFile: string
+) {
+  const packs = {} as Packs;
```
RA80533, line 1036: Should `Packs` be updated such that `{}` would be valid? This sort of pattern appears in a few other places here.

src/config-utils.ts
```diff
@@ -954,6 +1016,81 @@ async function loadConfig(
...
+const PACK_IDENTIFIER_PATTERN = (function () {
+  const alphaNumeric = "[a-z0-9]";
+  const alphaNumericDash = "[a-z0-9-]";
+  const component = `${alphaNumeric}(${alphaNumericDash}*${alphaNumeric})?`;
+  return new RegExp(`^${component}/${component}$`);
+})();
+
+// Exported for testing
+export function parsePacks(
+  packsByLanguage: string[] | Record<string, string[]> | undefined,
```
RA80533, line 1032: What is the use case for `packsByLanguage` to be `undefined`?

src/config-utils.ts
```diff
@@ -954,6 +1016,81 @@ async function loadConfig(
...
+        [languages[0]]: packsByLanguage,
+      };
+    } else {
+      // this is an error since multi-language analysis requires
+      // packs split by language
+      throw new Error(getPacksInvalidSplit(configFile));
+    }
+  }
+
+  for (const [lang, packsArr] of Object.entries(packsByLanguage)) {
```
RA80533, line 1055: Was this line intended to take place if `packsByLanguage` is of type `Record<string, string[]>`?

src/config-utils.ts
```diff
@@ -954,6 +1016,81 @@ async function loadConfig(
...
+    if (!Array.isArray(packsArr)) {
+      throw new Error(getPacksInvalid(configFile));
+    }
+    if (!languages.includes(lang as Language)) {
```
RA80533, line 1059: This type assertion is seemingly harmless on this line where it simply allows `lang` to be passed into what appears to be an array search method. It's afterwards, however, where issues arise due to the fact that `lang` is now asserted to be of type `Language`—even if it was not found in the languages array. If it was passed elsewhere, the type safety of code elsewhere would suffer because it would be assumed to be typed correctly rather than asserted.

src/config-utils.ts
```diff
@@ -954,6 +1016,81 @@ async function loadConfig(
...
+  if (!packsByLanguage) {
+    return packs;
+  }
+
+  if (Array.isArray(packsByLanguage)) {
+    if (languages.length === 1) {
+      // single language analysis, so language is implicit
+      packsByLanguage = {
+        [languages[0]]: packsByLanguage,
+      };
```
RA80533, line 1047: `packsByLanguage` was passed as a reference. Its reassignment will not extend outside the scope of the function.

src/config-utils.ts
aeisenberg: I had the type definition wrong. Originally, it was:

```typescript
export type Packs = Record<Partial<Language>, PackWithVersion[]>;
```

But it should have been:

```typescript
export type Packs = Partial<Record<Language, PackWithVersion[]>>;
```

src/config-utils.ts
aeisenberg: This parameter is coming directly from the `parsedYAML[PACKS_PROPERTY]` property, parsed from the file. This field is optional and hence possibly undefined.

src/config-utils.ts
aeisenberg: See https://github.com/github/codeql-action/blob/06687e95c8495c63469eb5f9b19c2ddca3aba075/src/config-utils.ts#L1045-L1048

In the case that `packsByLanguage` is of type `string[]`, it gets reassigned to something of type `Record<string, string[]>`. And there is only one key, the language of the database. And in the case `packsByLanguage` is of type `string[]` and there are multiple databases, this is an error.

src/config-utils.ts
aeisenberg: If `lang` is not in the array, that is an error. 

If `lang` is in the array, then we know `lang` is of type `Language` since we know that the array only allows things of type `Language` in it. Maybe I'm misunderstanding, but that seems safe to me.

src/config-utils.ts
aeisenberg: That's expected and what we want. We need to preserve the original user input and we also need to generate the `packs` object appropriately.

aeisenberg:
OK...I've added an integration test, which is now passing. And this PR is now ready for a proper review. For now, I'll keep it in the `packs/v1` branch until we want to do an official release.

src/config-utils.ts
RA80533: Oh, there's a conditional block right above that call site that it should be brought up inside.

src/config-utils.ts
```diff
@@ -927,7 +986,7 @@ async function loadConfig(
   }
 
   if (PATHS_PROPERTY in parsedYAML) {
-    if (!(parsedYAML[PATHS_PROPERTY] instanceof Array)) {
+    if (!Array.isArray(parsedYAML[PATHS_PROPERTY])) {
       throw new Error(getPathsInvalid(configFile));
     }
     for (const includePath of parsedYAML[PATHS_PROPERTY]!) {
```
RA80533, line 992: Move this into an `else` block.

src/analyze.ts
```diff
@@ -201,14 +213,25 @@ export async function runQueries(
...
+      if (packsWithVersion.length > 0) {
+        const { sarifFile, stdout } = await runQueryGroup(
+          language,
+          "packs",
+          createPackSuiteContents(packsWithVersion),
+          temporarySarifDir,
+          undefined
+        );
+        customAnalysisSummaries.push(stdout);
+        temporarySarifFiles.push(sarifFile);
```
edoardopirovano, line 233: This will need a bit of care with the merge as it non-trivially conflicts with https://github.com/github/codeql-action/pull/556 which is now merged (for instance, the `temporarySarifFiles` variable no longer exists).

src/analyze.ts
```diff
@@ -201,14 +213,25 @@ export async function runQueries(
...
+            createQuerySuiteContents(queries["custom"][i].queries),
             temporarySarifDir,
             queries["custom"][i].searchPath
           );
           customAnalysisSummaries.push(stdout);
           temporarySarifFiles.push(sarifFile);
         }
       }
+      if (packsWithVersion.length > 0) {
+        const { sarifFile, stdout } = await runQueryGroup(
```
edoardopirovano, line 225: Note the timings for queries run from packs will be included in the status report as part of the custom queries. This seems reasonable to me and I don't think we need a separate timing category.

edoardopirovano (commented):
I'd mostly already reviewed the draft version of this earlier, so only a few minor points.

src/analyze.ts
aeisenberg: OK...I'll merge with main. And, my new work messes up your beautiful log pruning that you've been working on. I'll fix that as well.

src/config-utils.ts
aeisenberg: The ts compiler isn't catching that `!Array.isArray(parsedYAML[PATHS_PROPERTY])` asserts that the use of the property below will never be falsy. I prefer avoiding type assertions, so I'll try to get rid of it.

src/codeql.ts
```diff
@@ -741,15 +760,66 @@ function getCodeQLForCmd(cmd: string): CodeQL {
       await new toolrunner.ToolRunner(cmd, args, {
         listeners: {
           stdout: (data: Buffer) => {
-            output += data.toString("utf8");
+            output += data.toString();
           },
         },
       }).exec();
```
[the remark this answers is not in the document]
aeisenberg: The default value for the encoding is `utf8`. See https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end, so this should work fine and it is what we are doing in other parts of this module.

src/analyze.ts
aeisenberg: I think I did this properly, but would like another look by you to be sure.

edoardopirovano:
@aeisenberg Looks like something went wrong with the rebase and this PR now has a bunch of unrelated commits in it?

aeisenberg:
Right. The problem is that I merged in main, but this PR is based off of `pack/v1`. I think I will retarget this PR so it's based on main.

edoardopirovano:
> Right. The problem is that I merged in main, but this PR is based off of `pack/v1`. I think I will retarget this PR so it's based on main.

Right, or if you don't want it to be in `main` yet you could also (separately from this PR) merge `main` into `pack/v1` and then target the PR onto that.

aeisenberg:
I think we can target main. At this point, the changes are mostly complete. Anything left over would be a bug in the initial implementation. I'd rather merge into main sooner. 

This should be backwards compatible.

.github/workflows/pr-checks.yml
```diff
@@ -100,6 +100,49 @@ jobs:
...
+        mkdir ../action
+        mv * .github ../action/
+        mv ../action/tests/multi-language-repo/{*,.github} .
+        mv ../action/.github/workflows .github
+    - uses: ./../action/init
+      with:
+        config-file: ".github/codeql/codeql-config-packaging.yml"
+        languages: javascript
+        # TODO: this is temporary until we have a release that includes the latest packaging work.
+        tools: https://github.com/dsp-testing/aeisenberg-codeql-action-packaging/releases/download/codeql-bundle-20210606/codeql-bundle-linux64.tar.gz
```
edoardopirovano, line 122: If you end up merging this after the CLI release I guess we can have that here. Also happy for it to (temporarily) stay like this, though.

edoardopirovano (approved):
This all looks good to me now!

PR #546, opened 2021-06-04
Add `--sarif-group-rules-by-pack` flag

edoardopirovano:
This adds the `--sarif-group-rules-by-pack` to invocations of `database analyze` which will result in the SARIF produced having rule objects grouped under the QL pack they correspond to rather than appearing at the top level.

- [ ] ~Confirm this change is backwards compatible with existing workflows.~ This change isn't _quite_ compatible with existing workflows if the users are consuming the SARIF in a way other than uploading it to Code Scanning, as the `tool.driver.rules` field disappears in favour of `<toolComponent>.rules` fields. As I understand it we don't support this use case, though.


adityasharad (approved):
Thanks!

Could you please add a change note? You can link to https://codeql.github.com/docs/codeql-cli/manual/database-analyze/#cmdoption-codeql-database-analyze-no-sarif-group-rules-by-pack or https://github.com/github/codeql-cli-binaries/blob/main/CHANGELOG.md#release-251-2021-04-19, or use the text there as a reference while writing the note. I want to make clear that this will not change most behaviour, but the rule objects will be in different places in the SARIF produced.

PR #548, opened 2021-06-04
Replace analyze with run-queries and interpret-results

edoardopirovano:
Currently, we call `database analyze` on the database several times, resulting in a number of SARIF files being created that we have to merge. This PR changes this to instead call `database run-queries` and then perform a call to `database interpret-results` to create a single SARIF file (per language).


src/analyze.ts
```diff
@@ -176,70 +175,43 @@ export async function runQueries(
...
+            language,
+            "builtin",
+            queries["builtin"],
+            undefined
+          )
         );
-        analysisSummaryBuiltIn = stdout;
-        await injectLinesOfCode(sarifFile, language, locPromise);
-
         statusReport[`analyze_builtin_queries_${language}_duration_ms`] =
```
adityasharad, line 189: Note this means the status report no longer includes result interpretation time. I think that's ok.
@robertbrignull should/could we have a separate telemetry category for interpretation, now that is a separate step?

src/codeql.ts
```diff
@@ -747,6 +773,71 @@ function getCodeQLForCmd(cmd: string): CodeQL {
...
+      addSnippetsFlag: string,
+      threadsFlag: string,
+      automationDetailsId: string | undefined
+    ): Promise<string> {
+      const args = [
+        "database",
+        "interpret-results",
+        threadsFlag,
+        "--format=sarif-latest",
+        "--print-metrics-summary",
```
adityasharad, line 820: This is going to conflict with your other PR. Make sure you include the pack grouping flag.

src/codeql.ts
```diff
@@ -747,6 +773,71 @@ function getCodeQLForCmd(cmd: string): CodeQL {
...
+        "-v",
+        `--output=${sarifFile}`,
+        addSnippetsFlag,
+        ...getExtraOptionsFromEnv(["database", "interpret-results"]),
+      ];
+      if (automationDetailsId !== undefined) {
+        args.push("--sarif-category", automationDetailsId);
```
adityasharad, line 827: @aeisenberg does the `automationDetailsId` passed in here ever end with a `run-id` component that we want to remove before treating it as the category?

src/analyze.ts
```diff
@@ -176,70 +175,43 @@ export async function runQueries(
...
-
+      if (ranCustom) {
         statusReport[`analyze_custom_queries_${language}_duration_ms`] =
           new Date().getTime() - startTimeCustom;
       }
-      logger.endGroup();
-
-      // Print the LoC baseline and the summary results from database analyze for the standard
-      // query suite and (if appropriate) each custom query suite.
-      logger.startGroup(`Analysis summary for ${language}`);
```
adityasharad, line 224: Looks like you have lost this log group and summary output. You still need to capture the stdout from `interpret-results` and log it here. The difference from the earlier code is that we only need to log one summary, rather than having separate summaries for builtin and custom queries.

`printLinesOfCodeSummary` only includes the baseline produced by the Action, not the summary produced by the CLI.

src/codeql.ts
```diff
@@ -747,6 +773,71 @@ function getCodeQLForCmd(cmd: string): CodeQL {
...
+      args.push(querySuite);
+      let output = "";
+      await new toolrunner.ToolRunner(cmd, args, {
+        listeners: {
+          stdout: (data: Buffer) => {
+            output += data.toString("utf8");
+          },
+        },
+      }).exec();
+      return output;
```
adityasharad, line 805: We're not using stdout from `run-queries`, so no need to capture or return it.

src/codeql.ts
aeisenberg: A `/` is always appended to the `--sarif-category` if it doesn't already end in one. It's handled in exactly the same way for `database analyze` as it is for `interpret results`. So, I think it's fine using the same logic for both.

src/analyze.test.ts
```diff
@@ -36,10 +36,17 @@ test("status report fields and search path setting", async (t) => {
 
     for (const language of Object.values(Language)) {
       setCodeQL({
-        databaseAnalyze: async (
```
aeisenberg, line 39: Can we completely remove `databaseAnalyze` now? The only place it was used before has been removed.

src/codeql.ts
adityasharad: I'm worried about the opposite direction: is the Action passing an `automationDetailsId` of the form `category/id` where we need to strip off the `id` before passing it to the CLI? If not, there's nothing to see here.

src/analyze.ts
edoardopirovano: Is it still needed? It looks like the stdout from the `interpret-results` call is showing up just fine in the Actions logs: https://github.com/github/codeql-action/runs/2748181217?check_suite_focus=true#step:5:2435

My understanding of the previous code was that it was there because otherwise the summary might have been further back in the logs (in particular if any custom queries were run) and we wanted to duplicate it at the end for convenience - since now `interpret-results` is always the last thing we do I think it isn't needed any more.

@aeisenberg Am I right in my understanding or is the `logger` output also used for something other than the Actions logs that needs to see that summary?

src/codeql.ts
edoardopirovano: Thanks for the reminder, I've rebased now making sure to include the flag.

src/analyze.ts
edoardopirovano: I've added it in since it seems potentially useful to have.

src/codeql.ts
edoardopirovano: I'm not entirely clear on what `automationDetailsId` is actually doing, but it is coming from user input (https://github.com/github/codeql-action/blob/242fd828aafa0b220c6a617507ca0eb48c4ec17c/src/analyze-action.ts#L88) so I'm not sure we have any guarantees about what's contained in it.

src/codeql.ts
edoardopirovano: Good point, done.

src/analyze.test.ts
```diff
@@ -36,10 +36,17 @@ test("status report fields and search path setting", async (t) => {
...
-          _,
-          sarifFile: string,
+        databaseRunQueries: async (
+          _db: string,
           searchPath: string | undefined
+        ) => {
+          searchPathsUsed.push(searchPath!);
```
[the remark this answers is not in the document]
edoardopirovano: Thanks, done.

src/analyze.ts
```diff
@@ -253,13 +225,29 @@ export async function runQueries(
 
   return statusReport;
 
+  async function runInterpretResults(
+    language: Language,
+    querySuites: string[],
```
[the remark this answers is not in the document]
edoardopirovano: Yep, changed here and elsewhere.

src/analyze.test.ts
edoardopirovano: Yep, removed.

src/codeql.ts
aeisenberg: > I'm worried about the opposite direction: is the Action passing an `automationDetailsId` of the form `category/id` where we need to strip off the `id` before passing it to the CLI? If not, there's nothing to see here.

There's nothing we need to do. The category is passed to the call to analyze if it is defined. And that category will have a `/` appended to it if one isn't on it already.

If no category is passed to analyze, then in the `upload-results` phase, the action will inject a calculated category. This category will always have a `/` on it. See: 

https://github.com/github/codeql-action/blob/3708898bf22a065449012e0b88abd0ecdb37ca57/src/upload-lib.ts#L54-L54

src/analyze.ts
aeisenberg: Hmmm...I wasn't aware of the duplication, but that makes sense. Though, now there are a lot of lines that are ungrouped (not exactly sure how it was before). Could you enclose the call to `interpret-results` in a logger group?

In fact, it would be really nice if we could put  run queries in a group as well.

aeisenberg: Now that I think about it, though, if you put `interpret-results`  in a group, then the summary table would be hidden, and I think that should be prominently displayed.

What do you think of both grouping `interpret-results` and duplicating the table outside of the group? I think that would make the logs more readable.

edoardopirovano: Seems like a good idea! Just pushed a commit with that, let's see what the logs in the PR checks look like but I agree it should neater 🙂

src/analyze.ts
```diff
@@ -208,14 +208,21 @@ export async function runQueries(
...
+        querySuitePaths,
+        sarifFile
+      );
       await injectLinesOfCode(sarifFile, language, locPromise);
       statusReport[`interpret_results_${language}_duration_ms`] =
         new Date().getTime() - startTimeInterpretResults;
-      printLinesOfCodeSummary(logger, language, await locPromise);
       logger.endGroup();
+      printLinesOfCodeSummary(logger, language, await locPromise);
+      logger.info(analysisSummary);
```
aeisenberg, line 225: Can you flip the output here?

```suggestion
      logger.info(analysisSummary);
      printLinesOfCodeSummary(logger, language, await locPromise);
```

aeisenberg (approved):
This looks great! I think it would be slightly better to flip the output (as in the suggestion). This is how it was before.

Also, I think the the _baseline_ comment is a bit awkward. We should probably change the text, but that could happen in another PR.

src/analyze.ts
adityasharad: I think that @AlonaHlobina has blog posts going out that include screenshots of the existing log groupings, so I was trying not to change it. This does make the summary more visible though, which is arguably even better than pointing users to the right group.

src/analyze.ts
robertbrignull: Sorry I missed this mention of me. What's done here should essentially be fine since the status report endpoint now ignores unknown fields, so it's ok to update the codeql-action first.

One thing though, could you please update the type definition at
https://github.com/github/codeql-action/blob/8e36bc2a6b646e44365338e137b78613a4bedfed/src/analyze.ts#L26
so that it includes the new fields. As you can see in this case that type is really only there for documentation purposes, but if it's there it would be nice if it's correct.

src/analyze.ts
AlonaHlobina: It is ok. I can include a new screenshot of it. The changelog is not published yet. Where can I see the new logs?

edoardopirovano: I've got another PR open that will further change the logs slightly, so if you need a screenshot now I would look at the run logs from that PR's checks, e.g. https://github.com/github/codeql-action/runs/2782645280?check_suite_focus=true#step:7:309

When that's merged to `main` (probably later today) the CodeQL Analysis workflows on `semmle-code` will also start producing the latest logs here: https://github.com/github/semmle-code/actions/workflows/codeql-analysis.yml

Issue #528, opened 2021-05-24
codeql-action/upload-sarif - JavaScript heap out of memory


jsjoeio:
We noticed recently `github/codeql-action/upload-sarif@v1` failing in our CI. 

```logs
Run github/codeql-action/upload-sarif@v1
  with:
    sarif_file: trivy-image-results.sarif
    checkout_path: /home/runner/work/code-server/code-server
    token: ***
    matrix: null
Uploading sarif files: ["trivy-image-results.sarif"]
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

<--- Last few GCs --->

[2047:0x42960e0]   211606 ms: Mark-sweep 1968.4 (1973.9) -> 1968.4 (1970.9) MB, 1146.9 / 0.0 ms  (average mu = 0.902, current mu = 0.000) last resort GC in old space requested
[2047:0x42960e0]   212790 ms: Mark-sweep 1968.4 (1970.9) -> 1968.4 (1970.9) MB, 1183.6 / 0.0 ms  (average mu = 0.811, current mu = 0.000) last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x1374fd9]
Security context: 0x1df0d8ac08a1 <JSObject>
    1: outputHash(aka outputHash) [0x23868395bba1] [/home/runner/work/_actions/github/codeql-action/v1/lib/fingerprints.js:~63] [pc=0x4e7bc427b85](this=0x2d8a3d4404a9 <undefined>)
    2: hash(aka hash) [0x4b2344f8471] [/home/runner/work/_actions/github/codeql-action/v1/lib/fingerprints.js:~41] [pc=0x4e7bc42400b](this=0x2d8a3d4404a9 <undefined>,0x23868395bbe1 <JSF...

 1: 0x9da7c0 node::Abort() [/home/runner/runners/2.278.0/externals/node12/bin/node]
 2: 0x9db976 node::OnFatalError(char const*, char const*) [/home/runner/runners/2.278.0/externals/node12/bin/node]
 3: 0xb39f1e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/home/runner/runners/2.278.0/externals/node12/bin/node]
[...]

PR #550, opened 2021-06-07
Fix out of memory in hash computation

edoardopirovano:
Closes https://github.com/github/codeql-action/issues/528 by making two changes:
1. As @adityasharad points out, we probably shouldn't try hashing files for results that don't have a line number, since these are probably not source code files. This will reduce the number of files we need to hash, and avoid us trying to hash large files like tarballs.
2. As @RA80533 points out, we could avoid the OOM even if we _do_ encounter a large file by computing the hash in a streaming manner so that we never need to store the whole thing in memory.

Either change individually would fix the issue above, but having both seems the most robust solution.


src/fingerprints.test.ts
```diff
@@ -92,16 +105,20 @@ test("hash", (t: ava.Assertions) => {
...
+      "e54938cc54b302f1:1",
+      "bb609acbe9138d60:1",
+      "1131fd5871777f34:1",
+      "5c482a0f8b35ea28:1",
+      "54517377da7028d2:1",
+      "2c644846cb18d53e:1",
+      "f1b89f20de0d133:1",
+      "c129715d7a2bc9a3:1",
+    ]
+  );
```
aeisenberg, line 121: I think this needs another test that includes a sarif file without line numbers.

aeisenberg (commented):
Looks great. One more test would be good and then 🚢 .

src/fingerprints.ts
```diff
@@ -113,7 +112,19 @@ export function hash(callback: hashCallback, input: string) {
...
     }
     updateHash(current);
-  }
+  };
+
+  await new Promise((fulfill) => {
+    const readStream = fs.createReadStream(filepath, "utf8");
+    readStream.on("close", fulfill);
+    readStream.on("end", () => {
+      processCharacter(65535);
```
RA80533, line 121: What is 65535?

src/fingerprints.test.ts
edoardopirovano: The `fingerprinting2.input.sarif` test file was already missing line numbers, so that's showing the new behaviour of omitting the fingerprints (see [change in expected output](https://github.com/github/codeql-action/pull/550/files#diff-bd73afb22e09c3541b7fcf055fa9d7faffff1656799978fdd5f06a71383536b4)). I'm not sure if that was testing any other behaviour that is now being missed, it didn't seem to be doing anything particularly different from the first test case to me.

aeisenberg: Ah...yes. Was re yon my phone and didn't see the whole file.

aeisenberg (approved):

src/fingerprints.ts
```diff
@@ -113,7 +112,19 @@ export function hash(callback: hashCallback, input: string) {
...
+    });
+    readStream.on("data", (data) => {
+      for (let i = 0; i < data.length; ++i)
+        processCharacter(data.charCodeAt(i));
+    });
+  });
```
RA80533, line 127: Be careful about mixing async-await and callback patterns. Error handling gets tricky because thrown errors don't necessarily correlate to promise rejections.

In this case, this could be rewritten entirely in async-await like so:

```ts
const readStream = fs.createReadStream(filepath, "utf8");

for await (const chunk of readable) {
  // Handle the chunk
}

processCharacter(65535);
```

RA80533: Since this appears to be used for iterating over lines, the Readline API could be useful for consuming the stream:

```ts
const reader = readline.createInterface({
  input: fs.createReadStream(filepath, "utf8")
});

for await (const line of reader) {
  // Handle the line
}

processCharacter(65535);
```

src/fingerprints.ts
RA80533: Yeah, it doesn't seem like a reasonable thing to change given that it would affect the hashes. It might be best to move it into a constant so it doesn't remain a magic number in that case.

src/fingerprints.ts
edoardopirovano: Thanks for the suggestions! I've applied your first one in https://github.com/github/codeql-action/pull/550, it is indeed nicer than what I wrote. The second one doesn't quite fit our use case nicely because we actually want the new lines in the hash.

RA80533: > […] we actually want the new lines in the hash.

Good catch. I figured it would contain the line endings since you're able to redirect the stream directly to {stdout|stderr}.write but apparently it doesn't. :thinking:

PR #563, opened 2021-06-16
Check if on default branch before uploading database

robertbrignull:
(Note this is merging into a feature branch, and not into `main`)

We only want to upload a CodeQL database if analyzing the default branch, and this will avoid unnecessary uploads and help ensure we always have a recent database uploaded. For now we don't care about other branches and don't want them polluting the cache of databases for the default branch.

I initially thought we would have to check this on the github side, but turns out we can easily find out the default branch of the repository from the action.

I have tested this by running this branch of the action from a branch, a PR, and on master, and confirmed the behaviour in each case.


rneatherway (approved):
Perfect, looks good

src/actions-util.ts
```diff
@@ -691,3 +691,19 @@ export function getRelativeScriptPath(): string {
...
+export async function isAnalyzingDefaultBranch(): Promise<boolean> {
+  // Get the current ref and trim and refs/heads/ prefix
+  let currentRef = await getRef();
+  currentRef = currentRef.startsWith("refs/heads/")
+    ? currentRef.substr("refs/heads/".length)
+    : currentRef;
+
+  const eventJson = JSON.parse(
+    fs.readFileSync(getRequiredEnvParam("GITHUB_EVENT_PATH"), "utf-8")
+  );
```
aeisenberg, line 706: A bit late on this comment, but I'd suggest wrapping this in a try catch block, just in case the file doesn't exist or it's malformed.

aeisenberg: Not a huge deal as the file is _supposed_ to be there, but this is just for safety.

robertbrignull: Good point, although what would it be able to do in this case? I think we'd be limited to just throwing an error with a better message, which is certainly still a valid thing to do.

aeisenberg: It might also be valid to return `false`, but probably safer to throw.

PR #573, opened 2021-06-18
Upload CodeQL databases

robertbrignull:
This adds the ability to upload the CodeQL database to github. These databases will be stored and can then be downloaded by users for running queries against. Currently though this feature is still in development and access will be restricted to github staff.

Most of this work has already been semi-reviewed on the `upload-database` branch, but it'll likely want a proper review now that we're going into the main branch. Also pay special attention to anything that could make this break for GHES/GHAE users using github connect.

I've also added [one more commit](https://github.com/github/codeql-action/commit/f3cd5fa0017f85a6a190a0aa9ae858d22c731a99) which adds an input to allow disabling this new behaviour. Although the upload only happens on github.com and only for repositories that are opted in to the new feature, I expect we'll still get requests for a way for users to disable this.

@aeisenberg, does this need a changelog entry? We do add a new input, but also for all regular users behaviour won't be changing either way because their repositories won't be opted in to the new feature. When we roll this out it won't be correlated with any particular codeql-action version.


rneatherway (approved):
Agreed we should wait for @aeisenberg to take a look. My feeling is that we don't need a changelog here because this isn't ready for end users yet.

aeisenberg (approved):
Looks fine to me. I'm leaning against a changelog as well. This is not a user facing feature (yet).

aeisenberg:
When query console becomes available to external users, we should add an entry.

PR #576, opened 2021-06-21
Handle errors during database upload

robertbrignull:
I've made a mistake here with my understanding of errors from octokit. I thought it always returned a response object, but that's not the case and for non-2XX responses it throws an exception instead. This manifested as https://github.com/github/codeql-action/runs/2873026409 which we didn't spot until it made it onto the `main` branch because that codepath doesn't execute during pull requests.

This change should be a fix for it. However this PR doesn't currently address the fact that we didn't catch it and there aren't tests for this area of the code. I'd like to also do that, ideally as part of this PR but could be afterwards instead.


robertbrignull:
I've recreated the `upload-database` branch so I've pushed this commit directly to there. No need for an extra pull request as it can be reviewed and tested before merging into `main`.

PR #578, opened 2021-06-21
Upload CodeQL databases - mk2

robertbrignull:
The first commit is a revert of https://github.com/github/codeql-action/pull/577 which is itself a revert of https://github.com/github/codeql-action/pull/573. Thus the first commit is the original contents we tried to merge in.

Not yet ready for review. I've merged in the fix from https://github.com/github/codeql-action/pull/576 which should mean the HTTP requests no longer crash the action. I'll keep this PR as a draft while I test it.


robertbrignull:
I tried running this branch on a test repo (that's not opted into database uploads) and it works and prints the nice message that uploading the database was skipped. I've made it only print the full ugly stack trace and request info in the case that it's not a 404. All the "expected" error cases are just logged at debug level so they won't surprise/confuse people, but then any unexpected cases like a 500 error will be printed slightly more visibly. I hope this will be a nice compromise.

Here's what it looks like
Screenshot 2021-06-21 at 12 31 34

and what it's like when the whole error is printed (this is from https://github.com/github/codeql-action/pull/578/commits/001318e810d3dd823a61c8cffbdc7a421513b186 but still shows what a 500 error would look like)
Screenshot 2021-06-21 at 12 31 58

robertbrignull:
I've finished writing tests. This turned out to be a huge pain. Either it's the testing or mocking library we're using, or just typescript in general, but I find doing mocking and sharing test code really painful in this repository. Anyway, here it is. The various code paths are now pretty well covered, and I've done more manual testing of this up-to-date branch.

The only non-test changes I've made are moving the `uploadDatabases` function out to its own file. You can't export and test methods from the actions-entrypoint files, but moving to a separate file is easy.

robertbrignull:
@rneatherway, @aeisenberg, this should now be ready for review again. As said above, the main fix here is https://github.com/github/codeql-action/pull/578/commits/001318e810d3dd823a61c8cffbdc7a421513b186 and then everything else is adding tests or other minor changes, but please do have a look through it all.
I've tested on a demo repo and I'm now pretty confident that won't cause a problem on `main` again.

src/database-upload.ts
```diff
@@ -0,0 +1,77 @@
...
+      owner: repositoryNwo.owner,
+      repo: repositoryNwo.repo,
+    });
+  } catch (e) {
+    if (util.isHTTPError(e) && e.status === 404) {
+      logger.debug(
+        "Repository is not opted in to database uploads. Skipping upload."
+      );
+    } else {
+      console.log(e);
```
aeisenberg, line 46: Why use console.log instead of something like logger.info. I see now that console.log is already being used elsewhere in the code, but not sure what the difference is.

aeisenberg (commented):
Looks good. Just some questions. Non-blocking if the answers don't require any changes.

src/database-upload.ts
```diff
@@ -0,0 +1,77 @@
...
+import { RepositoryNwo } from "./repository";
+import * as util from "./util";
+
+export async function uploadDatabases(
+  repositoryNwo: RepositoryNwo,
+  config: Config,
+  apiDetails: GitHubApiDetails,
+  logger: Logger
+): Promise<void> {
+  if (actionsUtil.getRequiredInput("upload-database") !== "true") {
```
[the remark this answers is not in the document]
robertbrignull: https://github.com/github/codeql-action/pull/578/files#diff-4f56858a6f315f152438a0c59f0d9e316f7065c75079efd37f54a28728fe14e4R40 should mean that is defaults to true and so will always upload unless the input is explicitly set to something else. This is how it behaved in my manual testing too.

I had to set the input explicitly each time in the tests otherwise it would fail but that shouldn't apply to production. I don't know if this technique of setting `required: true` but also setting a default value is ultimately the best, but it is something we do elsewhere too. I suppose we could alternatively have it not be required and set the default value in the code instead of in the yaml file, but that makes the default value harder to determine. Overall unless there's a strong objection I'd leave this as it is.

rneatherway (approved):
Looks good. I agree with you about difficulty testing some parts of this code. The only strategy that I have is to make as much of the code as pure (no side-effects) as possible and focus the testing on that.

For example, in this PR you could possibly have introduced a method for determining whether it's appropriate to upload the database, returning a boolean and a reason for not uploading. You could then at least avoid stubbing the logger.

src/database-upload.ts
robertbrignull: I think the main difference is that `console.log` will print the full stack trace, while `core.info` (which is called by `logger.info` on actions) doesn't and just prints the top message. I could be wrong here though. You're right though it is a bit of a pattern throughout the action and I count 12 cases of using `console.log` to print an exception. So I'm tempted to say lets also use it here, though we could investigate separate whether all of these could be `logger.info` instead and we wouldn't lose any functionality.

robertbrignull:
> For example, in this PR you could possibly have introduced a method for determining whether it's appropriate to upload the database, returning a boolean and a reason for not uploading. You could then at least avoid stubbing the logger.

robertbrignull:
I think something squiffy was going on with the merge commit. It was complaining about the javascript not being up to date when it was fine locally. I think this is likely a result of this PR being a revert of a revert, so we risk problems because commits could be already on the main branch, so merging does weird things.

I've done the nuclear option of `git reset --soft main` and then re-commit everything. This should avoid all of these issues now. You can see in https://github.com/github/codeql-action/compare/242b060e56857cd38751857f56e597d6e4001740..146c89790971fd60af8d5daca487d29b61d2e4f6 that my force-push didn't actually change any files.

robertbrignull:
Nope, it wasn't any of that. It was a file in `lib` for a source file that I introduced but then deleted, so therefore the file in `lib` shouldn't be there. I should think of the simple reason first before jumping to git shenanigans.  🤦

PR #591, opened 2021-06-23
Add new `packs` input to init action

aeisenberg:
This input allows users to specify which packs to run. It works in
unison with the packs block of the config file and it is similar to
how `queries` works. They both use `+` in the same way.

Note that the `#TODO` in the pr check is still around, but the CLI
is available. I will remove the TODO in the next commit.


.github/workflows/pr-checks.yml
```diff
@@ -143,6 +143,50 @@ jobs:
...
+      env:
+        TEST_MODE: true
+    - name: Assert Results
+      run: |
+        cd "$RUNNER_TEMP/results"
+        # We should have 3 hits from these rules
+        EXPECTED_RULES="javascript/example/empty-or-one-block javascript/example/empty-or-one-block javascript/example/two-block"
+
+        # use tr to replace newlines with spaces and xargs to trim leading and trailing whitespace
+        RULES="$(cat javascript.sarif | jq -r '.runs[0].results[].ruleId' | sort | tr "\n" " " | xargs)"
```
edoardopirovano, line 183: I also recently wrote a bit of code to parse SARIF results for a test ([ref](https://github.com/github/semmle-code/blob/main/odasa-buildutils/files/integration-tests/check-sarif.py)). Again, not for this PR but it might be worth thinking about a way to avoid having lots of ad-hoc solutions for this in different places. We could, for example, add a CLI command for this.

init/action.yml
```diff
@@ -22,6 +22,14 @@ inputs:
   queries:
     description: Comma-separated list of additional queries to run. By default, this overrides the same setting in a configuration file; prefix with "+" to use both sets of queries.
     required: false
+  packs:
+    description: >-
+      Comma-separated list of packs to run. Reference a pack in the format `scope/name[@version]`. If `version` is not
+      specified, then the latest version of the pack is used. By default, this overrides the same setting in a
+      configuration file; prefix with "+" to use both sets of packs.
+
+      This input is only available in single-language analyses.
```
edoardopirovano, line 31: I think this line should be expanded slightly to note that for multi-language analyses the user can specify packs in the configuration file. As it is currently written, a user might misunderstand this to mean that we can only run packs on single-language analyses.

src/config-utils.test.ts
```diff
@@ -1416,3 +1453,160 @@ test(invalidPackNameMacro, "c-/d");
...
+  "input with two languages",
+  parseInputAndConfigErrorMacro,
+  {},
+  "c/d",
+  [Language.cpp, Language.csharp],
+  /multi-language analysis/
+);
+
+test(
+  "input with + only",
```
edoardopirovano, line 1594: Hmm, I'm not so sure if this should be an error. We could interpret this as the user choosing to add no packs. This also seems reasonable, though, there's an argument that if the user specified something in this field they probably did mean to specify some packs so we should fail to give them a chance to notice this.

src/config-utils.ts
```diff
@@ -1068,12 +1075,44 @@ export function parsePacks(
   return packs;
 }
 
-function toPackWithVersion(packStr, configFile: string): PackWithVersion {
+function parsePacksInput(
```
edoardopirovano, line 1078: nit: `parsePacksFromInput` for consistency with `parsePacksFromConfig` above?

src/config-utils.ts
```diff
@@ -1068,12 +1075,44 @@ export function parsePacks(
...
+  packsInput: string | undefined,
+  languages: Language[]
+): Packs | undefined {
+  if (!packsInput?.trim()) {
+    return undefined;
+  }
+
+  if (languages.length > 1) {
+    throw new Error(
+      "Cannot specify a 'packs' input in a multi-language analysis. Use a codeql-config.yml file instead and specify packs by library."
```
edoardopirovano, line 1088:
```suggestion
      "Cannot specify a 'packs' input in a multi-language analysis. Use a codeql-config.yml file instead and specify packs by language."
```

src/config-utils.ts
```diff
@@ -1068,12 +1075,44 @@ export function parsePacks(
...
+    );
+  } else if (languages.length === 0) {
+    throw new Error("No languages specified. Cannot process the packs input.");
+  }
+
+  packsInput = packsInput.trim();
+  if (packsInput.startsWith("+")) {
+    packsInput = packsInput.substring(1).trim();
+    if (!packsInput) {
+      throw new Error("Remove the '+' from the packs input.");
```
edoardopirovano, line 1098: This could be a little more informative, e.g. _A + was used in the packs input to specify that you wished to add some packs to your CodeQL analysis. However, no packs were specified. Please either remove the + or specify some packs._

src/runner.ts
```diff
@@ -129,6 +130,14 @@ program
     "--queries <queries>",
     "Comma-separated list of additional queries to run. This overrides the same setting in a configuration file."
   )
+  .option(
+    "--packs <packs>",
+    `Comma-separated list of packs to run. Reference a pack in the format scope/name[@version]. If version is not
+    specified, then the latest version of the pack is used. By default, this overrides the same setting in a
+    configuration file; prefix with "+" to use both sets of packs.
+
+    This option is only available in single-language analyses.`
```
edoardopirovano, line 139: Is this generated from the above? If not my previous comment also applies here.

.github/workflows/pr-checks.yml
```diff
@@ -143,6 +143,50 @@ jobs:
...
+      run: |
+        mkdir ../action
+        mv * .github ../action/
+        mv ../action/tests/multi-language-repo/{*,.github} .
+        mv ../action/.github/workflows .github
+    - uses: ./../action/init
+      with:
+        config-file: ".github/codeql/codeql-config-packaging2.yml"
+        languages: javascript
+        packs: dsp-testing/codeql-pack1@0.0.4, dsp-testing/codeql-pack2
```
edoardopirovano, line 164: While the unit tests are reasonably convincing, it would be nice to also have an end-to-end test using `+` rather than only testing the (simpler) code path that doesn't have to combine the config file with the Action input.

edoardopirovano (commented):
Nice! This mostly looks good, a few comments inline.

.github/workflows/pr-checks.yml
```diff
@@ -143,6 +143,50 @@ jobs:
...
         fi
 
+  # tests that we can run packages through actions inputs
+  test-packaging-javascript-inputs:
+    needs: [check-js, check-node-modules]
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Move codeql-action
```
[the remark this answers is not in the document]
aeisenberg: Yep. That's a good idea.

.github/workflows/pr-checks.yml
aeisenberg: Possibly. Unless there's a real customer need for it, I'd be reluctant to add this as a CLI command. Though, we could create our own action, or some other tool that is shared across repos.

init/action.yml
aeisenberg: You're right. Though, hard to know what to add since the format in the yaml for multi-language is different. I don't want to go into too much detail. Ideally, I could add a link to online docs, but that's not available yet.

For now, I'll add a short, but incomplete description for what to do. I'll also add an `[Eperimental]` tag.

src/config-utils.test.ts
aeisenberg: That was my thinking. If someone has done this, it's _probably_ a mistake.

src/runner.ts
aeisenberg: Not generated. I'll do something similar as above.

.github/workflows/pr-checks.yml
aeisenberg: Sure. I'll add that.

.github/workflows/pr-checks.yml
```diff
@@ -187,6 +189,50 @@ jobs:
...
+      shell: bash
+      run: |
+        mkdir ../action
+        mv * .github ../action/
+        mv ../action/tests/multi-language-repo/{*,.github} .
+        mv ../action/.github/workflows .github
+    - uses: ./../action/init
+      with:
+        config-file: ".github/codeql/codeql-config-packaging3.yml"
+        packs: dsp-testing/codeql-pack1@0.0.4
```
edoardopirovano, line 206: I think you need a `+` here? Otherwise this LGTM now.

edoardopirovano (approved):

src/config-utils.ts
edoardopirovano: @aeisenberg Think you missed this comment? Unless you disagree, in which case no need to address it 🙂

aeisenberg: Right. Missed this. Thanks for pointing it out.

.github/workflows/pr-checks.yml
aeisenberg: Doh!

aeisenberg:
Build is failing because the version of the cli it is grabbing from the toolcache is not yet new enough to support all of the packaging commands.

PR #592, opened 2021-06-24
Improve error messages from CLI invocations

edoardopirovano:
This PR improves the error messages we get from CodeQL CLI invocations that went wrong. In particular, it captures the stderr of the program (up to a limit of 20,000 characters), and creates an error message that includes this as well as the command being run and the exit code. This error will bubble up to the top level where it should be caught and included in the status reports that get uploaded.


src/codeql.ts
```diff
@@ -884,14 +894,32 @@ export function getExtraOptions(
...
-  await new toolrunner.ToolRunner(cmd, args, {
+  let error = "";
+  const exitCode = await new toolrunner.ToolRunner(cmd, args, {
     listeners: {
       stdout: (data: Buffer) => {
         output += data.toString();
       },
+      stderr: (data: Buffer) => {
+        const toRead = Math.min(maxErrorSize - error.length, data.length);
+        error += data.toString("utf8", 0, toRead);
```
adityasharad, line 917: Unfortunately I don't think we're guaranteed UTF-8 encoded output. But that's hopefully rare, and partially incomplete errors are better than no errors.

adityasharad (commented):
Looks reasonable, though I think Andrew or Robert should take a look too.

src/codeql.ts
```diff
@@ -884,14 +894,32 @@ export function getExtraOptions(
...
 
+/*
+ * A constant defining the maximum number of characters we will keep from
+ * the programs stderr for logging. This serves two purposes:
+ * (1) It avoids an OOM if a program fails in a way that results it
+ *     printing many log lines.
+ * (2) It avoids us hitting the limit of how much data we can send in our
+ *     status reports on GitHub.com.
+ */
+const maxErrorSize = 20_000;
```
aeisenberg, line 905: Not a big deal, but is there a reason why you chose 20,000? Do you know what the max size for a status report is?

edoardopirovano: The max size of fields in the status report is 100kb and UTF8 characters can be up to four bytes so the max length of the message is 25k characters (assuming we're really unlucky and all the characters happen to require four bytes). I subtracted 5k to allow other space for the command invocation and arguments.

aeisenberg: Nice. Thanks for the explanation. Could you add something like that to the comment above just so this context doesn't get lost?

src/codeql.ts
edoardopirovano: Yeah I think it's reasonable to assume most of the errors we encounter will be UTF8. Certainly I think anything the CLI outputs typically is, unless a string we include in the error (file path, snippet from a QL file, etc.) isn't. I think it's okay if those cases don't get a perfect error message in the status reports.

src/codeql.ts
edoardopirovano: Yep, done!

aeisenberg (approved):
This is strictly better than we had before, and so I think it should be merged, but do you know if this fixes the error reporting problems discussed earlier?

edoardopirovano:
> This is strictly better than we had before, and so I think it should be merged, but do you know if this fixes the error reporting problems discussed earlier?

It will certainly make the error message better than what we had before. Whether they'll be _sufficiently_ better to actually gleam information from I think is something we'll have to establish once the data starts flowing in. Perhaps, we'll start to see patterns of common errors and we'll want to make special cases for them to distinguish them - for instance, clearly marking certain errors as user/configuration errors and others as CLI ones.

At the moment, however, we can't really do this because all the messages in status reports say the exact same (CLI failed with exit code 2) so we don't really know _what_ errors we need to distinguish. I suggest we wait a week or two from where this is merged and then take a look at what errors we have gotten and if/how we can improve them further.

src/codeql.ts
robertbrignull: > The max size of fields in the status report is 100kb

Just curious, where did this information come from? For my own benefit really as I didn't know what the limit was or where it comes from.

edoardopirovano: Ah, I misunderstood your comment on the issue as meaning that you knew 100kb was the limit rather than it being a suggestion. The database column allows storing up to 2GB apparently, but I'm not sure how much data the endpoint that receives status reports will accept.

robertbrignull: Ah, no worries. 25k characters is probably fine anyway, though you could increase the limit and it would be fine.

The API should accept large volumes. Not up to 2GB but I think the limit is around 400MB so easily high enough. There's also protobuf in the pipeline there but the limit on that is also big enough for anything we might want to send.

PR #594, opened 2021-06-24
Restore original getCodeQLActionRepository behaviour

simon-engledew:
In version 1.0.0 the logic for determining the `getCodeQLActionRepository` worked this way round:

https://github.com/github/codeql-action/blob/v1.0.0/src/codeql.ts#L139

In the latest version this logic was reversed during refactoring:

https://github.com/github/codeql-action/blob/v1.0.3/src/codeql.ts#L182

This is currently causing the codeql-runner to break with the error message:

`RUNNER_TEMP environment variable must be set`.

This PR restores the original behaviour.


aeisenberg (approved):
Thanks for catching and fixing this bug!

aeisenberg:
We didn't catch this earlier because all of our integration tests run within actions and the `RUNNER_TEMP` variable is already set.

PR #600, opened 2021-06-25
Run tests against nightly CLI bundles

edoardopirovano:
This PR changes the integration tests of the CodeQL Action to also run against a recent nightly build of the CodeQL bundle. This should help catch compatibility issues arising from changes in the Action earlier.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -267,22 +267,45 @@ jobs:
       with:
         tools: latest
         languages: javascript
-    - name: Compare default and latest CodeQL bundle versions
+    - name: Remove empty database
+      # allows us to run init a third time
+      run: |
+        rm -rf "$RUNNER_TEMP/codeql_databases"
+    - name: Get nightly release URL
+      id: get-url
+      env:
+        GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+      shell: bash
+      # This workflow step gets an unstable testing version of the CodeQL CLI. It should not be used outside of these tests.
+      run: |
+        export LATEST=`gh release list --repo dsp-testing/codeql-cli-nightlies -L 1 | cut -f 3`
+        echo "::set-output name=nightly-url::https://github.com/dsp-testing/codeql-cli-nightlies/releases/download/$LATEST/codeql-bundle.tar.gz"
+    - name: Init with a CodeQL bundle from a nightly release
+      id: init-nightly
+      uses: ./../action/init
+      with:
+        tools: ${{ steps.get-url.outputs.nightly-url }}
+        languages: javascript
+    - name: Compare CodeQL bundle versions
       id: compare
       env:
         CODEQL_DEFAULT: ${{ steps.init-default.outputs.codeql-path }}
         CODEQL_LATEST: ${{ steps.init-latest.outputs.codeql-path }}
+        CODEQL_NIGHTLY: ${{ steps.init-nightly.outputs.codeql-path }}
+        NIGHTLY_URL: ${{ steps.get-url.outputs.nightly-url }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -184,13 +195,8 @@ jobs:
...
         echo "Latest CodeQL bundle version is $CODEQL_VERSION_LATEST"
-        if [[ "$CODEQL_VERSION_DEFAULT" == "$CODEQL_VERSION_LATEST" ]]; then
-          # Just use `tools: null` to avoid duplication in the integration tests.
-          VERSIONS_JSON='[null]'
-        else
-          # Use both `tools: null` and `tools: latest` in the integration tests.
-          VERSIONS_JSON='[null, "latest"]'
-        fi
+        # Use both `tools: null` and `tools: (nightly URL)` in the integration tests.
+        VERSIONS_JSON='[null, "${{ steps.get-url.outputs.nightly-url }}"]'
```
adityasharad, line 199: I'd like to still be testing against `latest` because it exercises the hardcoded default configured within the Action repo, and sometimes may be different from both null and nightly. Can we matrix over all three?

adityasharad (changes requested):

.github/workflows/pr-checks.yml
edoardopirovano: Yep, done!

.github/workflows/pr-checks.yml
```diff
@@ -174,22 +174,44 @@ jobs:
...
         CODEQL_VERSION_LATEST="$("$CODEQL_LATEST" version --format terse)"
+        CODEQL_VERSION_NIGHTLY="$("$CODEQL_NIGHTLY" version --format terse)"
         echo "Default CodeQL bundle version is $CODEQL_VERSION_DEFAULT"
         echo "Latest CodeQL bundle version is $CODEQL_VERSION_LATEST"
+        echo "Nightly CodeQL bundle version is $CODEQL_VERSION_NIGHTLY"
         if [[ "$CODEQL_VERSION_DEFAULT" == "$CODEQL_VERSION_LATEST" ]]; then
-          # Just use `tools: null` to avoid duplication in the integration tests.
-          VERSIONS_JSON='[null]'
+          # Skip `tools: latest` since it would be the same as `tools: null`
+          VERSIONS_JSON='[null, "${{ steps.get-url.outputs.nightly-url }}"]'
```
adityasharad, line 211: Minor: You may want to pull this one into an env var too.

adityasharad (approved):
Looks good, thanks! Hopefully won't make too many matrixed jobs, otherwise we might have to split up the workflow. Merge after the nightly bundle is published.

.github/workflows/pr-checks.yml
edoardopirovano: Thanks, done.

edoardopirovano:
There's actually already a nightly release up from the workflow's testing, but the PR check is failing because https://github.com/github/semmle-code/pull/39627 isn't merged yet I think. We'll have to wait for the next nightly build after that's merged before this can go in 🙂

aeisenberg:
I'm a little bit confused. This looks like it only runs the `init` action with the nightly CLI. Should we be running analyze as well?

edoardopirovano:
> I'm a little bit confused. This looks like it only runs the `init` action with the nightly CLI. Should we be running analyze as well?

The way these tests are setup is a little weird. We use the `init` action because it gets the CodeQL CLI and puts it on disk so we can print out the versions that we are going to be running with in the later step. Then, these all get put in the `VERSIONS_JSON` matrix and all the later tests are run with all the versions in there. I'm not a huge fan of this structure, but it didn't seem necessary to change it.

PR #602, opened 2021-06-28
Support splitting of DB creation and query execution

edoardopirovano:
This PR makes a couple of changes that make it possible to write workflows that either don't create a CodeQL database, or don't run any CodeQL queries. In particular:
- The `analyze` step now supports a `skip-queries` parameter that tells it to not run any queries.
- The `analyze` step will now no longer error if a finalized database already exists. Instead, it will log an informational message, skip the finalize step, and proceed with query execution.


src/analyze-action.ts
```diff
@@ -161,7 +171,11 @@ async function run() {
     }
   }
 
-  await sendStatusReport(startedAt, stats);
+  if (runStats && uploadStats) {
+    await sendStatusReport(startedAt, { ...runStats, ...uploadStats });
+  } else if (runStats) {
+    await sendStatusReport(startedAt, { ...runStats });
```
robertbrignull, line 177: Ideally still send a successful status report here, even if we don't have `runStats`. You'll probably have to amend to types to in order to do that.

.github/workflows/pr-checks.yml
```diff
@@ -234,6 +232,54 @@ jobs:
...
+        mv * .github ../action/
+        mv ../action/tests/multi-language-repo/{*,.github} .
+        mv ../action/.github/workflows .github
+    - uses: ./../action/init
+      with:
+        config-file: ".github/codeql/codeql-config-packaging3.yml"
+        packs: +dsp-testing/codeql-pack1@0.0.4
+        languages: javascript
+        # TODO: this can be removed when cli v2.5.6 is released and available in the tool cache
+        tools: https://github.com/dsp-testing/aeisenberg-codeql-action-packaging/releases/download/codeql-bundle-20210615/codeql-bundle-linux64.tar.gz
```
aeisenberg, line 255:
```suggestion
        tools: latest
```

.github/workflows/pr-checks.yml
```diff
@@ -234,6 +232,54 @@ jobs:
...
+
+    - name: Build code
+      shell: bash
+      run: ./build.sh
+    - uses: ./../action/analyze
+      with:
+        skip-queries: true
+      env:
+        TEST_MODE: true
+    - uses: ./../action/analyze
```
aeisenberg, line 265: Before you run `analyze` a second time, you should check that there are indeed 0 results.

src/analyze-action.ts
```diff
@@ -82,15 +85,22 @@ async function run() {
...
-      util.getAddSnippetsFlag(actionsUtil.getRequiredInput("add-snippets")),
-      util.getThreadsFlag(actionsUtil.getOptionalInput("threads"), logger),
-      actionsUtil.getOptionalInput("category"),
-      config,
+    const threads = util.getThreadsFlag(
+      actionsUtil.getOptionalInput("threads"),
       logger
     );
+    await runFinalize(outputDir, threads, config, logger);
+    if (actionsUtil.getRequiredInput("skip-queries") !== "true") {
```
aeisenberg, line 93: Two comments:

1. `getRequiredInput` will fail if the input isn't there. I don't think that's what we want, since this input is optional.
2. Just a nit: Not sure how we are doing it for other boolean inputs, but do we want to support variants like `True`, `TRUE`, `yes`, `1`, etc?  Hmmm...and I see that you are using the same pattern we are using elsewhere. We should consider changing this, but in a separate PR.

src/analyze.ts
```diff
@@ -133,6 +137,25 @@ async function createdDBForScannedLanguages(
...
+) {
+  const dbPath = util.getCodeQLDatabasePath(config, language);
+  try {
+    const dbInfo = yaml.load(
+      fs.readFileSync(path.resolve(dbPath, "codeql-database.yml"), "utf8")
+    );
+    return !("inProgress" in dbInfo);
+  } catch (e) {
+    logger.warning(
+      `Could not check whether DB for ${language} was finalized. Assuming it is not.`
```
aeisenberg, line 153: nit: I think this reads better:

```suggestion
      `Could not check whether database for ${language} was finalized. Assuming it is not.`
```

src/analyze.ts
```diff
@@ -142,12 +165,18 @@ async function finalizeDatabaseCreation(
...
   for (const language of config.languages) {
-    logger.startGroup(`Finalizing ${language}`);
-    await codeql.finalizeDatabase(
-      util.getCodeQLDatabasePath(config, language),
-      threadsFlag
-    );
-    logger.endGroup();
+    if (dbIsFinalized(config, language, logger)) {
+      logger.info(
+        `There is already a finalized DB for ${language} at the location where the CodeQL Action places databases, so we did not create one.`
```
aeisenberg, line 170: nit: Same here: 

```suggestion
        `There is already a finalized database for ${language} at the location where the CodeQL Action places databases, so we did not create one.`
```

src/runner.ts
```diff
@@ -431,11 +431,13 @@ program
 
       const outputDir =
         cmd.outputDir || path.join(config.tempDir, "codeql-sarif");
-      await runAnalyze(
+      const threads = getThreadsFlag(cmd.threads, logger);
+      await runFinalize(outputDir, threads, config, logger);
+      await runQueries(
```
aeisenberg, line 436: So, we are not adding this functionality to the runner? This should be called out in the changelog.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## [UNRELEASED]
 
 - Fix `RUNNER_TEMP environment variable must be set` when using runner.
+- The `analyze` step now supports a `skip-queries` option to merely build the CodeQL database without analyzing.
```
aeisenberg, line 6: Can you add a link to this PR in the changelog entry for consistency?

Also, I think you should add something about handling already finalized databases.

aeisenberg (commented):
Nice.

I think it's great that we're running more unit-like integration tests on different options for the action. The `pr-checks` workflow is getting long. I wonder if we should be splitting these into a separate workflow.

src/analyze-action.ts
robertbrignull: I think using `getRequiredInput` is fine here because the input is defined as
```
  skip-queries:
    description: If this option is set, the CodeQL database will be built but no queries will be run on it. Thus, no results will be produced.
    required: false
    default: "false"
```
so if not given by the user then the actions runtime inserts the default value, so our code will always see a value present.

I did run into a problem during tests because then the value really will be missing and `getRequiredInput` will crash.

aeisenberg: Ah...got it. So, ignore my first comment. :)

.github/workflows/pr-checks.yml
edoardopirovano: Good point, done.

src/analyze-action.ts
edoardopirovano: In response to 2: I definitely agree we should support more variants of true. Let's leave it for a separate PR.

CHANGELOG.md
edoardopirovano: Done, though it is a rather long changelog entry now.

src/runner.ts
edoardopirovano: Yep, I don't think we need this in the runner (or at least, nobody has asked for it). I will address this in the changelog.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The `analyze` step of the Action now supports a `skip-queries` option to merely build the CodeQL database without analyzing. This functionality is not present in the runner. Additionally, the step will no longer fail if it encounters a finalized database, and will instead continue with query execution. [#602](https://github.com/github/codeql-action/pull/602)
```
aeisenberg, line 5: Nit: If you wanted to, you could turn this into 2 entries. This would make each entry shorter.

aeisenberg (approved):
LGTM.

CHANGELOG.md
edoardopirovano: Let's keep as is, it seems weird to have two entries referring to the same PR.

Issue #154, opened 2020-08-20
Support for locally present bundles


esbena:
I would like to run code-scanning with a custom code-scanning bundle
that already is present on my local disk.  I can currently do that by
making the following hack:

### 1 Install bundle in tool cache:

Use `toolCache.cacheDir` to install the bundle with a specially crafted ID:

```js
tc.cacheDir(localBundlePath, 'CodeQL', "0.0.0-XXX-DUMMY-BUNDLE-ID");
```

### 2 Convince codeql to look in the tool cache:

The codeql action is then invoked with a dummy URL that will make codeql.ts lookup the bundle with the specially crafted ID:

```yaml
        ...
        tools: dummy://codeql-bundle-XXX-DUMMY-BUNDLE-ID/dummy
        ...
```

---

This works, but I would like a more reliable solution.  I am not sure
what the best approach is for supporting my use case, but I am happy
to look into implementing a suggestion.

aibaars:
I think the `tools:` property could be made to allow `file://` URL.

edoardopirovano:
Agreed, this is something we also need for https://github.com/github/codeql-coreql-team/issues/1264 so I will be looking at it in the near future.

PR #606, opened 2021-06-28
Allow local instead of downloaded CodeQL

edoardopirovano:
Closes https://github.com/github/codeql-action/issues/154

This PR allows the `tools:` argument to accept an argument to specify a local CodeQL distribution instead of a remote one. This will be used internally for testing CodeQL distributions for compatibility with the Action. I do not think we need a changelog entry for this as it is not something users should be using.

_Reviewing note:_ This PR changed the indentation of a lot of a function that ended up inside of an `else`. Thus, the diff should be viewed with whitespace changes off for improved readability.


src/codeql.ts
```diff
@@ -312,102 +312,112 @@ export async function setupCodeQL(
...
   try {
-    // We use the special value of 'latest' to prioritize the version in the
-    // defaults over any pinned cached version.
-    const forceLatest = codeqlURL === "latest";
-    if (forceLatest) {
-      codeqlURL = undefined;
-    }
+    let codeqlFolder: string;
+    let codeqlURLVersion: string;
+    if (codeqlURL?.startsWith("file://")) {
```
adityasharad, line 317: We parse this in other code paths as a URL object. Would that make sense to do here? Then you can check the scheme is `file` and use the URL path to find the file, instead of string manipulation.

aeisenberg: It would be better to try to parse as a URL first, I think. `file://` is not technically correct (even though it is usually accepted). `file:/` and `file:///` is a correct file url. But, these details could be ignored by usng `new URL(codeqlURL).protocol` or something like that.

src/codeql.ts
```diff
@@ -312,102 +312,112 @@ export async function setupCodeQL(
...
+      codeqlFolder = await toolcache.extractTar(
+        codeqlURL.substr(7),
+        tempDir,
+        logger
+      );
+      codeqlURLVersion = "local";
```
aeisenberg, line 325: Don't know if we need to, but we _could_ get the codeql version by runnng `codeql --version`.

src/codeql.ts
```diff
@@ -312,102 +312,114 @@ export async function setupCodeQL(
...
-    // defaults over any pinned cached version.
-    const forceLatest = codeqlURL === "latest";
-    if (forceLatest) {
-      codeqlURL = undefined;
-    }
+    let codeqlFolder: string;
+    let codeqlURLVersion: string;
+    const parsedUrl: URL | undefined =
+      codeqlURL && codeqlURL !== "latest" ? new URL(codeqlURL) : undefined;
+    if (parsedUrl?.protocol === "file:") {
```
aeisenberg, line 319: Wonder if it would be better to just use a raw path to indicate a local file, rather than a `file:` url. Technically, relative file uris are not legal, so this: `file://../action/codeql-bundle.tar.gz` is invalid.

It's easy enough to distinguish a file path from an http(s) url (except for the degenerate case where someone has named a file `https://thisisafile`, which I don't think we need to handle). And doing it this way will be easier might be easier for users since they don't need to remember to use a `file:` protocol.

src/codeql.ts
edoardopirovano: As far as I can tell, the only use of this version number is that it goes in the status report and I think from the point of view of status reports, knowing that people are pinned to a local version of the CLI is probably more useful than having the exact version they happen to be on.

aeisenberg: Makes sense.

src/codeql.ts
edoardopirovano: Hmm, maybe let's go with @aeisenberg's suggestion below of not using `file://` URLs in the first place.

src/codeql.ts
edoardopirovano: Yeah, it seems reasonable to treat everything that doesn't look like a http(s) URL as a file path, let's do that.

aeisenberg (approved):

PR #607, opened 2021-06-29
Add a 'source-root' input to the init Action

mario-campos:


init/action.yml
```diff
@@ -38,6 +38,9 @@ inputs:
     description: Try to auto-install your python dependencies
     required: true
     default: 'true'
+  source-root:
+    description: Path to the root source-code directory, relative to the workspace.
```
aibaars, line 42: I saw you had trouble using `${{ github.workspace }}` in the description. In https://github.com/actions/checkout/blob/main/action.yml they use `$GITHUB_WORKSPACE` and it appears that
```
default: ${{ github.workspace }}
```
is also allowed.

src/init-action.ts
```diff
@@ -201,7 +203,12 @@ async function run() {
     const codeqlRam = process.env["CODEQL_RAM"] || "6500";
     core.exportVariable("CODEQL_RAM", codeqlRam);
 
-    const tracerConfig = await runInit(codeql, config);
+    const sourceRoot = path.join(
```
aibaars, line 206: Using `resolve` has the advantage that it "does the right thing" if someone specifies an absolute path for the source root.

```suggestion
    const sourceRoot = path.resolve(
```

init/action.yml
aibaars, line 42:
```suggestion
    description: Path of the root source code directory, relative to $GITHUB_WORKSPACE.
```

aibaars (commented):
This looks good to me, thanks!

robertbrignull (commented):
I'm worried there are two usages changed within the runner but only one in the init-action. It looks like there's a use of `getRequiredEnvParam("GITHUB_WORKSPACE")` passed to `initConfig` in `init-action.ts` that also needs changing.

Although, I'm actually not sure that changing the calls to `initConfig` is the right thing to do. This will change how local queries are resolved and I'm not sure if that is intended or not.

In general it depends how this option will be used. Is it for when you want to analyze the whole repository but the repository is checked out to a non-standard location? If so then this change looks good to me (minus my comment above). If instead the intention is to be able to analyze only a subdirectory of a repository, then I think this may need deeper thought to know if this will work or not.

Finally, what manual testing has been done of this? Have you tried running it in a workflow.

robertbrignull:
To follow on from my comment above, looking at how it works in `runner.ts` currently, the option is used to control where the repository being analyzed is checked out to, but we haven't done it properly. It seems wrong to me that we currently use `cmd.checkoutPath` in `initConfig` but don't use it when initializing the codeql databases. So to a large degree the changes in this PR are just fixing a bug that was present already 👍

aibaars:
> I'm worried there are two usages changed within the runner but only one in the init-action. It looks like there's a use of `getRequiredEnvParam("GITHUB_WORKSPACE")` passed to `initConfig` in `init-action.ts` that also needs changing.
> 
> Although, I'm actually not sure that changing the calls to `initConfig` is the right thing to do. This will change how local queries are resolved and I'm not sure if that is intended or not.

I think we should not change `initConfig` here. The `source-root` property is the equivalent of the `--source-root` flag of the CodeQL CLI and its only purpose it to inform CodeQL what the path to the "source files" is so it can produce the correct relative paths in the SARIF output.

robertbrignull:
> I think we should not change `initConfig` here.

I've now come around to the other way of thinking. I think we should change `initConfig` (in the sense that we should pass the checkout path to it) because that's what we're already doing in the runnner. You can see we pass `cmd.checkoutPath` to `initConfig` and that's currently the only place we use it in the runner command.

> The `source-root` property is the equivalent of the `--source-root` flag of the CodeQL CLI and its only purpose it to inform CodeQL what the path to the "source files" is so it can produce the correct relative paths in the SARIF output.

What does "path to the source files" mean? Is that the root of the github repository, or a subdirectory within a repository? When talking about relative paths in the SARIF output we need those to be relative to the root of the repository, don't we? If not then showing files in the code scanning UI will not work.

aibaars:
> What does "path to the source files" mean? Is that the root of the github repository, or a subdirectory within a repository? When talking about relative paths in the SARIF output we need those to be relative to the root of the repository, don't we? If not then showing files in the code scanning UI will not work.

Yes that's right. It should be the path to the root of the github repository being analysed, or perhaps more precisely the repository to which the SARIF file will be uploaded. The use-case for this feature is for complex build setups that for example require a bunch of repositories (the repo to be analysed an its dependencies) to be checked out in a particular folder hierarchy that does not match the `actions/checkout`'s defaults .

aibaars:
> I've now come around to the other way of thinking. I think we should change `initConfig` (in the sense that we should pass the checkout path to it) because that's what we're already doing in the runnner. You can see we pass `cmd.checkoutPath` to `initConfig` and that's currently the only place we use it in the runner command.
> 
I think there is no such thing as "the checkout path".  A user can organise the workspace folder structure in any way they like either by configuring `actions/checkout` or even cloning repos "manually". I think a user could store the config file in another repo that is checked out somewhere in the workspace, or perhaps even have it in a fixed place on a self-hosted worker.

mario-campos:
>Although, I'm actually not sure that changing the calls to `initConfig` is the right thing to do. This will change how local queries are resolved and I'm not sure if that is intended or not.

Yes, I deliberately did not change `initConfig` because I believe that when someone specifies a local query in the config file, `./` has the connotation of the root of the repository. It would be very confusing to find that it changes based on the `source-root` specified elsewhere in the workflow file, in my opinion.

>Is it for when you want to analyze the whole repository but the repository is checked out to a non-standard location?

Yes, exactly, which, admittedly, is a minor use-case, but one that I have seen already and one which is supported in the CLI with the `--source-root` and `--working-dir` flags.

>If instead the intention is to be able to analyze only a subdirectory of a repository, then I think this may need deeper thought to know if this will work or not.

My intention was not to replace `paths`/`paths-ignore`. Although, it seems it could somewhat be used to that effect.

>Finally, what manual testing has been done of this? Have you tried running it in a workflow.

Yes, you can find the workflow [here](https://github.com/mario-campos/animated-octo-barnacle/blob/master/.github/workflows/codeql-analysis.yml) and the results [there](https://github.com/mario-campos/animated-octo-barnacle/actions).

init/action.yml
mario-campos: If we use `path.resolve()`, as you suggested above, perhaps it should just be
```suggestion
    description: Path of the root source code directory.
```

aibaars: I would still leave in the `relative to $GITHUB_WORKSPACE` bit to make it clear how relative paths are interpreted. With `resolve` the path is still interpreted relative to the workspace. The main difference is how absolute paths are treated. `join` would append the workspace and the absolute path, while `resolve` returns just the absolute path.

So if you specify:
```
  source-root: foo/bar
```
You end up with
* `join` : `$GITHUB_WORKSPACE/foo/bar`
* `resolve`: `$GITHUB_WORKSPACE/foo/bar`

and if you specify:
```
  source-root: ${{ github.workspace }}/foo/bar
```
You end up with:
* `join` : `$GITHUB_WORKSPACE/$GITHUB_WORKSPACE/foo/bar`
* `resolve`: `$GITHUB_WORKSPACE/foo/bar`

I think the behaviour of `resolve` is more likely "what the user intended".

robertbrignull:
@mario-campos, sorry for the confusing comments above. You can likely ignore my comments and go with what @aibaars says. We had quite a long chat on this and I think we're now on the same page with this, so you can ignore anything contradictory that I said earlier.

init/action.yml
mario-campos: Yes, it is. That was in my first commit, but when I actually started to use it in my test, I found it inconvenient to always have to specify a `${{ github.workspace }}` prefix. Hence, why I made it relative to the repo in subsequent commits.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- The `init` step of the Action now supports a `source-root` input to be able to specify a path under the GitHub workspace to be the root source-code directory. [#607](https://github.com/github/codeql-action/pull/607)
```
adityasharad, line 5: Could we mention that the default is the `$GITHUB_WORKSPACE` root?

adityasharad (approved):
Looks sensible. Would you be willing to add a test, here or in a follow up PR?

CHANGELOG.md
mario-campos: Done.

mario-campos:
Yes, I can add tests. Where would I do that, though? I don't see a test file for _init.ts_.

adityasharad:
Hmm good point, we don't seem to have unit tests. Perhaps another test job in `pr-checks.yml`, that exercises the `source-root` field in the `init` step. I think you can copy the `test-packaging-javascript-config` job as a template, and remove the `packs` inputs there. Those tests use a simple multilanguage test folder checked into the repo, so you could point explicitly to that path as the source root.

mario-campos:
Huh. Any idea why a source root would be considered invalid (aside from it not existing)?

```
A fatal error occurred: Invalid source root: /home/runner/work/codeql-action/codeql-action/tests/multi-language-repo.
Error: Failure invoking /opt/hostedtoolcache/CodeQL/0.0.0-20210622/x64/codeql/codeql with arguments database,init,/home/runner/work/_temp/codeql_databases/javascript,--language=javascript,--source-root=/home/runner/work/codeql-action/codeql-action/tests/multi-language-repo.

Exit code 2 and error was:

A fatal error occurred: Invalid source root: /home/runner/work/codeql-action/codeql-action/tests/multi-language-repo.
```

.github/workflows/pr-checks.yml
```diff
@@ -894,3 +894,44 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Move codeql-action
+      shell: bash
+      run: |
+        mkdir ../action
+        mv * .github ../action/
+        mv ../action/tests/multi-language-repo/{*,.github} .
```
adityasharad, line 909: I suspect this is the problem: it will move all the contents of `multi-language-repo` into the current directory, and not preserve the `tests/multi-language-repo` enclosing directories.

One option to try might be to avoid these moves entirely, and set `source-root` as you've done below. The moves are needed for other tests because they scan the entire repo from the workspace root, but the whole point of `source-root` is to customise that behaviour.

.github/workflows/pr-checks.yml
```diff
@@ -894,3 +894,44 @@ jobs:
...
+        mv ../action/.github/workflows .github
+    - uses: ./../action/init
+      with:
+        config-file: ".github/codeql/codeql-config-packaging.yml"
```
adityasharad, line 913: I think you may be able to remove this, and keep the test focused on only the source-root feature.

.github/workflows/pr-checks.yml
```diff
@@ -894,3 +894,44 @@ jobs:
     - name: Build code
       run: ./build.sh
     - uses: ./../action/analyze
+
+  test-packaging-javascript-source-root:
```
adityasharad, line 898: Minor, but packaging is a different feature so let's try and keep the tests distinct.
```suggestion
  test-javascript-source-root:
```

.github/workflows/pr-checks.yml
mario-campos: Good call. That makes sense.

.github/workflows/pr-checks.yml
```diff
@@ -894,3 +894,22 @@ jobs:
...
+  test-javascript-source-root:
+    needs: [check-js, check-node-modules]
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./init
+      with:
+        languages: javascript
+        source-root: tests/multi-language-repo
```
adityasharad, line 907:
```suggestion
    - uses: ./init
      with:
        languages: javascript
        source-root: tests/multi-language-repo
    - uses: ./analyze
      with:
        skip-queries: true
        upload: false
```

adityasharad (approved):
One test addition to actually build the database, then I think it's good to go!

IssacAlegre (approved):

aibaars (approved):

PR #612, opened 2021-07-02
Update databases URL

robertbrignull:
We're changing the URL used to upload databases to include "codeql" as well as "code scanning". This should make it more clear what the URL is for and unnecessarily box us in if we want to add other aspects to this API. The API side has been updated to allow both the old and new URLs, so now we can change things over here.


src/database-upload.ts
```diff
@@ -60,7 +63,7 @@ export async function uploadDatabases(
     const payload = fs.readFileSync(databaseBundlePath);
     try {
       await client.request(
-        `PUT /repos/:owner/:repo/code-scanning/databases/${language}`,
+        `PUT /repos/:owner/:repo/code-scanning/codeql/databases/${language}`,
```
aeisenberg, line 66: Good catch, CodeQL. It's probably safe since we generate the config file ourselves, but maybe we want to check that the `language` is valid before using it.

aeisenberg (commented):
Looks fine, except for the code scanning warning.

src/database-upload.ts
robertbrignull: I wonder why it didn't catch this before this change. Is this a new query?

We could template it into the URL by doing `:language` like we do with `owner` and `repo`. Not sure why I didn't do that originally. That might give us some protection or it might just be doing simple string concatenation like we are here.

We could also validate that it's a valid language string. Would it be better to do that here, or at the point where we load the config? I expect CodeQL won't recognise that sanitisation step, so what do we then do to tell it that the language is safe to use?

aeisenberg: Hmmm...probably best to sanitize (or verify) the string on load (are we already doing that?), and then we can just mark this warning as a false positive. In retrospect, I think this is probably what was happening before and this line was marked as a false positive.

robertbrignull: So a test has confirmed what I thought which is that you can put literally anything in that array and it'll "happily" decode it. The only protection is the typescript type and that does nothing at runtime.

You're right, we generate this file ourselves and there is type checking when the object is first constructed. We then write it to disk and read it back. It's possible someone could rewrite the file between our actions but there's not much we could do except schema validating the file when reading it.

However, it's also worth noting an attacker doesn't gain anything by modifying this file and tricking us into making a HTTP request. They could just make the request from the actions runner themselves. Our action doesn't have any special information or context that isn't available to any process running on that machine.

Therefore, I think this becomes a correctness concern rather than a security concern, and therefore I think the current situation is fine as it is.

robertbrignull: > Hmmm...probably best to sanitize (or verify) the string on load (are we already doing that?), and then we can just mark this warning as a false positive. In retrospect, I think this is probably what was happening before and this line was marked as a false positive.

Sorry, I didn't see your response before posting just then. To confirm, we're not doing any validation currently when loading the config from disk. It's a straight `JSON.parse` and cast it to the right typescript type, so it's pretty fast and loose with safety.

I'd argue if we want to add checking there we should probably do it fully rather than something custom for just the languages field. That would be a slightly larger task than I'd like to do in this PR, but certainly something that would be beneficial and I wouldn't mind doing if you don't have the time.

aeisenberg: I appreciate your explanation. This all makes sense. I agree that there is low/no chance of a security error since all an attacker could do is make API calls that would fail, and this is only a correctness concern.

I think it's fine to get this in as-is and we can add proper checking to the loading of the config file later.

aeisenberg (approved):

PR #618, opened 2021-07-13
Install Python tools on self-hosted runners

edoardopirovano:
This PR changes the `init` step to also install the Python packages that the Action requires when running on self-hosted runners, as we already do on GitHub ones. The rationale given in https://github.com/github/codeql-action/pull/155 for not doing this was that _"we don't want our cli to be messing around with things in the system"_. On the one hand, this seems reasonable. However, I think we should try to make a best attempt to succeed with our analysis and in my opinion this means installing tools we require.

Notably, we are already messing around with things that are installed since the later step will install an arbitrary number of packages that are dependencies of the code so that we can analyse it. If we are happy to do this, I see no reason why we shouldn't also be happy to install the four Python packages that _we_ need.

This PR was prompted by https://github.com/github/codeql-action/issues/615, where the user's workflow failed due to this (and one other issue).


src/init.ts
```diff
@@ -237,7 +221,10 @@ export async function installPythonDeps(codeql: CodeQL, logger: Logger) {
   } catch (e) {
     logger.endGroup();
     logger.warning(
-      "We were unable to install your python dependencies. You can call this action with 'setup-python-dependencies: false' to disable this process."
+      "We were unable to automatically install your python dependencies. Please make sure any necessary " +
```
aeisenberg, line 224: Do you also want to log the error here?

aeisenberg (commented):
One small suggestion. And looks like you need to fix some conflicts.

src/init.ts
edoardopirovano: Thanks, done.

aeisenberg (approved):

aliscco:
__

PR #630, opened 2021-07-26
Enable dependabot automatic updates

edoardopirovano:
Closes https://github.com/github/codeql-coreql-team/issues/1344

This adds a new workflow (triggered by a label) that updates the checked-in dependencies on a PR branch. It also enables Dependabot and has it apply the relevant label to have the checked-in dependencies updated. Unfortunately, the workflow will not work on forks due to missing the permissions to push to other user's forks, so we can't use it on our own PRs. It should work on Dependabot's PRs, though, since it creates its PRs from the main repo.

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -1,6 +1,7 @@
 name: Update dependencies
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
 
 jobs:
   update:
@@ -32,6 +33,9 @@ jobs:
         git config --global user.email "github-actions@github.com"
         git config --global user.name "github-actions[bot]"
         git add node_modules
-        git commit -am "Update checked-in dependencies"
-        git push
+        if ! git commit -am "Update checked-in dependencies" ; then
+          echo "No changes detected, skipping pushing..."
+          exit 0
+        fi
+        git push origin "$BRANCH"
 
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -1,0 +1,37 @@
+name: Update dependencies
+on:
+  pull_request:
+
+jobs:
+  update:
+    name: Update dependencies
+    runs-on: macos-latest
+    if: contains(github.event.pull_request.labels.*.name, 'Update dependencies')
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Remove PR label
+      env:
+        REPOSITORY: '${{ github.repository }}'
+        PR_NUMBER: '${{ github.event.pull_request.number }}'
+        GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+      run: |
+        gh api "repos/$REPOSITORY/issues/$PR_NUMBER/labels/Update%20dependencies" -X DELETE
+
+    - name: Push updated dependencies
+      env:
+        BRANCH: '${{ github.head_ref }}'
+      run: |
+        git fetch
+        git checkout $BRANCH
+        sudo npm install --force -g npm@latest
+        npm install
+        npm ci
+        npm run removeNPMAbsolutePaths
+        git config --global user.email "github-actions@github.com"
+        git config --global user.name "github-actions[bot]"
+        git add node_modules
+        git commit -am "Update checked-in dependencies"
+        git push
+
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -30,12 +30,10 @@ jobs:
         npm install
         npm ci
         npm run removeNPMAbsolutePaths
-        git config --global user.email "github-actions@github.com"
-        git config --global user.name "github-actions[bot]"
-        git add node_modules
-        if ! git commit -am "Update checked-in dependencies" ; then
-          echo "No changes detected, skipping pushing..."
-          exit 0
+        if [ ! -z "$(git status --porcelain)" ]; then
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions[bot]"
+          git add node_modules
+          git commit -am "Update checked-in dependencies"
+          git push origin "$BRANCH"
         fi
-        git push origin "$BRANCH"
-
--- a/.github/depandabot.yml
+++ b/.github/depandabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "thursday" # Gives us a working day to merge this before our typical release
+    labels:
+      - "Update dependencies"
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -27,7 +27,7 @@ jobs:
       run: .github/workflows/script/check-js.sh
 
   check-node-modules:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/script/check-node-modules.sh
+++ b/.github/workflows/script/check-node-modules.sh
@@ -7,6 +7,7 @@ if [ ! -z "$(git status --porcelain)" ]; then
     >&2 echo "Failed: Repo should be clean before testing!"
     exit 1
 fi
+sudo npm install --force -g npm@latest
 # Reinstall modules and then clean to remove absolute paths
 # Use 'npm ci' instead of 'npm install' as this is intended to be reproducible
 npm ci
--- a/node_modules/.package-lock.json
+++ b/node_modules/.package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "codeql",
-  "version": "1.0.6",
+  "version": "1.0.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
@@ -1243,7 +1243,6 @@
       "dependencies": {
         "anymatch": "~3.1.1",
         "braces": "~3.0.2",
-        "fsevents": "~2.1.2",
         "glob-parent": "~5.1.0",
         "is-binary-path": "~2.1.0",
         "is-glob": "~4.0.1",
@@ -3220,9 +3219,6 @@
     "node_modules/jsonfile": {
       "version": "4.0.0",
       "license": "MIT",
-      "dependencies": {
-        "graceful-fs": "^4.1.6"
-      },
       "optionalDependencies": {
         "graceful-fs": "^4.1.6"
       }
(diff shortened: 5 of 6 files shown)


.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  update:
+    name: Update dependencies
+    runs-on: macos-latest
+    if: contains(github.event.pull_request.labels.*.name, 'Update dependencies')
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Remove PR label
```
robertbrignull, line 14: I wonder if it would be better to remove the label at the end. Currently if it fails after this step and you retrigger it then it will abort because the label isn't there. I suppose you could re-add the label, and possibly that'll also trigger the workflow, so that's also fine if you want to leave it this way.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+        GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+      run: |
+        gh api "repos/$REPOSITORY/issues/$PR_NUMBER/labels/Update%20dependencies" -X DELETE
+
+    - name: Push updated dependencies
+      env:
+        BRANCH: '${{ github.head_ref }}'
+      run: |
+        git fetch
+        git checkout $BRANCH
```
robertbrignull, line 27: I'm surprised this works. I thought you'd have to add `fetch-depth: 0` to the `actions/checkout` call, otherwise it makes a shallow checkout.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+        sudo npm install --force -g npm@latest
+        npm install
+        npm ci
+        npm run removeNPMAbsolutePaths
+        git config --global user.email "github-actions@github.com"
+        git config --global user.name "github-actions[bot]"
+        git add node_modules
+        git commit -am "Update checked-in dependencies"
```
robertbrignull, line 35: I think this will fail if there are no changes to make. Is it worth adding a check for this so it exits cleanly?

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+        git push
```
robertbrignull, line 36: I think this should be fine as-is, but for safety it might be nice to be more explicit and do `git push origin "$BRANCH"` just so it's clear it's not going to go rogue and push to `main`.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,37 @@
+name: Update dependencies
+on:
+  pull_request:
```
robertbrignull, line 3: I think unfortunately if this triggers on a dependabot PR it'll have read-only permissions and will therefore be unable to push the changes back to the branch. If you change this to `pull_request_target` then it should always run with write permissions.

See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target for more info.

If you make this change then note that the events that trigger this by default also become a bit more restricted to just `assigned`, `opened`, `synchronize`, or `reopened`. You might want to make this explicit in the workflow too. Maybe
```
on:
  pull_request_target:
    types: [opened, synchronize, reopened, labeled]
```
Basically just adding `labeled` as I think that could be useful.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+      run: |
+        gh api "repos/$REPOSITORY/issues/$PR_NUMBER/labels/Update%20dependencies" -X DELETE
+
+    - name: Push updated dependencies
+      env:
+        BRANCH: '${{ github.head_ref }}'
+      run: |
+        git fetch
+        git checkout $BRANCH
+        sudo npm install --force -g npm@latest
```
robertbrignull, line 28: Just an open question of is using latest better than using what's on the actions VMs. They come with something pretty modern anyway, and it may be more stable than going with the latest. Although I assume this is still the latest released version and not true bleeding edge, so it's probably fine either way.

Whatever we do here, it should match the NPM version used in https://github.com/github/codeql-action/blob/main/.github/workflows/pr-checks.yml

robertbrignull (commented):
Looks nice. Thanks for working this out. I just have a few comments but nothing major at all.

.github/workflows/update-dependencies.yml
edoardopirovano: This was actually unfortunately necessary because the version in the Actions VM is old enough that it uses v1 of the lockfile and cannot do `npm install` correctly with the v2 lockfile that is checked in. This isn't necessary for the PR checks there because they do not actually do `npm install` and the other `npm` commands are forwards-compatible with the new lockfiles, but I agree for consistency we should probably use the same version there - I will update this. Indeed `@latest` is the latest stable release rather than the bleeding edge pre-release version which is `@next`.

.github/workflows/update-dependencies.yml
edoardopirovano: This works and was previously necessary when I had just `git push` because unlike the `actions/checkout` call which checks out a detached `HEAD` the above actually checks out the branch and lets us push to it. Now that we've specified the branch in `push` I think we could use `actions/checkout` but I think leaving this is also okay (and might avoid someone confusion in the future about why just `push` doesn't work if they ever change it back).

.github/workflows/update-dependencies.yml
edoardopirovano: Yep, good idea, done.

.github/workflows/update-dependencies.yml
edoardopirovano: That does seem safer, done.

.github/workflows/update-dependencies.yml
edoardopirovano: Ah good point. Many thanks, I will change it to that.

.github/workflows/update-dependencies.yml
edoardopirovano: So my thinking with removing the label at the start is that it makes sure we do not retrigger this workflow when we push the commit in the later step. I'm not actually sure the user pushing the commit has permission to run workflows, so I don't think this would happen, but just in case that changes in future this seemed safer.

edoardopirovano:
I'm a little surprised the workflow didn't run on this PR since I added the Update dependencies label. Any ideas why it wouldn't be working @robertbrignull? Does the `pull_request_target` trigger work differently from the `pull_request` one and require the workflow to be in `main`?

robertbrignull:
> Does the `pull_request_target` trigger work differently from the `pull_request` one and require the workflow to be in `main`?

Quite likely. The idea is that it runs the version of the workflow from the base branch to avoid the PR being able to change the workflow.

.github/workflows/update-dependencies.yml
robertbrignull: You're right, it theoretically shouldn't be possible to trigger workflows from actions performs within a workflow. But makes sense to add another guard against this. 👍

edoardopirovano:
> > Does the `pull_request_target` trigger work differently from the `pull_request` one and require the workflow to be in `main`?
> 
> Quite likely. The idea is that it runs the version of the workflow from the base branch to avoid the PR being able to change the workflow.

Makes sense. Guess we can't test the updated workflow then, but since it hasn't changed too much from my first commit I'm sure it will be okay.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,41 @@
+name: Update dependencies
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
+
+jobs:
+  update:
+    name: Update dependencies
+    runs-on: macos-latest
```
robertbrignull, line 9: I didn't notice this first time around. Why mac instead of linux?

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,41 @@
...
+        git fetch
+        git checkout $BRANCH
+        sudo npm install --force -g npm@latest
+        npm install
+        npm ci
+        npm run removeNPMAbsolutePaths
+        git config --global user.email "github-actions@github.com"
+        git config --global user.name "github-actions[bot]"
+        git add node_modules
+        if ! git commit -am "Update checked-in dependencies" ; then
```
robertbrignull, line 36: `git commit` could conceivably fail for other reasons. I suppose though that if this workflow didn't do its job then we would quickly notice because the PR would still be failing PR checks. So it would not be too dangerous and we'd notice it.

You could alternatively copy what we do in the PR checks script to see if the workflow is dirty:
https://github.com/github/codeql-action/blob/934fb86c580816d8f762fb2bf9c0904086054516/.github/workflows/script/check-node-modules.sh#L16

.github/workflows/update-dependencies.yml
edoardopirovano: There's a macOS-specific dependency that doesn't get installed if `npm install` is run on a non-Mac machine so unfortunately at the moment it is not possible to update the checked-in dependencies from a Linux machine. This is documented here: https://github.com/github/codeql-action/blob/main/CONTRIBUTING.md#checking-in-compiled-artifacts-and-node_modules

I also changed the PR check that the dependencies are up to date to run on Mac since I think it makes sense to run it in the same environment as this new workflow - although that was working fine because (unlike `npm install`), `npm ci` doesn't remove the Mac dependency if it happens to be there.

.github/workflows/update-dependencies.yml
edoardopirovano: That seems better than using the exit code of `git commit`, thanks.

robertbrignull:
> Guess we can't test the updated workflow then, but since it hasn't changed too much from my first commit I'm sure it will be okay.

Probably the best way would be to push another branch with a slightly altered workflow that does run on PRs, or test it in a totally separate respository. Both of those options can be a bit of effort though. There's not too much risk here as the worst that can happen is that the workflow just doesn't work.

.github/workflows/update-dependencies.yml
robertbrignull: Oh dear, I'd forgotten about that. That is a shame but I agree it's necessary right now. Would be nice if we weren't in this situation though. Anyway, nothing to do here in this PR 👍

edoardopirovano:
> > Guess we can't test the updated workflow then, but since it hasn't changed too much from my first commit I'm sure it will be okay.
> 
> Probably the best way would be to push another branch with a slightly altered workflow that does run on PRs, or test it in a totally separate respository. Both of those options can be a bit of effort though. There's not too much risk here as the worst that can happen is that the workflow just doesn't work.

Given the low risk, I suggest we just wait and see if it works on Thursday. I tested my original commit and not that much has changed since it. Plus, testing it as you suggest while useful still wouldn't test some of the specifics of what happens when Dependabot opens the PR rather than me.

robertbrignull (approved):
Thanks for responding quickly to all the comments. I think this LGTM now. As you say, there's pretty low risk for this so we can just try it out.

PR #635, opened 2021-07-27
Update lockfile in mergeback

edoardopirovano:
The first mergeback workflow after https://github.com/github/codeql-action/pull/630 failed (https://github.com/github/codeql-action/pull/634) because the lockfile did not have the new version ([this commit](https://github.com/github/codeql-action/pull/634/commits/fb190722374162e20bbea5ef72a467e66d531efe) fixed it after I manually added the label). Adding the label on the PR should make sure that commit gets made automatically.

### DIFF
--- a/.github/workflows/post-release-mergeback.yml
+++ b/.github/workflows/post-release-mergeback.yml
@@ -119,5 +119,6 @@ jobs:
             --head "$NEW_BRANCH" \
             --base "$BASE_BRANCH" \
             --title "$PR_TITLE" \
+            --label "Update dependencies" \
             --body "$PR_BODY" \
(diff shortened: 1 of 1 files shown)


aeisenberg:
This seems like a bit of magic. Do you know why adding the label ensures the commit gets made?

edoardopirovano:
> This seems like a bit of magic. Do you know why adding the label ensures the commit gets made?

The label was added in https://github.com/github/codeql-action/pull/630 and triggers a workflow that adds a commit updating the checked-in dependencies.

aeisenberg (approved):
Ah.. perfect. I missed that. Thanks.

PR #663, opened 2021-07-29
Update runner building dependencies

edoardopirovano:
The dependencies we use when building the runner can't be automatically updated by Dependabot since they are in a `package.json` in a sub-folder rather than in the main one. Nonetheless, we should still occasionally update them. In particular, there is currently an alert on one of the transitive dependencies being affected by this CVE (https://github.com/advisories/GHSA-ww39-953v-wcq6), which I believe will be resolved by this update.

### DIFF
--- a/runner/package-lock.json
+++ b/runner/package-lock.json
@@ -4,227 +4,265 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@babel/parser": {
-      "version": "7.11.4",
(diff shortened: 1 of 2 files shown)


adityasharad (approved):
Can we teach Dependabot to look in the subfolder?

edoardopirovano:
> Can we teach Dependabot to look in the subfolder?

I'll look into it after this PR!

PR #699, opened 2021-08-11
Always run the `latest` tools PR checks so we can make them required

henrymercer:
This change runs integration tests with all three bundles, even if `tools: latest` would be the same as `tools: null` (i.e. omitting the `tools` property). This allows us to make integration tests against each one of these three `tools` configurations required status checks on PRs.

As future work, we could move the `"$CODEQL_VERSION_DEFAULT" == "$CODEQL_VERSION_LATEST"` logic into jobs running with `tools: latest` to minimise redundant work, or investigate depending on statuses rather than Actions jobs.  However right now the additional cost created by these redundant jobs doesn't seem to justify the complexity.


edoardopirovano (approved):
LGTM

PR #700, opened 2021-08-11
Reduce the number of PR checks that are run on `push`

henrymercer:
If we're running on push, then we can reintroduce the logic that skips running with `tools: latest` when it would be the same as running with `tools: null`.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -351,10 +351,17 @@ jobs:
         echo "Latest CodeQL bundle version is $CODEQL_VERSION_LATEST"
         echo "Nightly CodeQL bundle version is $CODEQL_VERSION_NIGHTLY"
 
-        # Run integration tests with all three bundles, even if `tools: latest` would be the same as
-        # `tools: null`. This allows us to make all three kinds of integration tests required status
-        # checks on PRs.
-        VERSIONS_JSON="[null, \"$NIGHTLY_URL\", \"latest\"]"
+        # If we're running on a pull request, run each integration test with all three bundles, even
+        # if `tools: latest` would be the same as `tools: null`. This allows us to make the
+        # integration test job for each of the three bundles a required status check.
+        #
+        # If we're running on push, then we can skip running with `tools: latest` when it would be
+        # the same as running with `tools: null`.
+        if [[ "$GITHUB_EVENT_NAME" == "pull_request" && "$CODEQL_VERSION_DEFAULT" == "$CODEQL_VERSION_LATEST" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -351,10 +351,17 @@ jobs:
...
-        # `tools: null`. This allows us to make all three kinds of integration tests required status
-        # checks on PRs.
-        VERSIONS_JSON="[null, \"$NIGHTLY_URL\", \"latest\"]"
+        # If we're running on a pull request, run each integration test with all three bundles, even
+        # if `tools: latest` would be the same as `tools: null`. This allows us to make the
+        # integration test job for each of the three bundles a required status check.
+        #
+        # If we're running on push, then we can skip running with `tools: latest` when it would be
+        # the same as running with `tools: null`.
+        if [[ -z "${{ github.event.pull_request }}" && "$CODEQL_VERSION_DEFAULT" == "$CODEQL_VERSION_LATEST" ]]; then
```
adityasharad, line 360: Minor: `"$GITHUB_EVENT_NAME" == "pull_request"` might be neater. I'm not generally a fan of interpolating Actions contexts into Bash when we have equivalent env vars available.

.github/workflows/pr-checks.yml
```diff
@@ -351,10 +351,17 @@ jobs:
...
+          VERSIONS_JSON="[null, \"$NIGHTLY_URL\"]"
+        else
+          VERSIONS_JSON="[null, \"$NIGHTLY_URL\", \"latest\"]"
+        fi
 
         # Output a JSON-encoded list with the distinct versions to test against.
```
adityasharad, line 366: Something else that might be useful: add a second output with the list of CodeQL versions, so we don't have to recompute them.

edoardopirovano (approved):
LGTM

adityasharad (approved):
Couple of minor suggestions.

.github/workflows/pr-checks.yml
henrymercer: Nice, I didn't know there was a handy env var for that :)

.github/workflows/pr-checks.yml
henrymercer: Do we actually use this anywhere else?  I can't see any other uses in the workflow, but may be missing something.

adityasharad: Not yet, but I found myself wanting to check it in other jobs (e.g. "only run the Ruby tests if the version >= 2.5.9").

henrymercer: I have a weak preference for adding the output when we have a use case for it, so we can test that it works.  Will punt on this for this PR unless you feel strongly.

PR #702, opened 2021-08-12
Pass `--ram` flag to `database finalize`

edoardopirovano:
This PR passes a RAM option to `database finalize`, using the same logic to determine what value to use as we do for `database run-queries`. Currently, without a RAM option the JVM will determine it's own limit which is much more conservative than what we would do (i.e. use most of the memory available) leading to OOM errors during TRAP import for certain projects.


edoardopirovano:
@cklin I opted for only using this option with CLI versions >=2.5.9 as, if I recall correctly, the flag was introduced in 2.5.8 but had a bug at that point. Is this correct?

src/util.ts
```diff
@@ -491,3 +492,10 @@ export class HTTPError extends Error {
 export function isHTTPError(arg: any): arg is HTTPError {
   return arg?.status !== undefined && Number.isInteger(arg.status);
 }
+
+export async function codeQlVersionAbove(
+  codeql: CodeQL,
+  requiredVersion: string
+): Promise<boolean> {
+  return semver.compare(await codeql.getVersion(), requiredVersion) >= 0;
```
henrymercer, line 500:
```suggestion
  return semver.gte(await codeql.getVersion(), requiredVersion);
```

henrymercer (approved):
Looks good modulo the discussion on the correct CodeQL version to check for before passing this flag.

cklin:
> @cklin I opted for only using this option with CLI versions >=2.5.9 as, if I recall correctly, the flag was introduced in 2.5.8 but had a bug at that point. Is this correct?

The bug fix https://github.com/github/semmle-code/pull/39696/commits/14b72e0439866c662cfdc1a9176ffad23b10a048 went into [2.5.8](https://github.com/github/semmle-code/releases/tag/codeql-cli%2Fv2.5.8).  And of course 2.5.9 works too.

PR #707, opened 2021-08-18
Improve handling of CLI versions

edoardopirovano:
This PR bundles together a few commits that improve how the Action handles what CLI version is being used. In particular:

1. We can cache the result of `codeql version` to avoid spawning the process many times. This will be important since the later changes will result in us using the version much more frequently.
2. A few flags that are only available in recent CodeQL CLIs are moved to be behind a version check, so we can be compatible with older versions. @henrymercer I'd be very keen to have your eyes on this commit to let me know if you foresee any issues with not passing those flags (we don't need full functionality, but we shouldn't break either).
3. The Action now checks that the CodeQL version is above a constant of `CODEQL_MINIMUM_VERSION` that records the oldest version we support. This is currently set to 2.3.2 when some breaking changes to how the tracing environment is configured occurred.
4. The PR checks now also test against the oldest supported version in addition to the version in the toolcache, the latest release, and the latest nightly build.

I think the integration tests workflow is in need of some refactoring now that I have added this additional complexity to it, but I propose delaying this to a later PR.


src/codeql.ts
```diff
@@ -775,14 +800,17 @@ function getCodeQLForCmd(cmd: string): CodeQL {
...
         `--output=${sarifFile}`,
         addSnippetsFlag,
         ...getExtraOptionsFromEnv(["database", "interpret-results"]),
       ];
+      if (await util.codeQlVersionAbove(this, CODEQL_VERSION_DIAGNOSTICS))
+        codeqlArgs.push("--print-diagnostics-summary");
+      if (await util.codeQlVersionAbove(this, CODEQL_VERSION_METRICS))
+        codeqlArgs.push("--print-metrics-summary");
+      if (await util.codeQlVersionAbove(this, CODEQL_VERSION_GROUP_RULES))
+        codeqlArgs.push("--sarif-group-rules-by-pack");
```
henrymercer, line 813: I think we want to gate the `--sarif-category` flag on 2.5.3 too.

src/codeql.ts
```diff
@@ -542,13 +561,19 @@ export function getCachedCodeQL(): CodeQL {
   return cachedCodeQL;
 }
 
-function getCodeQLForCmd(cmd: string): CodeQL {
-  return {
+async function getCodeQLForCmd(
+  cmd: string,
+  checkVersion: boolean
+): Promise<CodeQL> {
+  let cachedVersion: undefined | Promise<string> = undefined;
```
henrymercer, line 568: Minor, optional suggestion: We could also set `cachedVersion` in the constructor so it's a field on the class.  That may be slightly clearer.

henrymercer (commented):
> A few flags that are only available in recent CodeQL CLIs are moved to be behind a version check, so we can be compatible with older versions. @henrymercer I'd be very keen to have your eyes on this commit to let me know if you foresee any issues with not passing those flags (we don't need full functionality, but we shouldn't break either).

For diagnostics and metrics summaries, these are only printed when there are results from diagnostics or metrics query, so we handle anyway the situation where they're not printed.

For `--sarif-group-rules-by-pack`, Code Scanning should be able to handle receiving both SARIF that does and SARIF that doesn't group rules by the query pack they're from.  However I'm unsure whether `--sarif-category` works without `--sarif-group-rules-by-pack` — ccing @cannist to confirm since I think you worked on the implementation of this.

src/codeql.ts
```diff
@@ -178,7 +178,25 @@ let cachedCodeQL: CodeQL | undefined = undefined;
...
+
+/**
+ * Versions of CodeQL that version-flag certain functionality in the Action.
+ * For convenience, please keep these in descending order. Once a version
+ * flag is older than the oldest supported version above, it may be removed.
+ */
 const CODEQL_VERSION_RAM_FINALIZE = "2.5.8";
+const CODEQL_VERSION_DIAGNOSTICS = "2.5.6";
+const CODEQL_VERSION_METRICS = "2.5.5";
+const CODEQL_VERSION_GROUP_RULES = "2.5.1";
```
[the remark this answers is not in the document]
edoardopirovano: Ah, I was not aware of that. Yes, seems reasonable to require `2.5.5` for this in that case.

src/codeql.ts
edoardopirovano: Indeed we do as it didn't exist prior to this - added.

src/codeql.ts
edoardopirovano: I'm not sure I follow this suggestion - as I understand it, the object we are returning is not an instance of a class, it is just an anonymous object that happens to implement the `CodeQL` interface. As such there is no constructor that we could set this in. I'm not super familiar with TypeScript, though, so I might be missing something.

henrymercer: Sorry, for some reason I thought we had a `class` expression here.

starcke:
> However I'm unsure whether `--sarif-category` works without `--sarif-group-rules-by-pack`

I would be very surprised if these had an effect on each other, so I think this change is fine.

PR #721, opened 2021-08-26
Update integration test pack versions

aeisenberg:

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -164,7 +164,7 @@ jobs:
       with:
         config-file: ".github/codeql/codeql-config-packaging2.yml"
         languages: javascript
-        packs: dsp-testing/codeql-pack1@0.0.5, dsp-testing/codeql-pack2
+        packs: dsp-testing/codeql-pack1@0.1.0, dsp-testing/codeql-pack2
 
     - name: Build code
       shell: bash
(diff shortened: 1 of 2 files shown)


edoardopirovano (approved):
As a very temporary fix to unbreak the tests this seems okay, but we definitely need a more permanent solution that doesn't require us to keep manually bumping versions.

aeisenberg:
> I forced the failing tests to use the nightly build. Now they are passing.

In retrospect, this is misleading. I wasn't implying that this is a _good_ solution, or even something I'd like to commit. I'm planning on working on this tomorrow and finding out what is happening, because upgrades _should_ work.

@edoardopirovano as you said, this would only be temporary. If you are being blocked by failing builds, then it is ok to merge. But, I would prefer to explore deeper when I am able to.

edoardopirovano:
Agree that we should investigate why the upgrades aren't working. I'm not blocked on failing tests, so okay to not merge this and instead wait until we have a proper solution.

aeisenberg:
The 0.1.0 versions fail because the queries are too new and upgrades are not found. The bug is that we haven't updated `codeql resolve upgrades` to work with packs yet. 

I'll work on this next.

aeisenberg:
These tests will not be properly fixed until the next release of the CLI. This is because database upgrades is not yet package-aware. I'll fix that shortly, but it won't be available until the release.

aeisenberg:
Building a new nightly version of the CLI. I expect that once that version is available to the action, then this PR will pass and we can get the action back on track.

aeisenberg:
This latest version hard-codes the tools bundle being used by the packaging integration tests. This will work for now. However, we need to make sure that the version that is pointed to is not cleaned up by automation.

aeisenberg:
I made sure that the release is not cleaned up by automation by making it a non-prerelease. Now, it will always show on top of the releases list, which always keeps the 10 latest.

PR #732, opened 2021-09-06
Refactor PR checks

edoardopirovano:
This PR does a refactoring of how our PR checks work to try and make it a bit easier to understand the different checks and to ensure we get good coverage of different versions in all our tests. In particular, each check is now specified in an individual YAML file in the `pr-checks/checks` folder. This file has a name and description of the check, as well as the steps that are unique to that check. If the check is not designed to run with all supported versions of the CLI and/or operating systems, this is specified by fields in this YAML file as well.

A script `pr-checks/sync.py` reads in all these test specification files and builds up a workflow using all of them, inserting some common steps at the start of each check and matrixing each one over the combination of CLIs and/or operating systems that they support. If the test specifications claims to support everything, we use the following:

- CLI: 2.3.1 (oldest supported release), 2.4.6, 2.5.9, toolcache version, latest release, latest nightly.
- OS: Ubuntu, MacOS, Windows

This is similar to what the VS code extension tests against (recent versions and latest patch version of old minor versions), however instead of testing against the newest version of the 2.3 series we test against specifically 2.3.1 since it is the oldest version that the Action claims support for. We also don't test against anything in the 2.2 series since the Action does not support this.

A new PR check verifies that the generated workflows are in sync with the YAML files specifying the various checks. This check together with some other checks that do not fit the generating model (in particular, unit tests and tests of the runner) are in a separate workflow. All integration tests of the Action have been migrated to being generated.

The required checks will not pass on this PR because they have changed name - it will require force merging and updating the branch protection rules afterwards.

### DIFF
--- a/.github/workflows/script/verify-pr-checks.sh
+++ b/.github/workflows/script/verify-pr-checks.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -eu
+
+# Sanity check that repo is clean to start with
+if [ ! -z "$(git status --porcelain)" ]; then
+    # If we get a fail here then this workflow needs attention...
+    >&2 echo "Failed: Repo should be clean before testing!"
+    exit 1
+fi
+
+# Wipe the generated PR checks in case there are extra unnecessary files in there
+rm -rf .github/workflows/__*
+
+# Generate the PR checks
+cd pr-checks && python3 sync.py
+
+# Check that repo is still clean
+if [ ! -z "$(git status --porcelain)" ]; then
+    # If we get a fail here then the PR needs attention
+    git diff
+    git status
+    >&2 echo "Failed: PR checks are not up to date. Run 'cd pr-checks && python3 sync.py' to update"
+    exit 1
+fi
+echo "Success: PR checks are up to date"
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -1,0 +1,104 @@
+import ruamel.yaml
+import os
+
+defaultTestVersions = [
+    # The oldest supported CodeQL version: 2.3.1. If bumping, update `CODEQL_MINIMUM_VERSION` in `codeql.ts`
+    "stable-20201028",
+    # The last CodeQL release in the 2.4 series: 2.4.6.
+    "stable-20210319",
+    # The last CodeQL release in the 2.5 series: 2.5.9.
+    "stable-20210809",
+    # The version of CodeQL currently in the toolcache. Typically either the latest release or the one before.
+    "cached",
+    # The latest release of CodeQL.
+    "latest",
+    # A nightly build directly from the our private repo, built in the last 24 hours.
+    "nightly-latest"
+]
+defaultOperatingSystems = ["ubuntu-latest", "macos-latest", "windows-latest"]
+header = """# Warning: This file is generated automatically, and should not be modified.
+# Instead, please modify the template in the pr-checks directory and run:
+#     pip install ruamel.yaml && python3 sync.py
+# to regenerate this file.
+
+"""
+
+
+class NonAliasingRTRepresenter(ruamel.yaml.representer.RoundTripRepresenter):
+    def ignore_aliases(self, data):
+        return True
+
+
+def writeHeader(checkStream):
+    checkStream.write(header)
+
+
+yaml = ruamel.yaml.YAML()
+yaml.Representer = NonAliasingRTRepresenter
+allJobs = {}
+for file in os.listdir('checks'):
+    with open(f"checks/{file}", 'r') as checkStream:
+        checkSpecification = yaml.load(checkStream)
+
+    versions = defaultTestVersions
+    if 'versions' in checkSpecification:
+        versions = checkSpecification['versions']
+    operatingSystems = defaultOperatingSystems
+    if 'os' in checkSpecification:
+        operatingSystems = checkSpecification['os']
+
+    steps = [
+        {
+            'name': 'Check out repository',
+            'uses': 'actions/checkout@v2'
+        },
+        {
+            'name': 'Prepare test',
+            'id': 'prepare-test',
+            'uses': './.github/prepare-test',
+            'with': {
+                'version': '${{ matrix.version }}'
+            }
+        }
+    ]
+    steps.extend(checkSpecification['steps'])
+
+    checkJob = {
+        'strategy': {
+            'matrix': {
+                'version': versions,
+                'os': operatingSystems
+            }
+        },
+        'name': checkSpecification['name'],
+        'runs-on': '${{ matrix.os }}',
+        'steps': steps
+    }
+
+    for key in ["env", "container", "services"]:
+        if key in checkSpecification:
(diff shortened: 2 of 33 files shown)


adityasharad:
Thanks for taking this on! A few initial thoughts below, and I'd like to spend some time looking over it together to see if we can make things simpler.

- I like the idea of factoring out composite step actions, and think we could go even further to use them to share code between the various check jobs.
- I think there is room to split out a few more workflows rather than have everything in a single PR checks workflow. The Runner tests for example could live on their own. Possibly the integration tests too. This makes life much easier if you have to re-run a workflow due to transient errors.
- I'm not yet convinced by the added complexity of generating the PR check workflow from specifications -- my preference would be to use first-class Actions features (like composite steps) wherever we can, but not have to create our own specification layers on top of it, since the total number of checks we're dealing with here is not that large.

.github/workflows/script/verify-pr-checks.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Generate the PR checks
+cd pr-checks && python3 sync.py
+
+# Check that repo is still clean
+if [ ! -z "$(git status --porcelain)" ]; then
+    # If we get a fail here then the PR needs attention
+    git diff
+    git status
+    >&2 echo "Failed: PR checks are not up to date. Run 'cd pr-checks && python3 sync.py' to update"
```
aeisenberg, line 22: I'm a little confused here. Wasn't `cd pr-checks && python3 sync.py` run above? Should the message be something like this:

```suggestion
    >&2 echo "Failed: PR checks are not up to date. Commit the changes and run again"
```

aeisenberg: Oh...I see this is going to be run on CI, but the command to update needs to be run locally.

.github/workflows/__generated-pr-checks.yml
```diff
@@ -0,0 +1,254 @@
...
+    - v1
+  pull_request:
+    types:
+    - opened
+    - synchronize
+    - reopened
+    - ready_for_review
+    branches:
+    - main
+    - v1
```
aeisenberg, line 23: I see this workflow is run on PRs to v1 branch, but the original workflow is not. Should they be the same?

pr-checks/initial-steps.yml
```diff
@@ -0,0 +1,25 @@
...
+      export VERSION=`echo ${{ inputs.version }} | sed -e 's/^.*\-//'`
+      echo "::set-output name=tools-url::https://github.com/dsp-testing/codeql-cli-nightlies/releases/download/codeql-bundle-$VERSION-manual/codeql-bundle.tar.gz"
+    elif [[ ${{ inputs.version }} == *"stable"* ]]; then
+      export VERSION=`echo ${{ inputs.version }} | sed -e 's/^.*\-//'`
+      echo "::set-output name=tools-url::https://github.com/github/codeql-action/releases/download/codeql-bundle-$VERSION/codeql-bundle.tar.gz"
+    elif [[ ${{ inputs.version }} == "latest" ]]; then
+      echo "::set-output name=tools-url::latest"
+    elif [[ ${{ inputs.version }} == "cached" ]]; then
+      echo "::set-output name=tools-url::"
+    fi
```
aeisenberg, line 25: Maybe add a file `else` clause that just fails to ensure we don't accidentally add something here we don't mean to.

pr-checks/initial-steps.yml
```diff
@@ -0,0 +1,25 @@
+- name: Move codeql-action
```
aeisenberg, line 1: As `@adityasharad` suggests, I think this can be turned into a nested, composite action. This will make the generator simpler or even make it unnecessary.

aeisenberg (commented):
I would be leaning towards @adityasharad's suggestion that we try to avoid using a generator if possible. They are complex to maintain and it might be that we can create individual workflows for each of the generated actions.

That being said, I see 2 advantages with the generator:

1. A uniform way to declare matrix builds
2. Flexibility to add more complex, shared steps in the future (though, since the action is already fairly mature, I think there is a low likelihood of this needing to happen).

pr-checks/initial-steps.yml
edoardopirovano: Good point, done.

.github/workflows/__generated-pr-checks.yml
edoardopirovano: Aha, yes, I think we can just omit this bit and run on all PRs like the original workflow does.

edoardopirovano:
Thanks for the comments both, I've refactored the shared code into a step that the different checks can use.

I've also changed to having a workflow for each test rather than a workflow containing all of them so that things can be rerun individually in the case of transient errors.

I think there is some benefit to using a generator (which is now a bit simpler after the above changes). In particular:
1. It avoids repeating some boilerplate at the start of each check in which we'd have to specify the branches/PRs it runs on and some environment variables like `GITHUB_TOKEN` that we need in all the tests.
2. It provides a single place to update which versions of the CLI we are testing against by default (whilst still allowing some tests to override this).
3. Having a generator makes it easier to change how we run the tests. For instance, the change above to having a workflow per test instead of a workflow with all of them was fairly trivial to implement with just a few tweaks to the generator.

aeisenberg (approved):
Nice. 

Next step is to change the required checks so that they match with the new names.

And all currently open PRs will need to be rebased.

pr-checks/sync.py
```diff
@@ -0,0 +1,104 @@
...
+def writeHeader(checkStream):
+    checkStream.write(header)
+
+
+yaml = ruamel.yaml.YAML()
+yaml.Representer = NonAliasingRTRepresenter
+allJobs = {}
+for file in os.listdir('checks'):
+    with open(f"checks/{file}", 'r') as checkStream:
+        checkSpecification = yaml.load(checkStream)
```
adityasharad, line 41: Curious: does `ruamel` give us more functionality than the default `pyyaml` package, and are its `load/dump` functions equivalent to `yaml.safe_load/dump`?

edoardopirovano: I went for `ruamel` rather than `pyyaml` because `pyyaml` outputted multi-line strings (which we have a lot of since we use them for workflow steps that have more than one line) in a terrible way. In particular, instead of using `|` and then using line breaks, it would make a one-line string by replacing the line breaks with `\n`, i.e. instead of
```
steps: |
  line1
  line2
```
it would do:
```
steps: "line1\nline2"
```
This led to some very difficult to read workflows. I tried for a bit to get the right behaviour out of `pyyaml` but eventually gave up and decided to use `ruamel` instead.

---

Aha, a very good point. `ruamel`'s `load` is actually equivalent to `load` rather than `safe_load`. We need to add a `typ='safe'` to get it to behave like `safe_load`.

edoardopirovano: Actually, scrap that last point, the default `rt` ("round-trip") loader is forked from the safe loader so is also safe (see [here](https://stackoverflow.com/a/51318354)), and is needed to preserve line-breaks like we want.

PR #738, opened 2021-09-13
Add RAM and threads options to init action

cklin:


.github/workflows/pr-checks.yml
```diff
@@ -388,3 +388,42 @@ jobs:
...
+
+      - name: Build runner
+        run: |
+          cd runner
+          npm install
+          npm run build-runner
+
+      - name: Run init
+        run: |
+          runner/dist/codeql-runner-linux init --ram=230 --threads=1 --repository $GITHUB_REPOSITORY --languages java --github-url $GITHUB_SERVER_URL --github-auth ${{ github.token }}
```
edoardopirovano, line 408: I think I'd slightly prefer this to use something other than `1` for the threads as `1` could very easily be a default in which case we wouldn't really be checking anything. Maybe `3` as that seems very unlikely to ever be the default value.

src/init-action.ts
```diff
@@ -201,9 +203,18 @@ async function run() {
       );
     }
 
-    // Setup CODEQL_RAM flag (todo improve this https://github.com/github/dsp-code-scanning/issues/935)
-    const codeqlRam = process.env["CODEQL_RAM"] || "6500";
-    core.exportVariable("CODEQL_RAM", codeqlRam);
+    // Limit RAM and threads for extractors. When running extractors, the CodeQL CLI obeys the
+    // CODEQL_RAM and CODEQL_THREADS environment variables to decide how much RAM and how many
+    // threads it would ask extractors to use. See "codeql database trace-command" help text for
+    // the "--ram" and "--threads" options for details.
```
edoardopirovano, line 209: Perhaps we could link to the CLI docs here for convenience?

src/runner.ts
```diff
@@ -193,6 +205,16 @@ program
       const gitHubVersion = await getGitHubVersion(apiDetails);
       checkGitHubVersionInRange(gitHubVersion, logger, Mode.runner);
 
+      // Limit RAM and threads for extractors. When running extractors, the CodeQL CLI obeys the
+      // CODEQL_RAM and CODEQL_THREADS environment variables to decide how much RAM and how many
+      // threads it would ask extractors to use. See "codeql database trace-command" help text for
+      // the "--ram" and "--threads" options for details.
```
edoardopirovano, line 211: As above, might be worth linking to the CLI docs.

pr-checks/checks/extractor-ram-threads.yml
```diff
@@ -0,0 +1,29 @@
+name: "Extractor ram and threads options test"
+description: "Tests passing RAM and threads limits to extractors"
+versions: ["latest"]
+os: ["ubuntu-latest"]
+steps:
+  - uses: ./../action/init
+    with:
+      languages: java
+      ram: 230
+      threads: 1
```
edoardopirovano, line 10: As above, maybe try a different value here.

src/util.ts
```diff
@@ -123,17 +134,17 @@ export function getAddSnippetsFlag(
 }
 
 /**
- * Get the codeql `--threads` value specified for the `threads` input.
- * If no value was specified, all available threads will be used.
+ * Get the value of the codeql `--threads` flag specified for the `threads`
+ * input. If no value was specified, all available threads will be used.
  *
  * The value will be capped to the number of available CPUs.
```
edoardopirovano, line 140: I don't know how I feel about capping this. I note that the RAM is not capped and the user can request more RAM than is available on the machine thus causing an OOM. On the other hand, the threads _are_ capped even though using more threads than there are cores wouldn't cause us to crash (just give suboptimal performance). Additionally, there are valid reasons to have more threads than cores - for instance in CPUs with hyperthreading there is a benefit to having 2x the number of threads as cores.

If we do indeed want to cap the number of threads, it should definitely be documented in the user-facing description of the `threads` input rather than in a comment on the code.

edoardopirovano (commented):
Mostly looks good, some comments below.

.github/workflows/pr-checks.yml
cklin: This is a decision that I struggled with.

The main challenge here is that the number of threads used defaults to (and is capped at) the number of CPU cores available on the machine.  GitHub action runners have 2 CPUs, so both the default and the cap would be 2.  Passing any value larger than 1 means that the test will break if it runs on a system with different number of CPUs.

I would love to hear your thoughts.  For example, maybe we could assume that the runner will always have at least 2 CPUs and have two tests with threads set to 1 and 2?

cklin:
While we are here, there is a code scanning warning that we have the same inputs (`ram` and `threads`) in different actions (`init` and `analyze`) with different help texts.  In this case, they are indeed different:

- The `init` inputs limit resource usage during extraction, and
- The `analyze` inputs limit resource usage during query execution

Do you think there is more we can do to simplify or to clarify these different inputs?

src/runner.ts
cklin: Great suggestion!  Thanks!  Done.

src/init-action.ts
cklin: Great suggestion!  Thanks!  Done.

.github/workflows/pr-checks.yml
edoardopirovano: Aha, this is a good point. As I mentioned below, I'm actually not sure that capping this is a good idea, but for now let's leave this test as is 🙂

src/util.ts
edoardopirovano: I agree that we should invite a wider discussion on this outside the PR as it is technically a breaking behaviour change - some users might be passing 99 in and relying on the capping which would cause them to experience massive performance issues if we remove it.

Given this wasn't introduced by this PR, I don't think we need to block on it.

edoardopirovano:
Regarding your final point, I think I agree with Code Scanning that having the same option name of different steps and serving a different purpose may invite user confusion. It's too late to change the options on `analyze` as these are shipped but maybe these new options could be `extractor-{ram,threads}` instead?

adityasharad:
Which `threads` argument gets passed to `database finalize`?
I'm not sure users will know (or need to know) the context about these values going to extractors, so another option is to weaken the custom query and continue to explain carefully in the options how they are different for the two commands.

edoardopirovano:
`finalize` shares the arguments in the `analyze` step. A further option - make `init` the place where these should be specified and then use the ones from `init` _everywhere_ when they are available (storing them in the config for the `analyze` step to use later).

Obviously for backwards compatibility we still keep the inputs on `analyze` in the code, but we amend our documentation to remove all mention of them.

adityasharad:
That sounds good. We could respect the analyze arguments if they are provided separately (might be handy for debugging), but if they are absent just respect the values passed to init.

cklin:
> A further option - make `init` the place where these should be specified and then use the ones from `init` _everywhere_ when they are available (storing them in the config for the `analyze` step to use later).

Sounds like a plan.  I will revise this PR to do so.

> Obviously for backwards compatibility we still keep the inputs on `analyze` in the code, but we amend our documentation to remove all mention of them.

Can you point me to the documentation that need to be amended, just so that I do not miss it?

edoardopirovano:
> Can you point me to the documentation that need to be amended, just so that I do not miss it?

Huh, I'm actually not sure. I expected the README in this repository to mention it, but it does not.

In the `github/docs-internal` repository, I expected the page corresponding to https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning to mention this but actually it seems that it _also_ does not.

@adityasharad Any ideas where we might be mentioning this?

adityasharad:
That is the right place. I think we have never actually documented `ram` and `threads` outside this repo, though that is more by omission than by design.

PR #753, opened 2021-09-28
Use CLI's own baseline LOC counting

edoardopirovano:
Starting with CLI version 2.6.2, the CodeQL CLI does its own LOC counting which should perform better than the one the Action is using. I think this warrants having both code paths until we no longer need to support CLI versions before 2.6.2.

@aeisenberg I know you recently made some tweaks to the Action's LOC counting for Python. Before merging this PR, I'd be keen to have a chat so we can figure out if the same tweaks need to made to the CLI's counting, and if so carry them out.


src/analyze.ts
```diff
@@ -295,12 +299,17 @@ export async function runQueries(
...
       statusReport[`interpret_results_${language}_duration_ms`] =
         new Date().getTime() - startTimeInterpretResults;
       logger.endGroup();
       logger.info(analysisSummary);
-      printLinesOfCodeSummary(logger, language, await locPromise);
+      if (locPromise !== undefined) {
+        printLinesOfCodeSummary(logger, language, await locPromise);
+      } else {
+        logger.info(await runPrintLinesOfCode(language));
+      }
```
aeisenberg, line 312: Can we have a debug mode where we do the count using both ways and print both? Perhaps we can even print a warning or error if they differ by some delta. We could run this for our internal and integration tests. We would just do this until we have confidence that both counting mechanisms are behaving as we hope.

src/analyze.ts
```diff
@@ -333,6 +342,19 @@ export async function runQueries(
     );
   }
 
+  async function shouldInjectLinesOfCode() {
```
aeisenberg, line 345: I was initially confused by this name. I wasn't sure whether this was for the old or new style of line counting. 

And the more I think about it, I wonder if it would make sense to reverse the name here and assign the result to a local variable. So, maybe call it `cliCanCountLoC` or something like that.

Then, later you don't need to check if `locPromise` is undefined. Instead you check if the result of this function is true or false. I think this would make the behaviour more explicit and easier to read.

src/codeql.ts
```diff
@@ -860,6 +869,15 @@ async function getCodeQLForCmd(
       // capture stdout, which contains analysis summaries
       return await runTool(cmd, codeqlArgs);
     },
+    async databasePrintBaseline(databasePath: string): Promise<string> {
```
aeisenberg, line 872: So, to be clear, this command will not re-calculate the LoC? It will just print out metrics that were calculated earlier?

src/analyze.ts
edoardopirovano: I agree the name is confusing - `cliCanCountLoC` sounds much better. Unfortunately I think we do have to check if `locPromise` is `undefined` since otherwise using it in `injectLinesOfCode` doesn't type check 🙁

src/codeql.ts
edoardopirovano: Correct, this just loads the database metadata and prints the LoC count from it, it runs in `O(1)` time.

src/analyze.ts
aeisenberg: Hmmm....you _could_ assign `locPromise = Promise.resolve({})` instead of leaving it as undefined and then I think it would type check. Though, that's less nice as well.

src/analyze.ts
edoardopirovano: Interesting, that sounds reasonable, yes. How do you propose we enable this debug mode? Would it be an optional input parameter to the workflow or something else? It would be neat to be able to enable this mode when someone is running the workflow from `@main` but not when they are running it from `@v1` but I don't know if we have a way of distinguishing these two cases within the code.

An even better way to collect a lot of data about this would be to add this information into the status reports being sent to Turboscan, but I don't know if that's overkill - it would involve a gh/gh PR to add some new fields to the status report tables.

aeisenberg: Not sure if we need to send this data to gh/gh. That would be a bit of an overkill since we're really just trying to calibrate our line counting.

Does a `uses` block respect the environment variables set in the `env` block? If so, you could just check for a custom, hidden, internal environment variable to turn on this debug behaviour.

edoardopirovano: I believe it does, yes. Agree an environment variable is probably the easiest way forward - we can reserve my very overkill suggestion above for if we happen to encounter some issues that we think warrant a wider-scale collection of data.

src/analyze.ts
edoardopirovano: Hmm, also not ideal but I do think I like that _slightly_ more. Will change accordingly.

edoardopirovano:
Thanks for the review @aeisenberg! I believe I have addressed your comments in the new commit 🙂

pr-checks/sync.py
```diff
@@ -79,6 +79,8 @@ def writeHeader(checkStream):
         if key in checkSpecification:
             checkJob[key] = checkSpecification[key]
 
+    checkJob['env'] = checkJob.get('env', {})
+    checkJob['env']['DEBUG_MODE'] = True
```
aeisenberg, line 83: Either we should be using an env var that is specific to this action, or we should use a generic var that enables debug mode for the entire action, like `ACTIONS_RUNNER_DEBUG`. See https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging

I'd prefer the former. Maybe something like `INTERNAL_CODEQL_ACTION_DEBUG_LOC`.

adityasharad: We have `CODEQL_ENABLE_EXPERIMENTAL_FEATURES` (which is externally visible, because it's how you enable Ruby). Reuse that if you want.

I think we should also respect `ACTIONS_RUNNER_DEBUG` and `ACTIONS_STEP_DEBUG` if they're set. Those are quite useful.

edoardopirovano: Let's add a new internal variable `INTERNAL_CODEQL_ACTION_DEBUG_LOC`, but also do the debugging output if either `ACTIONS_RUNNER_DEBUG` or `ACTIONS_STEP_DEBUG` is set.

aeisenberg (approved):
Nice.

In a future commit, can you set `INTERNAL_CODEQL_ACTION_DEBUG_LOC` in the PR checks and we can see what gets printed? We should also do this for semmle-code.

PR #779, opened 2021-10-14
Do not artificially restrict the set of supported languages

esbena:
### Description

The hardcoded set of supported languages in `languages.ts` makes it hard to experiment with new languages as custom branch with a modified `languages.ts` is required for every new language. 

This PR makes it easier to experiment with new languages by making the `codeql-action` indifferent to the language choice, except for some *additional* support for known languages. 

### Implementation

The old implementation throws an early error if the user-provided language is not in the hardcoded set. As an alternative, a `Language` is now just a lowercase string, and an early error is only produced if that language is not in the output of `codeql resolve languages`.  The old `Language` enum is now named `KnownLanguage`, and is used in the places where *additional* support for a specific language is desired. 

This should be fully backwards compatible, as the implementation now should throw an early exception in fewer situations. This may be detrimental to how early and well users are informed of input mistakes. On the other hand, it is consistent with how the provided `codeql` and environment behave.

PMs may want to keep the tight coupling between the codeql-action and the GitHub-controlled CodeQL repositories, I do not.

### Testing

I will use this branch for some internal experiments for a short while. Hopefully, the many changes wont result in too many conflicts when we are ready to merge.

### DIFF
--- a/src/config-utils.test.ts
+++ b/src/config-utils.test.ts
@@ -974,7 +974,7 @@ test("No detected languages", async (t) => {
   });
 });
 
-test("Unknown languages", async (t) => {
+test("Unknown, unsupported languages", async (t) => {
   return await util.withTmpDir(async (tmpDir) => {
     const languages = "rubbish,english";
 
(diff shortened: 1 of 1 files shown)


walid97t (changes requested):
go ahead

tied2token (approved):

iamannalynleynes1206 (approved):

henrymercer:
Done in https://github.com/github/codeql-action/pull/2914

PR #781, opened 2021-10-18
Add an option to allow waiting until an analysis has been processed before finishing the Action.

chrisgavin:
This adds an option to allow waiting until the analysis has succeeded before continuing. This will be used to allow surfacing errors that have occurred while processing the uploaded SARIF file.

When this option is enabled, after uploading the SARIF file the Action will call the status endpoint every 5 seconds until the processing is complete. After 2 minutes, it will give up and continue anyway. I felt that there should be a timeout in case there is some issue with processing and that it shouldn't fail the Action since it's possible the analysis will eventually be processed.

I've not added the option to enable this behavior to the readme, as our hope is to test this out a bit and eventually make it the default at which point we'll remove the option.


src/upload-lib.ts
```diff
@@ -323,6 +327,9 @@ export function buildPayload(
   }
 }
 
+const STATUS_CHECK_FREQUENCY_MILLISECONDS = 5 * 1000;
+const STATUS_CHECK_TIMEOUT_MILLISECONDS = 2 * 60 * 1000;
```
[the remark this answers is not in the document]
chrisgavin: I wondered, but I'm not sure there's much benefit. We can probably choose a default value better than a user can because we know how long the processing is likely to take in the worst case.

analyze/action.yml
```diff
@@ -42,6 +42,9 @@ inputs:
     description: Whether to upload the resulting CodeQL database
     required: false
     default: "true"
+  wait-for-processing:
```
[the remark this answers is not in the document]
chrisgavin: Yep, that's effectively what the default is, but I've got no issue with making it explicit.

src/upload-lib.ts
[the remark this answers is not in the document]
cannist: Ok, let's leave it out initially. We can always add a configuration option later.

analyze/action.yml
```diff
@@ -52,6 +52,10 @@ inputs:
...
+    description: If true, the Action will wait for the uploaded SARIF to be processed before completing.
+    required: true
```
cannist, line 57: Should this be false since we're providing a default value?

upload-sarif/action.yml
```diff
@@ -20,6 +20,10 @@ inputs:
   category:
     description: String used by Code Scanning for matching the analyses
     required: false
+  wait-for-processing:
+    description: If true, the Action will wait for the uploaded SARIF to be processed before completing.
+    required: true
```
cannist, line 25: Should this be false since we provide a default?

src/upload-lib.ts
```diff
@@ -407,66 +406,78 @@ async function uploadFiles(
...
+    ) {
+      // If the analysis hasn't finished processing in the allotted time, we continue anyway rather than failing.
+      // It's possible the analysis will eventually finish processing, but it's not worth spending more Actions time waiting.
+      logger.warning(
+        "Timed out waiting for analysis to finish processing. Continuing."
+      );
+      break;
+    }
+    // We put the delay at the start of the loop, since it's unlikely that the analysis will have succeeded immediately. We might as well wait preemptively.
+    await util.delay(STATUS_CHECK_FREQUENCY_MILLISECONDS);
```
cannist, line 447: Do you think we should move this to the end of the loop, now that we're potentially doing the database upload before we start waiting?

src/upload-lib.ts
```diff
@@ -390,17 +397,89 @@ async function uploadFiles(
...
```
cannist, line 447: Should this go to the end of the loop, now that we potentially spent some time uploading the database?

cannist (commented):
LGTM, just some minor details.

src/upload-lib.ts
cannist: [Argh now that comment is here twice. GitHub's UI had made me think it had lost the first comment / I had not submitted it when I was looking at the single commit.]

analyze/action.yml
chrisgavin: I believe it being required means it cannot be null, which it won't be if we're providing a default.

upload-sarif/action.yml
chrisgavin: I believe it being required means it cannot be null, which it won't be if we're providing a default.

src/upload-lib.ts
chrisgavin: That probably does make sense, at least in the long run. I don't know how often we're uploading databases now, but presumably at some point we're hoping to upload them almost all the time.

aeisenberg (approved):
Looks reasonable to me, but maybe @cannist should have a look, too.

analyze/action.yml
cannist: Hmm, I always understood `required` here as "must be specified by the user". If it does not matter if we use `true` or `false` in the presence of a default, I think `false` still is less confusing to the reader.
That would also be in line with the example given in the [docs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs).

chrisgavin: There are also [other docs](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file) that show the opposite pattern; a required argument with a default value.

I tried in a test repository, and it seems like Actions doesn't really care about whether the value is "required" or not. Possibly it's a planned future feature, but without knowing how it will interpret the property it's hard to say if one way is any better than the other.

cannist: OK, it is your PR. 🙂 
I still prefer the other value but if actions do not fall over if the parameter is left unspecified by the user I do not mind too much.

cannist (approved):

PR #798, opened 2021-10-28
Add an option to upload some debugging artifacts

edoardopirovano:
This PR adds a new option `debug-artifacts` to the `analyze` step which will result in the following files being uploaded as artifacts in order to aid debugging:

1. Logs, including the tracer log for pre-multilanguage tracing. (After multi language tracing this is in the database logs, so does not need to be special cased).
2. The bundle(s) for the CodeQL database(s) produced.
3. The SARIF file(s) produced.


src/analyze-action.ts
```diff
@@ -106,6 +108,42 @@ async function run() {
         config,
         logger
       );
+
+      if (config.debugMode) {
```
aeisenberg, line 112: Is there ever a danger of running with `debugMode === true` on the runner? Presumably, if so, the runner would fail on the upload stage, but it would be nice if we could detect this and fail faster.

src/analyze-action.ts
```diff
@@ -1,6 +1,7 @@
 import * as fs from "fs";
 import * as path from "path";
 
+import * as artifact from "@actions/artifact";
```
aeisenberg, line 4: I think the `unguarded-action-lib.ql` query should be updated to include calls to `@actions/artifact`.

src/analyze-action.ts
aeisenberg: Note, this will be caught (if it is indeed a problem) by updating the `unguarded-action-lib.ql` query.

src/analyze-action.ts
```diff
@@ -188,6 +234,27 @@ async function run() {
   }
 }
 
+async function uploadDebugArtifact(toUpload: string[], rootDir: string) {
```
aeisenberg, line 237: nit: This uploads multiple artifacts, so:

```suggestion
async function uploadDebugArtifacts(toUpload: string[], rootDir: string) {
```

src/analyze-action.ts
```diff
@@ -188,6 +234,27 @@ async function run() {
...
+  await artifact.create().uploadArtifact(
+    DEBUG_ARTIFACT_NAME,
+    toUpload.map((file) => path.normalize(file)),
```
aeisenberg, line 240: Hmmm...artifacting lots of little files (eg- a database) can take a really long time. It looks like you're doing this in several places. Much better to zip things up first and then upload.

aeisenberg: Take a look and see if this really is a problem in the action, but I've encountered it before in semmle-code.

aeisenberg: Actually, you may already be zipping things, so maybe this is a noop.

aeisenberg:
Updating the query can happen in a different PR.

src/analyze-action.ts
edoardopirovano: Unless I'm misunderstanding that query, it will already be including these calls as they match
https://github.com/github/codeql-action/blob/4293754ed2f1f6fcf2f353cb6b59927f7b948562/queries/unguarded-action-lib.ql#L44
and are not specifically included in the whitelist
https://github.com/github/codeql-action/blob/4293754ed2f1f6fcf2f353cb6b59927f7b948562/queries/unguarded-action-lib.ql#L17-L20

src/analyze-action.ts
edoardopirovano: It should never be possible to be in debug mode when using the runner. In particular, note that the call to `initConfig` in `runner.ts` has gained a hard-coded value of `false` for the parameter specifying whether we are in debug mode.

src/analyze-action.ts
edoardopirovano: This is indeed a no-op, I think. The only place this a concern is databases since, as you say, they contain lots of little files. These are being zipped up using `codeql database bundle`. The SARIF and some of the logs (those that exist outside of a database) are being uploaded unzipped, but that is okay because they are only a couple of files.

src/analyze-action.ts
aeisenberg: Yep...sorry...you're right here.

aeisenberg (commented):
Nice. All my comments turned into noops. Thanks for explaining. The code looks fine to me. Just need to address merge conflicts and 🚢 .

edoardopirovano:
Thanks! Have squashed and rebased to address the merge conflicts.

aeisenberg (approved):

adityasharad:
I think this needs a changelog entry.

PR #801, opened 2021-10-29
Allow multiple uploads in a single job

aeisenberg:
They must all have a unique category.


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Allow multiple uploads in a single job as long as they have different categories. [#801](https://github.com/github/codeql-action/pull/ç)
```
adityasharad, line 5: Not quite the pull number you were going for, I think :)
```suggestion
- The `upload-sarif` action now allows multiple uploads in a single job, as long as they have different categories. [#801](https://github.com/github/codeql-action/pull/801)
```

src/upload-lib.ts
```diff
@@ -409,3 +400,19 @@ async function uploadFiles(
...
+
+export function validateUniqueCategory(category: string | undefined) {
+  if (util.isActions()) {
+    // This check only works on actions as env vars don't persist between calls to the runner
+    const sentinelEnvVar = `CODEQL_UPLOAD_SARIF + ${
+      category ? `_${category}` : ""
+    }`;
+    if (process.env[sentinelEnvVar]) {
+      throw new Error(
+        "Aborting upload: only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job per category. " +
```
adityasharad, line 412: Perhaps include the category in the error message, if one is set, or mention the default is being used if none is set.

src/upload-lib.ts
```diff
@@ -409,3 +400,19 @@ async function uploadFiles(
     num_results_in_sarif: numResultInSarif,
   };
 }
+
+export function validateUniqueCategory(category: string | undefined) {
+  if (util.isActions()) {
+    // This check only works on actions as env vars don't persist between calls to the runner
+    const sentinelEnvVar = `CODEQL_UPLOAD_SARIF + ${
+      category ? `_${category}` : ""
```
adityasharad, line 408: Is an arbitrary category string valid as an environment variable suffix?

src/upload-lib.test.ts
```diff
@@ -175,3 +175,14 @@ test("populateRunAutomationDetails", (t) => {
   );
   t.deepEqual(modifiedSarif, expectedSarif);
 });
+
+test("validateUniqueCategory", (t) => {
```
adityasharad, line 179: Can you also test two _different_ categories in these cases, and that it doesn't throw?

adityasharad (commented):
Thanks. I didn't know we had this mechanism and it's pretty neat to avoid user errors. Some suggestions for making that experience even nicer.

src/upload-lib.ts
aeisenberg: Oh...that's a good point. We'll need to escape this.

src/upload-lib.test.ts
aeisenberg: This is already being tested since each time you call `validateUniqueCategory`, an env var is created for that call.

aeisenberg:
Failing Debug artifact upload check is unrelated.

cklin (approved):

PR #804, opened 2021-11-02
Include custom query help in analysis results

cklin:


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - The `init` step of the Action now supports `ram` and `threads` inputs to limit resource use of CodeQL extractors. These inputs also serve as defaults to the subsequent `analyze` step, which finalizes the database and executes queries. [#738](https://github.com/github/codeql-action/pull/738)
+- When used with CodeQL bundle 2.7.1 or above, the action now includes custom query help (if it exists in Markdown files that have the same paths as the query files but with `.md` extension instead of `.ql`) in analysis results. [#804](https://github.com/github/codeql-action/pull/804)
```
adityasharad, line 6: How's this?
```suggestion
- When used with CodeQL 2.7.1 or above, the Action now includes custom query help in the analysis results uploaded to GitHub code scanning, if available. To add help text for a custom query, create a Markdown file next to the `.ql` file containing the query, using the same base name but the file extension `.md`. [#804](https://github.com/github/codeql-action/pull/804)
```

src/codeql.ts
```diff
@@ -875,6 +876,8 @@ async function getCodeQLForCmd(
         codeqlArgs.push("--print-metrics-summary");
       if (await util.codeQlVersionAbove(this, CODEQL_VERSION_GROUP_RULES))
         codeqlArgs.push("--sarif-group-rules-by-pack");
+      if (await util.codeQlVersionAbove(this, CODEQL_VERSION_CUSTOM_QUERY_HELP))
+        codeqlArgs.push("--sarif-add-query-help");
```
adityasharad, line 880: Could I trouble you to add a test for this, either as a unit test in `analyze.test.ts`, or an Actions integration test workflow that runs a custom query with a Markdown file next to it, and checks that the produced SARIF contains the Markdown help?

adityasharad (commented):
Code change looks good, some suggestions for change note and tests.

CHANGELOG.md
cklin: That is much better.  I applied it to the commit.

src/codeql.ts
cklin: An Actions integration test is not currently feasible because the version comparison with 2.7.1 would not succeed until next week.  I can certainly add a unit test to verify that the option is set for 2.7.1 but not 2.7.0.  I will follow up when the test is ready.

adityasharad: Good point, only the part of the test matrix that runs against a nightly bundle will work right now.

src/codeql.test.ts
```diff
@@ -400,3 +402,36 @@ test("getCodeQLActionRepository", (t) => {
...
 });
+
+test("databaseInterpretResults() does not set --sarif-add-query-help for 2.7.0", async (t) => {
+  const runnerConstructorStub = stubToolRunnerConstructor();
+  const codeqlObject = await codeql.getCodeQLForTesting();
+  sinon.stub(codeqlObject, "getVersion").resolves("2.7.0");
+  await codeqlObject.databaseInterpretResults("", [], "", "", "", "");
+  t.false(
+    runnerConstructorStub.firstCall.args[1].includes("--sarif-add-query-help"),
+    "--sarif-add-query-help is present"
```
adityasharad, line 413: This is the failure message? Perhaps "should not be present".

src/codeql.test.ts
```diff
@@ -400,3 +402,36 @@ test("getCodeQLActionRepository", (t) => {
...
+});
+
+test("databaseInterpretResults() sets --sarif-add-query-help for 2.7.1", async (t) => {
+  const runnerConstructorStub = stubToolRunnerConstructor();
+  const codeqlObject = await codeql.getCodeQLForTesting();
+  sinon.stub(codeqlObject, "getVersion").resolves("2.7.1");
+  await codeqlObject.databaseInterpretResults("", [], "", "", "", "");
+  t.true(
+    runnerConstructorStub.firstCall.args[1].includes("--sarif-add-query-help"),
+    "--sarif-add-query-help is present"
```
adityasharad, line 424: should be present

adityasharad (approved):

src/codeql.test.ts
cklin: Thank you for the suggestion.  I updated the failure message.

src/codeql.test.ts
cklin: Thank you for the suggestion.  I updated the failure message.

PR #809, opened 2021-11-04
Include matrix information in debug artifact name

edoardopirovano:
Adds the contents of `matrix` to the name of the debug artifact to avoid issues with clashing names in matrixed runs.


src/analyze-action.ts
```diff
@@ -237,8 +237,14 @@ async function run() {
 }
 
 async function uploadDebugArtifacts(toUpload: string[], rootDir: string) {
+  let suffix = "";
+  const matrix = actionsUtil.getRequiredInput("matrix");
+  if (matrix !== undefined && matrix !== "null") {
+    for (const entry of Object.entries(JSON.parse(matrix)).sort())
```
aeisenberg, line 243: How does sort work if the keys are arrays? I did a little test and it seems to work. Also, we only need this functionality to ensure our tests work, so even if the sorting is not stable, if our tests pass consistently, I'm not too worried.

aeisenberg (commented):
The uploaded artifacts look good. Is there any reason why we're not running these tests using windows?

aeisenberg (approved):
Meant to approve the first time, but clicked the wrong button. :)

src/analyze-action.ts
edoardopirovano: I believe the default way to compare two arrays in TypeScript is lexicographically. Since in this case the first element is the key of a JSON map, we know that all the arrays will differ in their first element (since the keys must be unique), so this is equivalent to sorting by the name of the key.

edoardopirovano:
> Is there any reason why we're not running these tests using windows?

Only that the bash script that checks the result wouldn't work on Windows (I think).

PR #817, opened 2021-11-15
Respect value of `LD_PRELOAD` given by the CLI

edoardopirovano:
When implementing its own version of sandwiched tracing, the Action was setting a value of `LD_PRELOAD`. This code stayed outside of the condition for using the new style of tracing (i.e. using sandwiched tracing form the CLI). So, the Action would try to overwrite the value of `LD_PRELOAD` given by the CLI. This should not have been the case - when using the CLI's sandwiched tracing we should use the value of `LD_PRELOAD` provided by the CLI so that the CLI is free to move/rename the libraries as it pleases.


src/tracer-config.ts
```diff
@@ -212,25 +212,25 @@ export async function getCombinedTracerConfig(
       );
     }
     mainTracerConfig = concatTracerConfigs(tracedLanguageConfigs, config);
-  }
```
adityasharad, line 215: I love how this PR is just moving a brace, but an extremely important one!
This does make me think we should factor the `else` block into a `getLegacyTracerConfig` function so that it's easy to spot what's going on. No need in this PR though.

adityasharad: Also it's good that the internal integration tests caught this, but why didn't the external tests catch this for 2.7.1?

adityasharad (approved):

src/tracer-config.ts
edoardopirovano: Agree that refactoring this might be sensible - let's leave that for later though. The tests against the nightly build will start failing soon if this isn't merged and I don't want this to hold up finishing off the 2.7.1 release tomorrow with a merge to `v1`.

2.7.1 did not include the renaming of the tracing mechanisms which only went into `main` earlier today and will end up in 2.7.2. So, with 2.7.1 the value of the `LD_PRELOAD` wasn't really being overwritten since it was just replaced with itself.

adityasharad: That is a relief. It was overwritten but the value was the same, and we changed that value in 2.7.2. Thank you for explaining!

PR #833, opened 2021-12-03
Deal with new dotnet version

robertbrignull:
A few CI checks are currently failing because dotnet is waiting for some user-input. e.g. https://github.com/github/codeql-action/runs/4381100962?check_suite_focus=true

This seems to be because of a change in the CodeQL CLI and how it handles dotnet under tracing in macos. This may get fixed in CodeQL but for now a workaround to get our CI working is to call `dotnet restore` when not traced by CodeQL so it can do it's first-time setup without interference.

### DIFF
--- a/.github/workflows/__debug-artifacts.yml
+++ b/.github/workflows/__debug-artifacts.yml
@@ -41,6 +41,8 @@ jobs:
       uses: ./.github/prepare-test
       with:
         version: ${{ matrix.version }}
+    - name: Initialize dotnet
+      run: dotnet restore
     - uses: ./../action/init
       with:
         tools: ${{ steps.prepare-test.outputs.tools-url }}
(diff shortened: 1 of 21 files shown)


edoardopirovano (approved):

edoardopirovano (approved):
New change LGTM. I wonder whether it might be easier to have `sync.py` insert that call at the start of every workflow. Needn't be done in this PR, though - for now let's get this merged to unblock our tests.

robertbrignull:
> New change LGTM. I wonder whether it might be easier to have `sync.py` insert that call at the start of every workflow. Needn't be done in this PR, though - for now let's get this merged to unblock our tests.

Thanks. I also think it would be better to just get it working for now. Having `sync.py` add that sounds nice, and would fit in with work on reducing the parallelism too.

edoardopirovano:
> Thanks. I also think it would be better to just get it working for now. Having `sync.py` add that sounds nice, and would fit in with work on reducing the parallelism too.

Agreed, I can look at doing that when I'm working on `sync.py` anyway. For now, let's get our CI working again 🙂

robertbrignull:
CI is passing so this looks like it works

PR #843, opened 2021-12-09
Start waiting for processing by default.

chrisgavin:
The option to wait for processing to complete has worked well during internal testing so I think we could now be at the point where we enable it by default.

- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary. (I don't think it's necessary, since most people will probably want to use this feature and it should only be turned off if it's causing issues.)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Update default CodeQL bundle version to 2.7.3. [#842](https://github.com/github/codeql-action/pull/842)
+- The `analyze` and `upload-sarif` actions will now wait up to 2 minutes for processing to complete so they can report any processing errors that occurred. This behavior can be disabled by setting the `wait-for-processing` action input to `"false"`.
```
cannist, line 6:
```suggestion
- The `analyze` and `upload-sarif` actions will now wait up to 2 minutes for processing to complete after they have uploaded the results so they can report any processing errors that occurred. This behavior can be disabled by setting the `wait-for-processing` action input to `"false"`.
```

cannist (commented):
Thank you for this PR, Chris, I had wanted to do this yesterday but did not get to it. 🙂 

Just one suggestion to have more context in the change note.

cannist (approved):

PR #844, opened 2021-12-09
Stop printing all output twice.

chrisgavin:
Maybe I'm missing some reason for this code existing, but it feels like all it does is duplicate the extractor output in the Actions logs. Can we remove it or have I missed something here?


chrisgavin:
Here's a somewhat noisy but hopefully still useful diff showing the effect of the change on a JavaScript build. https://www.diffchecker.com/GQYzoaIw

hmakholm (approved):

PR #853, opened 2021-12-13
Remove scheduled releases

aeisenberg:
The action will be released roughly every two weeks along
with the CodeQL CLI release process.

### DIFF
--- a/.github/workflows/update-release-branch.yml
+++ b/.github/workflows/update-release-branch.yml
@@ -1,7 +1,5 @@
 name: Update release branch
 on:
-  schedule:
-    - cron: 0 9 * * 1
   repository_dispatch:
     # Example of how to trigger this:
     # curl -H "Authorization: Bearer <token>" -X POST https://api.github.com/repos/github/codeql-action/dispatches -d '{"event_type":"update-release-branch"}'


.github/workflows/update-release-branch.yml
```diff
@@ -1,7 +1,5 @@
 name: Update release branch
 on:
-  schedule:
-    - cron: 0 9 * * 1
   repository_dispatch:
     # Example of how to trigger this:
     # curl -H "Authorization: Bearer <token>" -X POST https://api.github.com/repos/github/codeql-action/dispatches -d '{"event_type":"update-release-branch"}'
```
aeisenberg, line 5: Should we remove this as well?

```suggestion
```

.github/workflows/update-release-branch.yml
```diff
@@ -1,7 +1,5 @@
 name: Update release branch
 on:
-  schedule:
-    - cron: 0 9 * * 1
   repository_dispatch:
```
aeisenberg, line 3: Should we remove this as well?

```suggestion
```

edoardopirovano: I'm not aware of if/how this is used so I have no objections to removing it. I guess if something breaks we can just add it back in :)

edoardopirovano (approved):

PR #856, opened 2021-12-14
Feature flagging via the GitHub API

henrymercer:
This PR introduces feature flagging to the CodeQL Action via a GitHub API endpoint `/repositories/:repository_id/code-scanning/codeql-action/features` (see backlinked issue).  This API endpoint only exists on Dotcom, so for GHES and GHAE feature flags will all be marked as disabled.

This is a general purpose solution that'll allow us to add many feature flags to the CodeQL Action while only calling a single API endpoint.  In this PR we use this mechanism to replace the database uploading feature flags; in the future we'll use it to decide whether to run ML-powered queries.

I've written some unit tests for the feature flagging class, and verified that database upload works as expected on a repo with the feature flag enabled.  Ideas for further testing welcome.


src/feature-flags.ts
```diff
@@ -0,0 +1,116 @@
...
+          {
+            owner: repositoryNwo.owner,
+            repo: repositoryNwo.repo,
+          }
+        );
+        return response.data;
+      } catch (e) {
+        console.log(e);
+        this.logger.info(
+          `Disabling all feature flags due to unknown error: ${e}`
```
robertbrignull, line 80: Is this a bit dangerous? In case of a transient error, is it better to crash vs doing something unusual? If a user is running ML powered queries but then one of their workflows randomly doesn't run them because this endpoint encountered an error, it'll cause a lot of confusing alert churn as everything closes and reopens.

robertbrignull (commented):
I haven't reviewed in huge detail. I can do that tomorrow or leave it to the CodeQL team. I have a couple of initial comments though.

src/analyze-action-env.test.ts
```diff
@@ -25,6 +25,8 @@ test("analyze action with RAM & threads from environment variables", async (t) =
       .resolves({} as actionsUtil.StatusReportBase);
     sinon.stub(actionsUtil, "sendStatusReport").resolves(true);
     sinon.stub(configUtils, "getConfig").resolves({
+      // Use GHES so we don't try to call the feature flags API endpoint
+      gitHubVersion: { type: util.GitHubVariant.GHES, version: "3.0.0" },
```
aeisenberg, line 29: Can you explain why this is necessary? Is it to avoid making an actual HTTP request to the github api during tests? If so, then we should probably stub out the API itself and add an expectation that the endpoint is requested.

src/feature-flags.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  constructor(
+    private gitHubVersion: util.GitHubVersion,
+    private apiDetails: GitHubApiDetails,
+    private repositoryNwo: RepositoryNwo,
+    private logger: Logger
+  ) {}
+
+  getDatabaseUploadsEnabled(): Promise<boolean> {
+    return this.getFeatureFlag("database_uploads_enabled");
```
aeisenberg, line 31: I notice these constants are used in several places (at least in the tests). Can you extract them into a typescript enum?

aeisenberg: Ah...I see you have them as a `type` below. That's fine. I have a slight preference for enums, but have no reason other than aesthetics. So, it's fine as-is.

src/feature-flags.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return response || false;
+  }
+
+  private async getApiResponse(): Promise<FeatureFlagsApiResponse> {
+    const loadApiResponse = async () => {
+      // Do nothing when not running against github.com
+      if (this.gitHubVersion.type !== util.GitHubVariant.DOTCOM) {
+        this.logger.debug(
+          "Not running against github.com. Disabling all feature flags."
```
aeisenberg, line 61: Curious if this is always the case, or only for the feature flags we care about.

aeisenberg (commented):
Looks good. A few minor comments.

src/database-upload.test.ts
```diff
@@ -324,9 +256,13 @@ test("Don't crash if uploading a database fails", async (t) => {
     await uploadDatabases(
       testRepoName,
       getTestConfig(tmpDir),
+      featureFlags,
       testApiDetails,
       getRecordingLogger(loggedMessages)
     );
+
+    console.log(loggedMessages);
```
robertbrignull, line 264: Leftover console.log

robertbrignull (commented):
LGTM too once @aeisenberg is happy

src/analyze-action-env.test.ts
henrymercer: Yeah, this could be a bit more transparent.  Using GHES as our `gitHubVersion` will mean we don't call the feature flag API endpoint during the tests.  But stubbing the API is clearer — I'll do this.

src/feature-flags.ts
henrymercer: I agree enums are aesthetically a bit nicer.  Will update.

src/feature-flags.ts
henrymercer: The feature flags we have now are all Dotcom-specific.  I don't know of any immediate blockers to us using feature flags on GHES/GHAE in the same way, but haven't looked into it.

aeisenberg (approved):
Nice.

PR #857, opened 2021-12-15
Add support for running ML-powered queries for JS `security-extended` behind `ml_powered_queries` feature flag

henrymercer:
This PR builds on #856 to add support for running ML-powered queries.  Currently these queries would only be run for customers (a) performing JavaScript analysis, (b) running the `security-extended` (or `security-and-quality`) query suite, and (c) whose repos are opted into the `ml_powered_queries` feature flag.

The action happens in `src/config-utils.ts`: when all of these conditions hold, we add the `codeql/javascript-experimental-atm-queries` pack to the analysis. 

Since the security review for this feature is not yet complete, these queries will currently return no results.  We will not roll out the feature flag until this security review is complete to avoid needlessly increasing the runtime of customers' analyses by downloading the ML-powered query pack.

### Points for review

We want to enable ML-powered queries based in part on which query suite the customer is running. We resolve query suites into query paths in the `init` action, and the `analysis` action to my knowledge doesn't know what query suites the user requested. Therefore the `init` action is the natural point to test whether we should run ML-powered queries.

However this does necessitate an API call to test feature flag enablement in the `init` action as well as the existing one in the `analysis` action that's used for database upload.

Another option is saving some kind of information (e.g. feature flags or the query suites requested) to the config file in the `init` action and loading them in the `analysis` action.  I didn't go with this since it seems like overengineering for the benefit of one fewer API call, and the API call in consideration doesn't count towards rate limits, but what do reviewers think?


src/config-utils.ts
```diff
@@ -262,14 +263,33 @@ async function addBuiltinSuiteQueries(
...
+    languages.includes("javascript") &&
+    (found === "security-extended" || found === "security-and-quality") &&
+    (await featureFlags.getValue(FeatureFlag.MlPoweredQueriesEnabled))
+  ) {
+    if (!packs.javascript) {
+      packs.javascript = [];
+    }
+    packs.javascript.push({
+      packName: "codeql/javascript-experimental-atm-queries",
+      version: "0.0.2",
```
aeisenberg, line 289: Do you want to hard code this version? You don't want to have to update the action for every new release of ATM. I think changing this to `latest` would work.

src/init-action.ts
```diff
@@ -129,6 +130,18 @@ async function run() {
...
+    getRequiredEnvParam("GITHUB_REPOSITORY")
+  );
+
+  const featureFlags = new GitHubFeatureFlags(
+    gitHubVersion,
+    apiDetails,
+    repositoryNwo,
+    logger
+  );
+  void featureFlags.preloadFeatureFlags();
```
aeisenberg, line 143: Just making sure that this is safe. 

The idea is to kick off the API request before it's needed and without waiting for it so that by the time it's actually needed, there's less time needed to wait for the result?

In general, I don't think we are much concerned about the time it takes for a single request, but this looks fine if my understanding is correct.

aeisenberg (commented):
Couple of questions inline.

Also, to answer yours, I think it's fine to make the feature flags request in init AND analyze.

What do you think about adding a comment of storing the feature flags in the config file as a potential future improvement?

src/config-utils.ts
henrymercer: The intention here is to allow us to make a breaking change in the query pack that for instance requires a new CLI, but I agree that this is a little strict.  Would semver syntax like `^0.0.2` be supported here?

src/init-action.ts
henrymercer: > The idea is to kick off the API request before it's needed and without waiting for it so that by the time it's actually needed, there's less time needed to wait for the result?

That's the idea.  I agree the benefit is probably quite small.

src/config-utils.ts
aeisenberg: Careful here, I've been bitten by this before. `^0.0.2` means `>= 0.0.2` and `< 0.0.3`, which is probably not what you want.

`~0.0.2` means `>= 0.0.2` and `< 0.1.0`. I think this is what you want. [See npm's semver for more info.](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004)

aeisenberg (approved):
This is good to go in after updating the version range.

PR #861, opened 2021-12-31
Remove debugging output

edoardopirovano:
Looks like I accidentally committed some debugging output when I wrote this part of our CI. This PR removes it.

### DIFF
--- a/.github/prepare-test/action.yml
+++ b/.github/prepare-test/action.yml
@@ -28,7 +28,6 @@ runs:
           echo "::set-output name=tools-url::https://github.com/dsp-testing/codeql-cli-nightlies/releases/download/codeql-bundle-$VERSION-manual/codeql-bundle.tar.gz"
         elif [[ ${{ inputs.version }} == *"stable"* ]]; then
           export VERSION=`echo ${{ inputs.version }} | sed -e 's/^.*\-//'`
-          echo "Hello $VERSION"
           echo "::set-output name=tools-url::https://github.com/github/codeql-action/releases/download/codeql-bundle-$VERSION/codeql-bundle.tar.gz"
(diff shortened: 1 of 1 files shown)


aeisenberg (approved):

aeisenberg:
That's wild. Looks like failures on Python 2 on macos. Rerunning on main to see if this is a general problem: https://github.com/github/codeql-action/actions/runs/1633534088

aeisenberg:
Yep. Same jobs are failing on main. This must be related to some changes to the python 2 version that is bundled in the virtual environment. Will look deeper when I am back at work.

edoardopirovano:
Hmm, strange. I can't imagine this change causing any test failures so it's probably something that is failing on main too. Maybe a change to the Python version in the Mac OS Actions image is causing the test failures?

aeisenberg:
Here is the most recent passing job: https://github.com/github/codeql-action/runs/4668380714?check_suite_focus=true

You can see that the image used in the passing workflow is the same as the one used in the failing workflow. So, it must be some sort of python transitive dependency that has changed.

aeisenberg:
Difference is the version of virtualenv. For [the passing job](https://github.com/github/codeql-action/runs/4668380714?check_suite_focus=true#step:4:222), it is 20.11.2. For [the failing job](https://github.com/github/codeql-action/runs/4681503135?check_suite_focus=true#step:4:223), it is 20.12.1.

So, looks like we need to force virtualenv@20.11.2 for python 2 tests.

aeisenberg:
Let's see if this works. https://github.com/github/codeql-action/pull/862

edoardopirovano:
Sweet, thanks for fixing that!

PR #868, opened 2022-01-07
Make name of debugging artifact and DB within it configurable

edoardopirovano:
As requested by @esbena and @hvitved for use on DCA.

Adds two new options next to the `debugMode` one to control the name of the debugging artifact uploaded and the name of the database within that artifact. This is needed on DCA because it uses the Action multiple times within one workflow with different SHAs that it is comparing, so without this the debugging artifacts will overwrite each-other. It may also be useful to users who want their debugging artifacts to have more meaningful names.

@robertbrignull This changes the code that calls `codeql database bundle` which is also called by the code that uploads databases to Remote Queries. In the Remote Queries instance of this case, I have passed in just `language` as the name of the DB. This is what was (implicitly) being used before, so I believe this change to be behaviour-preserving from the point of view of Remote Queries. You may want to check it nonetheless.


robertbrignull (commented):
Thanks for the ping. The remote queries changes LGTM.

init/action.yml
```diff
@@ -59,6 +59,16 @@ inputs:
     description: Enable debugging mode. This will result in more output being produced which may be useful when debugging certain issues.
     required: false
     default: 'false'
+  debugArtifactName:
```
esbena, line 62: Should these new names be `kebab-case`?

edoardopirovano: Hmm, indeed that is the convention elsewhere. Let me update it.

src/util.ts
```diff
@@ -559,13 +564,11 @@ export async function codeQlVersionAbove(
...
-  codeql: CodeQL
+  codeql: CodeQL,
+  dbName: string
 ) {
   const databasePath = getCodeQLDatabasePath(config, language);
-  const databaseBundlePath = path.resolve(
-    config.dbLocation,
-    `${databasePath}.zip`
-  );
+  const databaseBundlePath = path.resolve(config.dbLocation, `${dbName}.zip`);
```
aeisenberg, line 571: So, this will now place the bundled db in a different location? I don't _think_ that's a problem, but it is a change and just want to be sure.

edoardopirovano: Indeed, that is correct. I _hope_ there's no code outside the `codeql-action` that is relying on where we happen to put the DB bundles before uploading them to an artifact. I don't think that's a use-case we've ever claimed to support, nor do I think we should be supporting it, so I think it is okay to break.

aeisenberg (approved):

PR #869, opened 2022-01-11
Refuse to run on Windows 11

edoardopirovano:
Refuse to run on Windows 11 and Windows Server 2022 which are currently unsupported. Sample failing run (obtained by modifying one of our CI checks to run on `windows-2022`): https://github.com/github/codeql-action/runs/4779298247


src/util.ts
```diff
@@ -588,3 +588,11 @@ export async function delay(milliseconds: number) {
 export function isGoodVersion(versionSpec: string) {
   return !BROKEN_VERSIONS.includes(versionSpec);
 }
+
+export function checkNotWindows11() {
+  if (os.platform() === "win32" && semver.gte(os.release(), "10.0.20348")) {
+    throw new Error(
+      "Unfortunately, CodeQL currently does not support Windows 11 or Windows Server 2022. Please modify your workflow to use an earlier version of Windows."
```
adityasharad, line 595:
```suggestion
      "CodeQL currently does not support Windows 11 or Windows Server 2022. Please modify your Actions workflow to use an earlier version of Windows for this job, for example by setting `runs-on: windows-2019`."
```

adityasharad:
I just remembered the suggestion I made on the corresponding CLI PR: non-compiled languages are perfectly fine, and we should avoid blocking them from running on Windows 11. Could we move this check later and error only if there is a compiled language in use?

edoardopirovano:
Good point, done @adityasharad. Sample error on the latest commit: https://github.com/github/codeql-action/runs/4779658824

src/init-action.ts
```diff
@@ -192,6 +193,11 @@ async function run() {
       logger
     );
 
+    if (config.languages.filter(isTracedLanguage).length > 0) {
```
adityasharad, line 196: Minor. Should be equivalent.
```suggestion
    if (config.languages.some(isTracedLanguage)) {
```

adityasharad (approved):

PR #876, opened 2022-01-19
Include better error message

aeisenberg:
When users call init multiple times. This PR addresses an error that we saw in our telemetry. It looks like there was a user who was calling the init action multiple times in the same job and getting an error. This seems like an error that new users will come across when they want to analyze databases for multiple languages.

This PR makes the error message clearer in this case.

@edoardopirovano suggested a more robust fix where we introduce a `UserError` and handle them differently from `Error`s. This PR does not implement the suggestion, but is something we can do in the future.


aeisenberg:
I'll update the changelog and then merge. Thanks for the review.

PR #878, opened 2022-01-21
Revert "Start waiting for processing by default."

chrisgavin:
This reverts commit b661ef1697d555e1f2ab5c42d3c60b5bb71b6cf2.

It looks like this change caused issues with analyses on pull requests from forks. This is due to the complicated nature of the permissions of workflows run on these pull requests. They are able to upload analyses for their own pull requests refs, but are not able to access other Code Scanning APIs such as the one required to get the status of a delivery.

This reverts adding this new functionality until we update the API to make it able to safely determine whether a pull request should be able to get the status of a particular delivery.

This is a hotfix on the v1 branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The feature to wait for SARIF processing to complete after upload has been disabled by default due to a bug in its interaction with pull requests from forks.
 
 ## 1.0.28 - 18 Jan 2022
 
--- a/analyze/action.yml
+++ b/analyze/action.yml
@@ -55,7 +55,7 @@ inputs:
   wait-for-processing:
     description: If true, the Action will wait for the uploaded SARIF to be processed before completing.
     required: true
-    default: "true"
+    default: "false"
   token:
     default: ${{ github.token }}
   matrix:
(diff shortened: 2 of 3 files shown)


edoardopirovano:
Could we add a changelog note mentioning this revert? We mentioned the original change in the log, so I think the revert should be too.

chrisgavin:
Sure! I've put this under a 1.0.29 release with today's date on. Let me know if this is not the correct way of doing a change note for a hotfix though.

edoardopirovano:
> Sure! I've put this under a 1.0.29 release with today's date on. Let me know if this is not the correct way of doing a change note for a hotfix though.

Oh I've just realized this PR is targeting `v1`. I think it should probably target `main` and the note should go in the "unreleased" heading. Then our automation can handle doing a release with the fix in and do the right things with regards to bumping the version number in all the right places.

chrisgavin:
Ah, okay that makes sense. I was a bit worried that would make it more difficult to do a hotfix because it would pull in other unrelated changes, but it actually looks like there aren't any, so that should work fine.

edoardopirovano:
> Ah, okay that makes sense. I was a bit worried that would make it more difficult to do a hotfix because it would pull in other unrelated changes, but it actually looks like there aren't any, so that should work fine.

Yeah, if the day ever comes that we have to do a hotfix directly on `v1` it will be a slight pain because we'll need to manually bump the version number in a few places then do a manual mergeback to `main`. No need for that here though, much easier to do the change on `main` and let the automation handle releasing.

edoardopirovano (approved):

cannist:
:+1:

PR #882, opened 2022-01-21
Ensure loadApiError is caught

aeisenberg:
And add a better error message.

By using `void` instead of `await`, any error thrown is not caught
by surrounding try-catch blocks.

I could continue to use `void` and explicitly handle any thrown errors
by using `.catch`, but most likely the time savings is minimal and
this makes the code more complex.


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Display a better error message when encountering a workflow that runs the `codeql-action/init` action multiple times. [#876](https://github.com/github/codeql-action/pull/876)
+- Add better error message is the workflow does not have security-events write permissions. [#882](https://github.com/github/codeql-action/pull/882)
```
henrymercer, line 6:
```suggestion
- Add better error message is the workflow does not have the `security-events: write` permission. [#882](https://github.com/github/codeql-action/pull/882)
```

src/feature-flags.ts
```diff
@@ -66,13 +66,25 @@ export class GitHubFeatureFlags implements FeatureFlags {
...
-        );
+        if (
+          e instanceof Error &&
+          e.message.includes("Resource not accessible by integration")
+        ) {
+          throw new Error(
+            `Resource not accessible by integration. This usually means that your ` +
+              `workflow is missing the required security-events write permissions. ` +
+              `See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions ` +
+              `for more information.`
```
henrymercer, line 77: I came up with:

```js
this.logger.warning(
  "This run of the CodeQL Action does not have permission to access Code Scanning API endpoints. " +
  "As a result, it will not be opted into any experimental features. " +
  "This could be because the Action is running on a pull request from a fork. If not, " +
  `please ensure the Action has the 'security-events: write' permission. Details: ${e}`
);
```

I think this probably should be a warning — we don't want to fail on PRs from forks.

It also might be useful to (a) mention the impact and (b) mention that this is expected behaviour in certain cases, e.g. PRs from forks.

src/feature-flags.ts
```diff
@@ -66,13 +66,25 @@ export class GitHubFeatureFlags implements FeatureFlags {
...
-        // Some feature flags, such as `ml_powered_queries_enabled` affect the produced alerts.
-        // Considering these feature flags disabled in the event of a transient error could
-        // therefore lead to alert churn. As a result, we crash if we cannot determine the value of
-        // the feature flags.
-        throw new Error(
-          `Encountered an error while trying to load feature flags: ${e}`
-        );
+        if (
+          e instanceof Error &&
+          e.message.includes("Resource not accessible by integration")
```
henrymercer, line 71: `util.isHTTPError(e) && e.status === 403` is slightly more general and guards against this message changing.

src/init-action.ts
```diff
@@ -143,9 +143,10 @@ async function run() {
     repositoryNwo,
     logger
   );
-  void featureFlags.preloadFeatureFlags();
 
   try {
+    await featureFlags.preloadFeatureFlags();
```
henrymercer, line 148: Can we still void this?  There's no point in preloading the flags if we void them.  Alternatively can we kick this promise off at the start of the `try` and await it at the end?

henrymercer (commented):
I was just looking at this too.  Couple of suggestions.

src/feature-flags.ts
aeisenberg: Thanks. Good points.

src/init-action.ts
aeisenberg: If we void it, then we need to explicitly handle any errors. This may be fine if the security events error should only be a warning. Maybe `Encountered an error while trying to load feature flags` should be a warning as well?

If either of these are warnings, then we will try to grab feature flags again later. Should we also set the feature flags to all falsy if we fail to grab them?

src/feature-flags.ts
aeisenberg: Nice.

src/init-action.ts
henrymercer: There's a comment in there about why we can't do that — in brief some feature flags affect alerts, so transient errors with the API endpoint could lead to alert churn if we don't crash.

aeisenberg: OK...so, we can't do it for non-403 errors, but we can for the case that we are trying to solve. 

So...my suggestion is for 403 errors, we can return falsy values, but for other errors, we must end the workflow run.

src/feature-flags.ts
aeisenberg: Actually...thinking about this now. If `security-events: write` is required anyway, wouldn't it be better to fail earlier?

src/init-action.ts
henrymercer: > So...my suggestion is for 403 errors, we can return falsy values, but for other errors, we must end the workflow run.

Yep, that sounds sensible to me.  I edited my comment above with a suggestion about telemetry, but I don't think it blocks the PR.

aeisenberg: > One of the reasons to be within the try though would be error telemetry. To get this, we could change the preload method to tryPreloadFeatureFlags and have it fail silently. The main disadvantage would probably be a few extra API requests in the event that the requests error with non 403s, which isn't the end of the world.

Sorry...a little confused here. The comment says that `ml_powered_queries_enabled` means we must fail hard if we can't get the feature flag. So, for non-403s, we would just get a single API failure and the run would end, right?

src/feature-flags.ts
henrymercer: Referring to #878:

> It looks like this change caused issues with analyses on pull requests from forks. This is due to the complicated nature of the permissions of workflows run on these pull requests. They are able to upload analyses for their own pull requests refs, but are not able to access other Code Scanning APIs such as the one required to get the status of a delivery.

I agree it's sensible to error in `init` if `security-events: write` is missing, since we wouldn't be able to upload the SARIF.  But we can get this error on a PR from a fork, even if `security-events: write` is present.

src/init-action.ts
henrymercer: I think if we void the `preloadFeatureFlags` promise (as we do in `main`), then a hard fail exception won't get captured by the `catch` and so we won't send an "init aborted" status report.  We have a couple of hard fail exceptions that are currently logged by that telemetry, and I was just thinking about how we could make `Encountered an error while trying to load feature flags` one of them.

src/feature-flags.ts
aeisenberg: OK....so it sounds like we should change your warning message above into an error that we throw, so the explanation is clear. And then make sure that error is caught and logged through telemetry. This way, the job fails quicker and with a clear reason why.

src/init-action.ts
aeisenberg: Based on our thread above, I think we should be throwing all errors from the function above. And as you suggest, we can also run the function at the start of the try block, and `await` it at the end.

src/feature-flags.ts
adityasharad: I think we need to be careful about that -- won't that start failing all runs on PRs from forks?

henrymercer: > OK....so it sounds like we should change your warning message above into an error that we throw, so the explanation is clear. And then make sure that error is caught and logged through telemetry. This way, the job fails quicker and with a clear reason why.

Sorry, that wasn't what I meant.  We have two separate cases:

1. a user hasn't specified `security-events: write` and is running Code Scanning on a standard `push` or `pull_request` event
2. a user has indeed specified `security-events: write`, but they're running Code Scanning on a PR from a fork, and so the GitHub token doesn't have permission to access the feature flags API.

For (1) I think it's sensible to error, as the workflow file is missing the `security-events: write` permission.

For (2) the workflow file might contain that permission.  I think we should log a warning to explain the impact of running on a fork, but we shouldn't error since that would break analyses on forks.  I'll investigate in parallel whether we can change the permissions required for the feature flags API, but that'll likely be a longer time horizon change.

aeisenberg: Is there a way to distinguish these two scenarios? If not, we'll have to avoid errors here.

henrymercer: Let's perhaps just do (2) for now.  We can do (1) in a separate PR, if it ends up being possible to check the permissions assigned in the workflow file.

aeisenberg: That makes sense.

src/init-action.ts
aeisenberg: Wait...this won't work. We are `await`ing the promise, only after it thows. We are going to have to add a rejection handler, capture the error, and then optionally throw it later. It will have to look something like this:

```js
async function errorFunction() {
  throw new Error()
}

async function wait() {
  return new Promise(res => setTimeout(res, 1000))
}

(async function() {
  try {
    let maybeError
    const promise = errorFunction().catch(e => maybeError = e)
    await wait()
    await promise

    if (maybeError) {
      throw maybeError
    }
  } catch (e) {
    console.log('error is caught')
  }

})()
```

henrymercer: Hmm, I'm not sure the complexity of that is worth the slight perf improvement from preloading the feature flags — perhaps we should remove the preloading.

aeisenberg: I think that would be fine. I don't think the API request is particularly long running...is it?

henrymercer: No, I'm not concerned.

edoardopirovano:
I haven't reviewed the code since Henry seems to have it covered (although happy to if you would like another set of eyes), but could I ask to hold off merging this until Tuesday since it seems like a potentially risky change to put into GHES 3.4 at the last minute?

aeisenberg:
Thanks for the reminder. I don't think this is a required fix for GHES 3.4. I will hold off on the merge.

henrymercer (approved):
A couple of suggestions, otherwise LGTM to get in after the GHES 3.4 release.

edoardopirovano:
I've done the release that is going into GHES 3.4 so please do not consider this blocked on that anymore 👍🏼

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Display a better error message when encountering a workflow that runs the `codeql-action/init` action multiple times. [#876](https://github.com/github/codeql-action/pull/876)
+- Add better error message is the workflow does not have the `security-events: write` permission. [#882](https://github.com/github/codeql-action/pull/882)
```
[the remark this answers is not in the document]
aeisenberg: I'd like to keep a mention of the `security-events: write` requirement and I think mentioning the warning around forks is important too.

src/feature-flags.test.ts
```diff
@@ -102,26 +102,6 @@ test("Feature flags are disabled if they're not returned in API response", async
   });
 });
 
-test("Feature flags exception is propagated if the API request errors", async (t) => {
```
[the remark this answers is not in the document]
aeisenberg: Sure. That works.

henrymercer (approved):
Thanks!

Issue #796, opened 2021-10-28
upload-sarif@v1 : master does not match /^refs\/(heads|pull|tags)\/.*$/.


totogtr:
Hi,

I'm facing an issue uploading a SARIF file using upload-sarif@v1 on a self-hosted runner.

Here is my code (in a composite action) : 
```
    - name: Scan image
      id: scan
      uses: sysdiglabs/scan-action@v3
      with:
       [...]

    - uses: github/codeql-action/upload-sarif@v1
      with:
        sarif_file: ${{ steps.scan.outputs.sarifReport }}
```

This works great when I push / PR to a branch called `develop` but an issue is happening when I apply the same composite action to `master` branch. It looks like for some reason the upload-sarif@v1 action doesn't call the API correctly ?

Here are the logs on develop (omitting env variables) : 
```
Run github/codeql-action/upload-sarif@v1
  with:
    sarif_file: ./sarif.json
    checkout_path: /home/azureuser/actions-runner-2/runner/_work/my-app/my-app
    token: ***
    matrix: {
    "node-version": "16.3.0",
    "env": {
      "name": "development"
    }
  }
  env:
    [...]
Uploading results
  Processing sarif files: ["./sarif.json"]
  Uploading results
  Successfully uploaded results
```
I can see the uploaded results on the security tab of the github repo.

Here are the logs on master: 
```
  Processing sarif files: ["./sarif.json"]
  Uploading results
  Error: Invalid request.
  
  master does not match /^refs\/(heads|pull|tags)\/.*$/.
  RequestError [HttpError]: Invalid request.
  
  master does not match /^refs\/(heads|pull|tags)\/.*$/.
      at /home/azureuser/actions-runner-2/runner/_work/_actions/github/codeql-action/v1/node_modules/@octokit/request/dist-node/index.js:66:23
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at async Job.doExecute (/home/azureuser/actions-runner-2/runner/_work/_actions/github/codeql-action/v1/node_modules/bottleneck/light.js:405:18) {
    name: 'HttpError',
    status: 422,
    headers: {
      'access-control-allow-origin': '*',
[...]

PR #889, opened 2022-01-24
Adds ref and SHA as inputs, and sarif-id as output

cw-acroteau:

This PR might be useful for those facing #796 by allowing to specify a ref, but is NOT a direct fix for it.


src/actions-util.ts
```diff
@@ -431,8 +431,15 @@ export function computeAutomationID(
 export async function getRef(): Promise<string> {
   // Will be in the form "refs/heads/master" on a push event
   // or in the form "refs/pull/N/merge" on a pull_request event
-  const ref = getRequiredEnvParam("GITHUB_REF");
-  const sha = getRequiredEnvParam("GITHUB_SHA");
+  const refInput = getOptionalInput("ref");
+  const ref = refInput || getRequiredEnvParam("GITHUB_REF");
+  const sha = getOptionalInput("sha") || getRequiredEnvParam("GITHUB_SHA");
```
aeisenberg, line 436: It should probably be an error if only one of these inputs are specified.

Also, I'm not sure what would happen if you specify a SHA that is not part of the current branch. Hopefully, code scanning would error out..

upload-sarif/action.yml
```diff
@@ -13,6 +13,12 @@ inputs:
     description: "The path at which the analyzed repository was checked out. Used to relativize any absolute paths in the uploaded SARIF file."
     required: false
     default: ${{ github.workspace }}
+  ref:
+    description: "The ref where results will be uploaded. If not provided, the Action will use the GITHUB_REF environment variable."
+    required: false
+  sha:
+    description: "The hash of the HEAD of the ref where results will be uploaded. If not provided, the Action will use the GITHUB_SHA environment variable."
+    required: false
```
aeisenberg, line 21: See suggestions for the `analyze` action.

upload-sarif/action.yml
```diff
@@ -24,6 +30,9 @@ inputs:
     description: If true, the Action will wait for the uploaded SARIF to be processed before completing.
     required: true
     default: "false"
+outputs:
+  sarif-id:
+    description: The ID of the uploaded sarif file.
```
aeisenberg, line 35:
```suggestion
    description: The ID of the uploaded SARIF file.
```

analyze/action.yml
```diff
@@ -63,6 +69,8 @@ inputs:
 outputs:
   db-locations:
     description: A map from language to absolute path for each database created by CodeQL.
+  sarif-id:
+    description: The ID of the uploaded sarif file.
```
aeisenberg, line 73:
```suggestion
    description: The ID of the uploaded SARIF file.
```

aeisenberg (changes requested):
Thank you for your PR. This is looking really good. I have a few questions inline and suggestions inline.

We will also need to create an integration test for this to ensure that the new inputs are working as expected in real workflows. I'm not sure if you will be able to run them due to restrictions on running CI in forks. I'll figure out how to proceed later.

aeisenberg:
Please include the compiled sources (`*.js` and `*.map` files) in this PR. Because of the way actions are run, all build artifacts must be checked into the repo.

src/actions-util.ts
cw-acroteau: 1. I've added a check that throws an error message, and a unit test to confirm it works. Integration test to come with the rest of the integration tests.

2. I was initially targeting the `upload-sarif` action when I added this, so I did really see it as two arbitrary values provided by the user, who would ultimately be responsible to confirm their validity.
I can see two ways to handle this:
   a. I can add an integration test for this specific use case, which would test that the code scanning would error out if an invalid ref is provided. If the error message is meaningful and relevant, we can rely on it.
   b. I can add a check using `git branch ${ref} --contains ${sha}` in getRef(), with a fallback if the git command can't be called (see line 86). I would of course add the corresponding unit/integration tests.
   
For 2., I would prefer option b. This way, this

cw-acroteau: 1. I've added a check that throws an error message, and a unit test to confirm it works. Integration test to come with the rest of the integration tests.

2. I was initially targeting the `upload-sarif` action when I added this, so I did really see it as two arbitrary values provided by the user, who would ultimately be responsible to confirm their validity.
I can see two ways to handle this:
   a. I can add an integration test for this specific use case, which would test that the code scanning would error out if an invalid ref is provided. If the error message is meaningful and relevant, we can rely on it.
   b. I can add a check using `git branch ${ref} --contains ${sha}` in getRef(), with a fallback if the git command can't be called (see line 86). I would of course add the corresponding unit/integration tests.
   
For 2., I would prefer option b. This way, this

cw-acroteau: 1. I've added a check that throws an error message, and a unit test to confirm it works. Integration test to come with the rest of the integration tests.

2. I was initially targeting the `upload-sarif` action when I added this, so I did really see it as two arbitrary values provided by the user, who would ultimately be responsible to confirm their validity.
I can see two ways to handle this:
   a. I can add an integration test for this specific use case, which would test that the code scanning would error out if an invalid ref is provided. If the error message is meaningful and relevant, we can rely on it.
   b. I can add a check using `git branch ${ref} --contains ${sha}` in getRef(), with a fallback if the git command can't be called (see line 86). I would of course add the corresponding unit/integration tests.
   
For 2., I would prefer option b. This way, this

cw-acroteau: EDIT: Option 1 would not be possible, because I realized integration tests can only be positive with the current setup.

So, I can either leave it as-is, without a test, or use option b.

cw-acroteau:
Hi @aeisenberg! Thanks for your feedback, I've included the compiled sources and fixed most inline suggestions.

I am trying to fix the integration tests, in the meantime there is one question remaining in inline suggestions.

FYI, Integration tests run in my fork. I had to enable workflows, then I created a [draft PR](https://github.com/cw-acroteau/codeql-action/pull/1) in there just so workflows are executed. This information might be useful for future contributors.

cw-acroteau:
All checks are green in my Draft, so as soon as running workflows is approved I think this can go forward.

pr-checks/checks/remote-config.yml
```diff
@@ -5,7 +5,7 @@ steps:
     with:
       tools: ${{ steps.prepare-test.outputs.tools-url }}
       languages: cpp,csharp,java,javascript,python
-      config-file: github/codeql-action/tests/multi-language-repo/.github/codeql/custom-queries.yml@${{ github.sha }}
+      config-file: ${{ github.repository }}/tests/multi-language-repo/.github/codeql/custom-queries.yml@${{ github.sha }}
```
aeisenberg, line 8: Did you change this so that you can run the CI jobs on your fork?

src/actions-util.ts
```diff
@@ -33,10 +33,10 @@ export function getRequiredInput(name: string): string {
  * This allows us to get stronger type checking of required/optional inputs
  * and make behaviour more consistent between actions and the runner.
  */
-export function getOptionalInput(name: string): string | undefined {
+export const getOptionalInput = function (name: string): string | undefined {
```
aeisenberg, line 36: Is there a reason for this change?

aeisenberg (commented):
This is looking really good. I'm running the CI checks again. Hopefully, they will all pass. Just one minor question inline, but it's not a blocker.

aeisenberg:
Everything looks good, but the new CI jobs are failing. I'm digging in to try to figure out why.

benmoss:
[sendStatusReport](https://github.com/github/codeql-action/blob/f1ca3385da22ea923e9ee86cf6d1452b5cd21879/src/actions-util.ts#L616) doesn't respect `TEST_MODE` like some other functions do, not sure why it is getting called here but it appears that the test doesn't have permissions to do a `PUT` against https://api.github.com/repos/github/codeql-action/code-scanning/analysis/status

pr-checks/checks/remote-config.yml
cw-acroteau: Yes, that's why. Other tests have this construct, so they could run in a fork, but not this one.

src/actions-util.ts
cw-acroteau: This was for unit tests, I could not stub a "function" with sinon. If there's a better way we can change it back to function.

cw-acroteau:
> [sendStatusReport](https://github.com/github/codeql-action/blob/f1ca3385da22ea923e9ee86cf6d1452b5cd21879/src/actions-util.ts#L616) doesn't respect `TEST_MODE` like some other functions do, not sure why it is getting called here but it appears that the test doesn't have permissions to do a `PUT` against https://api.github.com/repos/github/codeql-action/code-scanning/analysis/status

Should I add a TEST_MODE for sendStatusReport or something?

aeisenberg:
In the nonofked version, I see that the [permissions block](https://github.com/cw-acroteau/codeql-action/runs/4956659159?check_suite_focus=true#step:1:14) is more permissive, which explains why it is passing there.

aeisenberg:
Now things are starting to become clear. If I explicitly add `security-events: write` then the failing workflows will pass. The difference between the failing workflows and the passing ones is that the failing ones use a value for the `ref` property of the upload status report that is different from that of the PR. Upload status report is an internal API endpoint for telemetry into code scanning. I don't know if it is a bug that the endpoint sometimes accepts requests from user agents that do not have `security-events: write` permissions. I'm communicating with the team that maintains this part of the API.

Regardless, I think our test workflows _should_ have this permission anyway. Here is what I propose:

1. I will create a new PR that will add the `security-events: write` permission to all of our generated workflows.
2. Can you rebase your PR, @cw-acroteau, on top of the one I will create? And make sure to regenerate the two integration test workflows you created. Also, make sure to fix the merge conflict in `CHANGELOG.md`.

Thanks for helping me find this bug (or at least odd inconsistency) in the action.

cw-acroteau:
> 2. Can you rebase your PR, @cw-acroteau, on top of the one I will create? And make sure to regenerate the two integration test workflows you created. Also, make sure to fix the merge conflict in `CHANGELOG.md`.

Hi! I just did that, hope everything is fine. I'm waiting to confirm that the worflows will pass, if there's an issue I'll try to fix it. Thanks!

cw-acroteau:
Hi again @aeisenberg ! Can you please check if the checks are okay in a copy similar to https://github.com/github/codeql-action/pull/902 please? It fails here because of "RequestError [HttpError]: Resource not accessible by integration", but it might be due to the fork.

cw-acroteau:
Oh sorry I did a `git rebase`, and not changed the base branch. I thought this was what you meant.

aeisenberg:
No problem. I wasn't clear about it. I changed the branch and it all looks good. Yes, the reason why the new workflows are failing is because the security-events permission is always downgraded to `read` for forks. I don't think it will be possible to get it to pass from a fork.  I'm in discussion with the API maintainers to figure out if we should avoid uploading status reports for integration tests. This should allow your integration tests to pass.

aeisenberg:
They are in Europe, so I probably won't hear anything until tomorrow.

aeisenberg:
PRs from forks will always have their `security-events` permission downgraded to `read`. This makes sense because we don't want malicious PRs to come around and tamper with the security alerts for refs they are not associated with.

There is an exception to this. a PR from a fork can write security events to resources associated with that PR only. This allows the PR to report security concerns about itself without giving it the ability to overwrite other refs.

What this means is that using the standard PR workflow, these new `ref` and `sha` inputs will not be able to work in forks. We will need to document this accordingly.

It looks like there are ways around this using [`pull_request_target` ](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) or possibly [Send write tokens to workflows from pull requests[(https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks).  I haven't tried out either of these solutions yet, but they will probably work. However, there are added security considerations with this.

What this means is that we need to rebase on the latest tip of the `aeisenberg/permissions` in this PR (I'll do that shortly) and update the documentation of the inputs to point out the limitation with forks. This PR can then be merged. Later, we can  work on some proper documentation on how to use these inputs in forks.

aeisenberg:
I created a PR in your repo for these changes. If you are happy with them, please merge and then I can merge here.

aeisenberg:
This PR should not have been closed. It should have been retargeted to main. Let me see what happened.

aeisenberg:
@cw-acroteau I'm very sorry, but could you create a new PR targeting main? For some reason when #902 was merged, this PR was closed. Instead it should have been retargeted against main.

cw-acroteau:
> @cw-acroteau I'm very sorry, but could you create a new PR targeting main? For some reason when #902 was merged, this PR was closed. Instead it should have been retargeted against main.

@aeisenberg Please see #904

PR #907, opened 2022-02-03
Report ML-powered query enablement in the `init` status report

henrymercer:
Adds information about whether ML-powered JS queries are enabled to the `init` status report.  This will improve our observability by allowing us to compare metrics between runs with ML-powered JS queries on and off.

Specifically, we add a `ml_powered_js_queries` field to the `init` status report that contains:

- The version string if the analysis will use a specific version of the pack
- "latest" if the analysis will use the latest version of the pack
- "false" if the analysis won't run any ML-powered JS queries, or if the analysis will run multiple ML-powered JS query packs

We report this information in the `init` status report rather than the `analyze` status report so we can gather data about timeouts.

Example status report:

```jsonc
{
  "workflow_run_id": 1790730464,
  "workflow_name": "Code Scanning - Action",
  "job_name": "CodeQL-Build",
  "analysis_key": ".github/workflows/codeql-analysis.yml:CodeQL-Build",
  "commit_oid": "6a99ab511b076566d0b9259a4102ac7f5c4b3e0c",
  "ref": "refs/heads/main",
  "action_name": "init",
  "action_ref": "1cddec9558c522dd219349d2149a48c46e203a8a",
  "action_oid": "unknown",
  "started_at": "2022-02-03T17:08:30.984Z",
  "action_started_at": "2022-02-03T17:08:30.984Z",
  "status": "success",
  "completed_at": "2022-02-03T17:08:37.758Z",
  "matrix_vars": "null",
  "disable_default_queries": "",
  "languages": "javascript",
  "ml_powered_js_queries": "~0.0.2" /* NEW */,
  "paths": "",
  "paths_ignore": "",
  "queries": "security-extended",
  "tools_input": "latest",
  "tools_resolved_version": "20220120",
  "workflow_languages": "javascript"
}
```


src/util.ts
```diff
@@ -627,3 +627,28 @@ export function checkNotWindows11() {
...
+ *
+ * This function lives here rather than in `init-action.ts` so it's easier to test, since tests for
+ * `init-action.ts` would each need to live in their own file. See `analyze-action-env.ts` for an
+ * explanation as to why this is.
+ */
+export function getMlPoweredJsQueriesStatus(config: Config) {
+  const mlPoweredJsQueryPacks = (config.packs.javascript || []).filter(
+    (pack) => pack.packName === ML_POWERED_JS_QUERIES_PACK_NAME
+  );
+  return mlPoweredJsQueryPacks.length === 1
```
aeisenberg, line 651: It's probably an error if `ML_POWERED_JS_QUERIES_PACK_NAME` is specified more than once. Is this handled elsewhere? Otherwise maybe this function should throw, or at least issue a warning.

src/util.test.ts
```diff
@@ -291,3 +292,51 @@ async function mockStdInForAuthExpectError(
...
+  [[{ packName: util.ML_POWERED_JS_QUERIES_PACK_NAME }], "latest"],
+  [
+    [{ packName: util.ML_POWERED_JS_QUERIES_PACK_NAME, version: "0.0.2" }],
+    "0.0.2",
+  ],
+  [
+    [
+      { packName: "someOtherPack" },
+      { packName: util.ML_POWERED_JS_QUERIES_PACK_NAME },
+    ],
```
aeisenberg, line 307: Should include a test where `ML_POWERED_JS_QUERIES_PACK_NAME` appears more than once.

aeisenberg (commented):
Just a minor thing, otherwise looks good.

edoardopirovano:
I believe you also need a PR such as https://github.com/github/hydro-schemas/pull/1845 to add the new field to the database and also a `github/github` PR to propagate the fields?

src/util.ts
henrymercer: Good call.  I don't think the status report code is the right place to throw or issue a warning, but come to think of it, we probably want to modify the code in `config-utils.ts` that adds the ML-powered query pack such that it only adds the pack if the user hasn't already specified it.

henrymercer:
> I believe you also need a PR such as [github/hydro-schemas#1845](https://github.com/github/hydro-schemas/pull/1845) to add the new field to the database and also a `github/github` PR to propagate the fields?

Thanks, these are in the works (see the internal tracking issue)!

src/config-utils.ts
```diff
@@ -289,12 +289,18 @@ async function addBuiltinSuiteQueries(
...
-  // the ML-powered queries.
+  // If we're running the JavaScript security-extended analysis (or a superset of it), the repo is
+  // opted into the ML-powered queries beta, and a user hasn't already added the ML-powered query
+  // pack, then add the ML-powered query pack so that we run ML-powered queries.
   if (
     languages.includes("javascript") &&
     (found === "security-extended" || found === "security-and-quality") &&
+    !(
+      packs.javascript &&
+      packs.javascript.some(
```
aeisenberg, line 300: nit:

```suggestion
      packs.javascript?.some(
```

src/util.ts
```diff
@@ -627,3 +627,35 @@ export function checkNotWindows11() {
...
+  const mlPoweredJsQueryPacks = (config.packs.javascript || []).filter(
+    (pack) => pack.packName === ML_POWERED_JS_QUERIES_PACK_NAME
+  );
+  switch (mlPoweredJsQueryPacks.length) {
+    case 1:
+      return mlPoweredJsQueryPacks[0].version || "latest";
+    case 0:
+      return "false";
+    default:
+      return "multiple";
```
aeisenberg, line 659: I'm pretty sure that the action will happily run each version independently (but I haven't tried this out to be sure). Would it make more sense to do this:

```suggestion
      return mlPoweredJsQueryPacks.map(p => p.version).join(,);
```

Case 1 and the default could be combined.

henrymercer: We want to keep the cardinality of this field down.  In practice we can go a bit further than this — it would be sufficient to just capture a fixed set of version strings (i.e. those included in `addBuiltinSuiteQueries` in `config-utils.ts`.  I've pushed a commit that makes this change, and adds some comments explaining (a) that we're keeping the cardinality low and (b) why this is.

aeisenberg (approved):

PR #917, opened 2022-02-08
Fix CI jobs that broke as a result of `windows-latest` being upgraded to Windows Server 2022

henrymercer:
`windows-latest` is currently migrating from an image based on Windows Server 2019 to an image based on Windows Server 2022.  This breaks some of our CI jobs for two reasons:

1. Build tracing doesn't yet support Windows Server 2022
2. Python 2 isn't supported in the `windows-2022` image.

This PR addresses (1) by moving relevant jobs to use the `windows-2019` image and (2) by fixing a typo that prevented `setup-python` from downloading and installing Python 2 when it's not already in the Actions VM image.

### DIFF
--- a/.github/workflows/__analyze-ref-input.yml
+++ b/.github/workflows/__analyze-ref-input.yml
@@ -30,10 +30,7 @@ jobs:
         - cached
         - latest
         - nightly-latest
-        os:
-        - ubuntu-latest
-        - macos-latest
-        - windows-latest
+        os: [ubuntu-latest, macos-latest, windows-2019]
     name: "Analyze: 'ref' and 'sha' from inputs"
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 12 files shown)


henrymercer:
@github/codeql-action-reviewers Since this PR changes a few jobs from `windows-latest` to `windows-2019`, we'll need to modify the required checks before we merge this.

henrymercer:
I thought we'd already done a changelog note for Windows Server 2022, but seemingly we only did one for the CLI.  I agree we should do a changelog and update the readme, but I suggest we do so in a separate PR so we can get the checks fixed on `main` and unblock other PRs.

aeisenberg (approved):
That's fine by me. A changelog note as well as making a note in the README in a separate PR. Thanks for getting this done.

henrymercer:
Required checks updated.

PR #938, opened 2022-02-17
Respect `end-tracing` script instead of deleting one variable

edoardopirovano:
This _might_ fix an issue we've observed with the CLR tracing continuing to trace after we've supposedly stopped tracing. It also just seems the more correct thing to do anyway. The variable that was previously deleted (`ODASA_TRACER_CONFIGURATION`) is always unset by `end-tracing`, so we should only be doing more to try and halt tracing.


criemen (approved):
LGTM from my side, but a JS person should review as well.
I checked that for C#, `end-tracing.json` indeed contains (besides others)
```json
"COR_ENABLE_PROFILING":null,"COR_PROFILER":null,"COR_PROFILER_PATH_64":null,"CORECLR_ENABLE_PROFILING":null,"CORECLR_PROFILER":null
```

src/tracer-config.ts
```diff
@@ -20,6 +20,30 @@ const CRITICAL_TRACER_VARS = new Set([
...
+      "utf8"
+    )
+  );
+  for (const [key, value] of Object.entries(endTracingEnvVariables)) {
+    if (value !== null) {
+      process.env[key] = value;
+    } else {
+      delete process.env[key];
+    }
+  }
```
aeisenberg, line 44: Can you wrap this in a try-catch and rethrow with a good error message if the file doesn't exist or is malformed?

aeisenberg (commented):
Small suggestion, but I think it is important to help avoid potential problems with users' workflows.

edoardopirovano:
Thanks @aeisenberg, I've added some more error handling!

I note that it would be nice to have this for the _start_ of tracing as well. The natural thing to do would be to have a utility function that can parse either file and contains all the error handling. Unfortunately, it's not quite that easy because the types of the return value are different as the end tracing file can contain `null` values to unset environment variables whereas the start tracing one cannot, so we'd have to change the types of quite a few things all around the code that use the start tracing file.

I propose leaving that extension for a future PR since that seems to be growing the scope of this PR too much.

aeisenberg (approved):

PR #948, opened 2022-02-24
Revert "Merge pull request #869 from github/edoardo/windows-11-error"

edoardopirovano:
This reverts commit faa9ba73632fb4da04c4029e886c783459dbcc66, reversing changes made to 71112ab35de01202ecda3b2722caa7982547335e.

The last few versions of the CodeQL CLI have emitted errors about tracing incompatibility with Windows 11, so there is no need for the Action to continue to emit these. In addition, the CodeQL CLI will soon be compatible with Windows 11, so if we still had the error here in place when that happens we would be stopping users from using CodeQL on Windows 11 for no good reason.


criemen (approved):
LGTM from my side (but I'm not an actions reviewer, so maybe get a review from them, too)

henrymercer (commented):
Should we continue to raise errors for users using old versions of CodeQL with build tracing that's incompatible with Windows 11?

criemen:
I thought about this, but I don't think it's necessary:
These users would have to pin an old version of the CLI in their workflow, but use the codeql-action from `v1`/`main`. I think that's sufficiently uncommon to not keep this code around, as it's pure tech debt for us.
Especially as the code is relying on a hand-written list of which languages are traced and which are not, without querying the CLI.

edoardopirovano:
I'm inclined to agree. We already have so many version flags in the Action, if we can avoid adding another one that seems preferable.

henrymercer:
Makes sense 👍

henrymercer (approved):

PR #955, opened 2022-02-28
Delete the results directory in between runs.

chrisgavin:
The default results directory of the CodeQL Action is in a non-ephemeral directory. This means it will be persisted between runs. The CodeQL Action also does not keep track of what languages it has produced results for in this run. That means that if a language is present in one run, but absent in the next, the CodeQL Action will upload it regardless.

By deleting the results directory we ensure that stale results will never be used.

I couldn't find a particularly good place to add a test for this (which isn't that surprising, since it's very much an implementation detail), but I've tested manually and verified it's correctly no longer uploading stale results.


aeisenberg:
Thanks for creating this PR. I'll take a look shortly. It is likely that this fix will also address https://github.com/github/code-scanning/issues/5318

src/analyze.ts
```diff
@@ -423,7 +423,15 @@ export async function runFinalize(
     delete process.env[sharedEnv.ODASA_TRACER_CONFIGURATION];
   }
 
-  fs.mkdirSync(outputDir, { recursive: true });
+  // After switching to Node16, this entire block can be replaced with `await fs.promises.rm(outputDir, { recursive: true, force: true });`.
+  try {
+    await fs.promises.rmdir(outputDir, { recursive: true });
```
aeisenberg, line 428: On windows, this might be finicky. I usually add retries and retry delay option when recursively deleting directories on windows.

https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_rmdir_path_options_callback

```suggestion
    await fs.promises.rmdir(outputDir, { recursive: true, maxRetries: 5, retryDelay: 2000 });
```

The values above are pretty arbitrary, but might be good enough to avoid random errors.

chrisgavin: Makes sense, since Windows file locking can be pretty annoying. Interestingly `retryDelay` seems to only have been added in NodeJS v12.16.0, so I don't know how safe it is to depend on.

aeisenberg: Good point. I just verified locally, and it looks like node v10 will silently ignore `retryDelay` if it's passed.

chrisgavin: Hmm I can add it, but it's not recognized by TypeScript's typing declarations.

aeisenberg (approved):
👍🏼

ladykraken (commented):
cast

Issue #944, opened 2022-02-21
github/codeql-action/upload-sarif@v1 Failed to call git to get current commit, silently fails to upload the SARIF file and returns a false success status


HariSekhon:
I've just discovered that if `github/codeql-action/upload-sarif@v1` fails to get the git commit (relates to #58), it doesn't raise an error.

This is very bad for security scanning tools like static code analysis, because it likely means that nothing was scanned but gives a false positive success status for the workflow.

Output:

```
Run github/codeql-action/upload-sarif@v1
fatal: not a git repository (or any of the parent directories): .git
Failed to call git to get current commit. Continuing with data from environment or input: Error: The process '/usr/bin/git' failed with exit code 128
Error: The process '/usr/bin/git' failed with exit code 128
    at ExecState._setResult (/home/runner/work/_actions/github/codeql-action/v1/node_modules/@actions/exec/lib/toolrunner.js:592:25)
    at ExecState.CheckComplete (/home/runner/work/_actions/github/codeql-action/v1/node_modules/@actions/exec/lib/toolrunner.js:575:18)
    at ChildProcess.<anonymous> (/home/runner/work/_actions/github/codeql-action/v1/node_modules/@actions/exec/lib/toolrunner.js:469:27)
    at ChildProcess.emit (events.js:[21](https://github.com/HariSekhon/Templates/runs/5267697089?check_suite_focus=true#step:6:21)0:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Socket.<anonymous> (internal/child_process.js:4[30](https://github.com/HariSekhon/Templates/runs/5267697089?check_suite_focus=true#step:6:30):11)
    at Socket.emit (events.js:210:5)
    at Pipe.<anonymous> (net.js:659:12)
Uploading results
  Processing sarif files: ["results.sarif"]
  Uploading results
  Successfully uploaded results
```

Here is an example public workflow of mine showing the problem:

https://github.com/HariSekhon/Templates/runs/5267697089?check_suite_focus=true

I've corrected the error in my workflow so newer versions of this workflow do work (I missed the checkout step when porting to a reusable workflow for some reason), but the point is that the `github/codeql-action/upload-sarif@v1` action should fail when it detects that it is not in a git repo unless an optional flag is given to permit this condition as it will often be a serious error in the workflow, causing nothing to be scanned.

Any error in determining the git commit should result in an error code being propagated upwards and failing the entire workflow for safety to detect when you are not receiving security alerts due to not being in a checkout.

[...]

PR #956, opened 2022-02-28
Use the `checkout_path` for getting the commit oid

### LINKED ISSUE #952: `upload-sarif` action seems to always use `GITHUB_WORKSPACE` Git information to compute `commit_oid`
Uploading SARIF files using the `upload-sarif@v1` action can fail with:

```
Processing sarif files: ["scan.sarif"]
Uploading results
Error: commit not found
RequestError [HttpError]: commit not found
```

The problem here seems to be that the `commit_oid` parameter in the uploaded JSON is from an entirely different repository: the repository that's the "main" checkout (the one with empty `path` parameter to `actions/checkout@v2`), even with `checkout_path` specified.

<details>
<summary>Full log output</summary>

```
 Processing sarif files: ["jenkins-security-scan.sarif"]
 Uploading results
 Error: commit not found
 RequestError [HttpError]: commit not found
     at /home/runner/work/_actions/github/codeql-action/v1/node_modules/@octokit/request/dist-node/index.js:66:23
     at processTicksAndRejections (internal/process/task_queues.js:93:5)
     at async Job.doExecute (/home/runner/work/_actions/github/codeql-action/v1/node_modules/bottleneck/light.js:405:18) {
   name: 'HttpError',
   status: 404,
   headers: {
     'access-control-allow-origin': '*',
     'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',
     connection: 'close',
     'content-encoding': 'gzip',
     'content-security-policy': "default-src 'none'",
     'content-type': 'application/json; charset=utf-8',
     date: 'Fri, 25 Feb 2022 10:29:15 GMT',
     'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
     server: 'GitHub.com',
     'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
     'transfer-encoding': 'chunked',
     vary: 'Accept-Encoding, Accept, X-Requested-With',
     'x-content-type-options': 'nosniff',
     'x-frame-options': 'deny',
     'x-github-media-type': 'github.v3; format=json',
     'x-github-request-id': '0403:3D83:83138D:1507C61:6218AF7B',
     'x-ratelimit-limit': '1000',
     'x-ratelimit-remaining': '995',
     'x-ratelimit-reset': '1645785350',
     'x-ratelimit-resource': 'code_scanning_upload',
     'x-ratelimit-used': '5',
     'x-xss-protection': '0'
   },
   request: {
     method: 'PUT',
     url: 'https://api.github.com/repos/daniel-beck-org/sample-plugin/code-scanning/analysis',
     headers: {
       accept: 'application/vnd.github.v3+json',
       'user-agent': 'CodeQL-Action/1.1.3 octokit-core.js/3.1.2 Node.js/12.13.1 (linux; x64)',
       authorization: 'token [REDACTED]',
       'content-type': 'application/json; charset=utf-8'
     },
     body: '{"commit_oid":"92e0b0945a6334eeeb4c65a78c2be5a7767e3cc9","ref":"refs/heads/main","analysis_key":".github/workflows/jss.yaml:scan","analysis_name":"Jenkins Security Scan","sarif":"...","workflow_run_id":1898119060,"checkout_uri":"file:///home/runner/work/sample-plugin/sample-plugin","environment":"null","started_at":"2022-02-25T10:29:14.509Z","tool_names":["Jenkins Security Scan"]}',
     request: { agent: [Agent], hook: [Function: bound bound register] }
   },
   documentation_url: 'https://docs.github.com/rest'
 }
```

</details>

If no repository is checked out at this location (i.e. every `actions/checkout@v2` has a `path`), then the error is the same as in #944 (but at least the upload still happens, at least for a non-PR upload).

aeisenberg:
Ensure that repos checked out in a non-standard location can still be uploaded.

Here is a sample PR in a different repo that uses the `checkout_path`: https://github.com/aeisenberg/actions-test2/pull/4/checks

I created an integration test that validates that `checkout_path` is doing what it should. Looking at the logs, I can see that the test is doing the right thing, but I'm not able to get the test to fail when [`checkout_path` is not set](https://github.com/aeisenberg/actions-test2/pull/4/checks). So, this still needs a bit of work. Ideally, we would upload the sarif file and verify that the commit_oid is correct, but our tests cannot actually upload any sarif, so that makes checking a little bit harder.

**EDIT**: I think I have it. See the new integration test.

When running in test mode, payloads will not be uploaded. Instead, they
will be saved to disk so that they can be inspected later.

This change will also require a documentation fix.

I'm not entirely happy with this since to get this working, the init action requires a `source-path` input and the analyze action requires a `checkout_path`. But we can specify this in the doc changes.

Fixes #952


aeisenberg:
Still trying to figure out how to verify that the integration test is passing (other than by visually inspecting the log). We need to validate that the commitOid of the payload that _would_ have been invoked is the same as that of the second checkout. One possibility is that I can change the `determineMergeBaseCommitOid` and `getCommitOid` methods to write a file to disk with the payload parameters if the action is in debug mode. And the test can validate these parameters after the upload step.

pr-checks/checks/with-checkout-path.yml
```diff
@@ -0,0 +1,75 @@
...
+      path: x/y/z/some-path
+  - uses: ./../action/init
+    with:
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
+      # it's enough to test one compiled language and one interpreted language
+      languages: java,javascript
+      source-path: x/y/z/some-path/tests/multi-language-repo
+      debug: true
+  - name: Build code (non-windows)
+    shell:
```
cklin, line 22: Is this intentional?  Similarly for `Build code (windows)`.

pr-checks/checks/with-checkout-path.yml
```diff
@@ -0,0 +1,75 @@
...
+    if: ${{ !startsWith(matrix.os, 'windows-') }}
```
cklin, line 23: An alternative is to check `runner.os`.
https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context

pr-checks/checks/with-checkout-path.yml
aeisenberg: Hmmm...it should be `bash`. Here and below.

pr-checks/checks/with-checkout-path.yml
aeisenberg: Yes. That's true. It's cleaner to use runner.os. I'll make the change.

aeisenberg (commented):
Thanks for your review!

cklin (approved):

aeisenberg:
Fixes https://github.com/github/codeql-action/issues/944

aeisenberg:
@criemen I fixed the hang we were seeing by changing from java to csharp for `with-checkout-path.yml`. Not happy with that, and maybe we should take a deeper look into this.

aeisenberg:
Merging this PR since the previously hanging integration tests are unrelated.

PR #957, opened 2022-03-01
Second attempt to use `--codescanning-config` flag of CLI

edoardopirovano:
Reverts https://github.com/github/codeql-action/pull/936, thus re-introducing the changes from https://github.com/github/codeql-action/pull/830.

The previous attempt to roll this out failed due to some users having invalid properties in their config files and the CLI version of the parsing not being tolerant to unknown fields. This was fixed in 2.8.2, so we can try this again now.

The first commit is a clean revert of the revert. The second commit bumps the version flag so we don't try to use the CLI to parse the config file if it is a version from before 2.8.2 which contained the aforementioned fix.


aeisenberg (approved):
Similar to previous PR, which I approved. Looks fine to me.

PR #959, opened 2022-03-02
Add runner OS and arch to status report

cklin:

This PR adds the OS and architecture (as reported by the GitHub Action Runner environments `RUNNER_OS` and `RUNNER_ARCH`) to the status report.


src/actions-util.ts
```diff
@@ -643,6 +647,9 @@ export async function createStatusReportBase(
       workflowStartedAt
     );
   }
+  const runnerOs = process.env["RUNNER_OS"] || "";
+  const runnerArch = process.env["RUNNER_ARCH"] || "";
```
aeisenberg, line 651: We can expect that these env vars are available always. This is the pattern we are using elsewhere to get env vars.

```suggestion
  const runnerOs = getRequiredEnvParam("RUNNER_OS");
  const runnerArch = getRequiredEnvParam("RUNNER_ARCH");
```

aeisenberg (commented):
I'm pretty sure status reports will fail to upload because of the new fields until we can make the server side changes.

aeisenberg (approved):
Nice. Just be sure to get the hydro schema changes deployed before merging this.

PR #960, opened 2022-03-03
Introduce results-limit to fail instead of ignoring excessive results

andreaTP:
Hi, and thanks for this cool Action!

We want to use it in a relatively big repository and we found [this note](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning):

> SARIF upload supports a maximum of 5000 results per upload. Any results over this limit are ignored.

We want to enable rules over time and we want to have an hard limit instead of risking to have results to be silently ignored.

I have opened the PR directly instead of discussing over an Issue since the task seemed to be easy enough 🙂  .
This is my very first contribution to this repo, so, please, feel free to give me guidance on how to, possibly, move this PR ahead if we agree that this is a desired feature.


upload-sarif/action.yml
```diff
@@ -30,6 +30,9 @@ inputs:
     description: If true, the Action will wait for the uploaded SARIF to be processed before completing.
     required: true
     default: "false"
+  wait-for-processing:
```
aibaars, line 33: This property is already defined above.

aeisenberg:
Thank you for your PR. 

The 5000 result limit is set on the server and is not something we should be changing. I can see the value of adding a hard failure before uploading if the number of results > 5000. Can you tell me the use case for failing if the number is <= 5000?

upload-sarif/action.yml
andreaTP: fixed! thanks!

andreaTP:
Hi @aeisenberg ,
thanks a lot for the review!

I'm happy to change this PR to have `results-limit: boolean` and hardcode the limit to 5000 if this is the desired direction 🙂 .
WDYT?

andreaTP:
ok, went ahead and converted `results-limit` to a `boolean`, let me know if this works for you!

aeisenberg:
After discussing this with our team, it seems like using `wait-for-processing` will get you most of what you need. The only benefit with using `results-limit` is that you are notified slightly sooner that there is a problem. However, this difference is small enough that it is not worth adding a new input to support it.

We are deciding not to merge this PR. I'm happy to discuss more if you feel like there is something we've missed here.

andreaTP:
Thanks for explaining this @aeisenberg , can you confirm that setting `wait-for-processing` the upload will fail if the number of results goes over the 5K limit?
Is there any documentation I can look up to get this detail?

Thanks a lot in advance!

andreaTP:
Just found an additional comment:
https://github.com/github/codeql-action/blob/b1c781d3983c105067e87232b504af53d61f803b/CHANGELOG.md?plain=1#L44

Where can I find more information about?

aeisenberg:
I discussed this with our team internally. Currently, the `wait-for-processing` flag will not work on forks. Apologies for not noticing that earlier. This is something that is being actively worked on by the code scanning team. I can't make any promises, but I do expect a fix in the next week or so.

andreaTP:
Hi @aeisenberg !

Thanks for the answer, I think it's a good idea to open an issue to track this down for the community 🙂 

You can go ahead and scope it properly or I'll try to do my best on Monday 👍

aeisenberg:
Issue is here: https://github.com/github/codeql-action/issues/963

andreaTP:
Thanks 👍

PR #987, opened 2022-03-16
Include CodeQL CLI and action versions in status reports

cklin:
This PR extends status reports to include `codeql-action` version and CodeQL CLI version.  As part of the change it also moves CodeQL CLI version caching to `util.ts` to avoid introducing a circular dependency.


src/api-client.ts
```diff
@@ -85,7 +85,7 @@ let cachedGitHubVersion: GitHubVersion | undefined = undefined;
  */
 export async function getGitHubVersionActionsOnly(): Promise<GitHubVersion> {
   if (!util.isActions) {
-    throw new Error("This getGitHubVersion() function works only in an action");
+    throw new Error("getGitHubVersionActionsOnly() works only in an action");
```
henrymercer, line 88: Not related to this PR, but don't we need to call the function, as in `util.isActions()`?

src/codeql.ts
```diff
@@ -997,6 +999,14 @@ async function getCodeQLForCmd(
       await new toolrunner.ToolRunner(cmd, args).exec();
     },
   };
+  // To ensure that status reports include the CodeQL CLI version whereever
+  // possbile, we want to call getVersion(), which populates the version value
+  // used by status reporting, at the earliest opportunity. But invoking
+  // getVersion() directly here breaks tests that only pretend to create a
+  // CodeQL object. So instead we rely on the assumption that all non-test
+  // callers would set checkVersion to true, and util.codeQlVersionAbove()
```
henrymercer, line 1007: This seems OK, but we should add JSDoc to the constructor explaining that all non-test callers should set `checkVersion` to true.

cklin: That is a good suggestion.  Done.  Thank you!

PR #992, opened 2022-03-23
Status reporting: fix codeql_version field name

cklin:
This PR fixes the `StatusReportBase` field name for CodeQL CLI version.  It should be `codeql_version` instead of `codeql_cli_version`.


adityasharad (approved):
Have we got PR checks that would catch incorrect fields by failing the status reporting?

cklin:
> Have we got PR checks that would catch incorrect fields by failing the status reporting?

I am not sure if there is a way to do that.  I think the status reporting API endpoint is specifically configured to accept unknown fields, which is generally a good thing.  That does make it hard to catch field name errors—I found this one only because I was digging into the status report data and noticing that the CodeQL CLI version is always empty.

PR #995, opened 2022-03-24
Update release process to maintain both v2 and v1 releases

henrymercer:
This PR implements a new release process to maintain both v2 and v1 releases of the CodeQL Action, both of which receive updates to the analysis and other features from `main`.

The new release process is discussed in the updated `CONTRIBUTING.md` document.  We perform the v2 release via a `main -> v2` merge, and mergeback `v2 -> main` as we did before with the `v1` branch.  However what's new is that we also perform a `v2 -> v1` merge to ensure that the `v1` branch stays up to date with the latest changes while we need to continue supporting this release.

Please read the backlinked internal issue for information on how this was tested, a link to the design doc, and followup work. Commit-by-commit review recommended.

## Handling the `v2 -> v1` merge

The most difficult part of this process, and one that is liable to change in the future, is how we handle conflicts between the `v2` and `v1` branch during the `v2 -> v1` merge.  For instance, take the state just before the `v2 -> v1` merge while we're backporting a new 2.1.7 release to v1:

```
v2: commits from 2.1.6 --- mergeback patch version bump to 2.1.7 --- new changes to be released in 2.1.7
v1: commits from 2.1.6 --- change version number to 1.1.6
```

We need to address the conflict between `mergeback patch version bump to 2.1.7` and `change version number to 1.1.6`.  However, this is complicated, as the version number appears in three places (`package.json`, `package-lock.json`, and `node_modules/.package-lock.json`) and these files could conflict in other ways (e.g. different package versions between v1 and v2).

Our approach is to create a release prep branch from `v1` and revert the "change version number to 1.1.6" commit:

```
v1-release-prep: commits from 2.1.6 --- change version number to 1.1.6 --- revert "change version number to 1.1.6"
```

By reverting this commit, the release prep branch will have the 2.1.6 version, which means the "mergeback patch version bump to 2.1.7" commit from v2 can be merged in unproblematically.  Once this merge is done, we then update the version number to 1.1.7:

```
v1-release-prep: commits from 2.1.6 --- change version number to 1.1.6 --- revert "change version number to 1.1.6" --- merge v2 into v1-release-prep --- change version number to 1.1.7
```


.github/update-release-branch.py
```diff
@@ -212,19 +239,76 @@ def main():
...
+    print('Reverting the 1.x.x version number and changelog updates from the last release to avoid conflicts')
+    v1_update_commits = run_git('log', '--grep', '^Update version and changelog for v', '--format=%H').split()
+
+    if len(v1_update_commits) > 0:
+      print(f'  Reverting {v1_update_commits[0]}')
+      # Only revert the newest commit as older ones will already have been reverted in previous
+      # releases.
+      run_git('revert', v1_update_commits[0], '--no-edit')
+
+      # Also revert the "Update checked-in dependencies" commit created by Actions.
```
aeisenberg, line 262: Ah...I see...`update-release-branch.yml` is triggered on push to the v2 branch, and this ensures the version numbers are up to date. It also calls `update-release-branch.yml` from v2 to v1.

CONTRIBUTING.md
```diff
@@ -61,16 +61,22 @@ Here are a few things you can do that will increase the likelihood of your pull
...
-1. When the "Merge main into v1" pull request is merged into the `v1` branch, the "Tag release and merge back" workflow will create a mergeback PR.
-    This mergeback incorporates the changelog updates into `main`, tags the release using the merge commit of the "Merge main into v1" pull request, and bumps the patch version of the CodeQL Action.
+    Once you've checked off all but the last two of these, approve the PR and automerge it.
+1. When the "Merge main into v2" pull request is merged into the `v2` branch, the "Tag release and merge back" workflow will create a mergeback PR.
+    This mergeback incorporates the changelog updates into `main`, tags the release using the merge commit of the "Merge main into v2" pull request, and bumps the patch version of the CodeQL Action.
 
-    Approve the mergeback PR and automerge it. Once the mergeback has been merged into main, the release is complete.
+    Approve the mergeback PR and automerge it.
+1. When the "Merge main into v2" pull request is merged into the `v2` branch, the "Update release branch" workflow will create a "Merge v2 into v1" pull request to merge the changes since the last release into the `v1` release branch.
```
aeisenberg, line 74: Would it make sense at all to keep both the v2 and the v1 PRs open at the same time? This way if we needed to make a change to v1, we wouldn't need to make a new release.

.github/update-release-branch.py
```diff
@@ -80,43 +70,40 @@ def open_pr(repo, all_commits, short_main_sha, branch_name):
     body.append('')
     body.append('Contains the following commits not from a pull request:')
     for commit in commits_without_pull_requests:
-      body.append('- ' + commit.sha + ' - ' + get_truncated_commit_message(commit) + ' (@' + commit.author.login + ')')
+      author_description = ' (@' + commit.author.login + ')' if commit.author is not None else ''
+      body.append('- ' + commit.sha + ' - ' + get_truncated_commit_message(commit) + author_description)
 
   body.append('')
   body.append('Please review the following:')
   body.append(' - [ ] The CHANGELOG displays the correct version and date.')
```
cklin, line 78: Some of these checks seem to be relevant only for main-to-v2.  For example, there should be no need to match changelog to commits for v2-to-v1, as that should have already been done with main-to-v2.

Some of the checks, if we want to keep, might need tweaks.  For example the changelog version in v2-to-v1 will say 2.x.y, and that would be considered correct because the versions will be fixed up automatically later.

.github/update-release-branch.py
```diff
@@ -212,19 +239,76 @@ def main():
...
+    run_git('merge', f'{ORIGIN}/{args.source_branch}', '--no-edit')
+
+    # Migrate the package version number from a v2 version number to a v1 version number
+    print(f'Setting version number to {version}')
+    subprocess.run(['npm', 'version', version, '--no-git-tag-version'])
+    run_git('add', 'package.json', 'package-lock.json')
+
+    # Migrate the changelog notes from v2 version numbers to v1 version numbers
+    print('Migrating changelog notes from v2 to v1')
+    subprocess.run(['sed', '-i', 's/## 2\./## 1\./g', 'CHANGELOG.md'])
```
cklin, line 280: Please anchor the pattern to the beginning of line to reduce the chances of accidental match.

.github/workflows/post-release-mergeback.yml
```diff
@@ -15,6 +15,7 @@ on:
   push:
     branches:
       - v1
```
cklin, line 17: Should we remove the `v1` push trigger?  If I understand correctly the only mergeback is `v2` to `main`.

CONTRIBUTING.md
henrymercer: If we need to make a patch change to v1 that doesn't affect main or v2, we can do this by pushing to the "Merge v2 into v1" PR.  If the change needs to go into main or v2, then we would need to make a new release.

I'm not sure it makes sense to open a `v2 -> v1` PR before we've merged the `main -> v2` PR, since the v2 branch won't yet contain the necessary changes we want to pull in for the new release.  Or were you thinking about keeping different PRs open at the same time?

.github/workflows/post-release-mergeback.yml
henrymercer: `v2 -> main` is the only mergeback, however on a push `v1` we still want to tag the `v1` release.

.github/update-release-branch.py
```diff
@@ -179,23 +166,63 @@ def update_changelog(version):
...
+    help='GitHub token, typically from GitHub Actions.'
+  )
+  parser.add_argument(
+    '--repository-nwo',
+    type=str,
+    required=True,
+    help='The nwo of the repository, for example github/codeql-action.'
+  )
+  parser.add_argument(
+    '--source-branch',
```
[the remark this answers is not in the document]
henrymercer: I think this makes sense.  We try to be relatively general in the script, but I agree that exposing a smaller interface will likely make the script easier to maintain.

.github/update-release-branch.py
henrymercer: > Some of these checks seem to be relevant only for main-to-v2. For example, there should be no need to match changelog to commits for v2-to-v1, as that should have already been done with main-to-v2.

I think since the `v2 -> v1` merge is fairly complex, it would be good to continue to manually review the commits in the merge to check that nothing spurious has happened.  We could revisit this after a few release cycles when we've ironed out the kinks.

> Some of the checks, if we want to keep, might need tweaks. For example the changelog version in v2-to-v1 will say 2.x.y, and that would be considered correct because the versions will be fixed up automatically later.

At the point of opening the `v2 -> v1` PR, the changelog should have already been fixed up, so the changelog version should say 1.x.y.

Issue #996, opened 2022-03-25
CodeQL python setup fails when Python2 is not available


cdivitotawela:
There is a problem with https://github.com/github/codeql-action/blob/main/python-setup/install_tools.sh#L31 where the python setup fails when the runner does not have python2 installed. Expectation is that when python2 is not installed code execution is not go into the if condition. 

```sh
Installing collected packages: virtualenv-clone, pipenv
  Successfully installed pipenv-[202](https://github.com/ConsenSys/ethereum-prometheus-exporter/runs/5687749074?check_suite_focus=true#step:7:202)2.3.24 virtualenv-clone-0.5.7
  + + 
  command -v python2
  + + python2
  /runner/_work/_actions/github/codeql-action/v1/python-setup/install_tools.sh: 34: python2: not found
  curl --location --fail https://bootstrap.pypa.io/pip/2.7/get-pip.py
    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                   Dload  Upload   Total   Spent    Left  Speed
  
    0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
    0 1863k    0  4113    0     0   143k      0  0:00:12 --:--:--  0:00:12  143k
  curl: (23) Failed writing body (1354 != 1371)
```

PR #997, opened 2022-03-25
Fix python_setup/install_tool.sh when `python2` not present

cdivitotawela:
Fix for #996 

When the script  executed on a runner which does not have python2, script incorrectly execute the code inside the if condition. This script runs using /bin/sh in Ubuntu and cause the problem. Fix for this issue to use different syntax in output redirection.

Issue is reproduced using running the job on Ubuntu container. Fix also verified in the action https://github.com/cdivitotawela/codeql-issue/actions/runs/2038530172

Tested using action file https://github.com/cdivitotawela/codeql-issue/blob/main/.github/workflows/investigate.yaml

- [x] Confirm this change is backwards compatible with existing workflows. https://github.com/cdivitotawela/codeql-issue/actions/runs/2038530172


RasmusWL (approved):
Thanks. I guess I did not confirm that the `command -v` worked as intended back then. I have been able to reproduce with the code below (on Ubuntu 21.10)

```sh
#!/bin/sh

if command -v nonexistingcommand &> /dev/null; then
    echo "old true"
else
    echo "old false"
fi

if command -v nonexistingcommand > /dev/null 2>&1; then
    echo "new true"
else
    echo "new false"
fi
```

which outputs

```
old true
new false
```

I can't say I fully understand this (in particular what the old version actually did), but thanks for the fix :muscle:

edoardopirovano (approved):
Looks good to me too, thanks for the fix!

PR #1000, opened 2022-03-28
Run Actions on Node 16

henrymercer:
This PR starts running the `init`, `autobuild`, `analyze`, and `upload-sarif` Actions on Node 16.  This is a breaking change for customers using GitHub Connect or the Actions sync tool to pull in the latest version of the `v1` Action on older GHES versions (GHES 3.1–3.3), therefore we are bumping to a new major version with this change.

We'll need to revert these changes in the `v1` branch within the `v2 -> v1` merge during the next release process.


.github/update-release-branch.py
```diff
@@ -285,6 +285,9 @@ def main():
     print('Migrating changelog notes from v2 to v1')
     subprocess.run(['sed', '-i', 's/^## 2\./## 1./g', 'CHANGELOG.md'])
 
+    # Remove changelog notes from v2 that don't apply to v1
+    subprocess.run(['sed', '-i', '/^- The CodeQL Action now runs on Node\.js v16\./d', 'CHANGELOG.md'])
```
cklin, line 289: Do we need to run this command on every v2-to-v1 merge?

I am under the impression that the "CodeQL Action now runs on Node.js v16" changelog entry is only going to be in the first such merge, which will involve special manual handling anyway (as it involves reverting the node version back to 12).  Could this be done manually once as part of that special release process?

henrymercer: To avoid conflicts, we revert the commit on `v1` from the last release that updated the changelog notes, then migrate the changelog notes from scratch from the `v2` branch each time we generate the `v2 -> v1` merge (see the "Handling the `v2 -> v1` merge" section of https://github.com/github/codeql-action/pull/995).  As mentioned in the earlier PR, we're keen to replace this with something simpler if there's a way to avoid the merge conflicts without introducing more friction.

cklin: Here is what I envision happen the first time we make a v1 release from v2, using your proposed process but without this line.

When we run the process to release v1 from v2 for the first time, the v2-to-v1 merge PR would have the following commits (oldest first):
- Run Actions on Node 16
- Add PR check to validate compatibility with Node 12
- 1.1.6 (by github-actions[bot])
- main-to-v2 merge commit
- v2-to-v1 merge commit
- Update version and changelog for v1.1.6 (generated by update-release-branch.py)
  - The only change in this commit is to set 1.1.6 version numbers in package lock files and changelog

Of course we cannot merge the PR in this form because then v1 would be on node 16.  So you would perform some Git surgery so that "v2-to-v1 merge commit" (which has v1 HEAD and v2 HEAD as parents) include extra manual changes to keep v1 on node 12.  While you are at it, you also remove the "CodeQL Action now runs on Node.js v16" changelog entry.

So now v1 is what we expect it to be and we can release it.  For the next v1 release we will revert "Update version and changelog for v1.1.6", and that would not re-introduce the "CodeQL Action now runs on Node.js v16" changelog entry (because it was removed in "v2-to-v1 merge commit" for 1.1.6, not in "Update version and changelog for v1.1.6").

Does that match what you envision the process would be?

henrymercer: Ah, I think I understand what you meant now.

Perhaps I'm trying a bit too hard to avoid merge conflicts here, but my concern with the process that you suggest is that it involves the changelog diverging between `v1` and `v2` outside of the auto-generated "Update version and changelog for v1.1.6" commit.  As such when we revert this commit, the changelog doesn't return to a state that matches the state it was on an old commit on the `v2` branch.  Therefore we introduce the possibility for merge conflicts.

If we instead apply all the changelog differences in the auto-generated commit, we can revert this commit to get back to a changelog that is identical to an old state on `v2`.  I think by doing this we should be able to avoid any merge conflicts.

cklin (approved):

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The CodeQL Action now runs on Node.js v16. [#1000](https://github.com/github/codeql-action/pull/1000)
```
aeisenberg, line 5: I think you should specify that this is only for v2.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "codeql",
-  "version": "1.1.6",
```
aeisenberg, line 3: It is _a little_ confusing that the first version of the extension we are publishing is 2.1.6, but honestly, I think this is better than starting with 2.0.0 since we'd need to internally map between v1 versions and v2 versions.

aeisenberg (commented):
One small suggestion around the changelog.

package.json
henrymercer: Agreed.

PR #1001, opened 2022-03-28
Add PR check to validate compatibility with Node 12

henrymercer:
This PR adds a simple PR check that sets the `@types/node` dependency to version 12.12 (i.e. types for Node 12), then attempts to compile the Action.  This helps us guard against introducing new code that doesn't run with Node 12.

Example successful job: https://github.com/github/codeql-action/runs/5722503871?check_suite_focus=true
Example failed job after introducing some Node 16 only code: https://github.com/github/codeql-action/runs/5722718273?check_suite_focus=true

Note that we'll need to update the set of required checks after merging this PR.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -24,8 +24,31 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 45
 
+    strategy:
+      fail-fast: true
+      matrix:
+        node-types-version: [12.12, current]
+
     steps:
       - uses: actions/checkout@v2
+
+      - name: Update version of @types/node
+        if: matrix.node-types-version != 'current'
+        env:
+          NODE_TYPES_VERSION: ${{ matrix.node-types-version }}
+        run: |
+          export NODE_TYPES_VERSION="${NODE_TYPES_VERSION}"
+          contents=$(jq '.devDependencies."@types/node" = env.NODE_TYPES_VERSION' package.json)
+          echo "${contents}" > package.json
+          # Usually we need to run `npm install` on macOS, however we're not checking in the result
+          # here, so it's fine to run on Linux.
+          npm install
+
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions[bot]"
+          git add --all
+          git commit -am "Use @types/node=${NODE_TYPES_VERSION}"
+
       - name: Check generated JS
         run: .github/workflows/script/check-js.sh
 


.github/workflows/pr-checks.yml
```diff
@@ -24,8 +24,31 @@ jobs:
...
+
     steps:
       - uses: actions/checkout@v2
+
+      - name: Update version of @types/node
+        if: matrix.node-types-version != 'current'
+        env:
+          NODE_TYPES_VERSION: ${{ matrix.node-types-version }}
+        run: |
+          export NODE_TYPES_VERSION="${NODE_TYPES_VERSION}"
```
cklin, line 40: This command seems redundant.

.github/workflows/pr-checks.yml
```diff
@@ -24,8 +24,31 @@ jobs:
...
+          contents=$(jq '.devDependencies."@types/node" = env.NODE_TYPES_VERSION' package.json)
+          echo "${contents}" > package.json
+          # Usually we need to run `npm install` on macOS, however we're not checking in the result
```
cklin, line 43: It took me a while to fully understand this sentence.  Perhaps something like:

> Usually we run `npm install` on macOS to ensure that we picks up macOS-only dependencies. However ...

.github/workflows/pr-checks.yml
henrymercer: We need to export `NODE_TYPES_VERSION` so `jq` can pick up on it on the next line.  I'll add a comment to that effect.

aeisenberg (approved):

.github/workflows/pr-checks.yml
```diff
@@ -24,8 +24,33 @@ jobs:
...
       - uses: actions/checkout@v2
+
+      - name: Update version of @types/node
+        if: matrix.node-types-version != 'current'
+        env:
+          NODE_TYPES_VERSION: ${{ matrix.node-types-version }}
+        run: |
+          # Export `NODE_TYPES_VERSION` so it's available to jq
+          export NODE_TYPES_VERSION="${NODE_TYPES_VERSION}"
+          contents=$(jq '.devDependencies."@types/node" = env.NODE_TYPES_VERSION' package.json)
```
adityasharad, line 42: Do we need to quote this, to handle newlines or quotes within the JSON contents? Another option is to pipe it to a temporary file `jq '...' package.json > package.temp.json` and then copy that file into `package.json`.

.github/workflows/pr-checks.yml
```diff
@@ -24,8 +24,33 @@ jobs:
...
+          echo "${contents}" > package.json
+          # Usually we run `npm install` on macOS to ensure that we pick up macOS-only dependencies.
+          # However we're not checking in the updated lockfile here, so it's fine to run
+          # `npm install` on Linux.
+          npm install
+
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions[bot]"
+          git add --all
+          git commit -am "Use @types/node=${NODE_TYPES_VERSION}"
```
adityasharad, line 52: Do you need both `add --all` and `commit -a`? Also why do we need to commit these changes -- is there a check script that looks at Git state?

.github/workflows/pr-checks.yml
adityasharad: Another option is to pass the variable in to `jq`:
```bash
$(jq --arg NODE_TYPES_VERSION "$NODE_TYPES_VERSION" '.devDependencies."@types/node" = $NODE_TYPES_VERSION' package.json)
```

.github/workflows/pr-checks.yml
henrymercer: My understanding is that the right-hand side of an assignment doesn't need to be quoted in Bash.  That also works, though we would need a third command to delete the temporary file.

.github/workflows/pr-checks.yml
henrymercer: I think we need to run `commit -a` after `add --all` to ensure we stage any files that have been deleted.  Some further googling though reveals that `git add --all .` (with a dot) stages deleted files too.

.github/workflows/pr-checks.yml
cklin: Command substitution should not require quoting.

```bash
$ foo=$(printf "a\nb 'c' d\n\ne\n") ; echo "$foo"
a
b 'c' d

e
$
```

.github/workflows/pr-checks.yml
henrymercer: I'm ambivalent between the two here so will just go with the existing one.  Please shout if you have a preference.

.github/workflows/pr-checks.yml
adityasharad: Not that important, but so I understand: why do we need to stage and commit changes at all?

henrymercer: Ah sorry, I forgot to respond to that part of your comment. The reason is just because `check-js.sh` checks that the repo is clean before compiling the JS files.  (`check-js.sh` does this because it's designed such that it can also be used as a pre-push git hook).

adityasharad: Ok that makes sense, thanks!

PR #1010, opened 2022-03-30
Stop running ML-powered queries on Windows

henrymercer:
We have some evidence to suggest that some runs on Windows cause `CatastrophicError`s which abort the analysis.  This PR stops running ML-powered queries on Windows while we address this problem.


aeisenberg (approved):
That's too bad...looks reasonable to me.

PR #1012, opened 2022-03-30
Update major versions of Actions in README and workflows

henrymercer:
Update the README to use the latest major versions of `actions/checkout` and the Actions in this repo.  Also update our workflows to use the latest major versions of various first-party Actions.


cklin (approved):

cklin:
Thank you for the change!

A question just popped up in my mind.  Do we want these documentation changes in the v1 branch as well?  I can see arguments both ways...

henrymercer:
> Thank you for the change!
> 
> @henrymercer A question just popped up in my mind. Do we want these documentation changes in the v1 branch as well? I can see arguments both ways...

Good question.  For now I don’t think I see a good enough reason for the branches to diverge.  But very happy to reconsider.

PR #1039, opened 2022-04-22
python-setup: Check if `pip` is already installed for Python2

RasmusWL:
As mentioned in the code comment, double installing causes trouble when running the action twice for our internal testing (DCA)


cklin (approved):

python-setup/install_tools.sh
```diff
@@ -29,9 +29,16 @@ python3 -m pip install --user poetry!=1.0.10
...
-	# Setup Python 2 dependency installation tools.
-	# The Ubuntu 20.04 GHA environment does not come with a Python 2 pip
-	curl --location --fail https://bootstrap.pypa.io/pip/2.7/get-pip.py | python2
+	# Setup Python 2 dependency installation tools. The Ubuntu 20.04 GHA environment
+	# does not come with a Python 2 pip, but if it is already installed, don't try to
+	# install it again (since that causes problems).
+	#
+	# This might seem like a hypothetical situation, but it happens all the time in our
+	# internal testing where we run the action twice in a row.
+	if ! python2 -m pip --version; then
```
adityasharad, line 38: Curious -- how does this interact with `set -e` when the command fails?

cklin: I think bash exempts conditional checks (in `while`, `until`, `if`, `elif`, and so on) from `set -e` checks, so we should be ok here.

yoff (approved):
LGTM

PR #1050, opened 2022-04-28
Don't wait for processing in test mode

henrymercer:
In test mode, we don't upload results, so there's no point waiting for processing.  Disabling waiting for processing when we're in test mode speeds up most of our PR checks and removes Actions warnings from waiting for processing failing because we didn't upload any SARIF.


aeisenberg (approved):
Nice. Good catch.

PR #1051, opened 2022-04-28
Run ML-powered queries on Windows with CodeQL CLI 2.9.0+

henrymercer:
Versions 2.9.0+ of the CodeQL CLI run machine learning models from a temporary directory, which resolves an issue on Windows where TensorFlow models are not correctly loaded due to the absolute path of some of their files being greater than `MAX_PATH` (260 characters).

We should wait until all virtual environments including CodeQL CLI 2.9.0 have been fully rolled out before merging this PR to avoid alert churn from Code Scanning being run intermittently on the new and old bundles and therefore intermittently running with and without ML-powered queries on Windows. 


henrymercer:
Ready for review, but not ready to be merged yet per the PR description.

src/config-utils.test.ts
```diff
@@ -1807,42 +1807,64 @@ test(
...
+  "~0.2.0"
+);
+// Test that we don't inject an ML-powered query pack if the user has already specified one.
+test(
+  mlPoweredQueriesMacro,
+  "2.9.0",
+  true,
+  "codeql/javascript-experimental-atm-queries@0.0.1",
+  "security-and-quality",
+  process.platform === "win32" ? undefined : "0.0.1"
```
cklin, line 1869: I was not expecting this test to depend on whether the action runs on Windows, as the test specifies CLI version 2.9.0.  What am I missing?

henrymercer: That's a mistake.  Thanks!

henrymercer: (It might be a good idea to consider whether we want a CI job that runs our unit tests on Windows)

cklin (approved):

PR #1053, opened 2022-05-02
Add workflow to regenerate required checks

aeisenberg:
Update contributing guide. Ensure this workflow runs once a week.

### DIFF
--- a/.github/workflows/update-required-checks.yml
+++ b/.github/workflows/update-required-checks.yml
@@ -1,0 +1,43 @@
+
+# This job updates the required checks on the codeql-action repository based on the
+# checks performed on the most recent commit.
+
+name: Update required checks
+on:
+  schedule:
+    # 23:01 on Saturdays
+    - cron: "1 23 * * 6"
+  workflow_dispatch:
+
+jobs:
+  update-required-checks:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-required-checks.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Dump GitHub context
+      env:
+        GITHUB_CONTEXT: '${{ toJson(github) }}'
+      run: echo "$GITHUB_CONTEXT"
+
+    - name: Update checks
+      run: |
+        # Update the required checks based on the current branch.
+        # Typically, this will be main.
+        SHA="$(git rev-parse HEAD)"
```
cklin, line 28: I wonder whether this would really work (there is no checkout step in the workflow, and `HEAD` may not work in a single-commit checkout).  But I suppose we probably need to merge it before we can find out, as the branch protection rules probably would not allow repository updates from branches other than `main`.

cklin (approved):

.github/workflows/update-required-checks.yml
aeisenberg: Hmmm...now that I think about it, we can just get the SHA from the context. Let me change this.

PR #1057, opened 2022-05-05
Introduce a feature-flag to enable/disable lua-based tracing.

criemen:
This allows us to gradually roll out (or even roll back)
Lua-based tracing in case problems occur.

This should not be merged before we are sure that 2.9.2 is the right CLI version to target, but otherwise this PR can be reviewed already.


src/codeql.ts
```diff
@@ -732,6 +736,15 @@ async function getCodeQLForCmd(
           // because that always passes in a process name.
           extraArgs.push(`--trace-process-level=${processLevel || 3}`);
         }
+        if (
+          await util.codeQlVersionAbove(this, CODEQL_VERSION_LUA_TRACER_CONFIG)
+        ) {
+          if (await featureFlags.getValue(FeatureFlag.LuaTracerConfigEnabled)) {
+            extraArgs.push("--internal-use-lua-tracing");
+          } else {
+            extraArgs.push("--no-internal-use-lua-tracing");
```
henrymercer, line 745: Are we including this so we have the ability to roll back Lua tracing, even after we've made it on by default?

henrymercer (commented):
Generally looks good.  I think we should add some tests that the appropriate arguments get passed / not passed when we expect.  There are some examples of that for ML-powered queries in `config-utils.test.ts`.

henrymercer:
> I believe action rolls out new images per repository, so unless they roll back an image mid-rollout, a customer should not see their codeql version jumping back and forth.

If that's the case, that would simplify dealing with these types of PRs.  Do you have a source for this you could internally link me to?

criemen:
Rebased on top of main to address a conflict.
I'd like to merge this now, as 2.9.2 is indeed the right release. Rolling this out will have to wait once the new CLI is in the toolcache, though.

henrymercer (approved):

PR #1068, opened 2022-05-11
Run a couple more tests in test mode

henrymercer:
This reduces the number of API requests we make while running PR checks.


aeisenberg (approved):
I had no idea these tests were actually making network requests. Good catch.

henrymercer:
This isn't necessary as we already stub the status reporting code with sinon.

PR #1087, opened 2022-06-01
Run ML-powered query pack `~0.3.0` on v2.9.3+ of the CLI

TomBolton:
This PR starts running the following versions of the ML-powered query pack:

- Version `~0.3.0` of the query pack on v2.9.3+ of the CLI
- Version `~0.2.0` of the query pack on v2.8.4 <= CLI < v2.9.3
- Version `~0.1.0` of the query pack on CLI < v2.8.4


TomBolton:
@henrymercer is this how we would like to start running `~0.3.0`?

Specifically, I've assumed in this PR that if the CLI is below 2.9.3, but above 2.8.4, we use `~0.2.0`; otherwise we use `~0.1.0`. Is this correct?

aeisenberg (commented):
This looks generally sensible.

TomBolton:
This PR is _almost_ ready for another look, but one of the `Basic Checks and Runner` [tests is failing](https://github.com/github/codeql-action/runs/6864779785?check_suite_focus=true) with a duplicate test title and I'm not sure how to fix it, as the tests I added appear to be unique.

Screenshot 2022-06-13 at 17 28 41

aeisenberg (approved):
Looks reasonable, assuming all the tests pass.

src/config-utils.test.ts
```diff
@@ -1865,3 +1865,23 @@ test(
...
+// Test that the ~0.3.0 version of ML-powered queries are run on non-Windows platforms running
+// `security-extended` on version 2.9.3+ of the CodeQL CLI.
+test(
+  mlPoweredQueriesMacro,
+  "2.9.3",
+  true,
+  undefined,
+  "security-extended",
+  process.platform === "win32" ? undefined : "~0.3.0"
+);
```
henrymercer, line 1877: Hmm, this doesn't look right.  We should be running `~0.3.0` on _all_ platforms with v2.9.3+ of the CodeQL CLI.  We're not catching this in CI as we currently aren't running the unit tests on Windows — let's address that in a follow up PR.

henrymercer (changes requested):

src/config-utils.test.ts
TomBolton: I modified the test in [04a23ec](https://github.com/github/codeql-action/pull/1087/commits/04a23ec897a306bea2a38d44e8ef893829745138) - does this look right now?

henrymercer (approved):

henrymercer:
It looks like the multi-language repo checks are timing out.  Looking at [the history](https://github.com/github/codeql-action/actions/workflows/__multi-language-autodetect.yml) of this job, these checks are typically close to the timeout, so I think it'd be reasonable to bump the timeout here from 45 minutes to 60 minutes.

henrymercer:
I've raised the flaky tests internally.  It seems that another rerun has sorted things with one of the jobs, so I've kicked off another run of the other job too.

TomBolton:
Thanks for looking into it @henrymercer and doing the reruns!

Issue #1091, opened 2022-06-09
codeql-action/analyze@v2 - Deprecation warning from Node on use of `fs.rmdir`


vHeemstra:
When running the default CodeQL workflow, the log shows a deprecation warning from Node on the use of `fs.rmdir()`:

```
(node:1717) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```

See the log line [here](https://github.com/vHeemstra/fill-pot-po/runs/6804757597?check_suite_focus=true#step:5:682).

CodeQL version: `2.9.3`
GitHub Action: `github/codeql-action/analyze@v2`

PR #1095, opened 2022-06-13
Avoid use of rmdir

aeisenberg:
This is a deprecated method on node v16.

Fixes https://github.com/github/codeql-action/issues/1091


henrymercer (approved):
With this change we will no longer retry deleting the directory.  However this shouldn't be problematic since the Actions runtime environment isn't affected by the Windows ATP issues, and the runner no longer exists.

aeisenberg:
Yes, it's unclear if we ever actually needed it in the past.

PR #1096, opened 2022-06-14
Add the check-sarif action

aeisenberg:
Allows us to analyze and then check that certain queries were included
in the analysis and others were not.

This will be used in a future PR, but want to get this in first.


aeisenberg:
Still need to add a changelog note for this.

.github/check-sarif/index.js
```diff
@@ -0,0 +1,52 @@
...
+rules.forEach(rule => {
+  core.info(`${rule.id}: ${(rule.properties && rule.properties.name) || rule.name}`)
+})
+core.endGroup()
+
+core.startGroup('Full SARIF')
+core.info(JSON.stringify(sarif, null, 2))
+core.endGroup()
+
+function getInput(name) {
```
henrymercer, line 47: I think this is specifically for the comma-separated list of inputs, so therefore would suggest renaming to something like `getQueryIdsInput`

.github/check-sarif/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    description: |
+      Comma separated list of query ids that should be included in this SARIF file.
+
+  queries-not-run:
+    required: true
+    description: |
+      Comma separated list of query ids that should NOT be included in this SARIF file.
+
+runs:
+  using: node12
```
henrymercer, line 19:
```suggestion
  using: node16
```

.github/check-sarif/index.js
```diff
@@ -0,0 +1,52 @@
+'use strict'
```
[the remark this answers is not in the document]
aeisenberg: It would be challenging since we'd need to embed a compile step into the action, or run it using ts-node.

.github/check-sarif/action.yml
aeisenberg: I used node12 here since I wasn't sure if we ever needed to run this on older versions GHES. Do we ever run the action tests on it? If not necessary, then I can change this line.

henrymercer: We don't run the unit tests on GHES, and when we do, I expect we'll only want to run the `releases/v1` branch on versions of GHES that don't support Node 16.

henrymercer: We don't run the unit tests on GHES.  Perhaps when we do it'll be slightly easier to run this Action on `node12`, so I think it's OK to run on `node12` here.

.github/check-sarif/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Unexpected Queries
+const expectedQueriesNotRun = getInput('queries-not-run')
+
+const queriesThatShouldNotHaveRunButDid = expectedQueriesNotRun.reduce((acc, queryId) => {
+  if (rules.some(rule => rule.id === queryId)) {
+    acc.push(queryId)
+  }
+  return acc
+}, []);
```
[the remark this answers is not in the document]
aeisenberg: Sure

.github/workflows/expected-queries-runs.yml
```diff
@@ -0,0 +1,48 @@
...
+        output: ${{ runner.temp }}/results
+        upload-database: false
+        upload: false
+      env:
+        TEST_MODE: true
+
+    - name: Check Sarif
+      uses: ./../action/.github/check-sarif
+      with:
+        sarif-file:  ${{ runner.temp }}/results/javascript.sarif
```
henrymercer, line 46:
```suggestion
        sarif-file: ${{ runner.temp }}/results/javascript.sarif
```

.github/workflows/expected-queries-runs.yml
```diff
@@ -0,0 +1,48 @@
+name: Expected queries runs
```
henrymercer, line 1:
```suggestion
name: Check queries that ran
```

.github/check-sarif/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  sarif-file:
+    required: true
+    description: The sarif file to check
+
+  queries-run:
+    required: true
+    description: |
+      Comma separated list of query ids that should be included in this SARIF file.
+
+  queries-not-run:
```
henrymercer, line 13: These should maybe be "ran", but definitely a nit.

.github/check-sarif/index.js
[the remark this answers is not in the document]
henrymercer: Compiling it as part of the build seems like a sensible thing to do.  We don't have to do it now though — happy to investigate depending on how complex this workflow gets.

henrymercer (approved):

.github/check-sarif/index.js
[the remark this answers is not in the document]
aeisenberg: It does seem like overkill for a simple action like this. If this action gets more complex then I think it would make sense.

PR #1098, opened 2022-06-14
Add capability to filter queries

aeisenberg:
This change adds a `query-filters` property to the codeql-config file.

This property is an array of `exclude`/`include` entries for a query
suite. These filters are appended to the generated query suite files
and used to filter queries after they are selected.

A related change is that now, all pack references are run in a single
query suite, which has the query filters appended to them.


src/analyze.ts
```diff
@@ -392,32 +393,59 @@ export async function runQueries(
...
-        databasePath,
-        searchPath,
-        pack,
-        memoryFlag,
-        threadsFlag
-      );
+    // combine the list of packs into a query suite in order to run them all simultaneously.
+    const querySuite = packs
+      .map(convertPackToQuerySuiteEntry)
+      .concat(queryFilters as any[]);
```
henrymercer, line 407: Should `any[]` be `unknown[]`?

src/analyze.ts
```diff
@@ -392,32 +393,59 @@ export async function runQueries(
...
-  return queries.map((q: string) => `- query: ${q}`).join("\n");
+export function convertPackToQuerySuiteEntry(packStr: string) {
+  const pack = configUtils.parsePacksSpecification(packStr);
+  return {
+    qlpack: !pack.path ? pack.name : undefined,
+    from: pack.path ? pack.name : undefined,
+    version: pack.version,
+    query: pack.path?.endsWith(".ql") ? pack.path : undefined,
+    queries:
+      !pack.path?.endsWith(".ql") && !pack.path?.endsWith(".qls")
```
henrymercer, line 435: Let's hope nobody has a `queries.ql` directory :trollface:

.github/workflows/query-filters.yml
```diff
@@ -0,0 +1,97 @@
...
+      env:
+        TEST_MODE: true
+    - name: Check Sarif
+      uses: ./../action/.github/check-sarif
+      with:
+        sarif-file:  ${{ runner.temp }}/results/javascript.sarif
+        queries-run: js/zipslip
+        queries-not-run: js/path-injection
+    - name: Cleanup after test
+      run: rm -rf "$RUNNER_TEMP/results"
```
henrymercer, line 51: Could we avoid duplication here, perhaps with a composite workflow?  It would also be great to name the tests more descriptively, such that when they fail it's more immediately obvious what went wrong.

src/analyze.ts
aeisenberg: That's a type error. I'm just being a bit lazy here. I could specify the type of the return value of `convertPackToQuerySuiteEntry` and union that with `QueryFilter`. It just seemed like a bunch of extra typings that don't contribute anything to readability.

src/analyze.ts
aeisenberg: This would be from inside of a query suite. It's possible, but not particularly easy to figure out the filesystem location and check if this is a directory or a file. I think this is a rare enough occurrence that we don't need to account for it.

.github/workflows/query-filters.yml
aeisenberg: Do you mean running each test in a separate job? My hope was to run these in a single job so that we could theoretically add lots of them and avoid adding too many more jobs during each action run.  Or do you mean a composite action? That might work.

henrymercer: Yeah I agree with your rationale for running them in a single job.  I think if we use a composite action rather than a reusable workflow we should be able to run all these tests in a single job while improving code reuse a bit.

src/analyze.ts
henrymercer: I was mainly joking — the likelihood of hitting this doesn't really justify the IO cost.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add the ability to filter queries from a code scanning run by using the `query-filters` option in the code scanning configuration file.[#1098](https://github.com/github/codeql-action/pull/1098)
```
henrymercer, line 5:
```suggestion
- Add the ability to filter queries from a code scanning run by using the `query-filters` option in the code scanning configuration file. [#1098](https://github.com/github/codeql-action/pull/1098)
```

tests/multi-language-repo/.github/codeql/codeql-config-query-filters1.yml
```diff
@@ -0,0 +1,10 @@
+name: "CodeQL config 1"
```
henrymercer, line 1: Thanks for naming the tests in the workflow file.  We could potentially add those names here too.

tests/multi-language-repo/.github/codeql/codeql-config-query-filters3.yml
```diff
@@ -0,0 +1,35 @@
...
+  javascript:
+    - codeql/javascript-queries
+    - dsp-testing/codeql-pack1@1.0.0
+
+query-filters:
+# This should run js/path-injection and js/zipslip
+- include:
+    tags contain: external/cwe/cwe-022
+
+# Removes out js/path-injection
```
henrymercer, line 25:
```suggestion
# Removes js/path-injection
```

tests/multi-language-repo/.github/codeql/codeql-config-query-filters2.yml
```diff
@@ -0,0 +1,21 @@
...
+  javascript:
+    - codeql/javascript-queries
+    - dsp-testing/codeql-pack1@1.0.0
+
+query-filters:
+# This should run js/path-injection and js/zipslip
+- include:
+    tags contain: external/cwe/cwe-022
+
+# Removes out js/path-injection
```
henrymercer, line 15:
```suggestion
# Removes js/path-injection
```

tests/multi-language-repo/.github/codeql/codeql-config-query-filters1.yml
```diff
@@ -0,0 +1,10 @@
...
+
+query-filters:
+# This should run js/path-injection and js/zipslip
+- include:
+    tags contain: external/cwe/cwe-022
+
+# Removes out js/path-injection
```
henrymercer, line 8:
```suggestion
# Removes js/path-injection
```

.github/query-filter-test/action.yml
```diff
@@ -0,0 +1,52 @@
...
+        tools: ${{ inputs.tools }}
+        db-location: ${{ runner.temp }}/query-filter-test
+    - uses: ./../action/analyze
+      with:
+        output: ${{ runner.temp }}/results
+        upload-database: false
+        upload: false
+      env:
+        TEST_MODE: "true"
+    - name: Check Sarif
```
henrymercer, line 44: Nit: Rename to "SARIF" throughout

henrymercer (approved):
A few minor comments, but generally this looks good to me.

PR #1104, opened 2022-06-17
Remove toolcache decorator

henrymercer:
This decorator enabled us to use the functionality of the Actions toolcache within the runner too. Now that we no longer release the runner we no longer need it.


henrymercer:
Rebased to fix a merge conflict in one of the sourcemap files.

.github/workflows/script/check-js.sh
```diff
@@ -14,8 +14,8 @@ npm run-script build
 # Check that repo is still clean
 if [ ! -z "$(git status --porcelain)" ]; then
     # If we get a fail here then the PR needs attention
-    >&2 echo "Failed: JavaScript files are not up to date. Run 'npm run-script build' to update"
+    >&2 echo "Failed: JavaScript files are not up to date. Run 'rm -rf lib && npm run-script build' to update"
```
aeisenberg, line 17: minor: We should probably just move `rm -rf lib` into the npm build script itself.

aeisenberg (commented):
I sometimes run the runner locally in order to test out changes before pushing them. Does this PR essentially break the runner? (Even if so, it's not a reason _not_ to merge this change, but it's just something I'd like to be aware of.

.github/workflows/script/check-js.sh
henrymercer: Hmm, wouldn't that break incremental compilation though?

henrymercer: My understanding is `tsc --build` does some incremental stuff to speed itself up: https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript.  It's unfortunate `tsc` doesn't have support for deleting unneeded js/jsmap files, but I'd rather keep the distinction between a build and a clean build.  Personally I chain `npm run build` into other commands and end up running it fairly frequently, but only delete TypeScript files rarely.

aeisenberg (commented):
This would also be a good time to uninstall any unused dependencies.

Are these still being used?
```
import * as toolrunner from "@actions/exec/lib/toolrunner";
import * as io from "@actions/io";
import * as actionsToolcache from "@actions/tool-cache";
import * as safeWhich from "@chrisgavin/safe-which";
import del from "del";
import * as semver from "semver";
import { v4 as uuidV4 } from "uuid";
```

If not, then let's remove them from the node_modules. I guess it can be in a future PR, but let's not forget them.

henrymercer:
> This would also be a good time to uninstall any unused dependencies.

Good idea. I checked and these dependencies are all still being used.

src/codeql.ts
```diff
@@ -484,24 +469,25 @@ export async function setupCodeQL(
         logger.info(
           `Downloading CodeQL tools from ${codeqlURL}. This may take a while.`
         );
+
+        const dest = path.join(tempDir, uuidV4());
+        const finalHeaders = Object.assign(
+          { "User-Agent": "CodeQL Action" },
+          headers
+        );
```
aeisenberg, line 477: Was this header not being set previously? Is there any reason why you are creating a new object here and not setting `"User-Agent"` when you initially create the `headers` object?

eg-

```js
        const headers: OutgoingHttpHeaders = {
          accept: "application/octet-stream",
          "User-Agent": "CodeQL Action"
        };
```

aeisenberg (commented):
Minor comment. This can be merged without making the change, especially if there is a reason for doing this that I'm not seeing.

src/codeql.ts
henrymercer: This is just existing code being moved — we can take this as an opportunity to revisit it.  Since this PR has been open for a while, I'll merge without making the change since I'd like to get a confirmation that the toolcache isn't expecting this particular user agent.

PR #1110, opened 2022-06-21
Fix issue with required checks sync script

edoardopirovano:
This script didn't quite work on my machine because the update check that we want to filter out was called `update` and `jq` was matching in a case-sensitive way so did not filter it out with `contains("Update")`. Lets add another case to the filter to cover this, to avoid others hitting this issue.


aeisenberg (approved):
In the future we should reformat this line to make it easier to read. Maybe just by adding some new lines. Fine the way it is now.

PR #1116, opened 2022-06-22
Allow scans with packs for languages not being scanned

aeisenberg:
Previously, we were being too strict about checking that a pack's
language was being scanned. It was a failure if a pack language
was specified for a language not being scanned.


src/config-utils.ts
```diff
@@ -1168,7 +1168,8 @@ export function parsePacksFromConfig(
       throw new Error(getPacksInvalid(configFile));
     }
     if (!languages.includes(lang as Language)) {
-      throw new Error(getPacksRequireLanguage(lang, configFile));
+      // This particular language is not being analyzed in this run.
```
adityasharad, line 1171: I suggest a log message at least.

src/config-utils.ts
```diff
@@ -1168,7 +1168,8 @@ export function parsePacksFromConfig(
...
+      continue;
```
adityasharad, line 1172: This means that we won't fail on the pack with the wrong language, but we won't attempt to run the pack below either, right? A comment to that effect would be great.

src/config-utils.ts
aeisenberg: Done. Needed to pass the logger in from up the call chain.

Also, added a case where we actually do want to throw an error.

adityasharad: Would it make sense to have an integration test that resembles the customer setup you observed?

aeisenberg: Easiest thing to do is to augment the existing packaging tests to include other unused languages. I don't think we have any integration tests that are meant to fail (so we don't have the infrastructure check that an invalid language does indeed fail the analysis).

aeisenberg:
@adityasharad can I get another look at this PR from you? I'd like to get this in the next release.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- It is no longer an error to include an unscanned language in the `packs` block in the codeql configuration file. [#1116](https://github.com/github/codeql-action/pull/1116)
```
adityasharad, line 5: How's this?
```suggestion
- CodeQL query packs listed in the `packs` configuration field will be skipped if their target language is not being analyzed in the current Actions job. Previously, this would throw an error. [#1116](https://github.com/github/codeql-action/pull/1116)
```

adityasharad (approved):

PR #1120, opened 2022-06-24
Honor the Lua tracer FF for `database trace-command` invocations for scanned languages.

criemen:
In theory, a scanned language will not setup the build tracer, and so
shouldn't care about lua versus legacy tracing. However, `go` is a
special case where the autobuilder runs under the build tracer, that
then gets disabled immediately again, unless a special environment
variable is used.
Therefore, we need to thread through the feature flag to this
`database trace-command` invocation. For other scanned languages,
this should be a no-op, as no tracing is ever set up.

I couldn't find any tests for the existing functionality, so I couldn't extend them either. If anybody has some great ideas for how to test this change, I'm happy to write some tests for this, too.


src/analyze.ts
```diff
@@ -116,7 +117,8 @@ async function setupPythonExtractor(logger: Logger) {
 
 async function createdDBForScannedLanguages(
```
aeisenberg, line 118: Here's a fairly simple unit test you can create to test the FF.

1. change the signature of this function so that it accepts a `codeql` (and avoid creating the instance in the function). Note that the caller of this function already creates its own instance of `codeql`.
2. Change the visibility of this function to public.
3. Create a unit test that passes in a mock version of codeql. You can test on multiple "versions" and capture whether or not the `internal-use-lua-tracing` parameter is set.

We can pair on this if you like.

aeisenberg (commented):
Looks sensible. Though I think it shouldn't be too hard to create a unit test.

src/analyze.ts
aeisenberg: Hmmm...but the call to `setupPythonExtractor` might cause some problems.

aeisenberg: Ah...just don't use python as one of your test languages.

aeisenberg:
Another thing that I'm seeing is that these `Test Python Package Installation` tests are flaky. We should probably automatically retry them if they fail.

aeisenberg:
Oh...looks like this test is no longer flaky...it's just _failing_.

mikerjohnson27 (commented):
Looks good. All tests passed.

criemen:
I'll do so, thanks!

PR #1126, opened 2022-06-27
Python-Setup: run auto_install_packages.py with -B flag

aibaars:
This avoids creating a __pycache__ folder in the _actions folder, which
may cause file ownership problems on self-hosted runners
when run in a docker container.


aeisenberg:
Is there any performance improvements to creating a pycache folder? We _could_ use the `-X pycache_prefix=PATH` option instead to place them in the runner temp folder.

aeisenberg:
Ah...I see you already went over that in the internal issue. That option only exists in 3.8 or later. _If_ there is likely to be noticeable speedup when using the pycache, then we can first detect which python version we are using and only use the option for 3.8 or later.

I suspect that this is not the case, though and what you have is fine.

aeisenberg (approved):

aibaars:
> out of curiosity, why did you choose to revert the `PYTHONDONTWRITEBYTECODE=1` version of this fix?

I meant to mention that approach in the PR and ask whether it would be preferred or not. The drawback of setting the environment variable is that it affects all `python` invocations, not just the problematic one running the script in the setup-python action. I doubt it would cause any problems, but who knows.

PR #1132, opened 2022-07-01
Enable one-click debugging via the "Enable debug logging" option when re-running Actions jobs

henrymercer:
(As @adityasharad pointed out on the internal issue, it's actually three clicks, but you get the point :))

Users can now quickly debug a job that uses the CodeQL Action by re-running the job from the GitHub UI and selecting the "Enable debug logging" option.  Debugging mode is also automatically turned on if Actions step debugging is enabled on a repo by setting the `ACTIONS_STEP_DEBUG` secret to `true`.  This PR also brings all of our debugging functionality under the same banner, e.g. we print the CodeQL CLI debug logs when `debug: true`, not just when Actions step debugging is enabled.

Example run: the first run is the standard workflow, attempt 2 (i.e. the rerun) was triggered with the "Enable debug logging" option selected.


src/analyze-action.ts
```diff
@@ -260,7 +260,7 @@ async function run() {
       }
     }
 
-    if (core.isDebug() && config !== undefined) {
+    if (config !== undefined && config.debugMode) {
```
aeisenberg, line 263: minor:

```suggestion
    if (config?.debugMode) {
```

aeisenberg: Also, curious why `core.isDebug()` was removed? Is that overridden by `process.env["RUNNER_DEBUG"]`?

henrymercer: `core.isDebug()` checks for `process.env["RUNNER_DEBUG"] === 1`: https://github.com/actions/toolkit/blob/c5278cdd088a8ed6a87dbd5c80d7c1ae03beb6e5/packages/core/src/core.ts#L227, so this is subsumed by `config.debugMode`.

Come to think of it, we should call the toolkit method instead of looking at the environment variable directly — I'll make the change.

aeisenberg (approved):

PR #1139, opened 2022-07-12
Use concat instead of push around `listFolders`

aeisenberg:
This avoids stack overflows when using the spread operator on directories that have many, many children.


henrymercer (approved):
Looks good.  While we're here, perhaps it's worth auditing the rest of the spread operators in the codebase to make sure these aren't likely to cause a stack overflow.

aeisenberg:
I did a quick look and did not find any other likely dangerous locations.

PR #1145, opened 2022-07-18
Don't crash if we are unable to get a response from the feature-flag endpoint.

criemen:


edoardopirovano (approved):

src/feature-flags.ts
```diff
@@ -31,14 +31,21 @@ export class GitHubFeatureFlags implements FeatureFlags {
   ) {}
 
   async getValue(flag: FeatureFlag): Promise<boolean> {
-    const response = (await this.getApiResponse())[flag];
+    const response = await this.getApiResponse();
     if (response === undefined) {
+      this.logger.debug(
```
henrymercer, line 36: Could we add a test for this code path?  You can use `mockFeatureFlagApiEndpoint` to mock the API response as `undefined` (see examples in `feature-flags.test.ts`).

src/feature-flags.ts
```diff
@@ -31,14 +31,21 @@ export class GitHubFeatureFlags implements FeatureFlags {
...
+        `No feature flags API response for ${flag}, considering it disabled.`
+      );
+      return false;
+    }
+    const flag_value = response[flag];
```
henrymercer, line 41: Style nit: Use camel case variable names

henrymercer (approved):
A couple of suggestions but otherwise looks good.

henrymercer (approved):
Thanks!

PR #1149, opened 2022-07-21
autobuild: Update tests for C# on macOS

criemen:
- Add a test for the `autobuild` action using C# on macOS. This should work under SIP as long as the tracer is set up correctly, thanks to Actions' own mechanism. ~Without the `$CODEQL_RUNNER` prefix, the tracer will not be injected on MacOS, as we need the runner to circumvent SIP.~
- Fix the corresponding test for the Runner's `autobuild` step, which was failing. This test needs the `$CODEQL_RUNNER` prefix since the tracer is not yet active in the environment. The CLR tracer previously mitigated this problem, but no longer works with dotnet 6 in the virtual environment.

After releasing this change, the required checks in the branch protection has to be updated.


pr-checks/checks/autobuild-macos-dotnet.yml
```diff
@@ -0,0 +1,20 @@
+name: "Autobuild "
+description: "Tests that C# "
```
edoardopirovano, line 2: I think this line and the one above are incomplete.

edoardopirovano (approved):

src/codeql.ts
```diff
@@ -780,7 +780,12 @@ async function getCodeQLForCmd(
         "-Dmaven.wagon.http.pool=false",
       ].join(" ");
 
-      await runTool(autobuildCmd);
+      const runnerExe = process.env["CODEQL_RUNNER"];
```
adityasharad, line 783: Just to be sure: the tracer config step earlier injects this into the environment for future steps, not just `process.env`?

src/codeql.ts
```diff
@@ -780,7 +780,12 @@ async function getCodeQLForCmd(
...
+      if (runnerExe) {
+        await runTool(runnerExe, [autobuildCmd]);
+      } else {
+        await runTool(autobuildCmd);
```
adityasharad, line 787: A log message in this situation would be good.

pr-checks/checks/autobuild-macos-dotnet.yml
adityasharad, line 2: Unfinished text?

adityasharad (commented):
LGTM. Minor suggestions only.

pr-checks/checks/autobuild-macos-dotnet.yml
criemen: Yes, and the file name is wrong either.

src/codeql.ts
criemen: Yes, it ends up calling core.exportVariable which does exactly that.

aibaars:
I wonder why this change is necessary, and wouldn't workflows with custom build commands also need a similar tweak? The actions/runner already inserts an intermediate process, so CODEQL_RUNNER should not be needed. 

See also: https://github.com/actions/runner/pull/416

pr-checks/checks/autobuild-action.yml
```diff
@@ -0,0 +1,20 @@
+name: "autobuild-action"
+description: "Tests that the C# autobuild action works"
+versions: ["latest"]
+steps:
+  - uses: ./../action/init
+    with:
+      languages: csharp
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
+  - uses: ./../action/autobuild
```
adityasharad, line 9: Optional: Consider manually unsetting the CLR tracer environment variables here, to check this will keep working if we turn that off in future.

.github/workflows/__autobuild-macos-dotnet.yml
```diff
@@ -0,0 +1,64 @@
+# Warning: This file is generated automatically, and should not be modified.
```
adityasharad, line 1: I think you can delete this file now -- it has no entry in `pr-checks/checks` and so isn't regenerated, and I think it's effectively identical to the workflow above.

adityasharad (commented):
As discussed separately, we do not need to introduce the `CODEQL_RUNNER` prefix thanks to Actions' own protection mechanism, but we do want to keep the new PR check introduced here.

src/codeql.ts
```diff
@@ -780,12 +780,11 @@ async function getCodeQLForCmd(
         "-Dmaven.wagon.http.pool=false",
       ].join(" ");
 
-      const runnerExe = process.env["CODEQL_RUNNER"];
-      if (runnerExe) {
-        await runTool(runnerExe, [autobuildCmd]);
-      } else {
-        await runTool(autobuildCmd);
-      }
+      // We do not need to prefix $CODEQL_RUNNER here on macOS to bypass SIP,
```
aeisenberg, line 783: I feel like reading this comment 2 years from now will be confusing since it never directly explains why we ever needed to use the prefix here. Hopefully, in 2 years, the runner will be gone, but I think we can make this comment clearer to someone reading this code for the first time.

adityasharad: This is true -- the full explanation is in internal docs, but I will attempt to make this clearer.
Unfortunately this is a different notion of "runner" (a single executable packaged with the CLI, not the CodeQL Runner tool), which is not going away anytime soon.

adityasharad: Updated.

aeisenberg (approved):
Thanks. Clearer now.

adityasharad:
@aeisenberg failing test is fixed: https://github.com/github/codeql-action/runs/7547434836?check_suite_focus=true. I added an explicit step to source the tracing environment (apparently the Runner's autobuild command wasn't sufficiently loading the environment, or it was too late in the process tree).

The failing check is a 503 in a test that I haven't touched; I say we ignore it and get this merged.

aeisenberg:
Check is passing now. i'd say it is ok to merge.

PR #1152, opened 2022-07-27
CI: Add path filters to python-deps workflow

adityasharad:
No need to run this workflow on all PRs, only those that change the Python dependency installation mechanism.

Reviewer note: This will need an update to the branch protection rules, because these checks can no longer be required on all PRs.


aeisenberg (commented):
Maybe we should run this on a schedule as well (weekly?). And put in a workflow_dispatch trigger so we can run manually.

aeisenberg (approved):

PR #1158, opened 2022-07-28
CI: Move Runner checks into their own workflow

adityasharad:
Refactor the PR checks workflow into two workflows: PR checks and Runner checks.

This does not change the actual check jobs that are run.
It also does not change the expected check names (which rely only on the job name, not the workflow name).

This makes it easier to inspect workflow run summaries in the UI and to separately retry subsets of failed jobs in case of flakiness.

In future we will clean up the Runner checks, since this is a deprecated component.


aeisenberg (approved):

aeisenberg:
I think we just got unlucky with that check that timedout.

PR #1159, opened 2022-07-29
Move logs, SARIF, database bundle actions uploads to post: hooks

angelapwen:
Previously, even with debug mode on, if the `init` step failed we did not upload the appropriate Actions artifacts. This was because the artifacts were only uploaded in the `analyze` step. 

This change:
- moves the uploading of log files to a `post:` hook in the `init` step. Regardless of whether the entire workflow was successful or if any steps after `init` failed, we will upload whatever logfiles we have as an artifact.
- moves the uploading of database bundles to a `post:` hook in the `init` step. If the database has not been finalized, instead of running the CLI `database bundle` command, the action directly zips up everything in the database directory.
- moves the uploading of SARIF results (if they exist) to a `post:` hook in the `analyze` step. 

So far the change has been manually tested so that:
- when `init` fails, partial logs from `init` are uploaded as artifacts, and the partial database bundle is uploaded. 
- when `analyze` fails, logs from `init` and `analyze` are uploaded as artifacts; the partial database bundle is uploaded; and the SARIF upload succeeds if there is a SARIF file generated in the output folder of the action.
- on workflow success, all file types are successfully uploaded. 

Testing strategy:
- [x] unit tests for the scripts called by `post:` hooks
- [x] integration tests showing what happens when `init` and `analyze` fail; will write after #1160 is merged.


angelapwen:
I'm correctly hitting the case where the database isn't finalized now, and just need to implement the zipping of the partial database files. I see [documentation for the CLI command](https://codeql.github.com/docs/codeql-cli/manual/database-bundle/) but am still not exactly sure which files I'm looking to zip. I'll take a look at the CLI source code but would appreciate any pointers!

Also, I see I have a CI failure on debug artifact which looks legitimate, as that's the code I'm changing. It only failed on `ubuntu-latest, nightly-latest` though 👀

angelapwen:
I think perhaps splitting the PR check file into 2 jobs, one for running everything before checking artifact downloads, and one for just checking artifact downloads with a `needs` dependency on the former, might force things into the right order. Similar to the code blocks in the docs at https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions. I'll try to do this in a separate PR and see if it's at least a no-op without these new post: hook changes.

src/actions-util.ts
```diff
@@ -875,3 +876,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
 export function sanitizeArifactName(name: string): string {
   return name.replace(/[^a-zA-Z0-9_\\-]+/g, "");
 }
+
+// REVIEW: Not sure this is the best place for this function?
+export function doesDirectoryExist(dirPath: string) {
```
aeisenberg, line 881: Seems more appropriate to the `util.ts` module.

src/actions-util.ts
```diff
@@ -875,3 +876,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+  toUpload: string[],
+  rootDir: string,
+  artifactName: string
+) {
+  if (toUpload.length === 0) {
+    return;
+  }
+  let suffix = "";
+  const matrix = getRequiredInput("matrix");
+  if (matrix !== undefined && matrix !== "null") {
```
aeisenberg, line 900: Is `matrix === ''` ok? Probably not. This is more succinct and likely more correct:

```suggestion
  if (matrix) {
```

src/actions-util.ts
```diff
@@ -875,3 +876,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+    for (const entry of Object.entries(JSON.parse(matrix)).sort())
```
aeisenberg, line 901: What happens if `matrix` is not valid json? How is the error handled?

src/actions-util.ts
```diff
@@ -875,3 +876,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+      suffix += `-${entry[1]}`;
```
aeisenberg, line 902: Minor:

```suggestion
    for (const [, matrixVal] of Object.entries(JSON.parse(matrix)).sort())
      suffix += `-${matrixVal}`;
```

src/analyze-action-cleanup.ts
```diff
@@ -0,0 +1,56 @@
+import * as fs from "fs";
```
aeisenberg, line 1: As much as possible, this file should be small and do as little as possible. `uploadSarifDebugArtifact` and `run` should be moved to a separate module. This will make it easier to test it separately.

To make it even easier to unit test, you can pass in `actionsUtil.uploadDebugArtifacts` as a parameter to the `run` function. In a test, you can substitute a mock for this so you can test that uploading acts as you think it should without needing to do anything fancy with mocks and stubs.

src/init-action-cleanup.ts
```diff
@@ -0,0 +1,163 @@
...
+      );
+    } catch (error) {
+      core.info(
+        `Failed to upload database debug bundles for ${config.debugDatabaseName}-${language}: ${error}`
+      );
+    }
+  }
+}
+
+async function uploadLogsDebugArtifact(config: Config) {
```
aeisenberg, line 62: Same comment here about making this module as small as possible, and passing in `actionsUtil.uploadDebugArtifacts` as a function parameter.

src/actions-util.ts
angelapwen: Makes sense as I guess it has nothing to do with actions. Moved!

src/actions-util.ts
angelapwen: Thanks! I like that.

src/actions-util.ts
angelapwen: Hm — I actually didn't write this (just moved it to another section of the codebase) but that's a good question. I had thought that it was forced to be JSON from where it was declared in inputs:

https://github.com/github/codeql-action/blob/19d025e17ffd30b50d1638946eaed9fe971a85d4/init/action.yml#L14-L15

and 

https://github.com/github/codeql-action/blob/main/analyze/action.yml#L67-L68

but it seems like that's the default input and the user could theoretically have passed non-valid JSON. 

I am adding a `catch` block with an error message indicating that there was an error parsing the `matrix` input and the debug artifact will be uploaded without the `matrix` input in its name. It seems like the only thing this block of code does is add a suffix to the artifact name, for every value in the `matrix` input, so it should be able to still upload the artifact correctly (just without the appropriate suffix(es).

src/analyze-action-cleanup.ts
angelapwen: I was looking at other modules and it looked like `runWrapper()` and `run()` are usually declared in the same file (for the action being run). 

I can definitely move the `uploadSarifDebugArtifact` (and the other specific artifact upload methods) back to `actionsUtil` where I had them earlier. I moved them into the file where they were called because I thought it might be a bit odd to have a function in a util directory that's only called once. It does make sense that doing that would help with unit testing though!

angelapwen:
I've made most of the requested changes, and will work on the refactoring for unit tests. 

I also made the change to zip up the database bundle and it seems to be working as expected on a failed [run](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/runs/7608743902?check_suite_focus=true). 

The `core.info` logs state: 

```
db-javascript is not finalized. Uploading partial database bundle at /home/runner/work/_temp/codeql_databases/db-javascript-partial.zip...
```

and then the zip file is uploaded as part of the artifacts. I can't attach it here due to file size limits but here is the directory structure of the partial database bundle file when unzipped:
Screen Shot 2022-08-01 at 12 26 23 PM

src/analyze-action-cleanup.ts
angelapwen: I made this change but ended up passing `uploadSarifDebugArtifact` as a parameter to the `run` function rather than `uploadDebugArtifacts`, since `run` doesn't call `uploadDebugArtifacts` directly.

angelapwen:
I added `//TODO`s in locations where I am planning to unit test, with the description of what each unit test should do, to get feedback on the planned tests before/while I write them.

src/actions-util.ts
```diff
@@ -2,20 +2,32 @@ import * as fs from "fs";
 import * as os from "os";
 import * as path from "path";
 
+import * as artifact from "@actions/artifact";
 import * as core from "@actions/core";
 import * as toolrunner from "@actions/exec/lib/toolrunner";
 import * as safeWhich from "@chrisgavin/safe-which";
+import AdmZip from "adm-zip";
```
adityasharad, line 9: We already use the `zlib` library; I think we could reuse that in this situation rather than adding a dependency. One difference is that `zlib` produces `gzip`.

src/analyze-action-cleanup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import * as actionsUtil from "./actions-util";
+import { Config, getConfig } from "./config-utils";
+import { getActionsLogger } from "./logging";
+
+async function run(uploadSarifDebugArtifact: Function) {
+  const logger = getActionsLogger();
+
+  let config: Config | undefined = undefined;
+  config = await getConfig(actionsUtil.getTemporaryDirectory(), logger);
```
adityasharad, line 11: This should be enough, since you assign the variable right away.
```suggestion
  const config = await getConfig(actionsUtil.getTemporaryDirectory(), logger);
```

src/analyze-action-cleanup.ts
```diff
@@ -0,0 +1,34 @@
+import * as core from "@actions/core";
```
adityasharad, line 1: A top-level comment in this file would be good. Mention that it is the entry point for the `post`-hook of `analyze/action.yml`, which runs after the `analyze` action.

This makes me wonder whether we should call it `-post` rather than `-cleanup`, but that is bikeshedding of low importance :)

src/analyze-action-cleanup.ts
```diff
@@ -0,0 +1,34 @@
...
+  config = await getConfig(actionsUtil.getTemporaryDirectory(), logger);
+  if (config === undefined) {
+    throw new Error(
+      "Config file could not be found at expected location. Has the 'init' action been called?"
+    );
+  }
+
+  // Upload Actions SARIF artifacts for debugging
+  if (config?.debugMode) {
+    const outputDir = actionsUtil.getRequiredInput("output");
```
adityasharad, line 20: Consider printing a log message here, since we've just made a choice. Something like "Running in debug mode, will upload SARIF from <outputDir> as an Actions debugging artifact".

src/init-action-cleanup.ts
```diff
@@ -0,0 +1,43 @@
+import * as core from "@actions/core";
```
adityasharad, line 1: Similar: top-level doc comment explaining what this is the entry point for.

src/init-action-cleanup.ts
```diff
@@ -0,0 +1,43 @@
...
+
+async function run(
+  uploadDatabaseBundleDebugArtifact: Function,
+  uploadLogsDebugArtifact: Function,
+  uploadFinalLogsDebugArtifact: Function
+) {
+  const logger = getActionsLogger();
+
+  let config: Config | undefined = undefined;
+  config = await getConfig(actionsUtil.getTemporaryDirectory(), logger);
```
adityasharad, line 15:
```suggestion
  const config = await getConfig(actionsUtil.getTemporaryDirectory(), logger);
```

src/init-action-cleanup.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (config === undefined) {
+    throw new Error(
+      "Config file could not be found at expected location. Has the 'init' action been called?"
```
adityasharad, line 18:
```suggestion
      "Config file could not be found at expected location. Did the 'init' action fail to start?"
```

src/init-action-cleanup.ts
```diff
@@ -0,0 +1,43 @@
+import * as core from "@actions/core";
+
+import * as actionsUtil from "./actions-util";
+import { Config, getConfig } from "./config-utils";
+import { getActionsLogger } from "./logging";
+
+async function run(
+  uploadDatabaseBundleDebugArtifact: Function,
```
adityasharad, line 8: Are these `Function`'s so that we can pass stub versions for testing?

src/actions-util.ts
```diff
@@ -875,3 +887,174 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+      await uploadDebugArtifacts(
+        [databaseBundlePath],
+        config.dbLocation,
+        config.debugArtifactName
+      );
+      continue;
+    }
+    try {
+      // Otherwise run `codeql database bundle` command.
+      const toUpload: string[] = [];
```
adityasharad, line 1040: You only have one path, so no need for a list. Something like `const bundlePath = await bundleDb(...)` followed by await uploadDebugArtifacts([bundlePath], ...)` should do the trick.

src/actions-util.ts
```diff
@@ -875,3 +887,174 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+    }
+  }
+}
+
+export async function uploadDatabaseBundleDebugArtifact(
+  config: Config,
+  logger: Logger
+) {
+  for (const language of config.languages) {
+    if (!dbIsFinalized(config, language, logger)) {
```
adityasharad, line 1014: Extract this then-block into its own private function. Also comment why we need this -- because `codeql database bundle` does not support non-finalised DBs.

This makes it easier to see the two possible code paths for bundling.

src/actions-util.ts
```diff
@@ -875,3 +887,174 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+        } else if (entry.isDirectory()) {
+          walkLogFiles(path.resolve(dir, entry.name));
+        }
+      }
+    };
+    walkLogFiles(logsDirectory);
+  }
+}
+
+export async function uploadLogsDebugArtifact(config: Config) {
```
adityasharad, line 965: I am struggling to see how the `uploadLogs` and `uploadFinalLogs` functions differ -- can you explain the difference briefly? If they have to be separate, I would extract as much shared code as possible into helper functions, to make the difference more obvious.

src/actions-util.ts
```diff
@@ -875,3 +887,174 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
 export function sanitizeArifactName(name: string): string {
   return name.replace(/[^a-zA-Z0-9_\\-]+/g, "");
 }
+
+export async function uploadDebugArtifacts(
```
adityasharad, line 891: For my understanding, these are all moved from `analyze-action.ts`, and the only new logic is the alternative code path for zipping a non-finalised DB?

src/actions-util.ts
```diff
@@ -875,3 +887,174 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
 export function sanitizeArifactName(name: string): string {
   return name.replace(/[^a-zA-Z0-9_\\-]+/g, "");
 }
+
```
adityasharad, line 890: Minor: There is enough code here that I would consider having a separate utility file just for debug or debug-upload functions. Can be done in this PR or in a follow-up if you're making future changes in this area.

adityasharad (commented):
Good work! The overall logic looks sensible -- suggestions mostly about clarity and readability.

Let me know when you've added some tests and would like another look.

src/analyze-action-cleanup.ts
angelapwen: Makes sense, thanks!

src/analyze-action-cleanup.ts
angelapwen: ✅ that makes sense. I've added the comment to the `post: init` entry point as well, and renamed everything to `post` (was thinking something similar, uploading artifacts doesn't really constitute cleaning up 🤔)

src/init-action-cleanup.ts
angelapwen: Yes, @aeisenberg suggested that so I thought I'd give it at try 😸

src/actions-util.ts
angelapwen: Ah yes, I see! I copied this over without looking too closely. Changed.

src/actions-util.ts
angelapwen: Makes sense, thanks!

src/actions-util.ts
angelapwen: Yes, that's correct!

src/actions-util.ts
angelapwen: That makes sense — have made that change now. I named the file `debug-artifacts` but could bikeshed about whether `util` is needed in the name 🚲

src/init-action-cleanup.ts
angelapwen: Done — thanks! (and in the `post: analyze` as well)

src/init-action-cleanup.ts
angelapwen: Done!

src/analyze-action-cleanup.ts
angelapwen: Sounds good — I added a log message here as well as in the equivalent section of `init-action-post`.

src/actions-util.ts
angelapwen: I actually named them based off of the location in code in `analyze-action` they were in before. Upon taking a closer look I definitely misnamed them 😆 

`uploadFinalLogs` is actually more like `printDebugLogs` — it basically prints the entirety of the debug logs to the Actions output window. I've renamed it now and moved it back to `actions-util` as it's not really a debug artifact. 

I wonder if it's really necessary to both upload them to artifacts and print them? I guess a user would expect detailed logs directly in the Actions console, as well as in debug artifacts..

src/actions-util.ts
angelapwen: Ah interesting. I will give this a try!

angelapwen: Ah, I remembered that I had some trouble with `zlib` when I tried it out earlier because it seems to only compress streams of data, and it makes preserving the directory structure difficult. I looked around 👀 and see another package that can create zipped directories using `zlib` for compression: https://github.com/archiverjs/node-archiver 

But as this would also add a new dependency I'm not sure if it's worthwhile 🤔

angelapwen:
Quickly closing and re-opening to trigger new PR checks from #1160

angelapwen:
I am working on integration testing with a new PR check (Debug artifact upload after failure) and have a few thoughts:

- with the `init` step, I was previously only simulating failure by throwing an exception in the code itself. I tried to add `with: ram: 1` to the step but this seemed to be enough memory to pass the `init` step.
- with the `analyze` step, [I successfully fail the step mid-way with the `ram: 1` test, and the "Download and check" job successfully passes after I pass the `continue-on-error` flag to the "Upload" job](https://github.com/github/codeql-action/runs/7727023187?check_suite_focus=true) (otherwise it won't even run the Download job). But this still causes a red X in the PR checks which is likely very confusing to anyone else opening a pull request. The [Actions team suggests not making this PR check required](https://github.com/actions/toolkit/issues/399#issuecomment-607450398), but I'm not sure if that's worthwhile because the integration test could fail for another reason and we wouldn't necessarily know. 

Would appreciate thoughts!

adityasharad:
> with the init step, I was previously only simulating failure by throwing an exception in the code itself. I tried to add with: ram: 1 to the step but this seemed to be enough memory to pass the init step.

Let's not worry about this for now. We could try to convince `init` to fail by giving it malformed input arguments or config file, but I think `init` itself failing is an extreme case. We can focus for now on handling failures that occur after `init`. 

> with the analyze step, [I successfully fail the step mid-way with the ram: 1 test, and the "Download and check" job successfully passes after I pass the continue-on-error flag to the "Upload" job](https://github.com/github/codeql-action/runs/7727023187?check_suite_focus=true) (otherwise it won't even run the Download job). But this still causes a red X in the PR checks which is likely very confusing to anyone else opening a pull request. The https://github.com/actions/toolkit/issues/399#issuecomment-607450398, but I'm not sure if that's worthwhile because the integration test could fail for another reason and we wouldn't necessarily know.

Good observation. We don't have existing integration tests for expected failures, for exactly this reason.
A solution that works with branch protection is to make the new `Download and check debug artifacts` check required, but the `Upload debug artifacts` checks optional. This ensures PRs are blocked only if the debug artifact check is actually broken. However this does not save us from the ugly red X for `Upload debug artifacts`.

angelapwen:
> Let's not worry about this for now. We could try to convince `init` to fail by giving it malformed input arguments or config file, but I think `init` itself failing is an extreme case. We can focus for now on handling failures that occur after `init`.

Okay, that sounds good. 

> Good observation. We don't have existing integration tests for expected failures, for exactly this reason. A solution that works with branch protection is to make the new `Download and check debug artifacts` check required, but the `Upload debug artifacts` checks optional. This ensures PRs are blocked only if the debug artifact check is actually broken. However this does not save us from the ugly red X for `Upload debug artifacts`.

Yes, that makes sense. At least we could make sure that the download job succeeds. I could change the PR check for `Upload debug artifacts` to include [Failure Expected] or something.

angelapwen:
Screen Shot 2022-08-09 at 4 56 23 PM
What future PR checks will see. Hopefully this gets the message across? Maybe `Failure Intended` would be even more clear 😄

adityasharad:
That looks good. The other option I considered was turning this workflow off on PRs, and only running it on push events. Let's keep this for now, and if the "failing" check confuses other contributors we can adjust it.

adityasharad (commented):
Some minor suggestions while you polish up the tests.

src/actions-util.ts
```diff
@@ -873,37 +875,31 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+      if (entries.length === 0) {
+        core.info(`No debug logs found at directory ${logsDirectory}.`);
+      }
+      for (const entry of entries) {
+        if (entry.isFile()) {
+          core.startGroup(`CodeQL Debug Logs - ${language} - ${entry.name}`);
+          process.stdout.write(fs.readFileSync(path.resolve(dir, entry.name)));
+          core.endGroup();
+        } else if (entry.isDirectory()) {
+          walkLogFiles(path.resolve(dir, entry.name));
```
aeisenberg, line 899: Minor: I don't see the full path of the file names being printed anywhere, so maybe include a log entry for each directory that you read.

aeisenberg: Or maybe include the full path on line 895.

src/analyze-action-post.ts
```diff
@@ -0,0 +1,42 @@
...
+ * other `post:` hooks.
+ */
+import * as core from "@actions/core";
+
+import * as actionsUtil from "./actions-util";
+import { getConfig } from "./config-utils";
+import * as debugArtifacts from "./debug-artifacts";
+import { getActionsLogger } from "./logging";
+
+async function run(uploadSarifDebugArtifact: Function) {
```
aeisenberg, line 13: For easier testing, you can move this function to a different file and test that file instead. This way, you avoid invoking `runWrapper` by default.

src/debug-artifacts.ts
```diff
@@ -0,0 +1,172 @@
...
+        config.tempDir,
+        config.debugArtifactName
+      );
+    }
+  }
+}
+
+/**
+ * If a database has not been finalized, we cannot run the `codeql database bundle`
+ * command in the CLI because it will return an error. Instead we directly zip
```
aeisenberg, line 118: That's unfortunate and avoids a good use case for `codeql database bundle`. Do you know why it throws? I wonder if we could add a `--force` option (or something similar) to allow the command to succeed even if the database is malformed.

src/util.ts
```diff
@@ -756,3 +756,34 @@ export async function checkActionVersion(version: string) {
...
+  try {
+    const stats = fs.lstatSync(dirPath);
+    return stats.isDirectory();
+  } catch (e) {
+    return false;
+  }
+}
+
+/**
+ * Returns a list of files in a given directory.
```
aeisenberg, line 773:
```suggestion
 * Returns a recursive list of files in a given directory.
```

src/util.ts
```diff
@@ -756,3 +756,34 @@ export async function checkActionVersion(version: string) {
...
+ */
+export function listFolder(dir: string): string[] {
+  if (!doesDirectoryExist(dir)) {
+    return [];
+  }
+  const entries = fs.readdirSync(dir, { withFileTypes: true });
```
aeisenberg, line 779: What happens if `dir` is not a directory?

src/analyze-action-post.ts
```diff
@@ -0,0 +1,42 @@
...
+    const outputDir = actionsUtil.getRequiredInput("output");
+    await uploadSarifDebugArtifact(config, outputDir);
+  }
+}
+
+async function runWrapper() {
+  try {
+    await run(debugArtifacts.uploadSarifDebugArtifact);
+  } catch (error) {
+    core.setFailed(`analyze action cleanup failed: ${error}`);
```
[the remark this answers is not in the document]
angelapwen: Thanks! Will make this change locally along with the same change on `init` side.

src/actions-util.ts
```diff
@@ -872,6 +875,31 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
   return currentRef === defaultBranch;
 }
 
-export function sanitizeArifactName(name: string): string {
-  return name.replace(/[^a-zA-Z0-9_\\-]+/g, "");
+export async function printDebugLogs(config: Config) {
+  core.info("Debug mode is on. Printing CodeQL debug logs...");
```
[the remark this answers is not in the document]
angelapwen: Makes sense. It's actually already there so it would've printed twice!

src/debug-artifacts.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+export async function uploadDatabaseBundleDebugArtifact(
+  config: Config,
+  logger: Logger
+) {
+  for (const language of config.languages) {
+    if (!dbIsFinalized(config, language, logger)) {
+      await uploadPartialDatabaseBundle(config, language);
+      continue;
```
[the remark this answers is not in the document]
angelapwen: Ok, I did that and moved the common `uploadDebugArtifacts` line back into this main function, so the other functions simply create the database bundles.

src/util.test.ts
```diff
@@ -440,3 +441,47 @@ for (const [
     isActionsStub.restore();
   });
 }
+
+test("doesDirectoryExist", (t) => {
+  const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "does-dir-exist-"));
```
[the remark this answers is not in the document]
angelapwen: Ah! That makes it a lot simpler. Thanks!

src/util.ts
angelapwen: This should be covered in the `doesDirectoryExist` test — if it is a file or doesn't exist at all (or any error is thrown when determining the directory), `doesDirectoryExist` returns false. If it's a directory it returns true.

src/actions-util.ts
angelapwen: I'm going to include the full path as you mentioned. I don't think it's necessary to print every directory name because a lot of directories may not have debug logs.

src/debug-artifacts.ts
angelapwen: I did a little digging into the CLI source code, but I thought it'd be too complex to somehow catch the error thrown in the action and then do something different. A `--force` option would be interesting though and isn't something I'd considered. 

We'd like to get this change in this week but I can write up a follow-up issue on the `--force` option on the CLI side for our backlog.

src/analyze-action-post.ts
angelapwen: I've done this, though I didn't know what to call the other file (not sure what patterns we want to follow). I called it `analyze-action-post-helper` but would like to hear what you think!

angelapwen:
The remaining unit tests are up, requesting final review now 😄 

After this is merged, I will add the "Download and check debug artifacts after failure in analyze" PR check/job to the required PR checks for each branch.

src/analyze-action-post-helper.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as analyzeActionPostHelper from "./analyze-action-post-helper";
+import * as configUtils from "./config-utils";
+import { setupTests } from "./testing-utils";
+import * as util from "./util";
+
+setupTests(test);
+
+test("post: analyze action with debug mode off", async (t) => {
+  return await util.withTmpDir(async (tmpDir) => {
+    process.env["RUNNER_TEMP"] = tmpDir;
```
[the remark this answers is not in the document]
angelapwen: Hm — I see it used this way without restoring in several other existing tests like https://github.com/github/codeql-action/blob/a6d09016e7e432c9ce360fd7da4845178fe0575e/src/config-utils.test.ts#L626

Is there a reason this one would act differently from other tests?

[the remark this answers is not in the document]
aeisenberg: Since it's used elsewhere in this way, it's _probably_ fine. I would recommend in the future that we do what Aditya suggests in all the tests. The danger if we don't do this is there may be hard to spot dependencies between tests that use this environment variable.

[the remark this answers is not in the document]
angelapwen: I see we also have this helper function that is called quite a lot https://github.com/github/codeql-action/blob/a6d09016e7e432c9ce360fd7da4845178fe0575e/src/testing-utils.ts#L102

although we don't need the other two environment variables here.

src/analyze-action-post-helper.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    };
+    sinon.stub(configUtils, "getConfig").resolves({
+      debugMode: true,
+      gitHubVersion,
+      languages: [],
+      packs: [],
+    } as unknown as configUtils.Config);
+
+    const requiredInputStub = sinon.stub(actionsUtil, "getRequiredInput");
+    requiredInputStub.withArgs("output").returns("fake-output-dir");
```
angelapwen, line 49: On a related note, I see that in some places we use `inputStub.restore()` to clean up at the end of a test, but in others we don't. Is this necessary or best practice?

src/debug-artifacts.ts
aeisenberg: Right...there's no need to get this done now.

aeisenberg (approved):

src/analyze-action-post-helper.test.ts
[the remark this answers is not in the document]
angelapwen: Hm, does this line https://github.com/github/codeql-action/blob/cade2b5621f0d8a99851d6a77e035baf73b60445/src/testing-utils.ts#L96 effectively reset the environment vars after each test?

src/analyze-action-post-helper.test.ts
angelapwen: Same as above, I think this line https://github.com/github/codeql-action/blob/cade2b5621f0d8a99851d6a77e035baf73b60445/src/testing-utils.ts#L93 resets the `sinon` stubs after each test?

aeisenberg: Actually...we should do that everywhere. The vscode extension uses a different (and better) idiom, which we should move to in the action.

Instead of creating stubs directly using `sinon.stub`, use a sandbox: 
```
  let sandbox: sinon.SinonSandbox;
  beforeEach(() => {
    sandbox = sinon.createSandbox();
  });
  afterEach(() => {
    sandbox.restore();
  });
```

And create stubs using: `sandbox.stub()`. This ensures all stubs are removed after every test runs. In the action, we are not doing this. So far, it hasn't caused any problems, but it might later.

angelapwen: For now it looks like as long as we call 

https://github.com/github/codeql-action/blob/cade2b5621f0d8a99851d6a77e035baf73b60445/src/testing-utils.ts#L49

in each test file it should call `sinon.restore()` and also restore the environment variables, right?

src/analyze-action-post-helper.test.ts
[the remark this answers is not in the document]
aeisenberg: Yes...it does. I also see `sinon.restore();`, which makes my comment about using a sandbox irrelevant.

src/analyze-action-post-helper.test.ts
aeisenberg: Yes....I missed that. So, you can ignore my comment.

angelapwen: Ok, great! This was educational 🧑‍🎓

src/debug-artifacts.ts
angelapwen: Filed at https://github.com/github/codeql-core/issues/2629

PR #1160, opened 2022-07-29
Split debug artifacts PR check into two jobs

angelapwen:
The current debug artifacts PR check executes just after the `analyze` step, which works at the moment but fails in #1159 which moves the artifact uploading to `post:` hooks. This PR attempts to split the PR check into 2 jobs, so that the `post:` hooks from the first job complete before the job that attempts to download and check the artifacts. 

As the [sync.py](https://github.com/github/codeql-action/blob/main/pr-checks/sync.py) script doesn't yet work with a file that has multiple jobs, I stopped using the generator for this workflow.


aeisenberg:
Hmmm...or you can move `debug-artifacts.yml` out of the sync directory and avoid generating it altogether.

angelapwen:
> Hmmm...or you can move `debug-artifacts.yml` out of the sync directory and avoid generating it altogether.

Yeah, I think that will be easier. Will do that now.

.github/workflows/debug-artifacts.yml
```diff
@@ -0,0 +1,124 @@
...
+        - os: macos-latest
+          version: cached
+        - os: ubuntu-latest
+          version: latest
+        - os: macos-latest
+          version: latest
+        - os: ubuntu-latest
+          version: nightly-latest
+        - os: macos-latest
+          version: nightly-latest
```
angelapwen, line 46: I wasn't sure if there's a cleaner way to reuse the `strategy` field for both jobs 🤔

angelapwen (commented):
I changed the workflow to a non-generated workflow so that it can include 2 jobs, but kept all of the fields from the generated workflow. 

I'm not entirely sure how to test this to make sure I haven't messed up any of the syntax, though.

.github/workflows/debug-artifacts.yml
```diff
@@ -0,0 +1,124 @@
...
+          debug: true
+          debug-artifact-name: my-debug-artifacts
+          debug-database-name: my-db
+      - name: Build code
+        shell: bash
+        run: ./build.sh
+      - uses: ./../action/analyze
+        id: analysis  
+  download-and-check-artifacts:
+    strategy:
```
adityasharad, line 70: I think this job doesn't need to be matrixed. It can be a single job that runs on Linux, and iterates through all the possible values of `os` and `version` that we care about, within which it iterates through the languages and checks the artifacts.

adityasharad:
Re testing: because this workflow has a `pull_request` trigger that is satisfied by this PR, it should run as a PR check on this same PR.
However, looking at the Checks panel of this PR, we see that the `Debug artifact upload` checks are still pending, which means the modified workflow didn't run, suggesting a workflow syntax error. Debug by going to the [checks tab](https://github.com/github/codeql-action/pull/1160/checks), then the [run for this workflow](https://github.com/github/codeql-action/actions/runs/2773641769), then the [workflow file used for the run](https://github.com/github/codeql-action/actions/runs/2773641769/workflow), which shows syntax errors as annotations.

angelapwen:
> Re testing: because this workflow has a `pull_request` trigger that is satisfied by this PR, it should run as a PR check on this same PR. However, looking at the Checks panel of this PR, we see that the `Debug artifact upload` checks are still pending, which means the modified workflow didn't run, suggesting a workflow syntax error. Debug by going to the [checks tab](https://github.com/github/codeql-action/pull/1160/checks), then the [run for this workflow](https://github.com/github/codeql-action/actions/runs/2773641769), then the [workflow file used for the run](https://github.com/github/codeql-action/actions/runs/2773641769/workflow), which shows syntax errors as annotations.

Thanks! I didn't know that there was a view that showed syntax errors.

.github/workflows/debug-artifacts.yml
angelapwen: Yeah, that looks a lot cleaner. Thanks!!

.github/workflows/debug-artifacts.yml
angelapwen: I wasn't entirely sure how to iterate through all the possible versions of `os` and `version` without using the matrix. I changed this so that this job runs on only ubuntu, but the matrix is still used for the downloads. I'm not sure if that's how matrix is intended to be used 🤔 but it seemed the cleanest way I could find.

I also tried setting `env` variables in this step as the arrays of os and version, but there doesn't seem to be a clean way to iterate through arrays in Actions syntax.

Another potential option would be to set `env` variables and use bash to iterate through them rather than the built-in `uses:` syntax to download the artifact. Then we could call the `curl` the [API to download the appropriate artifacts](https://docs.github.com/en/rest/actions/artifacts#download-an-artifact). I'm not sure if that's preferable if the matrix version just works? 

Another side note is that we can actually remove the `name` bit altogether, and just download all artifacts from this workflow at once. The remaining check for SARIF/database bundle/log files should still work. I guess the concern is that when we potentially emit more debug artifacts in the future, we might be unnecessarily downloading them (but maybe they should be added to this PR check anyway).

angelapwen:
Hm... it seems that the `Download and check debug artifacts` job didn't run this time around. I found the [runs](https://github.com/github/codeql-action/runs/7629698713?check_suite_focus=true) but no syntax error annotations on the [workflow file](https://github.com/github/codeql-action/actions/runs/2781172199/workflow) used for the run.

.github/workflows/debug-artifacts.yml
adityasharad: > Another potential option would be to set env variables and use bash to iterate through them rather than the built-in uses: syntax to download the artifact. Then we could call the curl the [API to download the appropriate artifacts](https://docs.github.com/en/rest/actions/artifacts#download-an-artifact). I'm not sure if that's preferable if the matrix version just works?

Exactly what I had in mind. This is preferable over another matrix job, because it only spawns 1 job, not 12 parallel jobs :)

```bash
for os in ubuntu-latest macos-latest; do
  for version in stable-20210308 stable-20210319 stable-20210809 cached latest nightly-latest; do
    for language in cpp csharp go java javascript python; do
      # as before, but use $os and $version to identify the artifacts
```

It is a bit awkward to repeat the matrix configuration in the bash loops, and to keep both in sync. However this solution avoids us having to introduce additional jobs.

> Another side note is that we can actually remove the name bit altogether, and just download all artifacts from this workflow at once. 

Even better. This will download all artifacts at once, placing them in named subdirectories based on their artifact name (which in this case is derived from the matrix values). Then we use bash loops to check each named subdirectory.

angelapwen: Great! I've done this (had to first list out the subdirectories so I could see how they were named).

.github/workflows/debug-artifacts.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/download-artifact@v3
+      - name: Check expected artifacts exist
+        shell: bash
+        run: |
+          OPERATING_SYSTEMS="ubuntu-latest macos-latest"
+          VERSIONS="stable-20210308 stable-20210319 stable-20210809 cached latest nightly-latest"
+          LANGUAGES="cpp csharp go java javascript python"
+          for os in $OPERATING_SYSTEMS; do
+            for version in $VERSIONS; do
+              cd ./my-debug-artifacts-$os-$version
```
angelapwen, line 62: I also considered just checking for `my-debug-artifacts-$os-$version/$language.sarif`, etc, but this seemed a little more readable. Happy to change it if anyone feels strongly about not changing directory and then backing out of it every loop though.

angelapwen:
I think this PR is ready for final review! The download/check part of the job only runs on one machine and successfully checks all uploaded artifacts. See https://github.com/github/codeql-action/runs/7649251536?check_suite_focus=true

.github/workflows/debug-artifacts.yml
adityasharad: This looks good. (You can also use `pushd <dir>` and `popd` as alternatives to the `cd <dir>` and `cd ..`.)

adityasharad, line 62: Quotes not essential but a good habit.
I observe this `cd `will fail if the artifact dir doesn't exist, but I think that's ok and will give us a reasonable error message. We can improve it in future if necessary.
```suggestion
              cd "./my-debug-artifacts-$os-$version"
```

adityasharad (approved):
Looks good! Minor suggestions only.

.github/workflows/debug-artifacts.yml
angelapwen: Cool! Didn't know that. I like that.

angelapwen: Ah interesting.. I wonder if it fails if it will continue onto the next directory? Probably not..

adityasharad: It will not. This is because by default Actions runs `bash` steps with `set -e` (`set -eo pipefail` specifically; see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell), so any shell command exiting non-zero will fail the entire step.

You can turn this off in a given step by adding `set +e` at the beginning, but we don't need that here.

angelapwen: Interesting! Thanks!

angelapwen:
Hm.. it looks like this is blocking on the 12 expected checks, which are the checks that I removed in this PR 🤔 is there any way to force green?

PR #1164, opened 2022-08-03
Add dependency on `@actions/cache`

edoardopirovano:
This dependency isn't actually used yet, but will be needed for the TRAP caching feature that I am currently working on. I was originally planning to add it in the same PR where it gets used, but given how many changes to transitive dependencies this generates, I think it makes sense to do it before in a separate PR to make reviewing easier and also test this in isolation.


criemen (approved):
LGTM, but I'm not really sure how to review this PR, outside from the changes to package.json.

edoardopirovano:
> LGTM, but I'm not really sure how to review this PR, outside from the changes to package.json.

I don't think there's much more to review than that. [This](https://github.com/github/codeql-action/blob/main/.github/workflows/script/check-node-modules.sh) CI check will make sure everything else matches what's declared there.

PR #1170, opened 2022-08-05
Add option to override TRAP caching feature flag

edoardopirovano:
This is a no-op for now, but let's add it in so DCA can opt out already without any warnings.


esbena (commented):
Perhaps add a comment about this being a noop, otherwise curious users will stumble upon it and complain that it does not work.

esbena (approved):

PR #1175, opened 2022-08-10
Use the API URL from the environment if it is present.

chrisgavin:
Actions runners export a `GITHUB_API_URL` environment variable (and have done for around two years now). If we're running on Actions, it's more reliable to use that rather than trying to calculate the API URL ourselves.

It also means the Action then works on non-DotCom environments where the API URL doesn't start `/api/v3`.

On GitHub Actions we can assume this URL will always be available, though unfortunately we have to keep the code for deriving the API URL from the primary URL since this is still needed in the runner.


aeisenberg:
The runner is deprecated. The only reason why we still have the code around is because some of our integration tests rely on it. Unless the integration tests are failing, I don't think we need to keep the old code paths around.

src/api-client.ts
```diff
@@ -36,16 +38,21 @@ export const getApiClient = function (
   const auth =
     (allowExternal && apiDetails.externalRepoAuth) || apiDetails.auth;
   const retryingOctokit = githubUtils.GitHub.plugin(retry.retry);
+  let apiURL = apiDetails.apiURL;
+  if (!apiURL) {
+    apiURL = deriveApiUrl(apiDetails.url);
+  }
```
aeisenberg, line 44: Minor:

```suggestion
  const apiURL = apiDetails.apiURL || deriveApiUrl(apiDetails.url);
```

aeisenberg (approved):
Minor suggestion to make the code more idiomatic, but this is not necessary.

If all integration tests can pass with `deriveApiUrl` removed, then it can be deleted. Otherwise, we will need to wait until we remove the runner tech debt.

Approving now so you can finish this off one way or the other tomorrow while I am asleep.

PR #1177, opened 2022-08-12
Add `expect-error` input to force PR check green on expected failure

angelapwen:
In #1159 we introduced a PR check to confirm that when the `analyze` workflow fails, partial debug artifacts are still uploaded. This caused 2 PR checks that always failed with a red X, which was not ideal for user experience. 

This change adds an `expect-error` optional input to the `analyze` workflow, which is set by default to false. Only in the PR Check that intentionally causes a failure in `analyze` is it set to true (though it could also be used in future cases where we expect a failing job). If this parameter is true, we do not call `core.setFailed(...)` on the job so it succeeds. 

If `expect-error` is set in any run not created by this repo, or a fork of it, we throw an error.

If `expect-error` is set to true, but no error in the run was found, the job fails. 

Screen Shot 2022-08-12 at 12 35 50 PM


analyze/action.yml
```diff
@@ -66,6 +66,10 @@ inputs:
     default: ${{ github.token }}
   matrix:
     default: ${{ toJson(matrix) }}
+  expect-error:
+    description: For PR checks. If true, the Action will not set Actions status to failure.
```
angelapwen, line 70: May need to make this more descriptive so that our users don't set this to `true` when they don't mean to? I don't think I see a `hidden` option for input parameters.

aeisenberg: How about this?

```suggestion
    description: [Internal] If true, the Action will not set Actions status to failure.
```

src/analyze-action.ts
```diff
@@ -183,7 +183,10 @@ async function run() {
   } catch (origError) {
```
aeisenberg, line 183: Note that if an error _isn't_ thrown, we need to set the job to a failure if this input is set to true.

It's a little more complicated than that since I see there are two places where `core.setFailed` is being called. Do we need to call both of them? It looks like the one in `runWrapper` is higher up the call chain. _Maybe_ we don't need the one in `run`.

angelapwen: > Note that if an error isn't thrown, we need to set the job to a failure if this input is set to true.

I'm a little confused: if an error isn't thrown, then why do we need to set the job to a failure? In that case, the job should succeed and we want that, right? Or do you mean that if we `expect-error` that there _must_ be an error, otherwise there was a problem? That's interesting. I had interpreted it more like `force-success`

> It's a little more complicated than that since I see there are two places where core.setFailed is being called. Do we need to call both of them? It looks like the one in runWrapper is higher up the call chain. Maybe we don't need the one in run. 

I'm not sure either. I can play around with it.

analyze/action.yml
angelapwen: Hm, this caused some failures — I wonder if the bracket symbols aren't able to be parsed correctly.

angelapwen: I think adding double quotes around it resolved it.

adityasharad: Consider throwing an error inside the code if this is set but the repo is not `github/codeql-action` or a fork. (You should be able to get that information from the Actions event context.)

src/analyze-action.ts
aeisenberg: > Or do you mean that if we `expect-error` that there _must_ be an error, otherwise there was a problem? That's interesting. I had interpreted it more like `force-success`

That's what I mean. If we expect an error and there is none, then that's an error. :)

analyze/action.yml
```diff
@@ -67,7 +67,7 @@ inputs:
   matrix:
     default: ${{ toJson(matrix) }}
   expect-error:
-    description: For PR checks. If true, the Action will not set Actions status to failure.
+    description: "[Internal] If true, the Action will not set Actions status to failure."
```
aeisenberg, line 70: If this is an error outside of integration testing, then specify that in the docs. Maybe:

```suggestion
    description: "[Internal] It is an error to use this input outside of integration testing of the codeql-action."
```

(Unfortunate that actions does not support hidden inputs.)

src/analyze-action.ts
```diff
@@ -61,6 +61,13 @@ export async function sendStatusReport(
   await actionsUtil.sendStatusReport(statusReport);
 }
 
+function hasBadExpectErrorInput(): boolean {
+  return (
+    actionsUtil.getOptionalInput("expect-error") === "true" &&
```
aeisenberg, line 66: Perhaps test that this input exists and is set to a truthy value (ie- a non-empty string):

```suggestion
    !!actionsUtil.getOptionalInput("expect-error") &&
```

src/analyze-action.ts
```diff
@@ -183,7 +197,10 @@ async function run() {
   } catch (origError) {
     const error =
       origError instanceof Error ? origError : new Error(String(origError));
-    if (!actionsUtil.getOptionalInput("expect-error")) {
+    if (
+      actionsUtil.getOptionalInput("expect-error") === "false" ||
```
aeisenberg, line 201: Only accept `"true"`, all else is considered false.

```suggestion
      actionsUtil.getOptionalInput("expect-error") !== "true" ||
```

src/analyze-action.ts
```diff
@@ -217,7 +234,10 @@ async function runWrapper() {
   try {
     await runPromise;
   } catch (error) {
-    if (!actionsUtil.getOptionalInput("expect-error")) {
+    if (
+      actionsUtil.getOptionalInput("expect-error") === "false" ||
```
aeisenberg, line 238: Same as above.

src/analyze-action.ts
angelapwen: I don't know that this works, because `getOptionalInput` returns `"false"` if the input is unset (because it defaults to false). So if we check if it's truthy, it will return `true` even if it's not been set.

angelapwen: I don't know that this works, because `getOptionalInput` returns `"false"` if the input is unset (because it defaults to false). So if we check if it's truthy, it will return `true` even if it's not been set.

src/analyze-action.ts
angelapwen: Done, thank you!

angelapwen: Done, thank you!

src/analyze-action.ts
angelapwen: I've added this logic at the end of the outer `try` block (indicating that an error was not thrown). Thanks!!

analyze/action.yml
angelapwen: I've added this code — still doing a final test of:
- that if it's set in another repo, the error throws
- that if it's set in this repo, the test passes
- that if it's set in a fork of this repo, the test passes
- that if it's set in this repo but no error is found, an error indicating this is thrown

src/analyze-action.ts
aeisenberg: You can remove the default value perhaps?

aeisenberg: Or:

```suggestion
    actionsUtil.getOptionalInput("expect-error") !== "false" &&
```
Checks that it is set to a non-default value.

angelapwen: Hm, I actually thought that it should be by default `false` so that we wouldn't have to catch the unset/null case in every future use of the variable. Do we prefer the latter?

aeisenberg: Oops, our comments crossed the wire. I think you're right, and we should just check for a non-default value.

angelapwen: Oh nice, I hadn't even thought of that! Thanks. I'll do that.

src/analyze-action.ts
```diff
@@ -83,9 +83,11 @@ export async function sendStatusReport(
   }
 }
 
+// REVIEW: Instead of returning the boolean and calling the method twice
+// I could also update a top level variable in this file. Which is preferable?
```
aeisenberg, line 87: I think it's fine to call the method twice. It's not very expensive. But, it looks like there's a syntax error in this file now.

src/analyze-action.ts
```diff
@@ -220,11 +222,17 @@ async function run() {
...
+      core.setFailed(
+        `expect-error input was set to true but no error was thrown.`
+      );
+    }
   } catch (origError) {
     const error =
       origError instanceof Error ? origError : new Error(String(origError));
     if (
-      actionsUtil.getOptionalInput("expect-error") === "false" ||
+      actionsUtil.getOptionalInput("expect-error") !== "true" ||
```
aeisenberg, line 235: Maybe a bad merge here?

aeisenberg (commented):
I think all the bits are correct floating around somewhere, but there must have been a bad merge and there are some syntax errors.

src/analyze-action.ts
angelapwen: Thanks! I'll remove the comment.

src/analyze-action.ts
angelapwen: I think the failures are expected — I removed the `ram: 1` line in the last commit to make sure that if `expect-error` was true and no error was thrown, that it would throw `Error: expect-error input was set to true but no error was thrown.`

I'll put the line back now that I have that run 👍

angelapwen:
So far I have tested to confirm the following cases:
- when `expect-error` is set on a repo that's not this one, `expect-error input parameter is for internal use only. It should only be set by codeql-action or a fork.` is thrown. ([run](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/runs/7865553844?check_suite_focus=true))
- when `expect-error` is set on this repo, and no error is found, `expect-error input was set to true but no error was thrown.` is thrown. ([run](https://github.com/github/codeql-action/runs/7865620848?check_suite_focus=true))
- when `expect-error` is set on this repo and an error is thrown, the job succeeds and is green ([run](https://github.com/github/codeql-action/runs/7865395471?check_suite_focus=true))

I will test to make sure it does the right things in a fork and update.

angelapwen:
The fork logic works 👍 ([run](https://github.com/github/codeql-action/runs/7866508219?check_suite_focus=true)) 

I will do some comment and commit cleanup.

src/analyze-action.ts
angelapwen: Interestingly, I found that the outer `try` block wasn't very useful. This catch block with `origError` is where I was able to throw/catch/stop errors appropriately. I removed the duplicate logic from the outer `try` block because it wasn't doing anything.

aeisenberg (approved):
This is good. It's a nice and simple solution you came up with in the end.

angelapwen:
Thank you! I forgot to remove a couple of `core.info()` debug lines. Will remove them now, squash and merge, and then add these PR checks to the required set.

PR #1182, opened 2022-08-15
Add telemetry for TRAP caching

edoardopirovano:
This PR adds some telemetry for TRAP caching including the times and sizes of the cache uploads/downloads and the languages that we are using TRAP caching for. Almost all of the code change is adding a dependency that computes the recursive size of folders. This is a little jarring, but is I think simpler than writing that code ourselves which is potentially somewhat awkward. I'm definitely open to other suggestions on if there's a simpler way to get the size of the cache, though.


edoardopirovano:
Changing to draft as we need a difference dependency with a more open license.

edoardopirovano:
This is now ready for review again.

src/analyze-action.ts
```diff
@@ -29,13 +29,18 @@ interface AnalysisStatusReport
 
 interface FinishStatusReport
   extends actionsUtil.StatusReportBase,
-    AnalysisStatusReport {}
+    AnalysisStatusReport {
+  trap_cache_upload_size_bytes: number;
+  trap_cache_upload_duration_ms: number;
```
henrymercer, line 34: Let's document these too.

src/analyze-action.ts
```diff
@@ -29,13 +29,18 @@ interface AnalysisStatusReport
...
+}
 
 export async function sendStatusReport(
   startedAt: Date,
   config: Config | undefined,
   stats: AnalysisStatusReport | undefined,
-  error?: Error
+  error?: Error,
+  trapCacheUploadTime?: number,
+  didUploadTrapCaches?: boolean
```
henrymercer, line 43: Nit, optional: Move these into a separate interface, so we can (a) enforce that we set all the TRAP caching telemetry information when applicable, and (b) to make the TRAP caching disabled case slightly more succinct.

src/analyze-action.ts
```diff
@@ -163,8 +175,10 @@ async function run() {
     await uploadDatabases(repositoryNwo, config, apiDetails, logger);
 
     // Possibly upload the TRAP caches for later re-use
+    const trapCacheUploadStartTime = Date.now();
     const codeql = await getCodeQL(config.codeQLCmd);
-    await uploadTrapCaches(codeql, config, logger);
+    trapCacheUploadTime = Date.now() - trapCacheUploadStartTime;
+    didUploadTrapCaches = await uploadTrapCaches(codeql, config, logger);
```
henrymercer, line 181: Same comment here about the `performance` API.  Perhaps we should make the timing functionality a utility method.

src/trap-caching.ts
```diff
@@ -175,6 +184,23 @@ export async function getLanguagesSupportingCaching(
...
 
+export async function getTotalCacheSize(
+  trapCaches: Partial<Record<Language, string>>
+): Promise<number> {
+  const sizes = await Promise.all(
+    Object.values(trapCaches).map(async (cacheDir) => {
+      return new Promise<number>((resolve) => {
+        getFolderSize(cacheDir, (err, size) => {
+          // Ignore file system errors when getting the size. It's only used for telemetry anyway.
+          if (err) resolve(0);
```
henrymercer, line 195: True, but we'd probably still want to know — maybe this can be a `warning`?

src/init-action.ts
```diff
@@ -65,6 +66,12 @@ interface InitSuccessStatusReport extends StatusReportBase {
   tools_resolved_version: string;
   /** Comma-separated list of languages specified explicitly in the workflow file. */
   workflow_languages: string;
+  /** Comma-separated list of languages for which we are using TRAP caching. */
+  trap_cache_languages: string;
+  /** Size of TRAP caches that we downloaded, in megabytes. */
+  trap_cache_download_size_bytes: number;
```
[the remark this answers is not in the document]
edoardopirovano: Whoops, thanks.

src/config-utils.ts
```diff
@@ -1030,12 +1042,27 @@ export async function getDefaultConfig(
...
     augmentationProperties,
-    trapCaches: trapCachingEnabled
-      ? await downloadTrapCaches(codeQL, languages, logger)
-      : {},
+    trapCaches,
+    trapCacheDownloadTime,
   };
 }
 
+async function downloadCacheWithTime(
```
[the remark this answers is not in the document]
edoardopirovano: Done.

src/config-utils.ts
```diff
@@ -1030,12 +1042,27 @@ export async function getDefaultConfig(
...
+  codeQL: CodeQL,
+  languages: Language[],
+  logger: Logger
+) {
+  let trapCaches = {};
+  let trapCacheDownloadTime = 0;
+  if (trapCachingEnabled) {
+    const start = Date.now();
+    trapCaches = await downloadTrapCaches(codeQL, languages, logger);
+    trapCacheDownloadTime = Date.now() - start;
```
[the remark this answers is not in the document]
edoardopirovano: Nice, I didn't know about that. Done.

src/analyze-action.ts
edoardopirovano: Done.

src/analyze-action.ts
edoardopirovano: Good idea, refactored accordingly.

src/analyze-action.ts
edoardopirovano: Done here too. I think having this in two places is fine since it's only a couple of lines, but let's definitely consider a utility method if we use it in more places.

src/trap-caching.ts
edoardopirovano: Hmm, possibly yes. I'd shied away from doing that because threading through a `Logger` is a little awkward but have done that now.

package.json
```diff
@@ -38,6 +38,7 @@
     "fast-deep-equal": "^3.1.3",
     "file-url": "^3.0.0",
     "fs": "0.0.1-security",
+    "get-folder-size": "^2.0.1",
```
henrymercer, line 41: Why version 2 vs version 3?  It looks like version 3 uses a Promise-based API.

src/trap-caching.ts
```diff
@@ -175,6 +184,27 @@ export async function getLanguagesSupportingCaching(
...
+      return new Promise<number>((resolve) => {
+        getFolderSize(cacheDir, (err, size) => {
+          if (err) {
+            logger.warning(`Error getting size of ${cacheDir}: ${err}`);
+            resolve(0);
+          } else {
+            resolve(size);
+          }
+        });
+      });
```
henrymercer, line 202: If we can use v3 of the get-folder-size package, we should be able to rewrite this using an `await`.

henrymercer (approved):
Minor comment around the get-folder-size package, otherwise LGTM.

src/trap-caching.ts
edoardopirovano: See above.

package.json
henrymercer: Ah that's disappointing.  I suppose we could still wrap it in `util.promisify` to avoid the `Promise` constructor.

henrymercer: Ah that's disappointing but very reasonable of the maintainer.  I suppose we could still wrap it in `util.promisify` to avoid the `Promise` constructor.

PR #1185, opened 2022-08-16
Enable bypassing the cached CodeQL bundle via a feature flag

henrymercer:
In the case that a broken bundle makes its way through the release process, and we only make a rollback decision after that bundle has already made it into the toolcache, this gives us a mechanism to easily perform that rollback in a timely way.  See the internal issue for more details.


henrymercer:
> Could we get `pr-checks/sync.py` to add in `TEST_MODE` to the environment for all the tests so we don't have to remember to add it on every individual one? Or is there some tests for which it shouldn't be set? Also, I would weakly prefer it to be set at a higher level on the whole job rather than in individual steps, although again I guess this might not be possible if some steps shouldn't have it.

Hmm, we could modify the workflow generator to look for patterns like `uses: ./../action/init` and add a `TEST_MODE` env variable.  It would be nice if that was a bit more robust though.

We have one workflow that explicitly doesn't set `TEST_MODE`, but it's not autogenerated.
My main concern with setting `TEST_MODE` across the whole workflow is that this is a fairly generic environment variable name, and could potentially change the behaviour of other actions.  Perhaps we should look at renaming `TEST_MODE` to something more specific like `CODEQL_ACTION_TEST_MODE` first?

aeisenberg (approved):

edoardopirovano (approved):
Looks good! Thanks for doing a bit of drive-by cleaning up on the code as well 👍🏼

henrymercer:
I'll hold off on merging this until we've deployed the feature flag and I can do a final manual test in our internal testing org.

edoardopirovano (approved):

henrymercer:
Manual testing 👍

PR #1188, opened 2022-08-17
Round fields in TRAP caching telemetry to integers

edoardopirovano:
The status reporting API expects these to be integers.


src/analyze-action.ts
```diff
@@ -71,10 +71,9 @@ export async function sendStatusReport(
   if (config && didUploadTrapCaches) {
     const trapCacheUploadStatusReport: FinishWithTrapUploadStatusReport = {
       ...statusReport,
-      trap_cache_upload_duration_ms: trapCacheUploadTime || 0,
-      trap_cache_upload_size_bytes: await getTotalCacheSize(
-        config.trapCaches,
-        logger
+      trap_cache_upload_duration_ms: Math.round(trapCacheUploadTime || 0),
+      trap_cache_upload_size_bytes: Math.round(
```
henrymercer, line 75: You could optionally push the rounding in here, if you'll only need the rounded values.

henrymercer (approved):
I think keeping them as ints makes sense since it's consistent with the other duration metrics like `analyze_builtin_queries_cpp_duration_ms`.

src/analyze-action.ts
edoardopirovano: Hmm, yes, I considered doing that but decided that rounding at the point where we construct the status report was the safest way to make sure we only send an integer to the API. Let's leave it like this for now to unbreak things, and we can always shuffle this around at some other point.

PR #1190, opened 2022-08-17
Modify `expect-error` input checking to fix errors on main

henrymercer:
We're seeing some errors in `main` from the function that ensures that `expect-error` is only set to true by the CodeQL Action PR checks, for example [this run](https://github.com/github/codeql-action/actions/runs/2875586196).

There doesn't appear to be enough information in the logs to debug the problem, so I've modified the workflow to dump the contents of the GitHub event, and changed the checking behaviour to instead require the `TEST_MODE` environment variable to be set to `true`.  This has the benefit of also working in the case that someone clones the Action repo and pushes it up somewhere else without explicitly forking it.


src/actions-util.ts
```diff
@@ -901,17 +900,3 @@ export async function printDebugLogs(config: Config) {
     walkLogFiles(logsDirectory);
   }
 }
-
-// Returns whether workflow kicked off by codeql-action repo itself,
-// or a fork of it.
-export function isAnalyzingCodeQLActionRepoOrFork(): boolean {
-  const codeQLActionRepoUrl = `https://api.github.com/repos/${CODEQL_DEFAULT_ACTION_REPOSITORY}`;
```
henrymercer, line 908: Hmm actually I think the problem was that this should have been `https://github.com/${CODEQL_DEFAULT_ACTION_REPOSITORY}`.  But I have a mild preference for the new solution since that works with the clone that isn't a fork case.

edoardopirovano: Yeah, the new solution seems reasonable to me.

edoardopirovano (approved):

edoardopirovano (approved):

src/actions-util.ts
angelapwen: I tried this URL at first but it seemed like it was the `api.github.com` one when I was testing the PR check locally. I guess it's probably different in `main` vs in a PR check? 

I think your solution is better in any case as well 👍

henrymercer: Huh you're totally right — it's `api.github.com` on a `pull_request` trigger, and `github.com` on a `push` trigger.

angelapwen: Oh interesting.. I guess it would've worked if we'd checked for either. Do you mind pointing me to where you found this in the documentation? (if you found it there ha)

henrymercer: I just triggered some Actions to find out — see https://github.com/github/codeql-action/runs/7917327160?check_suite_focus=true where it's `github.com` on `push` and https://github.com/github/codeql-action/runs/7916384170?check_suite_focus=true where it's `api.github.com` on `ready_for_review`.

angelapwen: Ah yep, I see, the `Dump GitHub event` step gives it away. Thanks!

PR #1197, opened 2022-08-19
Make "Update release branch" workflow fail if `npm version` fails

henrymercer:
This fixes a bug in the "Update release branch" workflow that could occur when there are merge conflicts in `package.json`.  These merge conflicts cause the `npm version` command to fail, but this failure was not causing the workflow itself to fail.  As a result, we merged a `v2 -> v1` backport that did not update the version number in `package.json`.  The "Tag release and merge back" workflow subsequently did not update the `v1` release tag, so the `v1` release did not occur.

As a fix, we use `subprocess.check_output` instead of `subprocess.run` to check the exit code and fail the workflow if there are merge conflicts bad enough to prevent us from updating the package version.


.github/update-release-branch.py
```diff
@@ -88,6 +88,7 @@ def open_pr(
   body.append('')
   body.append('Please review the following:')
   if len(conflicted_files) > 0:
+    body.append('- The package.json file contains the correct version.')
```
edoardopirovano, line 91:
```suggestion
    body.append('- [ ] The `package.json` file contains the correct version.')
```

edoardopirovano (approved):

.github/update-release-branch.py
```diff
@@ -88,6 +88,7 @@ def open_pr(
   body.append('')
   body.append('Please review the following:')
   if len(conflicted_files) > 0:
+    body.append('- [ ] The `package.json` file contains the correct version.')
```
aeisenberg, line 91: For consistency with other lines.

```suggestion
    body.append(' - [ ] The `package.json` file contains the correct version.')
```

aeisenberg (approved):
LGTM, but would be good to add the whitespace change in first.

PR #1207, opened 2022-08-23
TRAP Caching: Unset missing cache rather than setting to `undefined`

edoardopirovano:
In other parts of the code, we expect the values in the map to not be `undefined`, so we are getting warnings like https://github.com/github/vscode-codeql/actions/runs/2911692423 if we fail to find a cache in the Actions cache because we are setting a value of `undefined` here. Let's just remove the language from the map entirely if we don't have a cache for it.


henrymercer (approved):
LGTM.  When we set the property `undefined`, it still appears in `Object.values(...)`.  Deleting it removes it from `Object.values(...)`, so we don't attempt to find its size.

PR #1209, opened 2022-08-23
Fix Go tracing on Windows, and fix tests

henrymercer:
### Fix Go tracing tests

We're currently supplying `CODEQL_EXTRACTOR_GO_BUILD_TRACING=true` rather than the expected `CODEQL_EXTRACTOR_GO_BUILD_TRACING=on` in the PR checks, therefore tracing isn't being tested.

This PR updates the PR checks to the expected `CODEQL_EXTRACTOR_GO_BUILD_TRACING=on` and also adds an error if it's set to `true`.  Making `true` a synonym of `on` is more difficult than it's worth as the Go autobuilder also looks at this environment variable, so whether `true` is a synonym or not would depend on the version of the CodeQL CLI.

### Fix Go tracing on Windows

After fixing the tests, we discovered a bug with Go tracing on Windows with the Lua tracer.  We have a fix here: https://github.com/github/codeql/pull/10147.  Until this fix is released, we disable Lua tracing on Windows when analyzing Go.


aeisenberg (approved):

henrymercer:
Rebased on main to address a merge conflict.

src/languages.ts
```diff
@@ -37,6 +37,13 @@ export function parseLanguage(language: string): Language | undefined {
 }
 
 export function isTracedLanguage(language: Language): boolean {
+  if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
```
aeisenberg, line 40: There's a chance that a workflow was using `true` for this value and passing even without build tracing being turned on. This could mean that some workflows could start failing. I guess these workflows would be passing for the wrong reasons, so it's ok to fail them.

src/languages.ts
```diff
@@ -37,6 +37,13 @@ export function parseLanguage(language: string): Language | undefined {
...
+    throw new Error(
+      "The CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable is set to an invalid value. " +
+        "To enable Go build tracing, please set it to 'on'."
+    );
+  }
```
aeisenberg, line 45: Alternatively, instead of throwing an error, we could set the env var to `on`.

```suggestion
  if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
    process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] = "on";
  }
```

Maybe add a warning, too.

henrymercer: Good point, let's do this.  We'll also need to make a `core.exportVariable` call to update the value of the environment variable in subsequent steps.

src/languages.ts
```diff
@@ -36,14 +40,23 @@
...
 
-export function isTracedLanguage(language: Language): boolean {
+export function isTracedLanguage(language: Language, logger: Logger): boolean {
+  if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
+    logger.warning(
+      "The CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable was set to 'true', but it must " +
+        "be 'on' to enable Go build tracing. Setting it to 'on'."
+    );
+    process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] = "on";
+    core.exportVariable("CODEQL_EXTRACTOR_GO_BUILD_TRACING", "on");
```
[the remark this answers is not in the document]
henrymercer: I've disabled this check in https://github.com/github/codeql-action/pull/1209/commits/182342cdd7fc5578803ce8c6c1650b273ac37850 as it's no longer useful now that we only use the runner for CI.

edoardopirovano (approved):
Seems reasonable. Thanks for the investigation and fix!

PR #1218, opened 2022-08-26
Move calls to pack download to the init action

aeisenberg:
This ensures all steps to gather queries happens in the init action. This is where checking out queries in other repos happens as well.

This change is a refactoring that facilitates future changes. In particular, this will help when implementing package download from GHES.


src/config-utils.ts
```diff
@@ -1680,6 +1682,12 @@ export async function initConfig(
     }
   }
 
+  // if using the codescanning config in the CLI, pack downloads
+  // happen in the CLI, so no need to do them here.
```
henrymercer, line 1686: If we're parsing the code scanning config in the CLI, do pack downloads also happen in the `init` Action?

src/config-utils.ts
```diff
@@ -1781,3 +1789,33 @@ export async function getConfig(
...
   return JSON.parse(configString);
 }
+
+export async function downloadPacks(
+  codeQL: CodeQL,
+  languages: Language[],
+  packs: Packs,
+  logger: Logger
+) {
+  let packsDownloaded = 0;
```
henrymercer, line 1799: Nit: Rename to `numPacksDownloaded`

src/config-utils.ts
```diff
@@ -1781,3 +1789,33 @@ export async function getConfig(
...
+          .join(", ")}`
+      );
+    }
+
+    if (packsDownloaded) {
+      logger.info(`Downloaded ${packsDownloaded} packs`);
+    } else {
+      logger.info("No packs to download");
+    }
+    logger.endGroup();
```
henrymercer, line 1819: Suggest we either have a log group for each language, or move this outside the loop over languages.

src/config-utils.ts
aeisenberg: Yes. That is the reason to move downloads from `analyze` to `init`. A secondary reason is that other downloads happen in init as well. For example, checking out third party repos with extra queries.

src/config-utils.ts
```diff
@@ -1781,3 +1790,33 @@ export async function getConfig(
...
+          .map((r) => `${r.name}@${r.version || "latest"}`)
+          .join(", ")}`
+      );
+    }
+
+    if (numPacksDownloaded > 0) {
+      logger.info(`Downloaded ${numPacksDownloaded} packs`);
+    } else {
+      logger.info("No packs to download");
+    }
```
henrymercer, line 1819: If `languages` is `javascript, ruby` and no packs are selected, I think this will currently produce output like:

```
(group) Downloading packs
No packs to download
No packs to download
```

If there are some packs selected, we'll produce:

```
(group) Downloading packs
Downloading custom packs for javascript
Downloaded 1 packs
Downloading custom packs for ruby
Downloaded 1 packs
```

Some suggestions:

- Add the language into the "Downloaded" / "No packs to download" log messages
- Consider fixing the pluralisation
- Consider renaming the log group to "Downloading custom packs"

henrymercer (approved):
A minor comment about the log messages, otherwise LGTM.

src/config-utils.ts
henrymercer: 👍

src/config-utils.ts
aeisenberg: Right...that should go outside of the loop as well. The output from the CLI command is already printed to the console, so we don't need to break down number of packs per language.

PR #1219, opened 2022-08-26
Autobuild Go in `analyze` if not already built

angelapwen:
This change will run the autobuilder for Golang in the `analyze` step if:
- the feature flag `golang_extraction_reconciliation_enabled` is on
- the autobuild step has not already run for Go
- the user has not provided manual build steps for Go

This change should be a no-op for now, as if a user has not provided manual build steps for Go, we treat Go as an interpreted language and the autobuilder will run even if users have not included the `autobuild` action. 

With some imminent changes where we will not treat Go as an interpreted language, we introduce this change to maintain backwards compatibility for users who may have removed the `autobuild` step and who have not provided manual build steps. In this case, we will still autobuild for them at the beginning of the `analyze` step. 


aeisenberg (commented):
Two minor things.

henrymercer:
I'll move this PR forward while Angela's on vacation.  I'll start by addressing Andrew's review comments and adding a PR check to test the new behavior.

src/analyze-action.ts
```diff
@@ -97,6 +101,81 @@ function hasBadExpectErrorInput(): boolean {
...
+  // Only proceed if the beta Go extraction reconciliation behavior is
+  // enabled.
+  if (
+    process.env["CODEQL_ACTION_RECONCILE_GO_EXTRACTION"] !== "true" &&
+    !(await featureFlags.getValue(
+      FeatureFlag.GolangExtractionReconciliationEnabled
+    ))
+  ) {
+    logger.debug(
+      "Won't run the Go autobuilder since Go extraction reconciliation is not enabled."
```
aeisenberg, line 155: Minor: If I see this line and I'm not extracting go in my analysis, I think I'd be a little confused. Can you move this block to the end of the `if` checks? As a bonus, `featureFlags.getValue` might issue an API call, so doing this last potentially saves time and avoids an unnecessary API call.

aeisenberg (commented):
Looks good. A minor comment.

aeisenberg (approved):

PR #1221, opened 2022-08-29
Add support for downloading packs from GHES

aeisenberg:
This change adds:

- new `registries` input in init action. This input is similar to the `registries` block in the `qlconfig.yml` file.
- Change the downloadPacks function so that it accepts new parameters: - registries block - api auth
- Generate a qlconfig.yml file with the registries block if one is supplied. Use this file when downloading packs.
- temporarily set the `GITHUB_TOKEN` and `CODEQL_REGISTRIES_AUTH` based on api auth

TODO:

1. integration test
2. handle pack downloads when the config is generated by the CLI


aeisenberg:
OK...this should work now. The integration test is not ideal since it won't fail if both the `registries-auth-tokens` input and the `registries` block are ignored. In this case it will just fall back to using the `token` input.

pr-checks/checks/init-with-registries.yml
```diff
@@ -0,0 +1,48 @@
...
+      config-file: ./.github/codeql/codeql-config-registries.yml
+      languages: javascript
+      registries: |
+        [
+          {
+            "url": "https://ghcr.io/v2/",
+            "packages": "*/*",
+            "token": "${{ secrets.GITHUB_TOKEN }}"
+          }
+        ]
```
henrymercer, line 24: Hopefully not too much of a bikeshed, but why JSON rather than YAML here?  YAML seems more consistent with other configuration files like the `registries` property in `qlconfig.yml` files and `codeql-workspace.yml` files.

aeisenberg: My thinking was that if we use JSON here, it's more explicit that this is another _thing_ and we can say "JSON encoded string". People know what that means. However, if we use yaml here, we have to say: "well...you need to add an extra `|` since inputs only accept strings, but want to parse yaml".

I do agree, though, that yaml would be easier to read. I am quite ambivalent about this and maybe we should get more feedback on this.

aeisenberg: @jf205 thoughts?

jf205: I think i prefer YAML. If we have with a nice clear example in the docs then i think this will be fine for users.

src/config-utils.ts
```diff
@@ -61,6 +61,23 @@ export interface UserConfig {
 
 export type QueryFilter = ExcludeQueryFilter | IncludeQueryFilter;
 
+export type RegistryConfig = SafeRegistryConfig & {
```
henrymercer, line 64: I think the terminology of safe vs unsafe is helpful in general, but two comments here:

- I think it might be more important to name the unsafe one unsafe vs the safe one safe, so we think more carefully whenever we see `Unsafe` in the code.
- It might be clearer (though more verbose) to use `RegistryConfigNoCredentials` and `RegistryConfigWithCredentials` for instance.

src/config-utils.ts
```diff
@@ -1686,14 +1704,35 @@ export async function initConfig(
...
   await saveConfig(config, logger);
   return config;
 }
 
+function parseRegistries(registriesInput: string | undefined) {
+  try {
+    return registriesInput ? yaml.l(registriesInput) : undefined;
+  } catch (e) {
+    throw new Error(
+      `Invalid registries input. Must be a JSON string, but got: ${
```
henrymercer, line 1729:
```suggestion
      `Invalid registries input. Must be a YAML string, but got: ${
```

src/config-utils.ts
```diff
@@ -1686,14 +1704,35 @@ export async function initConfig(
...
+        e instanceof Error ? e.message : String(e)
```
henrymercer, line 1730: This error message might contain credentials.  Actions will mask out any secrets, but perhaps it'd be safe practice to avoid printing the input in the first place.

src/config-utils.ts
```diff
@@ -1795,30 +1834,94 @@ export async function downloadPacks(
...
+        logger.info("No packs to download");
+      }
+      logger.endGroup();
+    }
+  );
+}
+
+function createRegistriesBlock(registries: RegistryConfig[]) {
+  // be sure to remove the `token` field from the registry before writing it to disk.
+  const safeRegistries = registries.map((registry) => ({
```
henrymercer, line 1893: Can we enforce the absence of `token` using the type system?

src/config-utils.ts
```diff
@@ -1795,30 +1834,94 @@ export async function downloadPacks(
...
+async function wrapEnvironment(
+  env: Record<string, string | undefined>,
+  operation: Function
+) {
+  // Remember the original env
+  const oldEnv = { ...process.env };
+
+  // Set the new env
+  for (const [key, value] of Object.entries(env)) {
+    // Ignore undefined keys
```
henrymercer, line 1912: I suppose we don't yet need to unset an environment variable, but it might be worth documenting that this function doesn't support that.

henrymercer (commented):
Mostly minor comments — this generally looks in good shape!

init/action.yml
```diff
@@ -10,9 +10,35 @@ inputs:
...
+            packages: */*
+            token: ${{ secrets.GHCR_TOKEN }}
+
+      The url property contains the url to the container registry you want to connect to.
+
+      The packages property contains a single entry or a list of globs specifying packages that can be found in the container registry. Order is important. Earlier entries will match before later entries.
+
+      The token property contains a connection token for this registry.
+
+      If this input is missing, the `token` input is used for all pack downloads.
```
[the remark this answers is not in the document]
aeisenberg: I am not sure what you mean here. If the `registries` input is missing, then we only connect to the current GitHub instance and we always use the `token` input to connect.

Would it be sufficient to say something like:

> If you only need to download packages from this GitHub instance, use the `token` input instead.

?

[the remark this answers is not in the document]
henrymercer: Thanks for the clarification.  I originally interpreted "the `token` input is used for all pack downloads" to mean that we would pass the `token` input to the registry for each pack.  Your suggestion sounds a lot clearer 👍

src/config-utils.ts
aeisenberg: I'll remove the original error message from new new error.

src/config-utils.ts
aeisenberg: I can add an explicit return type to the function, but there is no way to ensure the token is removed from the object using the type system alone.

init/action.yml
```diff
@@ -10,9 +10,37 @@ inputs:
...
+      Use this input only when you need to download CodeQL packages from another instance of GitHub. If you only need to download packages from this GitHub instance, use the token input instead.
+
+      A YAML string that defines the list of GitHub container registries to use for downloading packs. The string is in the following form (the | is required on the first line):
+
+      registries: |
+          - url: https://containers.GHEHOSTNAME1/v2/
+            packages:
+              - my-company/*
+              - my-company2/*
+            token: \$\{{ secrets.GHEHOSTNAME1_TOKEN }}
```
adityasharad, line 27: If you quote this string, can we avoid the backslash escapes? Then it looks more like the concrete example in your integration tests.

aeisenberg: Unfortunately, no. `"${{ secrets.GHEHOSTNAME1_TOKEN }}"` is still treated as an action expression, even with the quotes. I'm not even sure what the snippet above will look like.

init/action.yml
```diff
@@ -10,9 +10,37 @@ inputs:
...
+
+          - url: https://ghcr.io/v2/
+            packages: */*
+            token: \$\{{ secrets.GHCR_TOKEN }}
+
+      The url property contains the URL to the container registry you want to connect to.
+
+      The packages property contains a single entry or a list of globs specifying packages that can be found in the container registry. Order is important. Earlier entries will match before later entries.
+
+      The token property contains a connection token for this registry.
```
adityasharad, line 37:
```suggestion
      The `url` property contains the URL to the container registry you want to connect to.

      The `packages` property contains a single glob string or a list of glob strings, specifying which packages should be retrieved from this particular container registry. Order is important. Earlier entries will match before later entries.

      The `token` property contains a connection token for this registry.
```

src/config-utils.ts
```diff
@@ -1795,30 +1834,108 @@ export async function downloadPacks(
...
+ *
+ * @param env
+ * @param operation
+ */
+async function wrapEnvironment(
+  env: Record<string, string | undefined>,
+  operation: Function
+) {
+  // Remember the original env
+  const oldEnv = { ...process.env };
```
adityasharad, line 1922: Is this any less efficient than `Object.assign({}, process.env)`?

src/config-utils.ts
```diff
@@ -1795,30 +1834,108 @@ export async function downloadPacks(
...
-      const results = await codeQL.packDownload(packsWithVersion);
-      numPacksDownloaded += results.packs.length;
-      logger.info(
-        `Downloaded packs: ${results.packs
-          .map((r) => `${r.name}@${r.version || "latest"}`)
-          .join(", ")}`
-      );
+  let qlconfigFile: string | undefined;
+  let registriesAuthTokens: string | undefined;
+  if (registries) {
```
adityasharad, line 1844: As discussed live, you need to gate the new registries/auth behaviour based on the CodeQL version number, either 2.10.4 or 2.10.5, and give a suitable error message if the version is not supported.

adityasharad (commented):
Looks reasonable, and I think you've addressed @henrymercer's comments. Some minor suggestions, and one reminder to add a CodeQL version number gate.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ No user facing changes.
 ## 2.1.22 - 01 Sep 2022
 
 - Downloading CodeQL packs has been moved to the `init` step. Previously, CodeQL packs were downloaded during the `analyze` step. [#1218](https://github.com/github/codeql-action/pull/1218)
+- Allow CodeQL packs to be downloaded from GitHub Enterprise Server instances. [#1221](https://github.com/github/codeql-action/pull/1221)
```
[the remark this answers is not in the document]
aeisenberg: I see this entry is in the wrong version. I'll change that, too.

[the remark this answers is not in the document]
aeisenberg: I'd also like to link to the documentation when it is available, but that won't be out for a while.

src/config-utils.ts
aeisenberg: I'm not sure which is more efficient, but either way the difference would be small.

aeisenberg:
Thanks for the comments. I addressed them all.

src/config-utils.ts
```diff
@@ -1795,30 +1835,116 @@ export async function downloadPacks(
...
-          .map((r) => `${r.name}@${r.version || "latest"}`)
-          .join(", ")}`
+  let qlconfigFile: string | undefined;
+  let registriesAuthTokens: string | undefined;
+  if (registries) {
+    if (
+      !(await codeQlVersionAbove(codeQL, CODEQL_VERSION_GHES_PACK_DOWNLOAD))
+    ) {
+      throw new Error(
+        `'registries' input is not supported on CodeQL versions less than ${CODEQL_VERSION_GHES_PACK_DOWNLOAD}.`
```
adityasharad, line 1850: Just occurred to me: we may want to gate against the target GHES version too (if we're on GHES). Can be a follow-up; not critical for this PR.

adityasharad (approved):
Looks good! Assuming @jf205 is happy with the user-facing behaviour, this is good to go.

aeisenberg:
@adityasharad would you mind taking another look. I added some better error handling and logic that ensures the registry urls end in`/` to avoid the bug in 2.10.4.

adityasharad:
Code changes looks good; linter checks are grumbling.

PR #1228, opened 2022-09-01
Make Go a traced language when Go extraction reconciliation is enabled

henrymercer:
Based on #1219.

This PR treats Go as a traced language when the beta support for Go extraction reconciliation is enabled (either via an environment variable, for use in testing, or via the feature flag).  This has the effect of enabling tracing in `init` (and disabling it in `analyze`) when Go extraction reconciliation is enabled.

We add two PR checks to ensure that Go extraction works as expected for workflows with an autobuild step and workflows with custom build steps, in addition to the existing PR check from #1219 relating to (to be) legacy workflows with neither an autobuild step nor custom build steps. 


src/autobuild-action.ts
```diff
@@ -62,16 +72,35 @@ async function run() {
...
     }
 
-    const config = await config_utils.getConfig(
-      getTemporaryDirectory(),
+    const apiDetails = {
+      auth: getRequiredInput("token"),
+      externalRepoAuth: getOptionalInput("external-repository-token"),
+      url: getRequiredEnvParam("GITHUB_SERVER_URL"),
+      apiURL: getRequiredEnvParam("GITHUB_API_URL"),
+    };
```
aeisenberg, line 80: I think this would be simpler. `externalRepoAuth` isn't necessary here.

```suggestion
    const apiDetails = apiClient.getApiDetails();
```

src/languages.ts
```diff
@@ -40,7 +40,11 @@ export function parseLanguage(language: string): Language | undefined {
   return undefined;
 }
 
-export function isTracedLanguage(language: Language, logger: Logger): boolean {
+export function isTracedLanguage(
+  language: Language,
+  isGoExtractionReconciliationEnabled: boolean,
+  logger: Logger
+): boolean {
   if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
```
aeisenberg, line 48: Instead of spitting out this warning here, can we just check for `true` OR `on`?

I guess this assumes that the env var is not used elsewhere. I checked in the CLI, but I did not check in the go extractor.

henrymercer: The Go extractor also uses this env var to determine whether tracing is on or off, and it'll only accept `on`.  We could change that, but we'd still want to be compatible with older versions of the Go extractor, so I think the warning is simpler.  We'll be able to deprecate `CODEQL_EXTRACTOR_GO_BUILD_TRACING` once we fully roll out Go extraction reconciliation.

henrymercer: I think it probably makes sense to show a warning that we're changing the value of the environment variable vs silently updating it, but that would be an option.

src/analyze.test.ts
```diff
@@ -487,20 +488,30 @@ for (const options of [
...
+        // Call listener on `codeql resolve extractor`
+        if (args[0] === "resolve" && args[1] === "extractor") {
+          const func = execOptions.listeners!.stdout as (data: Buffer) => void;
+          t.truthy(func, "stdout listener is defined");
+          func(Buffer.from('"/path/to/extractor"'));
+        }
+        return runnerObjectStub;
+      }
+    );
+
```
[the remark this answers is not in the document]
henrymercer: I think unfortunately the aim is to test the part of that function that decides whether to set the Lua tracing flags.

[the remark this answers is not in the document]
henrymercer: (specifically https://github.com/github/codeql-action/blob/21530f507f59f30a57afd89722ee1f5dc7829df9/src/codeql.ts#L908)

src/autobuild-action.ts
henrymercer: We don't have an `apiClient` here I don't think, but we can certainly drop the `externalRepoAuth`.

aeisenberg: Right...minor point. You can make `getApiDetails` exported in `apiClient.ts`. Not a big deal either way.

henrymercer: Oh, I see what you mean now.  Let's do that.

aeisenberg (approved):
Optional: export `getApiDetails()` in `apiDetails.ts`.

aeisenberg (approved):

PR #1231, opened 2022-09-02
Fix running tests on Node 18.8

henrymercer:
This PR fixes errors like this:

```
  TypeError: Cannot destructure property 'payload' of 'sourceMap' as it is null.

  › file://node_modules/mem/dist/index.js:42:27
  › file://node_modules/mem/dist/index.js:42:27
  › Object.<anonymous> (src/analyze-action-post-helper.test.ts:12:5)

  ✖ src/analyze-action-post-helper.test.ts exited with a non-zero exit code: 1
```

when running tests on Node 18.8.  The fix is simple: update ava to v4.3.3.  This version of ava [adds](https://github.com/avajs/ava/releases/tag/v4.3.3) compatibility with Node 18.8.


adityasharad (approved):
Sounds reasonable. I had hoped Dependency Review showed me the other dependencies that were updated in the lockfile as a result, but it's clear what you're primarily changing here.

aeisenberg (approved):

PR #1234, opened 2022-09-05
TRAP Caching: Be tolerant to not finding the extractor

edoardopirovano:
See also https://github.com/github/codeql/pull/10292

In some internal use-cases where we hack in an extra extractor by adding a `--search-path` to various CLI calls, we may not actually find it in this CLI call if the `--search-path` hasn't been added in. Since TRAP caching is non-essential, let's be tolerant here and just skip languages whose extractor we fail to find.


src/trap-caching.ts
```diff
@@ -162,7 +162,11 @@ export async function getLanguagesSupportingCaching(
     return result;
   const resolveResult = await codeql.betterResolveLanguages();
   outer: for (const lang of languages) {
-    if (resolveResult.extractors[lang].length !== 1) continue;
+    if (
+      resolveResult.extractors[lang] === undefined ||
+      resolveResult.extractors[lang].length !== 1
```
henrymercer, line 167: TIL: I didn't realise `resolve languages` would completely omit the `extractors` property.

You could also achieve this a little more succinctly with `?.length > 0`.

henrymercer: Also, would we want some kind of logging when we fail to find the extractor here?

henrymercer (approved):

src/trap-caching.ts
edoardopirovano: Thanks, I've improved the logging around when we hit these various cases. Note that the `extractors` property isn't completely omitted - it's just that `extractors[lang]` isn't set because in this particular case we can't find an extractor for the language since the custom search path for the QL4QL extractor isn't configured.

PR #1238, opened 2022-09-07
Add workflowEventName util function

Daverlo:
For the 1 click setup feature of Code Scanning we are planning on use the dynamic workflows internal api. Triggering an action run this way changes the content of `GITHUB_EVENT_NAME` to dynamic. As we will be triggering this run after a push/pr and the action does some work in those cases, we will be providing the real event name under the env var `CODESCANNING_EVENT_NAME`. 


src/actions-util.ts
```diff
@@ -821,9 +821,21 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
 }
 
+export function workflowEventName() {
+  // If the original event is dynamic CODESCANNING_EVENT_NAME will contain the right info (push/pull_request)
+  if (process.env["GITHUB_EVENT_NAME"] === "dynamic") {
+    const value = process.env["CODESCANNING_EVENT_NAME"];
+    if (value === undefined || value.length === 0) {
+      return process.env["GITHUB_EVENT_NAME"];
+    }
+    return value
```
aeisenberg: Minor: I think the build will fail because of this.

```suggestion
    return value;
```

aeisenberg (approved):

PR #1242, opened 2022-09-12
Go extraction reconciliation: Ensure backwards compatibility for multi-language builds

henrymercer:
Additionally autobuild Go in the autobuild Action to ensure backwards compatibility for users performing a multi-language build within a single job. For example, consider a user with the following workflow file:

```yml
- uses: github/codeql-action/init@v2
  with:
    languages: go, java
- uses: github/codeql-action/autobuild@v2
- uses: github/codeql-action/analyze@v2
```

- With Go extraction disabled, we will run the Java autobuilder in the autobuild Action, ensuring we extract both Java and Go code.
- With Go extraction enabled, taking the previous behavior we'd run the Go autobuilder, since Go is first on the list of languages. We wouldn't run the Java autobuilder at all and so we'd only extract Go code.

We therefore introduce a special case here such that we'll autobuild Go in addition to the primary non-Go traced language in the autobuild Action. This special case behavior should be removed as part of the next major version of the CodeQL Action.


src/autobuild.ts
```diff
@@ -19,26 +19,77 @@ export async function determineAutobuildLanguage(
...
+   *   autobuilder, since Go is first on the list of languages. We wouldn't run
+   *   the Java autobuilder at all and so we'd only extract Go code.
+   *
+   * We therefore introduce a special case here such that we'll autobuild Go
+   * in addition to the primary non-Go traced language in the autobuild Action.
+   *
+   * This special case behavior should be removed as part of the next major
+   * version of the CodeQL Action.
+   */
+  const autobuildLanguagesNoGo = autobuildLanguages.filter(
```
angelapwen, line 57: minor: `autobuildLanguagesWithoutGo` seems to make more sense to me?

src/autobuild.ts
```diff
@@ -19,26 +19,77 @@ export async function determineAutobuildLanguage(
...
+  // that should be fixed remain on a repo since they are linked to SARIF
+  // categories that are no longer updated.
+  if (autobuildLanguagesNoGo.length > 1) {
     logger.warning(
-      `We will only automatically build ${language} code. If you wish to scan ${autobuildLanguages
+      `We will only automatically build ${languages.join(
+        " and "
+      )} code. If you wish to scan ${autobuildLanguagesNoGo
         .slice(1)
         .join(" and ")}, you must replace this call with custom build steps.`
```
angelapwen, line 88: Not a change from this PR, but it's a little unclear to me what "this call" refers to.

angelapwen (commented):
Overall this makes sense to me. Seems unavoidable if we want to maintain full backwards compatibility. Happy to re-review/approve when out of draft mode.

angelapwen (approved):
Looks good to me! Approving for now, but I believe we are waiting until the latest release to go in before merging.

PR #1244, opened 2022-09-13
Remove the lua tracer feature flag check from the codeql-action.

criemen:
Always defer to the CLI on the Lua tracer state from now on.

This will turn off Lua tracing when combined with a CLI that has Lua tracing disabled by default, but as Lua tracing and legacy tracing on these CLIs are 100% compatible, that does not pose a problem.
Furthermore, the reduction of complexity in our code is worth turning off Lua tracing for those few users.

Querying the feature flag is becoming soon pointless, as dotcom will always return `true` anyways.


src/codeql.ts
```diff
@@ -800,29 +796,26 @@ async function getCodeQLForCmd(
...
-                isGoExtractionReconciliationEnabled,
-                logger
-              ) ||
-              process.platform !== "win32" ||
-              (await util.codeQlVersionAbove(
-                this,
-                CODEQL_VERSION_LUA_TRACING_GO_WINDOWS_FIXED
-              )))
-          ) {
-            extraArgs.push("--internal-use-lua-tracing");
```
cklin, line 822: Since we are no longer explicitly setting `--internal-use-lua-tracing`, does it mean that the modified action would work only with CLI versions that defaults to Lua tracing?

criemen: No, the action works with all CLI versions, but on some CLI versions the action will now use the legacy tracer instead.
As both tracers are 100% compatible, this is not an issue.
On later CLI versions, where the Lua tracer has gained additional capabilities, we also use it by default.

criemen:
Now that you mention it, I think we need to do the opposite and ensure this PR makes it into GHES, otherwise C# tracing will be broken on GHES:

Without the PR, our logic will detect that we're running on GHES, which disables all feature flags.
Therefore, we will pass in `--no-internal-use-lua-tracing`.
On `rc/3.7`, unlike `main`, that flag still has an effect, and disable Lua tracing.
However, we have already deleted the CLR tracer, so the legacy tracer for C# is in a bad shape. Furthermore, we really want to ship the C# improvements also to our GHES customers.

henrymercer:
Ah, I wasn't aware we'd deleted the CLR tracer in `rc/3.7`.  In that case, it sounds like we need to get this in.  AIUI 2.10.5 enables the Lua tracer by default, so the behaviour here of dropping the `--internal-use-lua-tracing` flag will be to use Lua tracing.

henrymercer (approved):

PR #1256, opened 2022-09-16
python-setup: Allow newest `virtualenv`

RasmusWL:
Context for previous version is
https://github.com/github/codeql-action/pull/862

Locally, I was able to install `20.15.1` with Python2.

I don't see any reason why python3 version should be restricted.

### DIFF
--- a/python-setup/install_tools.ps1
+++ b/python-setup/install_tools.ps1
@@ -5,8 +5,8 @@ py -3 -m pip install --user --upgrade pip setuptools wheel
 
 # virtualenv is a bit nicer for setting up virtual environment, since it will provide up-to-date versions of
 # pip/setuptools/wheel which basic `python3 -m venv venv` won't
-py -2 -m pip install --user 'virtualenv<20.11'
-py -3 -m pip install --user 'virtualenv<20.11'
+py -2 -m pip install --user 'virtualenv!=20.12.0'
+py -3 -m pip install --user virtualenv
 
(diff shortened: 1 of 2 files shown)


aeisenberg (commented):
Thanks for following up on this. I think this looks fine, but I just want to clarify my understanding.

With this change, the `install_tools` scripts will install the latest version of virtualenv that is not 20.12.0 (on python2) and the latest version on python3?

Since the latest version is now 20.15 and going up, the range restriction is not being used.

RasmusWL:
> With this change, the `install_tools` scripts will install the latest version of virtualenv that is not 20.12.0 (on python2) and the latest version on python3?
> 
> Since the latest version is now 20.15 and going up, the range restriction is not being used.

Correct on both. I guess we could remove the range restriction entirely, if you prefer :blush:

yoff (commented):
This is fine as far as I can tell. Is there a way to test it?

RasmusWL:
> This is fine as far as I can tell. Is there a way to test it?

sure, see the jobs at https://github.com/github/codeql-action/actions/runs/3070050275

yoff:
> > This is fine as far as I can tell. Is there a way to test it?
> 
> sure, see the jobs at https://github.com/github/codeql-action/actions/runs/3070050275

Ah, thanks 👍

yoff (approved):
LGTM

henrymercer (approved):

Issue #1249, opened 2022-09-15
Exception in configuring venv when the project has a `setup.py` file


mrc0mmand:
Hey!

When playing around with python CodeQL analysis I stumbled across an exception during the init phase:

```
  Installing collected packages: virtualenv-clone, pipenv
  Successfully installed pipenv-2022.9.8 virtualenv-clone-0.5.7
  + command -v python2
  /usr/bin/python3 -B /home/runner/work/_actions/github/codeql-action/v2/python-setup/auto_install_packages.py /opt/hostedtoolcache/CodeQL/0.0.0-20220825/x64/codeql
  Source root appears to be the real root.
  Will try to guess Python version, as it was not specified in `lgtm.yml`
  Trying to guess Python version based on Trove classifiers in setup.py
  Found Trove classifiers for both Python 2 and Python 3 in setup.py -- will use Python 3
  Found setup.py, will install package with pip in editable mode
  Creating venv in /home/runner/work/python-systemd/codeql-action-python-autoinstall
  + python3 -m virtualenv /home/runner/work/python-systemd/codeql-action-python-autoinstall
  /home/runner/.local/lib/python3.10/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
    warnings.warn(
  created virtual environment CPython3.10.4.final.0-64 in 491ms
    creator CPython3Posix(dest=/home/runner/work/python-systemd/codeql-action-python-autoinstall, clear=False, no_vcs_ignore=False, global=False)
    seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/runner/.local/share/virtualenv)
      added seed packages: pip==21.3.1, setuptools==58.3.0, wheel==0.37.0
    activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
  + /home/runner/work/python-systemd/codeql-action-python-autoinstall/bin/pip install -e .
  Traceback (most recent call last):
    File "/home/runner/work/_actions/github/codeql-action/v2/python-setup/auto_install_packages.py", line 175, in <module>
      python_executable_path = install_packages(codeql_base_dir)
    File "/home/runner/work/_actions/github/codeql-action/v2/python-setup/auto_install_packages.py", line 163, in install_packages
      return install_with_setup_py(version)
[...]

PR #1257, opened 2022-09-19
python-setup: Fix venv creation in Ubuntu 22.04

RasmusWL:
Fixes https://github.com/github/codeql-action/issues/1249

As described [here](https://askubuntu.com/questions/1406304/virtualenv-installs-envs-into-local-bin-instead-of-bin), when using Ubuntu 22.04 with new enough versions of `setuptools` (60.0.0+), the virtual environment created with `virtualenv` will put binaries in `<venv-path>/local/bin` instead of `<venv-path>/bin`.

Next release after `20.16.5` of `virtualenv` will include a fix for this ([PR](https://github.com/pypa/virtualenv/pull/2415)).

### DIFF
--- a/python-setup/auto_install_packages.py
+++ b/python-setup/auto_install_packages.py
@@ -5,6 +5,7 @@ import os
 import subprocess
 from tempfile import mkdtemp
 from typing import Optional
+import shutil
 
 import extractor_version
 
@@ -153,6 +154,17 @@ def install_packages(codeql_base_dir) -> Optional[str]:
 
     # get_extractor_version returns the Python version the extractor thinks this repo is using
     version = extractor_version.get_extractor_version(codeql_base_dir, quiet=False)
+
+    if version == 2 and not sys.platform.startswith('win32'):
+        # On Ubuntu 22.04 'python2' is not available by default. We want to give a slightly better
+        # error message than a traceback + `No such file or directory: 'python2'`
+        if shutil.which("python2") is None:
+            sys.exit(
+                "package installation failed: we detected this code as Python 2, but 'python2' executable was not available."
+                "To enable automatic package installation, please install 'python2' before the 'github/codeql-action/init' step, "
+                "such as running 'sudo apt install python2' (Ubuntu 22.04)."
+                "If your code is not Python 2, but actually Python 3, please file a bug report at https://github.com/github/codeql-action/issues/new"
+            )
 
     if os.path.exists('requirements.txt'):
         print('Found requirements.txt, will install packages with pip', flush=True)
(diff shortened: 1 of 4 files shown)


yoff (dismissed):
From how I read the given context (thanks!) this looks fine to me.

RasmusWL:
> Could you add `ubuntu-22.04` to the matrix in `python-deps.yml` (and any other relevant tests)?

Done :+1: (but only for the python-setup tests)

henrymercer:
I don't think Python 2 is preinstalled in the `ubuntu-22.04` virtual environment.  Would it be reasonable to add the combination of `python_version: 2` and `os: ubuntu-22.04` to the `exclude` listing in the matrix, or should we install Python 2 using `actions/setup-python`?

RasmusWL:
> I don't think Python 2 is preinstalled in the `ubuntu-22.04` virtual environment. Would it be reasonable to add the combination of `python_version: 2` and `os: ubuntu-22.04` to the `exclude` listing in the matrix, or should we install Python 2 using `actions/setup-python`?

we have a bit of an internal discussion on how to handle this. Let me get back to this PR once we have figured out what to do.

RasmusWL:
this should fail the Ubuntu 22.04 test, but just want to be able to point to a log file to show that the logic works before disabling the test. If only I had designed the test system to be more flexible, maybe we could have kept the test alive under a form of expected failure mode :thinking:

RasmusWL:
Example of failure here: https://github.com/github/codeql-action/actions/runs/3097944471/jobs/5015231587#step:4:140 (revealed that I had forgotten a few spaces)

python-setup/auto_install_packages.py
```diff
@@ -153,6 +154,19 @@ def install_packages(codeql_base_dir) -> Optional[str]:
...
+
+    if version == 2 and not sys.platform.startswith('win32'):
+        # On Ubuntu 22.04 'python2' is not available by default. We want to give a slightly better
+        # error message than a traceback + `No such file or directory: 'python2'`
+        if shutil.which("python2") is None:
+            sys.exit(
+                "package installation failed: we detected this code as Python 2, but 'python2' executable was not available. "
+                "To enable automatic package installation, please install 'python2' before the 'github/codeql-action/init' step, "
+                "such as running 'sudo apt install python2' (Ubuntu 22.04). "
+                "If your code is not Python 2, but actually Python 3, please file a bug report at https://github.com/github/codeql-action/issues/new"
```
henrymercer, line 168:
```suggestion
                "Python package installation failed: we detected this code as Python 2, but the 'python2' executable was not available. "
                "To enable automatic package installation, please install 'python2' before the 'github/codeql-action/init' step, "
                "for example by running 'sudo apt install python2' (Ubuntu 22.04). "
                "If your code is not Python 2, but actually Python 3, please file a bug report at https://github.com/github/codeql-action/issues/new"
```

henrymercer (approved):
Looks good to me, just a copy editing suggestion

henrymercer (approved):

henrymercer:
Taking this out of draft since that reruns the PR checks.

PR #1259, opened 2022-09-19
Correctly report CodeQL version when using cache

angelapwen:
A one-line change to correctly update the version reported to our telemetry when we used a cached CodeQL bundle. Also updates the test suite to check the return value of the `setupCodeQL` function, which is what feeds into the version reported to our telemetry.


adityasharad:
> I am unsure how to test this without waiting for the next CodeQL release and monitoring our telemetry during the period of time when the older cached version is used.

There are some unit tests that call `setupCodeQL`, but use it to configure CodeQL without checking the return value. See if you can extend those, or write simpler new ones, so that they check the return value from this function, while mocking different conditions of the toolcache/GitHub API/user config.

> I'm also not sure if this warrants a changelog post because the only visible changes are to internal users viewing our telemetry.

Nothing user-facing, so indeed no change note.

adityasharad:
> I am unsure how to test this without waiting for the next CodeQL release and monitoring our telemetry during the period of time when the older cached version is used.

There are some unit tests that call `setupCodeQL`, but use it to configure CodeQL without checking the return value. See if you can extend those, or write simpler new ones, so that they check the `toolsVersion` in the return value from this function, while mocking different conditions of the toolcache/GitHub API/user config.

> I'm also not sure if this warrants a changelog post because the only visible changes are to internal users viewing our telemetry.

Nothing user-facing, so indeed no change note.

adityasharad:
> I am unsure how to test this without waiting for the next CodeQL release and monitoring our telemetry during the period of time when the older cached version is used.

I see some unit tests that call `setupCodeQL`, but use it to configure CodeQL without checking the return value. See if you can extend those, or write simpler new ones, so that they check the `toolsVersion` in the return value from this function, while mocking different conditions of the toolcache/GitHub API/user config. By mocking these conditions and checking the `toolsVersion`, you'll be able to simulate what gets reported to the telemetry and make sure we are handling the various possible cases, without having to actually reach that situation during a real workflow run.

> I'm also not sure if this warrants a changelog post because the only visible changes are to internal users viewing our telemetry.

Nothing user-facing, so indeed no change note.

julioAk47 (approved):

aeisenberg:
You _could_ test this by running the the action in a testing repo using this branch. But, the changes look reasonable to me. I suggest we merge and then check if it is working.

aeisenberg (approved):

PR #1265, opened 2022-09-23
TRAP Caching: Skip uploading of small caches

edoardopirovano:
Our telemetry shows that in a fair amount of cases we get some very small TRAP caches because the repository contains only a very small cacheable amount of the language being analyzed. In this case, we get slight performance regressions because the extra network calls just aren't worth doing for so little information. Let's set a minimum size of cache that we will consider worth uploading, so that we don't waste time uploading (and then redownloading) caches that are just too small to be worth it.

Empirically, 10MB is a good threshold for this because it is around 2MB more than the size of the TRAP for the JavaScript standard library, so this represents us actually caching some useful things beyond this. Other languages might have a different threshold for what is considered a big enough cache to be useful - we can explore that further (guided by our telemetry) once we get to implementing this feature for other languages.

There'll still be some small caches in the internal repos that had the feature flag on, but these should disappear on their own after around a week since that's the retention time of the GitHub Actions cache.


src/util.ts
```diff
@@ -836,3 +838,23 @@ export async function isGoExtractionReconciliationEnabled(
...
+
+/**
+ * Get the size a folder in bytes. This will log any filesystem errors
+ * as a warning and then return 0.
+ *
+ * @param cacheDir A directory to get the size of.
+ * @param logger A logger to log any errors to.
+ * @returns The size in bytes of the folder, or 0 if errors occurred.
+ */
+export async function getFolderBytes(
```
henrymercer, line 850: We might want to consider including `try` in the name and returning a `Promise<number | undefined>`.  This would make it a little more explicit that this function will fail without an exception.

henrymercer (commented):
Seems sensible.  I think this could warrant a changenote.

src/util.ts
edoardopirovano: Seems sensible, done.

edoardopirovano:
> Seems sensible. I think this could warrant a changenote.

Since we haven't enabled this feature yet for any external users, nor made any promises about the specifics of its behaviour, I think it's reasonable to put in a change like this without another change note. Happy to add one if you disagree, though.

PR #1269, opened 2022-09-26
Update unset environment variables PR check

angelapwen:
The updated Go reconciliation changes surfaced a failing PR check that, upon closer look, was meant to fail previously due to a bug in the CLI in versions < 2.5.1 that didn't propagate environment variables needed by the tracer. This was[ fixed in v2.5.1](https://github.com/github/codeql-cli-binaries/releases/tag/v2.5.1).  

Therefore this change splits the PR check into two:
- checks Java only after version 2.5.1 (ie. not for `stable-20210308` which was v2.4.5 or `stable-20210319` which was v2.4.6). 
- checks all other languages for all other CLI versions, as we do now.


.github/workflows/unset-environment-old-cli.yml
```diff
@@ -0,0 +1,80 @@
+name: PR Check - Test unsetting environment variables for CLI version < 2.5
```
henrymercer, line 1: Can we add a comment explaining why behaviour of the older CLIs is different, similar to what you said in the PR description?

henrymercer (approved):
This looks good.  We could also consider removing support for earlier versions of the CLI — these are quite old now — but I'm happy with this approach.

.github/workflows/unset-environment-old-cli.yml
```diff
@@ -0,0 +1,84 @@
...
+    - name: Build code
+      shell: bash
+      run: env -i PATH="$PATH" HOME="$HOME" ./build.sh
+    - uses: ./../action/analyze
+      id: analysis
+      env:
+        TEST_MODE: true
+    - shell: bash
+      run: |
+        CPP_DB=${{ fromJson(steps.analysis.outputs.db-locations).cpp }}
```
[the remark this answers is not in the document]
angelapwen: Have done this!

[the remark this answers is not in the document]
angelapwen: I'm curious — why does `CPP_DB` not need the quotes here, but you've added them to its usage later on? Is it because it shouldn't be added while it's being assigned to?

.github/workflows/unset-environment-new-cli.yml
```diff
@@ -55,34 +55,40 @@ jobs:
         TEST_MODE: true
     - shell: bash
       run: |
-        CPP_DB=${{ fromJson(steps.analysis.outputs.db-locations).cpp }}
+        CPP_DB="${{ fromJson(steps.analysis.outputs.db-locations).cpp }}"
         if [[ ! -d $CPP_DB ]] || [[ ! $CPP_DB == ${{ runner.temp }}/customDbLocation/* ]]; then
```
aeisenberg, line 59: You should also quote the variable references here:

```suggestion
        if [[ ! -d "$CPP_DB" ]] || [[ ! "$CPP_DB" == "${{ runner.temp }}/customDbLocation/*" ]]; then
```

Also, below.

.github/workflows/unset-environment-new-cli.yml
```diff
@@ -55,34 +55,40 @@ jobs:
...
-          echo "Did not create a database for CPP, or created it in the wrong location."
+          echo "Did not create a database for CPP, or created it in the wrong location." \
+            "Expected location was '${ runner.temp }'/customDbLocation/* but actual was '${ CPP_DB }'"
```
aeisenberg, line 61: Careful...`${ CPP_DB }` is not valid. Also, I think `runner.temp` is not correct. You want the actions expression, not a bash variable.

```suggestion
            "Expected location was '${{ runner.temp }}/customDbLocation/*' but actual was '${CPP_DB}'."
```

adityasharad: There is also a corresponding env var, which is better since you're already in Bash: `${RUNNER_TEMP}`.

.github/workflows/unset-environment-old-cli.yml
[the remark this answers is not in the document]
aeisenberg: When you are defining a bash variable, you don't need quotes around its name, only its value. When you are using a bash variable, it should always be quoted.

If you don't quote at the appropriate times, whitespace will be considered separate arguments.

.github/workflows/unset-environment-old-cli.yml
```diff
@@ -0,0 +1,84 @@
...
+        if [[ ! -d $CPP_DB ]] || [[ ! $CPP_DB == ${{ runner.temp }}/customDbLocation/* ]]; then
+          echo "Did not create a database for CPP, or created it in the wrong location."
```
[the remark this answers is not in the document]
angelapwen: Interesting, I actually read the test differently — I thought expected would be at `${{ runner.temp }}/customDbLocation/*` and actual would be `$CPP_DB`.

.github/workflows/unset-environment-new-cli.yml
angelapwen: Just noticed this as well, should be done now.

.github/workflows/unset-environment-old-cli.yml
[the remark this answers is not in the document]
angelapwen: Thank you!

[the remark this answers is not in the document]
aeisenberg: Here's an example. Let's say `${{ fromJson(steps.analysis.outputs.db-locations).cpp }}` resolves to `/path/with a space`.

This command:

```
CPP_DB=${{ fromJson(steps.analysis.outputs.db-locations).cpp }}
```

will be interpreted as setting the variable `CPP_DB/path/with` and apply this to a command `a` with an argument `space`.

```
CPP_DB="${{ fromJson(steps.analysis.outputs.db-locations).cpp }}"
```

Will simply set the `CPP_DB` variable to `/path/with a space`. Bash is arcane and learning all these rules takes a long time.

.github/workflows/unset-environment-old-cli.yml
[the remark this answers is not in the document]
aeisenberg: Ah...apologies. You might be right here.

.github/workflows/unset-environment-new-cli.yml
angelapwen: Ah, I see, I had gotten the Actions expression syntax mixed up with bash's. I have made those changes!

.github/workflows/unset-environment-old-cli.yml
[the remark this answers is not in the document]
angelapwen: Thanks for the example, that helps! I didn't realize that. Makes sense as to why the variable doesn't need quotes when it's being defined, then.

aeisenberg (approved):
Nice!

angelapwen:
Hm.. the test is failing with

```
Did not create a database for CPP, or created it in the wrong location. Expected location was '/home/runner/work/_temp/customDbLocation/*' but actual was '/home/runner/work/_temp/customDbLocation/cpp'
```

On the bright side, got to test the error message. Do I need the `{}` operators because of the double quotes, ie. `"${CPP_DB}"` rather than `"CPP_DB"`?

.github/workflows/unset-environment-old-cli.yml
```diff
@@ -56,32 +56,32 @@ jobs:
...
-          echo "Did not create a database for Go, or created it in the wrong location." \
+        if [[ ! -d "$GO_DB" ]] || [[ ! "$GO_DB" == "${RUNNER_TEMP}/customDbLocation/"* ]]; then
+          echo "::error::Did not create a database for Go, or created it in the wrong location." \
             "Expected location was '${RUNNER_TEMP}/customDbLocation/*' but actual was '${GO_DB}'"
           exit 1
         fi
         JAVASCRIPT_DB="${{ fromJson(steps.analysis.outputs.db-locations).javascript }}"
-        if [[ ! -d "$JAVASCRIPT_DB" ]] || [[ ! "$JAVASCRIPT_DB" == "${RUNNER_TEMP}/customDbLocation/*" ]]; then
-          echo "Did not create a database for Javascript, or created it in the wrong location." \
+        if [[ ! -d "$JAVASCRIPT_DB" ]] || [[ ! "$JAVASCRIPT_DB" == "${RUNNER_TEMP}/customDbLocation/"* ]]; then
```
aeisenberg, line 77: Is there a reason why `*` is outside of the quotes? I don't _think_ it changes the meaning. Also, does `$JAVASCRIPT_DB` end in `*`? I guess so, or else the workflow would be failing.

angelapwen: Yes, it's a wildcard character when outside quotes. @adityasharad and I debugged the reason it failed just now and it was because `*` was in the quotes and not considered a wildcard 😄

.github/workflows/unset-environment-new-cli.yml
```diff
@@ -56,38 +56,38 @@ jobs:
...
-          echo "Did not create a database for Java, or created it in the wrong location." \
+        if [[ ! -d "$JAVA_DB" ]] || [[ ! "$JAVA_DB" == "${RUNNER_TEMP}/customDbLocation/"* ]]; then
+          echo "::error::Did not create a database for Java, or created it in the wrong location." \
             "Expected location was '${RUNNER_TEMP}/customDbLocation/*' but actual was '${JAVA_DB}'"
           exit 1
         fi
         JAVASCRIPT_DB="${{ fromJson(steps.analysis.outputs.db-locations).javascript }}"
-        if [[ ! -d "$JAVASCRIPT_DB" ]] || [[ ! "$JAVASCRIPT_DB" == "${RUNNER_TEMP}/customDbLocation/*" ]]; then
-          echo "Did not create a database for Javascript, or created it in the wrong location." \
+        if [[ ! -d "$JAVASCRIPT_DB" ]] || [[ ! "$JAVASCRIPT_DB" == "${RUNNER_TEMP}/customDbLocation/"*]]; then
```
aeisenberg, line 83: Fixes a bash syntax error.

```suggestion
        if [[ ! -d "$JAVASCRIPT_DB" ]] || [[ ! "$JAVASCRIPT_DB" == "${RUNNER_TEMP}/customDbLocation/"* ]]; then
```

.github/workflows/unset-environment-old-cli.yml
adityasharad: Just looked at this one together -- you need the `*` outside the quotes so that Bash glob-expands it, otherwise it gets treated as a literal `*` character.

aeisenberg: Bash is awesome. :) Thanks for explaining.

angelapwen:
There were some flaky runner tests that I had to re-run, but it's ready for review again.

.github/workflows/unset-environment-new-cli.yml
```diff
@@ -0,0 +1,95 @@
...
+    - reopened
+    - ready_for_review
+  workflow_dispatch: {}
+jobs:
+  unset-environment:
+    strategy:
+      matrix:
+        include:
+        - os: ubuntu-latest
+          version: stable-20210809
```
aeisenberg, line 25: Not much we can do about this, but we will need to remember to change this when we no longer support CLI v2.5.9.

.github/workflows/unset-environment-old-cli.yml
```diff
@@ -0,0 +1,89 @@
...
+      shell: bash
+      run: env -i PATH="$PATH" HOME="$HOME" ./build.sh
+    - uses: ./../action/analyze
+      id: analysis
+      env:
+        TEST_MODE: true
+    - shell: bash
+      run: |
+        CPP_DB="${{ fromJson(steps.analysis.outputs.db-locations).cpp }}"
+        if [[ ! -d "$CPP_DB" ]] || [[ ! "$CPP_DB" == "${RUNNER_TEMP}/customDbLocation/"* ]]; then
```
aeisenberg, line 59: So, why exactly do we need to glob here? Is it because we are not sure what the last segement of the directory will be named? And presumably, if there are two entries in the directory, this test will always fail? What if there is a whitespace in this directory segment? I think the test wuld fail, but can this conceivably happen?

.github/workflows/unset-environment-new-cli.yml
angelapwen: Yes, agreed (it's also present in all the other PR checks so we can change them all at once).

.github/workflows/unset-environment-old-cli.yml
angelapwen: >So, why exactly do we need to glob here? Is it because we are not sure what the last segement of the directory will be named? 

I'm not exactly sure (I didn't write this test originally) — I think we could remove the `*` and replace it with the name of the directory, which AFAIK is just the language (`cpp` for example)

> And presumably, if there are two entries in the directory, this test will always fail? 

I don't think this is true, because currently the `${RUNNER_TEMP}/customDbLocation/` directory should have all the subdirectories named by each language in them. 

> What if there is a whitespace in this directory segment? I think the test wuld fail, but can this conceivably happen?

Which directory segment do you mean — the very last one that is just the language? I don't think it can happen, but if it did happen, I think it would fail too. Maybe it would just be better to explicitly write out the language rather than use the `*` at all.

aeisenberg: Wild...I didn't know that. Just tried it out. `if [[  "$var" == * ]];` will match if _any_ file in the current directory matches `$var`.

It would definitely be clearer if you replaced `*` with the language name, but I don't think it's a blocker.

aeisenberg (approved):

PR #1280, opened 2022-09-30
TRAP Caching: Add timeouts to upload/download operations

edoardopirovano:
Per @henrymercer's suggestion as he observed [one instance](https://github.com/github/codeql-action/actions/runs/3152962897) where TRAP cache downloading hanged and caused a whole run to fail, this PR adds timeouts to both the download and upload operations.


src/trap-caching.ts
```diff
@@ -24,6 +24,11 @@ const CACHE_SIZE_MB = 1024;
 // cache for us to consider it worth uploading.
 const MINIMUM_CACHE_MB_TO_UPLOAD = 10;
 
+// This constant sets the maximum number of milliseconds to wait for
```
henrymercer, line 27:
```suggestion
// The maximum number of milliseconds to wait for
```

src/util.ts
```diff
@@ -878,3 +878,29 @@ export async function tryGetFolderBytes(
...
+ * Run a promise for a given amount of time, and if it doesn't resolve within
+ * that time, return undefined instead after logging an informational message.
+ *
+ * @param timeoutMs The timeout in milliseconds.
+ * @param promise The promise to run.
+ * @param errorMessage An informational message to log if the promise times out.
+ * @param logger A logger to log the informational message to.
+ * @returns The result of the promise, or undefined if the promise times out.
+ */
+export async function withTimeout<T>(
```
henrymercer, line 892: Minor, optional: Another slightly more general option here would be specifying an `onTimeout` input.

src/trap-caching.ts
```diff
@@ -161,7 +171,13 @@ export async function uploadTrapCaches(
     logger.info(`Uploading TRAP cache to Actions cache with key ${key}`);
     toAwait.push(cache.saveCache([cacheDir], key));
   }
-  await Promise.all(toAwait);
+  await withTimeout(
+    // We upload all caches in parallel, so multiply the timeout by the number of caches
+    MAX_CACHE_OPERATION_MS * toAwait.length,
```
henrymercer, line 176: Minor: I can see how this would make sense if we uploaded them in series, but not in parallel.  Perhaps we could say something like "This should finish at least as quickly as if we uploaded each cache serially".

henrymercer (approved):
LGTM, just some minor comments.

jbj:
Will we get sufficient telemetry to tell how often the timeout is reached?

src/trap-caching.ts
edoardopirovano: Thanks, I've amended this and also reworded the text that follows it a bit.

src/util.ts
edoardopirovano: Good idea, I've generalised this per your suggestion.

src/trap-caching.ts
edoardopirovano: Hmm, looking back at this I'm not sure there's much to be gained by running the uploads in parallel. They'll be competing for IO and bandwidth anyway, so I highly doubt we gain any performance from this and it makes the timeout behaviour a bit confusing. Let's just upload them serially and use an individual timeout for each one.

edoardopirovano:
> Will we get sufficient telemetry to tell how often the timeout is reached?

Our telemetry will record an upload/download time that's very near to the timeout. This should allow us to pick out runs where we timed out fairly easily by filtering on telemetry rows that are near this value. I suppose there might be a few runs that finished _just_ before the timeout that might mistakenly get attributed as timeouts by this. If we find that becomes an issue, we can add an extra telemetry field that records whether we hit a timeout, but as long as that's fairly rare (which I expect to be, because most normal upload/downloads we've seen thus far are very far from the timeout) I don't think it warrants an extra field and the associated complexity that threading that through the code entails.

henrymercer:
> Our telemetry will record an upload/download time that's very near to the timeout.

IIRC we measure the aggregate upload / download time over all languages.  So I think we could tell whether we timed out when analyzing a single language, but not when analyzing multiple languages (unless they all timed out).  It should be an unlikely case though, so I wouldn't personally push for extra telemetry fields at this time.

src/trap-caching.ts
```diff
@@ -159,9 +172,16 @@ export async function uploadTrapCaches(
...
     logger.info(`Uploading TRAP cache to Actions cache with key ${key}`);
-    toAwait.push(cache.saveCache([cacheDir], key));
+    await withTimeout(
+      MAX_CACHE_OPERATION_MS,
+      cache.saveCache([cacheDir], key),
+      () => {
+        logger.info(
+          `Timed out waiting for TRAP cache for ${language} to upload, will continue without uploading`
+        );
+      }
```
henrymercer, line 182: Minor, optional: You can remove the curly braces here and in other calls like this.

henrymercer (approved):

edoardopirovano:
> So I think we could tell whether we timed out when analyzing a single language, but not when analyzing multiple languages (unless they all timed out).

Yes, it will be a little tricker to infer from the timings whether timeouts occurred when there's more than one language. We may want to consider adding an extra field then, although I hope this will be rare enough to not warrant it.

src/trap-caching.ts
edoardopirovano: Let's leave them in. I personally prefer how it reads with the curly braces, since it makes it clearer that this is part of the callback. If we ever want to add a second statement here (e.g. to record a telemetry field saying we timed out), then it will also make the diff then cleaner.

PR #1281, opened 2022-09-30
Add logging statements declaring state of the cli_config_file_enabled

aeisenberg:
Minor change to insert a log statement to keep track of the new feature flag. It's possible to determine this otherwise, but this makes it easier to spot.

Also, [Avoid using single value as array](https://github.com/github/codeql-action/pull/1281/commits/59fbe348619eb3b019f408cc24f2e8293f40dd2d)

The user config parser in the CLI doesn't yet support it.


src/analyze.ts
```diff
@@ -239,6 +239,17 @@ export async function runQueries(
   }
 
   const codeql = await getCodeQL(config.codeQLCmd);
+
+  if (await util.useCodeScanningConfigInCli(codeql, featureFlags)) {
+    logger.info(
+      "Code Scanning configuration file being processed in the codeql-action."
```
adityasharad, line 245: Wrong way round!

src/config-utils.ts
```diff
@@ -1705,6 +1705,9 @@ export async function initConfig(
   // happen in the CLI during the `database init` command, so no need
   // to download them here.
   if (!(await useCodeScanningConfigInCli(codeQL, featureFlags))) {
+    logger.info(
```
adityasharad, line 1708: Maybe log within the check function itself?

aeisenberg: I'd prefer avoiding that since the function is called several times and it would mean the log message is printed multiple times. I just want it once for `init` and once for `analyze`.

src/analyze.ts
aeisenberg: Arrrgh!

aeisenberg:
Oh...need to change how the query filters tests work. Some of them are explicitly expecting the config to be processed by the action and others the CLI. I will need to change how the env var is working for them.

src/util.ts
```diff
@@ -817,6 +817,22 @@ export async function useCodeScanningConfigInCli(
   return await codeQlVersionAbove(codeql, CODEQL_VERSION_CONFIG_FILES);
 }
 
+export async function logCodeScanningConfgInCli(
```
adityasharad, line 820: Minor typo in name.

adityasharad (approved):

src/config-utils.ts
```diff
@@ -1704,6 +1705,8 @@ export async function initConfig(
   // When using the codescanning config in the CLI, pack downloads
   // happen in the CLI during the `database init` command, so no need
   // to download them here.
+  await logCodeScanningConfigInCli(codeQL, featureFlags, logger);
+
   if (!(await useCodeScanningConfigInCli(codeQL, featureFlags))) {
```
adityasharad, line 1710: This will hit the feature flags endpoint twice right? (And I think we don't attempt to add cache parameters to the API requests.) Can we make it call only once? Not sure of the cleanest way; perhaps the `logCodeScanningConfigInCli` function should just accept the boolean result of the `use` function, without calling `use`.

aeisenberg: Hmmm...you're right. Feature flags values are not being cached. Rather than implementing a workaround for `useCodeScanningConfigInCli` only, I think it makes more sense to simply add the caching (it's not too much code).

What do you think about my doing this in a separate PR?

aeisenberg: Wait!!!! It's a little bit hidden, but the response _is_ cached: https://github.com/aeisenberg/codeql-action/blob/5960ce11900cc253a1ac219eb7cd41769b00cebf/src/feature-flags.ts#L98-L99

adityasharad: Oh fancy! Nothing to worry about then.

adityasharad (approved):

PR #1282, opened 2022-10-04
Convert "Invalid source root" errors to UserErrors

aeisenberg:
This is an internal change. No changelog note required.


src/init.ts
```diff
@@ -117,9 +117,10 @@ export async function runInit(
       }
     }
   } catch (e) {
-    // Handle the situation where init is called twice
-    // for the same database in the same job.
+    // Check if we need to re-throw the error as a UserError in order to avoid
+    // counting this error towards our internal error budget.
     if (
```
adityasharad, line 122: Time to extract these conditions into their own `shouldReportAsUserError` helper function?

aeisenberg: sure

aeisenberg:
Rebasing in order to pick up a fix in integration tests from https://github.com/github/codeql-action/pull/1281.

src/init.ts
```diff
@@ -144,6 +127,40 @@ export async function runInit(
...
+      `Is the "init" action called twice in the same job? ${e.message}`
+    );
+  } else if (
+    // Version of CodeQL CLI is incompatible with this version of the CodeQL Action
+    e.message?.includes("is not compatible with this CodeQL CLI") ||
+    // Expected source location for database creation does not exist
+    e.message?.includes("Invalid source root")
+  ) {
+    return new util.UserError(e.message);
+  } else {
```
adityasharad, line 159: Similar, this can just be a final fall-through return statement.

adityasharad (dismissed):
Looks good, minor suggestions only.

src/init.ts
```diff
@@ -144,6 +127,42 @@ export async function runInit(
...
+ * Possibly convert this error into a UserError in order to avoid
+ * counting this error towards our internal error budget.
+ *
+ * @param e The error to possibly convert to a UserError.
+ *
+ * @returns A UserError if the error is a known error that can be
+ *         attributed to the user, otherwise the original error.
+ */
+function processError(e: any): Error {
+  if (!(e instanceof Error)) {
```
angelapwen, line 140: Just curious: when would `e` not be an `instanceof Error` if it was sent from the `try/catch` block? Or is this just good practice in any error handling function?

src/init.ts
```diff
@@ -144,6 +127,42 @@ export async function runInit(
...
+    return e;
+  }
+
+  if (
+    // Init action called twice
+    e.message?.includes("Refusing to create databases") &&
+    e.message.includes("exists and is not an empty directory.")
```
angelapwen, line 147:
```suggestion
    e.message?.includes("exists and is not an empty directory.")
```

angelapwen (commented):
Just a question/suggestion, not approving yet because it'll probably be dismissed by another commit anyway.

src/init.ts
aeisenberg: In javascript, you can "throw" anything. So you are not guaranteed that a catch clause is actually catching an `Error` instance. In practice, you should only throw errors. Typescript is just being pedantic here, and so is this code.

src/init.ts
aeisenberg: Oops. This commit is going to fail since the generated code is not up to date.

angelapwen: Ah, I see, that makes sense!

angelapwen (approved):

PR #1284, opened 2022-10-05
Refactor handling of feature flags

aeisenberg:
This commit centralizes how feature flags are handled. All feature flags must now add an entry in the `featureFlagConfig` dictionary. This dictionary associates the flag with an environment variable name and optionally a minimum version for CodeQL.

The new logic is:

- if the environment variable is set to false: disabled
- if the minimum version requirement specified and not met: disabled
- if a minimum version requirement is specified, but a codeql instance is not passed in: throw
- if the environment variable is set to true: enable
- Otherwise check feature flag enablement from the server


src/feature-flags.test.ts
```diff
@@ -42,14 +48,11 @@ const ALL_FEATURE_FLAGS_DISABLED_VARIANTS: Array<{
...
     await withTmpDir(async (tmpDir) => {
-      setupActionsVars(tmpDir, tmpDir);
-
       const loggedMessages = [];
-      const featureFlags = new GitHubFeatureFlags(
-        variant.gitHubVersion,
-        testApiDetails,
-        testRepositoryNwo,
-        getRecordingLogger(loggedMessages)
+      const featureFlags = setUpTmpDir(
```
henrymercer, line 52: Minor: Change to `setUpTest` instead so it's more obvious why this returns some feature flags

src/feature-flags.test.ts
```diff
@@ -153,34 +141,128 @@ test("Feature flags exception is propagated if the API request errors", async (t
...
+  test(`Feature flag '${featureFlag}' is enabled if the associated environment variable is true`, async (t) => {
+    await withTmpDir(async (tmpDir) => {
+      const featureFlags = setUpTmpDir(tmpDir);
+
+      // set all feature flags to false
       const expectedFeatureFlags: { [flag: string]: boolean } = {};
-      for (const f of FEATURE_FLAGS) {
+      for (const f of Object.keys(featureFlagConfig)) {
         expectedFeatureFlags[f] = false;
       }
```
henrymercer, line 175: Minor: Extract this into a utility method (`createFeatureFlags(f: FeatureFlag => boolean)`?)

src/feature-flags.ts
```diff
@@ -15,6 +16,32 @@ export enum FeatureFlag {
   CliConfigFileEnabled = "cli_config_file_enabled",
 }
 
+export const featureFlagConfig: Record<
```
henrymercer, line 19: Nit: alphabetical order please

src/feature-flags.ts
```diff
@@ -33,12 +60,35 @@ export class GitHubFeatureFlags implements FeatureFlags {
     private logger: Logger
   ) {}
 
-  async getValue(flag: FeatureFlag): Promise<boolean> {
+  async getValue(flag: FeatureFlag, codeql?: CodeQL): Promise<boolean> {
```
henrymercer, line 63: Q: Should it be possible to disable any given feature flag via an Action input?  Or is that too much plumbing?

henrymercer (commented):
Thanks for doing this.  As a general piece of feedback, I wonder whether it would be appropriate to rename `FeatureFlags` to `Features`.  This would reflect a distinction between "feature flags", which are flippers on the monolith, and "features", which are functions in the CodeQL Action whose enablement depends on environment variables, CLI versions, and feature flags.

src/feature-flags.test.ts
```diff
@@ -86,7 +85,7 @@ test("API response missing", async (t) => {
       t.assert((await featureFlags.getValue(flag)) === false);
     }
 
-    for (const featureFlag of ["ml_powered_queries_enabled"]) {
+    for (const featureFlag of Object.keys(featureFlagConfig)) {
```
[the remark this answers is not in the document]
aeisenberg: My opinion is that the advantage is small to doing all, but it does ensure that we handle all flags the same. Also, I don't think the cost of running against all flags is high, so I would lean towards doing it this way.

src/feature-flags.ts
```diff
@@ -33,12 +60,35 @@ export class GitHubFeatureFlags implements FeatureFlags {
...
+      process.env[featureFlagConfig[flag].envVar] || ""
+    ).toLocaleLowerCase();
+
+    // Do not use this feature if user explicitly disables it via an environment variable.
+    if (envVar === "false") {
+      return false;
+    }
+
+    // Never use this feature if the CLI version explicitly can't support it.
+    const minimumVersion = featureFlagConfig[flag].minimumVersion;
```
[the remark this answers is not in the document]
aeisenberg: So, we throw if a `minimumVersion` is set, but no codeql is passed in? I think that makes sense.

src/feature-flags.ts
aeisenberg: I feel like that's too much plumbing. Users should not generally be made aware of a feature flag. Allowing them to toggle them on or off is really just a back door to be used in extreme cases and testing. There are already a lot of inputs for our actions. And doing this would mean we'd need to add an input both for `init` and `analyze` (and possibly `upload-sarif` too).

src/feature-flags.ts
```diff
@@ -15,6 +16,32 @@ export enum FeatureFlag {
...
+export const featureFlagConfig: Record<
+  FeatureFlag,
+  { envVar: string; minimumVersion: string | undefined }
+> = {
+  [FeatureFlag.BypassToolcacheEnabled]: {
+    envVar: "CODEQL_BYPASS_TOOLCACHE",
+    minimumVersion: undefined,
+  },
+  [FeatureFlag.MlPoweredQueriesEnabled]: {
+    envVar: "CODEQL_VERSION_ML_POWERED_QUERIES",
```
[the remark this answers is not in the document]
aeisenberg: Would changing the env var affect existing repos?

aeisenberg:
> I wonder whether it would be appropriate to rename `FeatureFlags` to `Features`. 

Sure, that makes sense. Maybe `FeatureEnablement` is better?

src/feature-flags.ts
henrymercer: I suppose if a user really wants to disable a feature flag, they can always set the environment variable.

src/feature-flags.ts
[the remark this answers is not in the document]
henrymercer: That was never an environment variable.  I think before it was a static constant in the code.

src/feature-flags.ts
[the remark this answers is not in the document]
henrymercer: It would probably be theoretically possible to enforce this via the type system, but potentially tricky to do in practice.

src/feature-flags.ts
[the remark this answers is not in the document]
aeisenberg: Ah...then I will change.

src/feature-flags.test.ts
```diff
@@ -153,34 +141,128 @@ test("Feature flags exception is propagated if the API request errors", async (t
...
+
+        // set all feature flags to true
+        const expectedFeatureFlags: { [flag: string]: boolean } = {};
+        for (const f of Object.keys(featureFlagConfig)) {
+          expectedFeatureFlags[f] = true;
+        }
+        mockFeatureFlagApiEndpoint(200, expectedFeatureFlags);
+
+        // feature flag should be enabled initially (ignoring the minimum CLI version)
+        t.assert(await featureFlags.getValue(featureFlag as FeatureFlag));
```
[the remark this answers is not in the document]
aeisenberg: Since I will be changing behaviour so that calling `getValue` without a codeql instance but with a minimum value specified will be an error, this check no longer makes sense.

henrymercer:
> Sure, that makes sense. Maybe `FeatureEnablement` is better?

Sure, that works too.

src/feature-flags.ts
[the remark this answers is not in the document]
aeisenberg: I don't think we should change this to being language specific since the FF on github.com is not language specific. This could cause confusion.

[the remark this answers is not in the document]
henrymercer: Makes sense.  I think we'll want to avoid encouraging anyone to use this environment variable outside of PR checks, but that's probably true anyway.

aeisenberg:
Actually, I like `Features` better since the names are used for more than just enablement.

aeisenberg:
Addressed all comments. Please look at the two commits individually since the second commit adds a lot of noise through a renaming.

src/feature-flags.ts
```diff
@@ -1,27 +1,54 @@
 import { getApiClient, GitHubApiDetails } from "./api-client";
+import { CodeQL } from "./codeql";
 import { Logger } from "./logging";
 import { RepositoryNwo } from "./repository";
 import * as util from "./util";
 
 export interface FeatureFlags {
-  getValue(flag: FeatureFlag): Promise<boolean>;
+  getValue(flag: Feature, codeql?: CodeQL): Promise<boolean>;
 }
```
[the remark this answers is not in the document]
aeisenberg: This is a little bit of bike shedding, but I thought this should remain `FeatureFlags` since it really is an interface that checks whether a `Feature` is turned on or off. Also, the concrete instantiation of this is `GitHubFeatureFlags`.

src/feature-flags.test.ts
```diff
@@ -142,9 +130,9 @@ test("Feature flags exception is propagated if the API request errors", async (t
 });
 
 for (const featureFlag of Object.keys(featureFlagConfig)) {
```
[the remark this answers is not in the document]
aeisenberg: I guess there might be a point in the future where we have no more feature flags with minimum versions (that would be glorious! but unlikely), and if so a bunch of these tests would no longer run. I can add a test that will fail if there are no more `minimumVersion` flags in order to remind us that we need to change the testing structure.

src/feature-flags.ts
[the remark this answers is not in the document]
henrymercer: My thinking was that `FeatureFlags.getValue(x)` tells us whether the feature x should be enabled, which as of this PR is no longer the same as whether the feature flag on the monolith is enabled for x.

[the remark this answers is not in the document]
henrymercer: Agree we're in bikeshedding territory here, but this interface is also public facing and is really about feature enablement, not feature flags.  Couple of suggestions:
- Rename this interface to `Features`, and the class that implements it to `GithubFeatures`
- Rename this interface to `FeatureEnablement`

src/feature-flags.ts
```diff
@@ -1,44 +1,141 @@
...
+   * @param codeql An optional CodeQL object. If provided, and a `minimumVersion` is specified for the
+   *        feature flag, the version of the CodeQL CLI will be checked against the minimum version.
+   *        If the version is less than the minimum version, the feature flag will be considered
+   *        disabled. If not provided, and a `minimumVersion` is specified for the feature flag, the
+   *        this function will throw.
+   * @returns true if the feature flag is enabled, false otherwise.
+   *
+   * @throws if a `minimumVersion` is specified for the feature flag, and `codeql` is not provided.
+   */
+  async getValue(flag: Feature, codeql?: CodeQL): Promise<boolean> {
```
henrymercer, line 87: Consistency: Rename `flag` -> `feature`

src/feature-flags.ts
```diff
@@ -1,44 +1,141 @@
...
+    if (!codeql && featureConfig[flag].minimumVersion) {
+      throw new Error(
+        `Internal error: A minimum version is specified for feature flag ${flag}, but no instance of CodeQL was provided.`
```
henrymercer, line 90:
```suggestion
        `Internal error: A minimum version is specified for feature ${feature}, but no instance of CodeQL was provided.`
```

src/feature-flags.ts
```diff
@@ -1,44 +1,141 @@
...
+      );
+    }
+
+    // Bypassing the toolcache is disabled in test mode.
+    if (flag === Feature.BypassToolcacheEnabled && util.isInTestMode()) {
+      return false;
+    }
```
henrymercer, line 97: Optional: Make this a property of the `featureConfig` record.

src/feature-flags.ts
```diff
@@ -53,51 +150,52 @@ export class GitHubFeatureFlags implements FeatureFlags {
       );
       return false;
     }
-    return flagValue;
+    return flagValue || false;
```
henrymercer, line 153: For my understanding: When would we need the `||` here?  We already tackled the `undefined` case above.

src/feature-flags.ts
```diff
@@ -53,51 +150,52 @@ export class GitHubFeatureFlags implements FeatureFlags {
...
+      this.cachedApiResponse || (await this.loadApiResponse());
+    this.cachedApiResponse = apiResponse;
+    return apiResponse;
+  }
+
+  private async loadApiResponse() {
+    // Do nothing when not running against github.com
+    if (this.gitHubVersion.type !== util.GitHubVariant.DOTCOM) {
+      this.logger.debug(
+        "Not running against github.com. Disabling all feature flags."
```
henrymercer, line 167:
```suggestion
        "Not running against github.com. Feature flags will only be enabled if they are enabled explicitly, for example through an environment variable."
```

src/feature-flags.ts
```diff
@@ -106,7 +204,7 @@ export class GitHubFeatureFlags implements FeatureFlags {
  *
  * This should be only used within tests.
  */
-export function createFeatureFlags(enabledFlags: FeatureFlag[]): FeatureFlags {
+export function createFeatureFlags(enabledFlags: Feature[]): FeatureFlags {
```
henrymercer, line 207: Rename this if you go with my suggestion to rename the `FeatureFlags` interface

src/init-action.ts
```diff
@@ -157,7 +157,7 @@ async function run() {
     getRequiredEnvParam("GITHUB_REPOSITORY")
   );
 
-  const featureFlags = new GitHubFeatureFlags(
+  const featureFlags = new Features(
```
henrymercer, line 160: These variable assignments ought to be renamed too.  It's a shame it's quite a lot of noise.

src/init-action.ts
```diff
@@ -321,8 +321,10 @@ async function getTrapCachingEnabled(
   featureFlags: FeatureFlags
 ): Promise<boolean> {
   const trapCaching = getOptionalInput("trap-caching");
-  if (trapCaching !== undefined) return trapCaching === "true";
-  return await featureFlags.getValue(FeatureFlag.TrapCachingEnabled);
+  if (trapCaching !== undefined) {
+    return trapCaching === "true";
+  }
+  return await featureFlags.getValue(Feature.TrapCachingEnabled);
```
henrymercer, line 327: Optional

```suggestion
  return trapCaching === "true" || (await featureFlags.getValue(Feature.TrapCachingEnabled));
```

henrymercer (commented):
This is looking generally in good shape to me.  Some follow up suggestions, some of which are optional.

src/feature-flags.ts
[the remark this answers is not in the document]
aeisenberg: All right. A slightly more complex structure now, but I think it is better:

- Interface is `FeatureEnablement`
- The public facing class is `Features`
- This delegates to another internal class `GitHubFeatureFlags` for the portion of the enablement checking that needs to get information from github.

src/feature-flags.ts
aeisenberg: We are getting this information from an API. I just wanted to ensure the value was coerced into a boolean. I should have put:

```
    return !!flagValue;
```

and I will change.

src/feature-flags.ts
aeisenberg: Going to use `toggleable features` instead of feature flags since that seems more appropriate. "Feature flags" is only when they are coming from github. "Features" is the more general concept.

However, here you are suggesting we mention environment variables, but we don't have any place where we actually document them, so it might be confusing. So, I think it would be better if we just do:

```
Not running against github.com. Disabling all toggleable features.
```

angelapwen:
From PR body — 
> The new logic is:
...
> if the minimum version requirement specified and met: disabled

Should the above be `if the minimum version requirement specified is not met`?

src/feature-flags.ts
```diff
@@ -1,115 +1,200 @@
 import { getApiClient, GitHubApiDetails } from "./api-client";
+import { CodeQL } from "./codeql";
 import { Logger } from "./logging";
 import { RepositoryNwo } from "./repository";
 import * as util from "./util";
 
-export interface FeatureFlags {
-  getValue(flag: FeatureFlag): Promise<boolean>;
+export interface FeatureEnablement {
+  getValue(feaature: Feature, codeql?: CodeQL): Promise<boolean>;
```
angelapwen, line 8:
```suggestion
  getValue(feature: Feature, codeql?: CodeQL): Promise<boolean>;
```

(+autogenerated code)

src/feature-flags.ts
```diff
@@ -1,115 +1,200 @@
...
+    } catch (e) {
+      if (util.isHTTPError(e) && e.status === 403) {
+        this.logger.warning(
+          "This run of the CodeQL Action does not have permission to access Code Scanning API endpoints. " +
+            "As a result, it will not be opted into any experimental features. " +
+            "This could be because the Action is running on a pull request from a fork. If not, " +
+            `please ensure the Action has the 'security-events: write' permission. Details: ${e}`
+        );
+      } else {
+        // Some feature, such as `ml_powered_queries_enabled` affect the produced alerts.
```
angelapwen, line 190:
```suggestion
        // Some features, such as `ml_powered_queries_enabled` affect the produced alerts.
```

src/feature-flags.ts
```diff
@@ -1,115 +1,200 @@
...
+        // Considering these feature disabled in the event of a transient error could
```
angelapwen, line 191:
```suggestion
        // Considering these features disabled in the event of a transient error could
```

src/feature-flags.test.ts
```diff
@@ -70,117 +75,242 @@ for (const variant of ALL_FEATURE_FLAGS_DISABLED_VARIANTS) {
...
-      async () => featureFlags.getValue(FeatureFlag.MlPoweredQueriesEnabled),
+      async () =>
+        featureFlags.getValue(
+          Feature.MlPoweredQueriesEnabled,
+          includeCodeQlIfRequired(Feature.MlPoweredQueriesEnabled)
+        ),
       {
         message:
-          "Encountered an error while trying to load feature flags: Error: some error message",
+          "Encountered an error while trying to determine feature enablement: Error: some error message",
```
angelapwen, line 125: Should `some error message` be populated? 🤔

src/feature-flags.test.ts
```diff
@@ -70,117 +75,242 @@ for (const variant of ALL_FEATURE_FLAGS_DISABLED_VARIANTS) {
...
+      });
+    });
+  }
+}
+
+// If we ever run into a situation where we no longer have any feature flags that
+// specify a minimum version, then we will have a bunch of code no longer being
+// tested. This is unlikely, and this test will fail if that happens.
+// If we do end up in that situation, then we should consider adding a synthetic
+// feature flag with a minium version that is only used for tests.
```
angelapwen, line 268:
```suggestion
// feature flag with a minimum version that is only used for tests.
```

angelapwen (commented):
Just small comment typo nits 😸 

Thanks for the code comments + clear variable naming!

src/feature-flags.test.ts
aeisenberg: No. This is coming from a mocked api response in the test.

aeisenberg:
Done!

src/feature-flags.test.ts
angelapwen: Ah, I see!

src/codeql.test.ts
```diff
@@ -76,7 +80,7 @@ async function mockApiAndSetupCodeQL({
   version,
 }: {
   apiDetails?: GitHubApiDetails;
-  featureFlags?: FeatureFlags;
+  featureFlags?: FeatureEnablement;
```
henrymercer, line 83: This should be `features` or `featureEnablement`

src/codeql.test.ts
```diff
@@ -505,47 +509,41 @@ const injectedConfigMacro = test.macro({
...
+        tempDir,
+        augmentationProperties,
+      };
+
+      await codeqlObject.databaseInitCluster(
+        thisStubConfig,
+        "",
+        undefined,
+        undefined,
+        createFeatureFlags([Feature.CliConfigFileEnabled]),
```
henrymercer, line 531: This should be `createFeatures` or `createFeatureEnablement`

src/feature-flags.test.ts
```diff
@@ -70,117 +75,242 @@ for (const variant of ALL_FEATURE_FLAGS_DISABLED_VARIANTS) {
...
+      // retrieve the values of the actual feature flags
+      const actualFeatureFlags: { [flag: string]: boolean } = {};
+      for (const f of Object.keys(featureConfig)) {
+        actualFeatureFlags[f] = await featureFlags.getValue(
+          f as Feature,
+          includeCodeQlIfRequired(f)
+        );
+      }
 
+      // Alls flags should be false except the one we're testing
```
henrymercer, line 152:
```suggestion
      // All flags should be false except the one we're testing
```

src/feature-flags.test.ts
```diff
@@ -42,26 +48,25 @@ const ALL_FEATURE_FLAGS_DISABLED_VARIANTS: Array<{
...
     await withTmpDir(async (tmpDir) => {
-      setupActionsVars(tmpDir, tmpDir);
-
       const loggedMessages = [];
-      const featureFlags = new GitHubFeatureFlags(
-        variant.gitHubVersion,
-        testApiDetails,
-        testRepositoryNwo,
-        getRecordingLogger(loggedMessages)
+      const featureFlags = setUpTests(
```
henrymercer, line 52: I think this should be `features` / `featureEnablement` throughout.

src/feature-flags.ts
```diff
@@ -1,115 +1,200 @@
...
+      gitHubVersion,
+      apiDetails,
+      repositoryNwo,
+      logger
+    );
+  }
+
+  /**
+   *
+   * @param feature The feature flag to check.
```
henrymercer, line 77:
```suggestion
   * @param feature The feature to check.
```

src/feature-flags.ts
```diff
@@ -1,115 +1,200 @@
...
+    if (envVar === "true") {
+      return true;
+    }
+
+    // Ask the GitHub API if the feature is enabled.
+    return await this.gitHubFeatureFlags.getValue(feature);
+  }
+}
+
+class GitHubFeatureFlags implements FeatureEnablement {
```
henrymercer, line 126: Perhaps add a doc comment here saying something like "Feature enablement based solely on the GitHub API. Most of the time you'll want to use `Features` instead."

src/feature-flags.ts
```diff
@@ -1,115 +1,200 @@
 import { getApiClient, GitHubApiDetails } from "./api-client";
+import { CodeQL } from "./codeql";
```
henrymercer, line 2: Consider renaming this file to `features` or `feature-enablement`.

henrymercer (commented):
There are a bunch of uses of `featureFlags` across the codebase that probably ought to become `features` or `featureEnablement`.  I could be convinced to merge this and clean those up as follow up though.

aeisenberg:
Let's just get it all done now.

henrymercer:
> Let's just get it all done now.

Sounds good to me!  I've pushed a commit to another branch with some more renaming suggestions to avoid needing to make them all in the UI :) https://github.com/github/codeql-action/compare/aeisenberg/ff-refactoring...henrymercer/more-ff-renaming?expand=1.

henrymercer (dismissed):
Partial review up to the penultimate commit, since I authored the last one.

aeisenberg:
I reviewed the last commit. 🤠

aeisenberg:
@henrymercer, can you give another approval? Not sure why your previous one was dismissed.

henrymercer (approved):

PR #1293, opened 2022-10-11
Fix `withTimeout` helper function

edoardopirovano:
Noticed while looking at logs during the rollout of TRAP caching.

It turns out that I had slightly misunderstood the semantics of `Promise.race`. While it returns the value of the first promise that finishes, it does not actually halt the other promise (there is, as far as I now understand things, no way to do this in JavaScript). This means we were eventually calling the timeout callback even if the task completed. This isn't a big deal because the timeout callback just logged an informational message. Still, it's not ideal and creates confusing logs since we log a message about the download timing out some time after it has in fact already finished.

This PR fixes that, and adds a unit test that would've caught this. It's still not ideal that when we _do_ timeout we don't halt the download in the background but I think that is okay since if we've timed out it probably means that has got stuck somehow so isn't using network resources. Actually cancelling the download would require re-writing parts of the Actions cache library to support cancellation, which is outside of our team's scope (although it's certainly a sensible feature request we should consider filing with the Actions team).


jbj:
> It's still not ideal that when we _do_ timeout we don't halt the download in the background but I think that is okay since if we've timed out it probably means that has got stuck somehow so isn't using network resources.

If the download keeps being stuck in the background, will the NodeJS runtime then be allowed to shut down when the action is done with all its foreground tasks? Can you verify this?

jbj:
Would it help to call `process.exit` at the end of `runWrapper`?

edoardopirovano:
> Would it help to call `process.exit` at the end of `runWrapper`?

This works, but is a little scary because we might also terminate other async tasks that are still pending.

I had a conversation with @henrymercer about this, and he spotted that https://github.com/actions/toolkit/pull/1140 added a timeout option for the downloading, which seems like a better idea than our workaround. I've bumped the version of the library we pull in so that we have that PR and started using that new option. The timeout is per-segment rather than for the whole cache, but as we are bounding our caches to 1GB they will always fit in one segment (which is 2GB on a 64-bit machine) so this isn't an issue.

For the uploading, there is no such option. I propose we leave the workaround in place there. It's not ideal that we won't terminate the workflow until we hit the timeout for the run, but at least with our workaround the code will run to completion before we hang which means that results will get uploaded to the Code Scanning, the database will make it to MRVA if applicable, etc. This should still hopefully be a rare case, and we can monitor with our telemetry to see if it becomes a concern and then perhaps think about more aggressive solutions like using `process.exit()` or spawning a new process for the upload that we can kill independently.

In the future, of course, the ideal solution would be for the Actions library to just expose a timeout option in the upload case too.

aeisenberg (approved):
This looks reasonable.

src/util.ts
```diff
@@ -899,6 +899,9 @@ export async function tryGetFolderBytes(
  * Run a promise for a given amount of time, and if it doesn't resolve within
  * that time, call the provided callback and then return undefined.
  *
+ * Note that this does NOT cancel the original promise, so that promise will
+ * continue in the background even after the timeout has expired.
```
henrymercer, line 903:
```suggestion
 * Important: This does NOT cancel the original promise, so that promise will continue in the
 * background even after the timeout has expired. If the original promise hangs, then this will
 * prevent the process terminating.
```

henrymercer (approved):
Small suggestion around documenting some of what we learnt.  Unfortunately accepting this requires rebuilding the Action and reapproving the PR, so feel free to skip it.

src/util.ts
edoardopirovano: Thanks, that is indeed better. Will do this in a follow-up to avoid dismissing the approvals here.

PR #1307, opened 2022-10-17
Fix merge conflicts with v2 into main

aeisenberg:
This commit merges the main into releases/v2 so that we can avoid conflicts when we do a mergeback from releases/v2 into main.


henrymercer (commented):
This seems to merge new changes in `main` into `releases/v2` too, meaning we are effectively releasing them, just without bumping the version number / changelog and publishing the tag.

Would it make more sense to merge in just the changes from the last release (and possibly the mergeback too if required), such that this diff is empty?

aeisenberg:
Ah..yes. Let me do this again.

PR #1321, opened 2022-10-25
Actions status report: Send testing_environment.

criemen:
The testing environment is taken from the environment variable `CODEQL_ACTION_TESTING_ENVIRONMENT`.

I guess the main question for the reviewers is if the environment variable should be set again using the actions environment mechanism to preserve it for future steps.
But I think we get away with a top-level [env](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env) block, or a per-job env entry that we can use.

CC @esbena 


henrymercer (approved):
> I guess the main question for the reviewers is if the [env](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env)ironment variable should be set again using the actions environment mechanism to preserve it for future steps.

I like this idea — it sounds like this would be one line of code that could prevent some headaches.

criemen:
Rebased on top of main and added the call to `exportVariable`.

src/actions-util.ts
```diff
@@ -600,6 +600,12 @@ export interface StatusReportBase {
   completed_at?: string;
   /** State this action is currently in. */
   status: ActionStatus;
+  /**
+   * Testing environment: Set if non-production environment.
+   * The server accepts one of the following values:
+   *  ["", "qa-rc", "qa-rc-1", "qa-rc-2", "qa-experiment-1", "qa-experiment-2", "qa-experiment-3"].
```
henrymercer, line 606:
```suggestion
   *  `["", "qa-rc", "qa-rc-1", "qa-rc-2", "qa-experiment-1", "qa-experiment-2", "qa-experiment-3"]`.
```

henrymercer (dismissed):

henrymercer (approved):

PR #1322, opened 2022-10-25
Deprecate Go extraction reconciliation feature flag and `CODEQL_EXTRACTOR_GO_BUILD_TRACING` for custom builds

angelapwen:
This PR removes the `CODEQL_EXTRACTOR_GO_BUILD_TRACING` from all tests as well as another unused environment variable, `CODEQL_ACTION_RECONCILE_TRACING`. Users who previously had `CODEQL_EXTRACTOR_GO_BUILD_TRACING` set with custom Go builds will be shown a warning suggesting that they remove the environment variable as it no longer has effect. 

Additionally, this PR deprecates the use of the Go extraction reconciliation feature flag so that its functionality is always on (ie. Go will always be treated as a traced language). A separate PR removing the flag from dotcom will be merged after this PR.


src/languages.ts
```diff
@@ -45,22 +43,15 @@ export function isTracedLanguage(
   isGoExtractionReconciliationEnabled: boolean,
   logger: Logger
 ): boolean {
-  if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
+  if ("CODEQL_EXTRACTOR_GO_BUILD_TRACING" in process.env) {
     logger.warning(
-      "The CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable was set to 'true', but it must " +
-        "be 'on' to enable Go build tracing. Setting it to 'on'."
+      "The CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable is deprecated and its behavior is now the default behavior."
```
angelapwen, line 48: Some questions: should this be a `logger.info()`? And is this an appropriate message or should we be more explicit about suggesting they remove it from their workflow?

pr-checks/checks/go-reconciled-tracing-autobuilder.yml
```diff
@@ -2,7 +2,6 @@ name: "Go: Reconciled tracing with autobuilder"
 description: "Checks that Go reconciled tracing works when using an autobuilder step"
 os: ["ubuntu-latest", "macos-latest"]
 env:
-  CODEQL_ACTION_RECONCILE_GO_EXTRACTION: "true"
```
henrymercer, line 5: Now that Go reconciled tracing is fully rolled out, our matrix of configurations that we need to test has reduced.  I think we should therefore be able to switch to three PR checks analyzing Go: (a) workflow with custom build steps, (b) workflow with autobuild step, and (c) legacy workflow.

src/languages.ts
henrymercer: I think `logger.warning` is fine — `logger.info` probably won't ever be seen.

Suggestion around the message, let's loop Alona in on this too before merging:

```suggestion
      "Go build tracing is now enabled by default, so the CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable which was previously used to manually enable Go build tracing is now deprecated. We recommend that you remove this environment variable from your workflow."
```

src/languages.ts
```diff
@@ -45,22 +43,15 @@ export function isTracedLanguage(
...
   }
 
-  const shouldTraceGo =
-    process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "on" ||
-    isGoExtractionReconciliationEnabled;
-
   return (
     ["cpp", "java", "csharp", "swift"].includes(language) ||
-    (shouldTraceGo && language === Language.go)
+    (isGoExtractionReconciliationEnabled && language === Language.go)
```
henrymercer, line 54: Will you remove the feature flag check in a future PR?

henrymercer (dismissed):

henrymercer:
> I'm not sure if the variable deprecation should be part of the changelog here or not 🤔

I think it would make sense to add a changelog note in the PR that removes the feature flag check similar to [the one we used before starting rollout](https://github.com/github/codeql-action/blob/main/CHANGELOG.md#2127---06-oct-2022), saying that we have now rolled out this change.  You could then mention the variable deprecation there too — it won't affect most users but the Actions changelog can be more detailed.

angelapwen:
> > I'm not sure if the variable deprecation should be part of the changelog here or not 🤔
> 
> I think it would make sense to add a changelog note in the PR that removes the feature flag check similar to [the one we used before starting rollout](https://github.com/github/codeql-action/blob/main/CHANGELOG.md#2127---06-oct-2022), saying that we have now rolled out this change. You could then mention the variable deprecation there too — it won't affect most users but the Actions changelog can be more detailed.

👍 I'll make a note to do so in that PR

src/languages.ts
henrymercer: Makes sense.  I think we might want to hold off on this PR then until we remove the feature flag, such that we can create a consistent changelog note that encompasses both.

angelapwen: I'll go ahead and make this back into a draft PR for now 👍

pr-checks/checks/go-reconciled-tracing-autobuilder.yml
angelapwen: Done — I removed the extraneous PR checks in this PR and removed mention of "extraction reconciliation" 

FYI there is still a 4th Go-related PR check I left in, which checks custom queries for Go.

src/languages.ts
angelapwen: Adding the feature flag removal to this PR, and putting up a changelog note for both now.

src/languages.ts
```diff
@@ -40,38 +38,16 @@ export function parseLanguage(language: string): Language | undefined {
...
-  isGoExtractionReconciliationEnabled: boolean,
-  logger: Logger
-): boolean {
-  if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
+export function isTracedLanguage(language: Language, logger: Logger): boolean {
+  if ("CODEQL_EXTRACTOR_GO_BUILD_TRACING" in process.env) {
     logger.warning(
-      "The CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable was set to 'true', but it must " +
-        "be 'on' to enable Go build tracing. Setting it to 'on'."
+      "Go build tracing is now enabled by default, so the CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable which was previously used to manually enable Go build tracing is now deprecated. We recommend that you remove this environment variable from your workflow."
```
angelapwen, line 44: @AlonaHlobina — could you weigh in on the wording of this error message we'll surface to users if they have enabled the now deprecated prior environment variable? It'll surface as a warning for them in their workflow logs. Thank you!!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- We have completed rollout of the feature described in [CodeQL Action version 2.1.27](#2127---06-oct-2022): the way Go code is extracted for analysis is now consistent with that of other compiled languages such as C/C++, C#, and Java. The `CODEQL_EXTRACTOR_GO_BUILD_TRACING` environment variable, which was previously used to manually enable this behavior, is now deprecated; if your code scanning workflow uses this environment variable, we recommend that you remove it. Otherwise, you do not need to alter your code scanning workflows.
```
henrymercer, line 5:
```suggestion
- Go is now analyzed in the same way as other compiled languages such as C/C++, C#, and Java. This completes the rollout of the feature described in [CodeQL Action version 2.1.27](#2127---06-oct-2022). As a result, the `CODEQL_EXTRACTOR_GO_BUILD_TRACING` environment variable, which was previously used to manually enable this behavior, is now deprecated. If your code scanning workflow uses this environment variable, we recommend that you remove it. Otherwise, you do not need to alter your code scanning workflows.
```

src/languages.ts
```diff
@@ -40,38 +38,16 @@ export function parseLanguage(language: string): Language | undefined {
   return undefined;
 }
 
-export function isTracedLanguage(
-  language: Language,
-  isGoExtractionReconciliationEnabled: boolean,
-  logger: Logger
-): boolean {
-  if (process.env["CODEQL_EXTRACTOR_GO_BUILD_TRACING"] === "true") {
+export function isTracedLanguage(language: Language, logger: Logger): boolean {
```
henrymercer, line 41: Note that we still need to export `core.exportVariable("CODEQL_EXTRACTOR_GO_BUILD_TRACING", "on")` to tell the Go extractor that we're using build tracing.

angelapwen: Oh, I see — I wonder why the relevant [PR check](https://github.com/github/codeql-action/actions/runs/3397883575) didn't fail though.

henrymercer: Good question.  Let's file an issue to make a PR check that only passes when the extractor is in tracing mode.  One way to do that would be to change the Go tracing with autobuilder check to assert that the autobuilder outputs "Tracing enabled".  Ideally though we'd create a test repo that we can only analyze properly when the extractor is in tracing mode.

angelapwen: Filed at https://github.com/github/codeql-core/issues/2944 and I added the `core.exportVariable(...)` line back to this method. It's a bit awkward because I wouldn't expect a method called `isTracedLanguage` to do that, but because `isTracedLanguage` is called in so many different places I guess it makes sense to keep it here.

henrymercer: Good point about `isTracedLanguage`.  We could export the variable somewhere else, such as in `init-action.ts`.

angelapwen: Yeah, that's fair! We already check for Python analysis and install dependencies in `init-action` so I feel that makes sense. Updated.

src/languages.ts
angelapwen: Moved the message to another file — will tag you there instead!

src/init-action.ts
```diff
@@ -218,6 +218,15 @@ async function run() {
       logger
     );
 
+    if ("CODEQL_EXTRACTOR_GO_BUILD_TRACING" in process.env) {
+      logger.warning(
+        "Go build tracing is now enabled by default, so the CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable which was previously used to manually enable Go build tracing is now deprecated. We recommend that you remove this environment variable from your workflow."
```
angelapwen, line 223: @AlonaHlobina — could you weigh in on the wording of this error message we'll surface to users if they have enabled the now deprecated prior environment variable? It'll surface as a warning for them in their workflow logs. Thank you!!

angelapwen:
I ended up finding an error in the Debug Artifacts Upload PR check because of this change — I hadn't added the `actions/setup-go@v3` step in those PR checks so there wasn't any Go source code found for artifact upload. I didn't look into why it was previously passing, but as the PR checks are all passing now I think this is ready for re-review (also pending PM approval of the user-facing warning)

src/init-action.ts
AlonaHlobina: I notice that we do not use the term `build tracing` anywhere in docs. Is it right? Is it possible to align the terminology we use for an error with the documentation?
Also, Is it possible to add links to the doc? Soon we will have the tool status page, and the idea is to make all errors super clear and actionable. I like that you say what we recommend users to do `We recommend that you remove this environment variable from your workflow.`It would be great to add a way to read more about it.

CHANGELOG.md
```diff
@@ -14,6 +15,7 @@ No user facing changes.
 
 ## 2.1.29 - 26 Oct 2022
 
+- We have completed rollout of the feature described in [CodeQL Action version 2.1.27](#2127---06-oct-2022): the way Go code is extracted for analysis is now consistent with that of other compiled languages such as C/C++, C#, and Java. The `CODEQL_EXTRACTOR_GO_BUILD_TRACING` environment variable, which was previously used to manually enable this behavior, is now deprecated; if your code scanning workflow uses this environment variable, we recommend that you remove it. Otherwise, you do not need to alter your code scanning workflows.
```
henrymercer, line 18: I think this was accidentally duplicated.

```suggestion
```

src/init-action.ts
henrymercer: > I notice that we do not use the term build tracing anywhere in docs. Is it right? Is it possible to align the terminology we use for an error with the documentation?

I suppose we could say "Go is now analyzed in the same way as other compiled languages like C/C++, C#, and Java by default, so the CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable which was previously used to manually enable this functionality is now deprecated."

> Also, Is it possible to add links to the doc? Soon we will have the tool status page, and the idea is to make all errors super clear and actionable. I like that you say what we recommend users to do `We recommend that you remove this environment variable from your workflow.`It would be great to add a way to read more about it.

We could potentially link to https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages.  We don't mention the `CODEQL_EXTRACTOR_GO_BUILD_TRACING` environment variable in the docs though, since this was never a first-class feature.

henrymercer (commented):
Very nice!  Just a couple of comments around the docs.

src/init-action.ts
AlonaHlobina: What's about:

> Go is now analyzed as compiled language. The CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable is deprecated. We recommend that you remove this environment variable from your workflow. Read [more](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages).

henrymercer: It's not too obvious to me why the environment variable is deprecated reading the above.  Also we can't do Markdown links unfortunately.

How about:

"Go is now analyzed as a compiled language by default, so the CODEQL_EXTRACTOR_GO_BUILD_TRACING environment variable is deprecated. We recommend that you remove this environment variable from your workflow. For more information, see https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages"?

CHANGELOG.md
angelapwen: Thanks!

src/init-action.ts
angelapwen: I have updated the PR with the most recent suggestion from Henry above!

henrymercer (dismissed):
LGTM, subject to @AlonaHlobina's approval of the warning message.

src/init-action.ts
AlonaHlobina: Yes, the latest suggestion from Henry is fine 👍

.github/workflows/debug-artifacts-failure.yml
```diff
@@ -35,6 +35,9 @@ jobs:
         uses: ./.github/prepare-test
         with:
           version: latest
+      - uses: actions/setup-go@v3
+        with:
+        go-version: ^1.13.1
```
henrymercer, line 40: I think this is why these checks aren't running.
```suggestion
          go-version: ^1.13.1
```

.github/workflows/debug-artifacts.yml
```diff
@@ -55,6 +55,9 @@ jobs:
         uses: ./.github/prepare-test
         with:
           version: ${{ matrix.version }}
+      - uses: actions/setup-go@v3
+        with:
+        go-version: ^1.13.1
```
henrymercer, line 60: I think this is why these checks aren't running.
```suggestion
          go-version: ^1.13.1
```

angelapwen:
Huh.. the debug artifact upload PR check is [failing on Go](https://github.com/github/codeql-action/actions/runs/3430458803/jobs/5717448325) again (previously I thought it was because I hadn't included the Go setup action). Now that I've taken a closer look, it appropriately enters the section of code in `analyze` that attempts to autobuild if it hasn't already built the code (the legacy workflow).

It seems to encounter this error when it tries to build:
```
  2022/11/09 17:40:43 Warning: encountered errors extracting package `./...`:
  2022/11/09 17:40:43   -: pattern ./...: directory prefix . does not contain main module or its selected dependencies
  2022/11/09 17:40:43 Done extracting types for package ./....
  2022/11/09 17:40:43 Done processing dependencies.
  2022/11/09 17:40:43 Starting to extract packages.
  2022/11/09 17:40:43 Done extracting packages.
```

Interestingly though, the legacy workflow PR check goes through the same steps but doesn't encounter this warning, and successfully extracts the Go code: https://github.com/github/codeql-action/actions/runs/3430458810/jobs/5717457044 

I'm not quite sure what's causing the discrepancy 🔍

angelapwen:
Ok — I resolved some merge conflicts, made some fixes to PR checks that were failing due to the new Go build command being added to `build.sh`, and re-ran possibly flaky tests. I think this PR is ready for review again — all the "expected" checks are ones that I have removed/replaced as of this change.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [UNRELEASED]
 
+- Go is now analyzed in the same way as other compiled languages such as C/C++, C#, and Java. This completes the rollout of the feature described in [CodeQL Action version 2.1.27](#2127---06-oct-2022).
+- Update the ML-powered additional query pack for JavaScript to version 0.4.0.
+- Go is now analyzed in the same way as other compiled languages such as C/C++, C#, and Java. This completes the rollout of the feature described in [CodeQL Action version 2.1.27](#2127---06-oct-2022). As a result, the `CODEQL_EXTRACTOR_GO_BUILD_TRACING` environment variable, which was previously used to manually enable this behavior, is now deprecated. If your code scanning workflow uses this environment variable, we recommend that you remove it. Otherwise, you do not need to alter your code scanning workflows.
 No user facing changes.
```
henrymercer, line 8: These items went into 2.1.32

```suggestion
```

henrymercer (dismissed):
Looks good modulo the CHANGELOG.

CHANGELOG.md
angelapwen: Oh, hm, I could have sworn that I fixed this in a merge conflict. Thank you for catching!

henrymercer (approved):

PR #1330, opened 2022-10-27
Make PR checks compatible with the latest version of the `ubuntu-latest` runner image

henrymercer:
This PR updates the PR checks to handle the recent update to the `ubuntu-latest` runner image.  This is being moved over to Ubuntu 22.04, which is only compatible with CodeQL CLI 2.7.3 and later.  Some of our PR checks use very old CLIs, and this PR updates them to run on the `ubuntu-20.04` image instead.

Planned work:
- Introducing a process to run our checks against new runner images before the `-latest` image changes.
- In the future, deprecating support for versions of the CLI earlier than 2.7.3.
- Potentially introducing a more helpful error message into the Action when we hit this case.  Impact depends on how many customers are running the latest Action (so they'll benefit from the new error message) against a very old CLI (older than 2.7.3).

### DIFF
--- a/.github/workflows/unset-environment-new-cli.yml
+++ b/.github/workflows/unset-environment-new-cli.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         include:
-        - os: ubuntu-latest
+        - os: ubuntu-20.04
           version: stable-20210809
         - os: ubuntu-latest
           version: cached
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -22,8 +22,8 @@ def isCompatibleWithLatestImages(version):
     if version in ["cached", "latest", "nightly-latest"]:
         return True
     date = version.split("-")[1]
-    # The first version of the CodeQL CLI compatible with the latest runner images is 2.7.3.
-    # This appears in CodeQL Bundle version codeql-bundle-20211208.
+    # The first version of the CodeQL CLI compatible with `ubuntu-22.04` and `windows-2022` is
+    # 2.7.3. This appears in CodeQL Bundle version codeql-bundle-20211208.
     return date >= "20211208"
 
 
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -1,6 +1,7 @@
 import ruamel.yaml
 import os
 
+# The default set of CodeQL Bundle versions to use for the PR checks.
 defaultTestVersions = [
     # The oldest supported CodeQL version: 2.4.5. If bumping, update `CODEQL_MINIMUM_VERSION` in `codeql.ts`
     "stable-20210308",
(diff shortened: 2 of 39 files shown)


pr-checks/sync.py
```diff
@@ -15,14 +16,32 @@
...
     "nightly-latest"
 ]
-defaultOperatingSystems = ["ubuntu-latest", "macos-latest", "windows-2019"]
+
+
+def isCompatibleWithLatestImages(version):
+    if version in ["cached", "latest", "nightly-latest"]:
+        return True
+    date = version.split("-")[1]
+    # The first version of the CodeQL CLI compatible with the latest runner images is 2.7.3.
```
aeisenberg, line 25: Can you be more specific here than "latest"? In the future, the "latest" will be different from what it is today.

aeisenberg (commented):
Thanks for looking into this.

administrateur-Sexflixxx (approved):

.github/workflows/unset-environment-new-cli.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         include:
-        - os: ubuntu-latest
+        - os: ubuntu-20.04
           version: stable-20210809
         - os: ubuntu-latest
```
aeisenberg, line 26: Should this be changed to ubuntu-22.04? Or maybe above should be changed to ubuntu-latest (I think I'd prefer this). In general, we should not have to specify 22.04 (using latest instead) and when the next version of ubuntu comes out, some of our tests may break, but that's a good thing since we will be aware of this and can fix proactively.

henrymercer: Could you clarify what you mean by "Or maybe above should be changed to ubuntu-latest"?

aeisenberg: Sorry...that wasn't very clear. I am suggesting that everywhere there is `ubuntu-22.04` now, we should change to `ubuntu-latest`. This was when `ubuntu-latest` is changed again, we will know if anything is broken.

In particular, I was thinking here:

https://github.com/github/codeql-action/pull/1330/files/993ca05cd7741fd4e48fe6b8a3b28f102174f4b6#diff-9bec90e4ad9a179ec1c8e1266c2ddc2fc0fbd73db55720b3425cfdf667b89e14R116

and here:

https://github.com/github/codeql-action/pull/1330/files/993ca05cd7741fd4e48fe6b8a3b28f102174f4b6#diff-9bec90e4ad9a179ec1c8e1266c2ddc2fc0fbd73db55720b3425cfdf667b89e14R116

And in other places in `.github/workflows/python-deps.yml`.

henrymercer: Ah, I see what you mean now.  I think you're generally right, but my hesitancy for `python-deps.yml` specifically is that these jobs were previously testing against both `ubuntu-20.04` and `ubuntu-22.04`, and the migration of `ubuntu-latest` to `ubuntu-22.04` isn't fully rolled out.  It would be possible for `ubuntu-latest` to start pointing to `ubuntu-20.04` again before the rollout completes, which would cause us to silently lose test coverage here.  I'd be in favour to switching to `ubuntu-latest` once the rollout completes.

aeisenberg: > I'd be in favour to switching to ubuntu-latest once the rollout completes.

Sure. I'm guessing you're tracking this in https://github.com/github/codeql-core/issues/2909?

aeisenberg (approved):
This is fine as long as we use https://github.com/github/codeql-core/issues/2909 to remember to move to ubuntu-latest when 22.04 is fully rolled out.

PR #1334, opened 2022-10-28
Add a better error message for users of CodeQL CLI 2.7.2 and earlier running on `ubuntu-22.04`

henrymercer:
Improves the error message for users running (a) CLI 2.7.2 and earlier and (b) `ubuntu-22.04`, to which `ubuntu-latest` is now being migrated. Previously this was "undefined symbol: __libc_dlopen_mode, version GLIBC_PRIVATE". Now we give some guidance around glibc versions and using the `ubuntu-20.04` runner image.

Example workflow run with error: https://github.com/github/codeql-action/actions/runs/3347488560/jobs/5545531438


src/codeql.ts
```diff
@@ -742,15 +754,39 @@ async function getCodeQLForCmd(
...
+          tracerEnvJs,
+          envFile,
+        ]);
+      } catch (e) {
+        if (
+          e instanceof CommandInvocationError &&
+          e.output.includes(
+            "undefined symbol: __libc_dlopen_mode, version GLIBC_PRIVATE"
+          ) &&
+          process.platform === "linux" &&
```
aeisenberg, line 773: Minor: Is this line necessary? If we get this error and the platform is not linux, then what does that mean?

My guess is that it can't really happen.

src/codeql.ts
```diff
@@ -742,15 +754,39 @@ async function getCodeQLForCmd(
...
+            this,
+            CODEQL_VERSION_TRACING_GLIBC_2_34
+          ))
+        ) {
+          throw new util.UserError(
+            "The CodeQL CLI is incompatible with the version of glibc on your system. " +
+              `Please upgrade to CodeQL CLI version ${CODEQL_VERSION_TRACING_GLIBC_2_34} or ` +
+              "later. If you cannot upgrade to a newer version of the CodeQL CLI, you can " +
+              "alternatively use glibc 2.33 or earlier, for instance by running your workflow on " +
+              `the "ubuntu-20.04" runner image.`
```
aeisenberg, line 784: Does this sound better?

```suggestion
              "run your workflow on another image such as the "ubuntu-20.04" runner that has glibc 2.33 or earlier installed.`
```

src/codeql.ts
```diff
@@ -1259,7 +1295,7 @@ async function runTool(cmd: string, args: string[] = []) {
     ignoreReturnCode: true,
   }).exec();
   if (exitCode !== 0)
-    throw new CommandInvocationError(cmd, args, exitCode, error);
+    throw new CommandInvocationError(cmd, args, exitCode, error, output);
```
aeisenberg, line 1298: Looking at the definition of `CommandInvocationError`, the constructor only takes 4 args. Is this change right?

src/codeql.ts
henrymercer: I'm not aware of how this could happen on non-Linux operating systems.  My thinking was if we get this error and the platform is Linux, I think we want to know about it in our telemetry, so we don't want to make it a `UserError`.

src/codeql.ts
henrymercer: I added an additional argument in this PR so we can examine the stdout too.

aeisenberg (approved):

PR #1336, opened 2022-10-31
Handle concurrency waste out-of-the-box

IOrlandoni:
### Contents
Default implementation shown in readme does not handle concurrency in any way, wasting minutes running scans on each and every commit when it's unnecessary.

This PR proposes that the default implementation cancels in-progress runs of the workflow for the same github ref or PR. Meaning that, for example, quickly successive commits to an open pull request will not end with multiple completed runs, and only complete the latest commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,11 @@ on:
     #        *  * * * *
     - cron: '30 1 * * 0'
 
+# We cancel any in-progess scans on the same ref/pr, as they are unnecessary.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   CodeQL-Build:
     # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest


adityasharad:
Thank you for the suggestion! 

Before we start making this the default recommendation in our readme and workflow template, there are some details we (the GitHub team maintaining the CodeQL Action) would like to figure out internally, in order to ensure a smooth experience when using this feature.

The two main potential risks we know of so far are:
- On large repos, repeatedly pushing new commits to a PR will keep cancelling the existing run and starting fresh runs, creating a long gap before the PR author sees even an initial set of code scanning results.
- Cancelled runs create notifications for the commit author, which may be a noisy user experience.

There may be other points of friction that we haven't fully understood yet.

I am going to close this PR for now, since we are not likely to merge it right away, but I want to be clear that this is a valuable suggestion we are taking seriously, and we have tracked it internally. We will spend some more time to understand the potential risks/friction of using this feature in a code scanning workflow, reduce that friction where possible, and then document it for our users so they can make an informed decision on whether to opt in. Until then, please feel free to use the `concurrency` feature within your own code scanning workflows, and let us know if you have any further feedback about the experience.

PR #1349, opened 2022-11-07
Improve experience when init fails before generating a config file

henrymercer:
Suppose a customer has a run where the init Action failed before saving a config file. When the customer opens their Actions logs, the UI currently focuses on the post init step, since this is the last step that failed.

Demoting the error in the post init Action to a warning means that the UI will instead focus on the `init` step, which is more useful for debugging what went wrong.


angelapwen (approved):
Makes sense, thank you!

PR #685, opened 2021-08-05
Add PR check job for Ruby beta

adityasharad:
Add a separate job that analyses the multilanguage test repo's Ruby code.
For now, run this only with the latest released build of CodeQL pointed to in `defaults.json`.
The nightly build and the cached builds don't support Ruby yet.

In future, we can update this and other PR checks to test a wider range of cases.

### Merge / deployment checklist

- [ ] Confirm this change is backwards compatible with existing workflows.
- [ ] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [ ] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #1350, opened 2022-11-07
Support Swift for private beta

angelapwen:
This PR:
- Updates the autobuild script path for Swift (it is expected that it lives in the `experimental` folder)
- Adds PR checks to test Swift analysis with custom builds and autobuild with the expected `CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT` environment variable set; it looks like all other mechanisms are already in place (ie. Swift is already in the list of traced languages) so am testing that. I noticed it was missing in the linguist to language prefixes map so added it there. Note that custom builds are supported on ubuntu and MacOS but autobuild is only supported on MacOS for the time being. 
- Adds Swift interpretation and analysis timings to the status report on the CodeQL Action side. Note that we have not yet added the fields to the data warehouse, nor are we consuming it on github.com; this change will be a no-op until those changes are merged. 

Adding Swift to the [languages.test.ts file](https://github.com/github/codeql-action/pull/1322/files#diff-bab9ba94239436262696bc41b4415725ba3a7ff7e9a0c65b3de2f44c009ca4eb) will have been done in https://github.com/github/codeql-action/pull/1322 so I didn't add it to this PR to prevent the would-be merge conflict from refactoring the test. 

As a follow-up, we should add a PR check to test that Swift can be auto-detected once the corresponding CLI change makes it to a release. 


henrymercer (commented):
Looks good!  Could we check that language autodetection works for Swift too by adding a some more assertions to the `pr-checks/checks/multi-language-autodetect.yml` test?

pr-checks/checks/test-swift.yml
```diff
@@ -0,0 +1,28 @@
+name: "Swift analysis"
+description: "Tests creation of a Swift database"
+versions: ["latest", "cached", "nightly-latest"]
+operatingSystems: ["ubuntu", "macos"]
+env:
+  CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT: "true"
```
henrymercer, line 6: We'll probably want to add `DOTNET_GENERATE_ASPNET_CERTIFICATE: "false"` here to suppress the interactive prompt on macOS.

pr-checks/checks/test-swift.yml
```diff
@@ -0,0 +1,28 @@
...
+steps:
+  - uses: ./../action/init
+    with:
+      languages: swift
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
+    env:
+      TEST_MODE: true
+  - name: Build code
+    shell: bash
+    run: ./build.sh
```
henrymercer, line 16: It might be useful to test the autobuild Action for Swift too / instead, since we added some custom logic to find it.

angelapwen:
> Looks good! Could we check that language autodetection works for Swift too by adding a some more assertions to the `pr-checks/checks/multi-language-autodetect.yml` test?

I actually made a separate Swift PR check because I saw that we deliberately made a separate Ruby check from the multilanguage test when Ruby was being rolled out for beta (https://github.com/github/codeql-action/pull/685/files). In fact, Ruby is still a separate PR check and not in `multi-language-autodetect.yml` test probably because we forgot to add it back once we GA'd it. Do you think we should just combine them all into 1 check now, even though Swift is going to be in beta?

To clarify though, I believe the autodetection itself is being tested with the Ruby check, and the Swift check I added. They're both finding the Ruby/Swift code in the `tests/multi-language-repo` directory and detecting/building them with the build script appropriately; just in a separate test from the PR check that tests for all repos. I could rename it to `test-swift-autodetect.yml` to clarify.

henrymercer:
> I actually made a separate Swift PR check because I saw that we deliberately made a separate Ruby check from the multilanguage test when Ruby was being rolled out for beta (https://github.com/github/codeql-action/pull/685/files).

The separate PR check makes sense, as we can validate that Swift _isn't_ enabled when the experimental environment variable isn't set, because the autodetect languages test doesn't build a Swift DB.

I think the thing that's missing from both Ruby and Swift is a test that the autodetection mechanism works, i.e. if you don't pass a `languages: ` input to the `init` Action, then we pick up on the Swift code in the repo when the experimental environment variable is set.

pr-checks/checks/test-swift.yml
angelapwen: I've added it to all the new checks, but I don't fully understand why we need to include this env variable on some checks but not others (so far they're only enabled on the Go tracing ones).

pr-checks/checks/test-swift.yml
angelapwen: Done!

angelapwen:
Ah, makes sense!

I made the following changes:
- added a separate Swift autodetect PR check
- added a separate Swift autobuild PR check
- added a separate Swift check that doesn't set the experimental environment variable and makes sure the Swift database isn't built (I thought it didn't make too much sense to check this inside the multilanguage one, but we could theoretically do this check and the corresponding Ruby check in that test).

pr-checks/checks/test-swift.yml
adityasharad: Conceptually, any build that runs `dotnet` under the CodeQL tracer will need this variable set.

However, the C# tracing configuration already sets this variable on its own, which should explain why you don't need to manually set it in the test workflows that run only C#.

As far as I know the Go tracing configuration does not set this variable (because it's not expecting to run `dotnet` builds -- maybe we should add it anyway). However, the Go autobuilder is very permissive in what build scripts it looks for, and will call our multilanguage test build script that includes `dotnet`. So for the Go test workflows we need to set this variable explicitly to avoid the unrelated `dotnet` build script from freezing during a Go traced build.

angelapwen: A ha, I see. So because the Swift autobuild script doesn't call the multilanguage test build script, we shouldn't need it here (that explains why the PR checks were passing before I added the variable).

pr-checks/checks/ruby-autodetect.yml
```diff
@@ -0,0 +1,24 @@
+name: "Ruby analysis using autodetect"
```
aeisenberg, line 1: We already have the `multi-language-autodetect.yml` test. Does it make more sense to merge this test and the swift autobuild test into that one?

aeisenberg: Hmmm...I see there is the swift autodetect and swift autbuild tests below. Only one of these can be merged into `multi-language-autodetect.yml` The other would have to remain separate

angelapwen: I don't think we want to merge these tests in, because both depend on (separate) experimental flags while the languages are in beta. But we'll merge them in once the languages are out of beta. I can add that as a comment.

angelapwen:
Reverting to draft because I hadn't realized some features we're testing were not/may not yet be implemented on the CLI side (Swift autodetection + Swift autobuild)

angelapwen:
Autobuild and custom build PR checks now work 😄 

Waiting on CLI side autodetection of Swift, and then hopefully that PR check will pass and this PR will be ready for review.

AlexDenisov:
> Waiting on CLI side autodetection of Swift, and then hopefully that PR check will pass and this PR will be ready for review.

What is "CLI side autodetection of Swift"?

AlexDenisov:
Also, the failing job' log contains the following (_modified by hand for readability_):
```
/opt/hostedtoolcache/CodeQL/0.0.0-20221024/x64/codeql/codeql database init \
--db-cluster /home/runner/work/_temp/codeql_databases \
--source-root=/home/runner/work/codeql-action/codeql-action \
--language=javascript --language=python --language=java \
--language=csharp --language=ruby --language=cpp --language=go \
--begin-tracing --trace-process-name=Runner.Worker.exe \
--codescanning-config=/home/runner/work/_temp/user-config.yaml
```
note the absence of Swift in the list of languages.

henrymercer:
> What is "CLI side autodetection of Swift"?

Language autodetection is about figuring out what languages to analyze when users don't provide a `languages` input to the `init` Action.  Generally we advise leaving this blank so users will automatically benefit from support for new languages, as in the snippet below.

```yml
      # Initializes the CodeQL tools for scanning.
      - name: Initialize CodeQL
        uses: github/codeql-action/init@v2
        # Override language selection by uncommenting this and choosing your languages
        # with:
        #   languages: go, javascript, csharp, python, cpp, java
```

We are in the process of migrating some functionality from the CodeQL Action to the CLI, including language autodetection.  As a result, the functionality is different between the CodeQL Action and the CLI.  Language autodetection for Swift for instance works in the CodeQL Action but not the CLI.

My current thoughts about this is that Swift language autodetection not working in the CLI should not block merging this PR — we should disable it with `CODEQL_PASS_CONFIG_TO_CLI=false` in the PR check for now and add a comment explaining that it is not currently implemented.

henrymercer:
Looking at the other failing jobs, "PR Check - Swift analysis using a custom build command" is hanging on macOS due to the .NET certificate prompt.  We need to add `DOTNET_GENERATE_ASPNET_CERTIFICATE=false` to all checks that run `build.sh` either manually or through the autobuilder to avoid this, and I think adding this environment variable to this check should make the check pass.

henrymercer:
Looks like we're not autodetecting Swift in the Action either — I think this is because the language auto-detection uses the languages on `main`.  We could add the Swift code to `multi-language-repo` in a separate PR to introduce some Swift on `main` — not sure if there's a more elegant solution.

angelapwen:
This PR should be ready to merge but is blocked on https://github.com/github/codeql-action/pull/1357 (or some other fix to the node modules PR check)

henrymercer (approved):
Very nice, thanks for your diligence with the PR checks!  I think we just need to merge in `main` to get the checked-in dependencies check passing.

PR #1357, opened 2022-11-10
Update node modules PR check with fixed `npm` version

angelapwen:
We used to always use the `latest` version of `npm` in this PR check, but the most recent version of `npm` would create a different `package-lock.json` file that causes the PR check to fail. We are unable to update this `package-lock.json` file because it would not be backwards compatible, so instead we are pinning the PR check to version `8.19.2`.


.github/workflows/pr-checks.yml
```diff
@@ -58,6 +58,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
```
aeisenberg, line 61: Should there be an explicit version here? Otherwise, the version can change without us having control.

angelapwen: You're right, the README for the Action says as much: `However, it is recommended to always specify Node.js version and don't rely on the system one.` https://github.com/actions/setup-node 

Which version should we use — the one currently on the runner? `16.18.0`

angelapwen: Huh.. still fails that check without the explicit version. I'll try with `16.18.0`

angelapwen: Still failing 😢

aeisenberg: Oh....I see what's happening. The `check-node-modules.sh` script is forcing the use of the _latest_ npm version, which is now 9.1.1.  If you run this locally:

```
npm install --force -g npm@latest
npm ci
npm run removeNPMAbsolutePaths
```

On a mac, you will see the same changes. Just run that locally and commit the results. I think that would work.

aeisenberg: So, the explicit `setup-node` step does nothing.

angelapwen: Ah... let me give it a try.

.github/workflows/script/check-node-modules.sh
```diff
@@ -15,7 +15,7 @@ npm run removeNPMAbsolutePaths
 # Check that repo is still clean
 if [ ! -z "$(git status --porcelain)" ]; then
     # If we get a fail here then the PR needs attention
-    >&2 echo "Failed: node_modules are not up to date. Run 'npm ci && npm run removeNPMAbsolutePaths' on a macOS machine to update. Note it is important this command is run on macOS and not any other operating system as there is one dependency (fsevents) that is needed for macOS and may not be installed if the command is run on a Windows or Linux machine."
+    >&2 echo "Failed: node_modules are not up to date. Run 'npm ci && npm run removeNPMAbsolutePaths' on a macOS machine to update and commit. Note it is important this command is run on macOS and not any other operating system as there is one dependency (fsevents) that is needed for macOS and may not be installed if the command is run on a Windows or Linux machine. If there are no changes, run 'npm install --force -g npm@latest' with root permissions first."
```
aeisenberg, line 18: Are we going to pin to v8.19.2? I don't think we can move to package-lock v3 while we also need to support node v12.

henrymercer: The latest v8 release (8.19.3) of npm is compatible with Node 12.  I _think_ package-lock v3 should be compatible with Node 12 since the package-lock is only consumed by npm

angelapwen:
Closing as this PR check was fixed in https://github.com/github/codeql-action/pull/1348

PR #1366, opened 2022-11-14
Update lockfiles to v3

henrymercer:
Lockfile v3 works with npm v7 and later ([source](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#lockfileversion)).  We're currently using npm v8, so we can simplify our lockfiles a bit by updating the lockfile version.  Also add an `.npmrc` specifying which lockfile version to use, such that both npm v8 (used for compatibility with checks that run on Node 12) and npm v9 (which contributors might be running on their local machine) will generate v3 lockfiles.


aeisenberg (approved):
I think this will work. Our tests for v1 do not rely on lockfile v2.

PR #1368, opened 2022-11-15
Read CODE_SCANNING_REF env variable

marcogario:
Depending on how we trigger the action run, sometimes the GITHUB_REF is not set.
This seems to be a special variable and we are unable to override it in anyway.

The solution we propose here is to have a dedicated extra environment variable that we can consider.

*Why not use the existing params*: There is a restriction on the existing `ref` param that requires the `sha` to be set. This is not something we need (and it is a stronger requirement than we care about). Moreover, when trying to use these parameters we ran into some warning messages so we are not clear on how well supported they are.

This is the minimal patch that serves our purpose, please provide feedback on how to shape it better.


edoardopirovano (approved):

henrymercer (approved):

aeisenberg (approved):
Looks good.

It would be nice if we had a `getOptionalEnvVar` function for cases like this. But, this is outside the scope of this PR.

PR #1370, opened 2022-11-15
Support Kotlin for public beta

angelapwen:
Kotlin will enter public beta with the release of 2.11.4. This change:

- adds it to the list of languages that a user could specify in the `init` action, aliasing it to Java.
- checks in Kotlin test code and a Kotlin build command to our test custom build, so that its extraction and autobuild will be tested along with existing Kotlin tests
- adds support for a feature flag, `disable_kotlin_cli_enabled`, that will allow us to mass-enable the `CODEQL_EXTRACTOR_JAVA_AGENT_DISABLE_KOTLIN` flag (ie. disable Kotlin extraction). The flag is not yet implemented on GitHub.com so the code is a no-op at the moment.


src/feature-flags.ts
```diff
@@ -11,8 +11,8 @@ export interface FeatureEnablement {
 export enum Feature {
   BypassToolcacheEnabled = "bypass_toolcache_enabled",
   CliConfigFileEnabled = "cli_config_file_enabled",
+  DisableKotlinEnabled = "disable_kotlin_enabled",
   FileBaselineInformationEnabled = "file_baseline_information_enabled",
-  GolangExtractionReconciliationEnabled = "golang_extraction_reconciliation_enabled",
```
adityasharad, line 15: Maybe remove this in a separate PR?

src/languages.ts
```diff
@@ -15,6 +15,7 @@ const LANGUAGE_ALIASES: { [lang: string]: Language } = {
   c: Language.cpp,
   "c++": Language.cpp,
   "c#": Language.csharp,
+  kotlin: Language.java,
```
adityasharad, line 18: Interesting, I didn't know we recognised these additional names! Generally we want to treat Kotlin as a part of enabling `java` rather than a new language, but this gives us flexibility to handle user misconfiguration gracefully.

cc @turbo in case you have concerns.

adityasharad (commented):
Nice. Thank you for looking into this!

src/feature-flags.ts
angelapwen: Yep, can do!

src/languages.ts
angelapwen: I won't add it to the changelog to surface in that case, but it might be nice to keep it in here to handle the misconfiguration as you said 😄

henrymercer: Make sure to mirror this change in the CLI too!

angelapwen: I actually thought that I wouldn't need to mirror in the CLI because by the time we call CLI commands from the Action,  it would've been "translated" via aliasing to Java 🤔

henrymercer: True.  Without any CLI changes, `codeql database create --language kotlin` won't work.  But that might be fine — we want to avoid confusion and it might be confusing seeing `java` in all of the log messages when you've just called `codeql database create --language kotlin`.

angelapwen: Hm yes, I think this is what I was thinking — only accepting the value here to catch the code scanning case, but not in the CLI. I don't think we allow `codeql database create --language typescript` (or `c` `c++ `c#`) but here we are allowing those configurations in the Action, so I thought it should follow ~= the same pattern.

angelapwen:
I decided to remove the PR check that tests that when the `CODEQL_EXTRACTOR_JAVA_AGENT_DISABLE_KOTLIN` option is passed to the CLI, analysis will fail — when it's disabled, our Java analysis actually is able to still run off of the Java test code checked in. I think it's probably not necessary to test here as it should be tested in the CLI.

src/feature-flags.ts
```diff
@@ -11,6 +11,7 @@ export interface FeatureEnablement {
 export enum Feature {
   BypassToolcacheEnabled = "bypass_toolcache_enabled",
   CliConfigFileEnabled = "cli_config_file_enabled",
+  DisableKotlinEnabled = "disable_kotlin_enabled",
```
henrymercer, line 14: Optional: I wonder whether `KotlinEnabled` is clearer — with the current name we'd end up with a `DisableKotlinDisabled` flag in the monolith.  This would mirror the current usage of `MlPoweredQueriesEnabled` in that it would be used for disablement only.

tests/multi-language-repo/main.kt
```diff
@@ -0,0 +1,5 @@
+fun main() {
+    if (true) {
+        println("Hello, World!")
+    }
+}
```
henrymercer, line 5: Nit: new line at eof

henrymercer (dismissed):
An optional comment, otherwise LGTM

src/feature-flags.ts
angelapwen: Ha, I just made that exact comment on the monolith PR: https://github.com/github/github/pull/247129 

I hadn't realized that was how `MlPoweredQueriesEnabled` worked. I'll look into making the change.

henrymercer: Your call :)

angelapwen: I've made the change on both sides!

henrymercer: Actually I just realised that renaming has more implications — it'll disable Kotlin by default on GHES 🤦

We could add some more logic to handle this, but it's probably simpler to stick with the previous name.

angelapwen: Oh yes, I see what you mean.. nice catch! I'll revert

henrymercer (approved):

angelapwen:
I am updating the feature flag name to include `_cli` because the code scanning experiences team has another similar-sounding Kotlin feature flag in the monolith repo. Sorry, needs re-review again!

src/feature-flags.ts
```diff
@@ -24,6 +25,10 @@ export const featureConfig: Record<
     envVar: "CODEQL_BYPASS_TOOLCACHE",
     minimumVersion: undefined,
   },
+  [Feature.DisableKotlinCliEnabled]: {
+    envVar: "CODEQL_DISABLE_KOTLIN_CLI",
```
henrymercer, line 29: Optional: I find the name of this flag a little confusing, for instance it containing "CLI" yet not doing anything if we pass it to the CLI.  Can we come up with a clearer name?  Maybe something like `DisableKotlinAnalysis` for the `Feature` enum and `CODEQL_ACTION_DISABLE_KOTLIN_ANALYSIS` for the environment variable?  What do you think?

henrymercer: As an aside, I think prefixing our feature env vars with `CODEQL_ACTION_` in general would be clearer, but that's slightly risky so I suggest we don't do that in this PR.

angelapwen: Hm... then I'll leave the environment variable as `CODEQL_DISABLE_KOTLIN_ANALYSIS` for now, and we can prefix them all together later 👍 I do agree though — including `ACTION` would make things a lot clearer.

In this specific case we are actually passing (a different variable) to the CodeQL CLI so it's not _too_ misleading, and I don't expect us to use this feature environment variable anyway because we might as well use the `CODEQL_EXTRACTOR_JAVA_AGENT_DISABLE_KOTLIN` environment variable that the CLI directly accepts.

henrymercer (approved):

PR #1373, opened 2022-11-16
TRAP Caching: Re-introduce workaround for download timeout

edoardopirovano:
Looking at our telemetry, I found [a run](https://github.com/zooniverse/panoptes/actions/runs/3420941934/jobs/5696413442#step:3:259) where the download timeout provided by the Actions library didn't do what we expect. In particular, looking at the timestamps shows that while the _downloading_ happened quickly, the check to see if the key was present in the cache took over 15 minutes (much longer than our 2 minute timeout). Clearly, this is problematic. While it isn't a widespread problem as far as I can see in our telemetry, an outage in the Actions cache could cause this to break all CodeQL runs which is not good.

Let's use our workaround that wraps the entire call in a timeout here too like we already do for uploading.


henrymercer (commented):
Since we download trap caches in the init Action, do we need to add `await checkForTimeout();` to its entrypoint to prevent things hanging?

edoardopirovano:
> Since we download trap caches in the init Action, do we need to add `await checkForTimeout();` to its entrypoint to prevent things hanging?

Indeed we do, good point. I've amended the PR to do this.

henrymercer (approved):

PR #1358, opened 2022-11-11
Bump minimum CodeQL bundle version to 2.6.3

henrymercer:
Now that GHES 3.2 is deprecated, we can clean up a bunch of tech debt in the Action by bumping our minimum CodeQL Bundle version to 2.6.3, the first version that went into GHES 3.3.

The highlight here is that the Action no longer needs to count lines of code itself, since now all supported versions of the CLI can do that.  This allows us to remove a bunch of code and 14 dependencies!

### Merge / deployment checklist

- [ ] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #1374, opened 2022-11-16
Set testing environment for CodeQL workflow

henrymercer:
Previously we simply disabled uploading the status reports for all PR checks except ones that were there specifically to test the status report functionality.  However when there's a lot of development activity on the CodeQL Action repo and the Actions runtime is being flaky, it's valuable being able to completely exclude CodeQL Action development from telemetry queries.

As drive-by improvements, this PR also makes `TEST_MODE` specific to the CodeQL Action and enables it in the `check-codescanning-config` Action.  We also remove the `INTERNAL_CODEQL_ACTION_DEBUG_LOC` variable, which is redundant as of #1358.


angelapwen (approved):
Thanks for this, `CODEQL_ACTION_TEST_MODE` is much clearer to me 👍

PR #1384, opened 2022-11-21
Cache feature flags on disk

aeisenberg:
This will allow feature flags to be shared across steps in the same job, avoiding an error we saw earlier where the init action had the flag enabled, but the analyze step had it disabled.

This uses the runner's temp folder to cache the flags file, which will stick around until the job completes.


src/feature-flags.test.ts
```diff
@@ -285,6 +289,57 @@
...
+      fs.readFileSync(cachedFeatureFlags, "utf8")
+    );
+    t.deepEqual(actualFeatureEnablement, expectedFeatureEnablement);
+
+    // now test that we actually use the feature flag cache instead of the server
+    actualFeatureEnablement[Feature.CliConfigFileEnabled] = false;
+    fs.writeFileSync(
+      cachedFeatureFlags,
+      JSON.stringify(actualFeatureEnablement)
+    );
```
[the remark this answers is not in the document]
angelapwen: Huh.. should we be checking files we _write_ to?

src/feature-flags.ts
```diff
@@ -66,11 +71,13 @@ export class Features implements FeatureEnablement {
   constructor(
     gitHubVersion: util.GitHubVersion,
     repositoryNwo: RepositoryNwo,
+    tempDir: string,
```
angelapwen, line 74: Could we call this `localCacheDir` perhaps?

src/feature-flags.ts
```diff
@@ -55,6 +58,8 @@ export const featureConfig: Record<
  */
 type GitHubFeatureFlagsApiResponse = Partial<Record<Feature, boolean>>;
 
+export const FEATURE_FLAGS_FILE_NAME = "feature-flags.json";
```
angelapwen, line 61: Wondering if this could be renamed to include `local` or `cache` somehow so it's easier to remember the difference between this file and the remote server-side flags 🤔

angelapwen (approved):
Thanks!! Makes sense, just a couple of naming thoughts

src/feature-flags.ts
aeisenberg: I have a slight preference for keeping the name as-is since we use this variable name throughout the code base and it always means the same thing. Changing the name would make it harder to keep track that `tempDir` here is the same as `config.tempDir` in most cases (except in tests).

src/feature-flags.test.ts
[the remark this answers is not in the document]
aeisenberg: This does feel like an FP. I will contact the team.

src/feature-flags.ts
angelapwen: Ah sorry, I didn't realize this was the same directory. Disregard 😄

angelapwen (approved):
Thanks! Extra test looks 👍

PR #1387, opened 2022-11-22
Use Swift version 5.7 in PR checks

henrymercer:
The version installed in the latest runner image, 5.7.1, is not yet supported.

cc @AlexDenisov 

### DIFF
--- a/pr-checks/checks/export-file-baseline-information.yml
+++ b/pr-checks/checks/export-file-baseline-information.yml
@@ -2,6 +2,10 @@ name: "Export file baseline information"
 description: "Tests that file baseline information is exported when the feature is enabled"
 versions: ["nightly-latest"]
 steps:
+  - uses: swift-actions/setup-swift@v1
+    if: "!startsWith(matrix.os, 'windows')"
+    with:
+      swift-version: 5.7
   - uses: ./../action/init
     with:
       languages: javascript
--- a/pr-checks/checks/export-file-baseline-information.yml
+++ b/pr-checks/checks/export-file-baseline-information.yml
@@ -3,7 +3,7 @@ description: "Tests that file baseline information is exported when the feature 
 versions: ["nightly-latest"]
 steps:
   - uses: swift-actions/setup-swift@5cdaa9161ad1f55ae39a5ea1784ef96de72f95d9
-    if: "!startsWith(matrix.os, 'windows')"
+    if: runner.os != 'Windows'
     with:
       swift-version: "5.7"
   - uses: ./../action/init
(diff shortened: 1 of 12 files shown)


pr-checks/checks/export-file-baseline-information.yml
```diff
@@ -2,6 +2,10 @@ name: "Export file baseline information"
 description: "Tests that file baseline information is exported when the feature is enabled"
 versions: ["nightly-latest"]
 steps:
+  - uses: swift-actions/setup-swift@v1
```
aeisenberg, line 5: Should we pin this to a SHA since this is not a first party action? (Also below.)

henrymercer: Good idea, done.

angelapwen (approved):
✔️ 

I wonder if I should also include this step in the CodeQL Action integration test on the CLI-side (just adding the check for Swift here https://github.com/github/semmle-code/pull/44368/files#)

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -4,15 +4,23 @@ operatingSystems: ["ubuntu", "macos"]
 env:
   CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT: "true" # Remove when Swift is GA.
 steps:
+  - uses: swift-actions/setup-swift@5cdaa9161ad1f55ae39a5ea1784ef96de72f95d9
+    if: "!startsWith(matrix.os, 'windows')"
+    with:
+      swift-version: 5.7
```
angelapwen, line 10: Actually, should this be `"5.7"` rather than `5.7`? Noticed as I add it to the CodeQL Action integration test in the CLI.

henrymercer:
> I wonder if I should also include this step in the CodeQL Action integration test on the CLI-side (just adding the check for Swift here https://github.com/github/semmle-code/pull/44368/files#)

👍 That makes sense to me

pr-checks/checks/multi-language-autodetect.yml
henrymercer: It's equivalent, since 5.7 doesn't contain any characters that have a special meaning in yaml

angelapwen: Ah ok — workflow generator on the CLI side is a bit stricter then!

henrymercer: We're ok here since the Actions runtime will automatically convert the number to a string, but good question!

henrymercer: On reflection, it's probably better practice to use a string, so let's do that.

angelapwen (approved):

angelapwen (approved):

PR #1390, opened 2022-11-23
Remove some dead code

henrymercer:
The Action requires GHES 3.3, so we can safely remove this fallback logic for GHES 2.22 and GHES 3.0.


aeisenberg (commented):
Looks like `GITHUB_ACTION_REPOSITORY` is not being set during the integration tests. See:

https://github.com/github/codeql-action/actions/runs/3532744788/jobs/5927496032#step:4:14

aeisenberg:
Otherwise looks reasonable.

henrymercer:
So `GITHUB_ACTION_REPOSITORY` is only set when calling an Action from another repository.  We still need to populate it manually if we're running locally.

I've updated the PR to only remove the code that deals with the fact that GHES 2.22 didn't populate `GITHUB_ACTION_REPOSITORY` even when the Action is called from anothe repository.

aeisenberg (approved):
LGTM assuming all jobs pass.

PR #1392, opened 2022-11-23
Add functionality for parsing Action inputs from a workflow file

henrymercer:
In order to upload SARIF files for code scanning runs where the `init` step failed, we need to be able to handle the case where the `analyze` Action did not run.  In this case, the `post-analyze` step will not be run, and therefore we don't know the inputs to `analyze`.  (In particular, we care about `category`, `upload`, and `checkout_path`.)

This PR implements some functionality to make a best effort attempt to work out the value of an input to an Action given the workflow file and information that's available to the `init` step like matrix variables.  I recommend reviewing the first commit separately since it separates out workflow-related code into a separate file `workflow.ts`.


src/workflow.ts
```diff
@@ -290,3 +292,43 @@ export function getWorkflowRunID(): number {
...
+      "Could not get category input since workflow.jobs was undefined."
+    );
+  }
+  const categories: string[] = Object.values(workflow.jobs)
+    .map((job) => getAnalyzeSteps(job).map((step) => step.with?.category))
+    .flat()
+    .filter((category) => category !== undefined)
+    .map((category) => category!);
+
+  if (categories.length === 0) {
```
aeisenberg: Not common, but I've seen the pattern where there are multiple analyze steps, one for each language (maybe disabled if the language is matrixed). If we know the language of the current analyze step from the input, and we can match it with the `language` property of the `with` block, we might be able to get more precision.

To be fair, though, I don't know how common this really is.

src/workflow.ts
```diff
@@ -325,10 +335,18 @@ export function getCategoryInput(workflow: Workflow): string | undefined {
       "Could not get category input since multiple categories were specified by the analysis step."
     );
   }
-  if (categories[0].includes("${{")) {
+
+  // Make a basic attempt to substitute matrix variables
+  // First normalize by removing whitespace
+  let category = categories[0].replace(/\${{\s+/, "${{").replace(/\s+}}/, "}}");
```
aeisenberg: Maybe rare since by default, unless there are explicit quotes, yaml will trim whitespace.

```suggestion
  let category = categories[0].trim().replace(/\${{\s+/, "${{").replace(/\s+}}/, "}}");
```

henrymercer: You're right about explicit quotes.  I think if we do have leading / trailing whitespace at this point, then that's the category we'd submit to code scanning in the event of a successful run, so that's what we need to submit for a failed run too.

src/workflow.ts
henrymercer: Hmm, I think this might be the kind of thing we would want to measure and then address as follow up.  Things are already hacky enough without evaluating `if` conditions :)

henrymercer: Though we don't limit ourself to the current job — that seems like a worthwhile optimization to handle workflows with a job per category.

aeisenberg (commented):
OK. This will now throw if we can't determine what the actual input value is. I don't actually see where getting the category is being used. Is that coming in a later PR? We should make sure that the action doesn't fail if it can't find the category (most likely, the action will already have failed for some other reason, so we need to at least make sure that this failure doesn't hide the original failure).

henrymercer:
This functionality is just for https://github.com/github/codeql-action/pull/1393 — ideally we'd minimise the number of places where we need to do this hacky parsing of the workflow.  We wrap it in a try-catch block there to avoid hiding the original failure as you mention.  I think the `tryOrThrow` names should make it fairly obvious that these functions should generally be surrounded by try-catch blocks — is there anything else we should do to avoid future consumers creating bugs?

aeisenberg:
I think that all makes sense. I'll review #1393 later today. Maybe just adding a comment to the function suggesting that it should be wrapped in a try-catch.

aeisenberg (approved):
LGTM. Optionally, add a comment to the `tryOrThrow...` functions to suggest wrapping in try-catch.

aeisenberg (approved):

PR #1393, opened 2022-11-23
Submit SARIF for failed runs too

henrymercer:
This PR helps us better surface failed code scanning workflow runs by attempting to submit a SARIF file containing `executionSuccessful: false` in some of these cases.  We take special care to ensure that we respect SARIF categories and `upload: false` options to avoid stale tips.  This is made more difficult by the fact that these are only passed to the `analyze` Action.  We want to handle cases where the workflow fails before the `analyze` step, and we don't have access to these inputs in the post-init step.  Therefore we resort to a best-effort attempt at working out what the inputs were based on the workflow file and the variables we do have access to in the post-init step.

We don't to cover every case where the workflow fails here.  Cases where the workflow run is cancelled or times out won't be possible to handle — we'd need a different mechanism to notify Code Scanning about these.  However this PR will cover many of the cases where workflows fail so we can start providing a better experience.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [UNRELEASED]
 
+- Add a step that tries to upload a SARIF file for the workflow run when that workflow run fails. This will help better surface failed code scanning workflow runs. [#1393](https://github.com/github/codeql-action/pull/1393)
+
+## 2.1.34 - 25 Nov 2022
```
aeisenberg, line 7: Is not in main already?

pr-checks/checks/submit-sarif-failure.yml
```diff
@@ -0,0 +1,32 @@
...
+operatingSystems: ["ubuntu"]
+
+env:
+  # Internal-only environment variable used to indicate that the post-init Action
+  # should expect to upload a SARIF file for the failed run.
+  CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF: true
+  # Make sure the uploading SARIF files feature is enabled.
+  CODEQL_ACTION_UPLOAD_FAILED_SARIF: true
+  # Upload the failed SARIF file as an integration test of the API endpoint.
+  CODEQL_ACTION_TEST_MODE: false
```
aeisenberg, line 13: So, we are not actually testing that the endpoint was called, only that it didn't fail. I'm not sure how we could do this better. We _could_ add a step before the init step that runs an action with a post run property. This post run, will run after `init`'s post run. But from here, it's not immediately apparent to see how we can check if the request was actually sent.

aeisenberg (commented):
Partial review only.

CHANGELOG.md
aeisenberg: Oh. I see this PR isn't merging into main.

src/feature-flags.ts
```diff
@@ -19,6 +19,7 @@ export enum Feature {
   FileBaselineInformationEnabled = "file_baseline_information_enabled",
   MlPoweredQueriesEnabled = "ml_powered_queries_enabled",
   TrapCachingEnabled = "trap_caching_enabled",
+  UploadFailedSarifEnabled = "upload_failed_sarif_enabled",
```
aeisenberg, line 22: This is not currently a FF. Are you planning on making one, or is this for local testing only?

src/init-action-post-helper.ts
```diff
@@ -1,25 +1,103 @@
...
+    ))
+  ) {
+    logger.debug("Uploading failed SARIF is disabled.");
+    return;
+  }
+  const workflow = await getWorkflow();
+  const jobName = getRequiredEnvParam("GITHUB_JOB");
+  const matrix = parseMatrixInput(actionsUtil.getRequiredInput("matrix"));
+  if (
+    getUploadInputOrThrow(workflow, jobName, matrix) !== "true" ||
```
aeisenberg, line 46: Is this wrapped in a catch block so it doesn't prevent the post-step from completing?

src/workflow.ts
```diff
@@ -405,3 +409,49 @@ export function getUploadInputOrThrow(
...
 }
+
+/**
+ * Makes a best effort attempt to retrieve the wait-for-processing input for the
+ * particular job, given a set of matrix variables.
+ *
+ * @returns the wait-for-processing input
+ * @throws an error if the wait-for-processing input could not be determined
+ */
+export function getWaitForProcessingInputOrThrow(
```
aeisenberg, line 420: All of the "...OrThrow" functions make me nostalgic for the java `throws` keyword.

src/init-action-post-helper.ts
aeisenberg: Oh...I think it is. I was code reviewing commit-by-commit and I see you added the catch block later.

pr-checks/checks/submit-sarif-failure.yml
henrymercer: We could do something with a proxy to spy on the HTTP requests, but this kind of thing probably makes more sense as a unit test.  I can add some.

src/feature-flags.ts
henrymercer: I'll create one once this PR is ready to merge so we can roll this out gradually.

src/init-action-post-helper.test.ts
```diff
@@ -75,3 +84,87 @@ test("post: init action with debug mode on", async (t) => {
     t.assert(printDebugLogsSpy.called);
   });
 });
+
+test("uploads failed SARIF run for typical workflow", async (t) => {
```
aeisenberg, line 88: Nice.

src/upload-lib.ts
```diff
@@ -427,6 +430,28 @@ export async function waitForProcessing(
...
+        status === "failed" &&
+        Array.isArray(response.data.errors) &&
+        response.data.errors.length === 1 &&
+        response.data.errors[0].toString().startsWith("unsuccessful execution")
+      ) {
+        logger.debug(
+          "Successfully uploaded a SARIF file for the unsuccessful execution. Received expected " +
+            '"unsuccessful execution" error, and no other errors.'
+        );
+        break;
```
aeisenberg, line 445: So, this is the happy path for when `isUnsuccessfulExecution`? We expect that code scanning will return with an error? That's a little confusing, but I guess it is a code scanning decision and outside of this change.

src/upload-lib.ts
```diff
@@ -427,6 +430,28 @@ export async function waitForProcessing(
...
+      } else if (options.isUnsuccessfulExecution && status !== "pending") {
+        throw new Error(
+          `${
+            "Did not receive expected 'unsuccessful execution' error from the API. " +
```
aeisenberg, line 449: This is an error that will probably be user facing. Can you clarify it a bit? Also, looking at the logic above, it might be that the `unsuccessful execution` message is received, but _also_ other messages are.

src/upload-lib.ts
henrymercer: We're going to further discuss the semantics of this API, but currently, given that we don't want unsuccessful runs to close all the alerts (due to transient errors), throwing an error is currently the best way to integrate unsuccessful runs into the processing API.

src/upload-lib.ts
henrymercer: Good point, done.  This will be surfaced as a warning in the logs for the post-init step, or an error if it occurs in our PR checks.

aeisenberg: It looks like this hasn't changed. Is there a missing commit?

henrymercer: Huh, the commits are on https://github.com/github/codeql-action/tree/henrymercer/report-failed-runs, but they're not visible on this PR.  I think we may need to merge https://github.com/github/codeql-action/pull/1392 first.

src/upload-lib.ts
```diff
@@ -386,60 +385,120 @@ async function uploadFiles(
...
+    );
+  } else {
+    throw new Error(
+      `${
+        "Failed to upload a SARIF file for the unsuccessful execution. Code scanning status " +
+        "information for the repository may be out of date as a result. "
+      }${
+        status === "failed"
+          ? `Processing errors: ${response.data.errors}`
+          : 'Encountered no processing errors, but expected to receive an "unsuccessful execution" error.'
```
aeisenberg, line 498: This message makes sense to me since I know what's going on in the background, but I can imagine users would be confused if they saw it. I'm not sure the best message here, but maybe something simpler like:

```suggestion
          : 'Encountered unexpected status from server.'
```
or
```suggestion
          : 'Encountered unexpected message from server.'
```

henrymercer: OK, let's make the detailed message into a debug log.

aeisenberg (approved):
Nice!

PR #1394, opened 2022-11-23
Add a way to bypass the toolcache for kotlin and swift

aeisenberg:
This works by moving the logic to check for toolcache bypass out of
creating the codeql instance. The logic now _may_ perform an API request
in order to check what languages are in the repository. This check is
redundant because the same call is being made later in the action when
the actual list of languages is calculated.

Commit-by-commit review is suggested.

Even though it is inefficient to add this extra API call, I am leaning towards leaving it in because refactoring to perform only one API call would make the code messier, I ensure that the extra API call is made only if necessary, and we expect that this workaround will be temporary.


aeisenberg:
I don't think this requires a change note.

src/config-utils.ts
```diff
@@ -924,10 +929,14 @@ async function getLanguages(
...
     const parsedLanguage = parseLanguage(language);
+    const dealiasedLanguage =
+      parsedLanguage && parsedLanguage in LANGUAGE_ALIASES
+        ? LANGUAGE_ALIASES[parsedLanguage]
+        : (parsedLanguage as Language);
     if (parsedLanguage === undefined) {
       unknownLanguages.push(language);
-    } else if (parsedLanguages.indexOf(parsedLanguage) === -1) {
-      parsedLanguages.push(parsedLanguage);
+    } else if (parsedLanguages.includes(dealiasedLanguage)) {
```
adityasharad: Should this be negated?

src/feature-flags.ts
```diff
@@ -18,6 +18,7 @@ export enum Feature {
   FileBaselineInformationEnabled = "file_baseline_information_enabled",
   MlPoweredQueriesEnabled = "ml_powered_queries_enabled",
   TrapCachingEnabled = "trap_caching_enabled",
+  BypassToolcacheKotlinSwiftEnabled = "bypass_toolcache_kotlin_switft_enabled",
```
adityasharad, line 21: Typo!
```suggestion
  BypassToolcacheKotlinSwiftEnabled = "bypass_toolcache_kotlin_swift_enabled",
```

src/feature-flags.ts
```diff
@@ -48,6 +49,10 @@ export const featureConfig: Record<
     envVar: "CODEQL_TRAP_CACHING",
     minimumVersion: undefined,
   },
+  [Feature.BypassToolcacheKotlinSwiftEnabled]: {
+    envVar: "CODEQL_BYPASS_TOOLCACHE_KOTLIN_SWIFT",
+    minimumVersion: undefined,
```
adityasharad, line 54: Should we set this to 2.11.4, or leave it undefined?

src/util.ts
```diff
@@ -832,3 +834,43 @@ export function isHostedRunner() {
...
+  // Check if the toolcache is disabled for all languages
+  if (await featuresEnablement.getValue(Feature.BypassToolcacheEnabled)) {
+    return true;
+  }
+
+  // Check if the toolcache is disabled for kotlin and swift.
+  if (
+    await featuresEnablement.getValue(Feature.BypassToolcacheKotlinSwiftEnabled)
+  ) {
+    // Now check to see if kotlin or swift is one of the languages being analyzed.
```
adityasharad, line 867: Add a log message here.

src/languages.ts
```diff
@@ -31,7 +35,7 @@ export function parseLanguage(language: string): Language | undefined {
 
   // Check language aliases
   if (language in LANGUAGE_ALIASES) {
-    return LANGUAGE_ALIASES[language];
```
adityasharad, line 34: Are there any other callers of `parseLanguage` where we need to make sure the caller looks up the return value int the language aliases map?

src/languages.ts
```diff
@@ -11,16 +11,20 @@ export enum Language {
...
   typescript: Language.javascript,
 };
 
+export type LanguageOrAlias = Language | keyof typeof LANGUAGE_ALIASES;
+
+export const KOTLIN_SWIFT_BYPASS = ["kotlin", "swift"];
+
 // Translate from user input or GitHub's API names for languages to CodeQL's names for languages
-export function parseLanguage(language: string): Language | undefined {
+export function parseLanguage(language: string): LanguageOrAlias | undefined {
```
adityasharad, line 27: Comment is no longer accurate right? And should we name this `asKnownLanguage`?

src/config-utils.ts
aeisenberg: Yes. Fixed.

src/feature-flags.ts
aeisenberg: Eek!

src/feature-flags.ts
aeisenberg: We can't set to 2.11.4 since we won't know the CLI version until we invoke the CLI and we can't invoke the CLI until we need to know if we bypass the toolcache.

Dependency is circular.

src/languages.ts
aeisenberg: No...but I found an issue (caught by failing tests). I was setting the alias too late and filtering out known good languages.

src/feature-flags.ts
adityasharad: Aha! Yes makes sense. Leave a comment so we remember this.

aeisenberg:
I'm going to add a few more tests here since this logic is getting tricky.

src/feature-flags.ts
angelapwen: While you're editing — I think we were keeping this list alphabetized (or at least, I was 😆)

src/testing-utils.ts
```diff
@@ -162,6 +162,26 @@ export function mockFeatureFlagApiEndpoint(
   sinon.stub(apiClient, "getApiClient").value(() => client);
 }
 
+export function mockLangaugesInRepo(languages: string[]) {
```
angelapwen, line 165:
```suggestion
export function mockLanguagesInRepo(languages: string[]) {
```

harmless typo 😸

aeisenberg:
I think I addressed all comments, added more tests, and fixed the bug I found. Hopefully, all tests will pass.

adityasharad (commented):
One more suggestion, otherwise looks reasonable. I will let Henry or Angela do a more detailed review however.

src/config-utils.test.ts
```diff
@@ -1761,7 +1766,10 @@ parseInputAndConfigMacro.title = (providedTitle: string) =>
 
 const mockLogger = {
   info: (message: string) => {
-    console.log(message);
+    console.log("info:", message);
+  },
+  debug: (message: string) => {
+    console.log("debug:", message);
```
henrymercer, line 1772: Minor, non-blocking: We can also use `getRunnerLogger(true)` here

src/config-utils.test.ts
```diff
@@ -2449,3 +2457,119 @@ test("downloadPacks-with-registries fails with invalid registries block", async
...
+    const languages = args.codeqlResolvedLanguages.reduce(
+      (acc, lang) => ({
+        ...acc,
+        [lang]: true,
+      }),
+      {}
+    );
+    const codeQL = setCodeQL({
+      resolveLanguages: () => Promise.resolve(languages),
+    });
```
henrymercer, line 2561: Minor, non-blocking: Combine these two sets of tests to avoid duplicating this code (e.g. wrap in `throwsAsync` iff `expectedError` is set).

src/config-utils.ts
```diff
@@ -890,27 +896,27 @@ async function getLanguagesInRepo(
...
+
+  if (autodetected) {
     const availableLanguages = await codeQL.resolveLanguages();
     languages = languages.filter((value) => value in availableLanguages);
-    logger.info(
-      `Automatically detected languages: ${JSON.stringify(languages)}`
-    );
+    logger.info(`Automatically detected languages: ${languages.join(", ")}`);
+  } else {
+    logger.info(`Languages from configuration: ${languages.join(", ")}`);
```
henrymercer, line 919: Optional: Could add a comment here saying it's an error to include an unsupported language in the `languages` input to `init`.

src/feature-flags.ts
```diff
@@ -26,6 +27,14 @@ export const featureConfig: Record<
 > = {
   [Feature.BypassToolcacheEnabled]: {
     envVar: "CODEQL_BYPASS_TOOLCACHE",
+    // Cannot specify a minimum version because this flag is checked before we have
+    // access to the CodeQL instance.
+    minimumVersion: undefined,
```
henrymercer, line 32: I think it's also true to say that it doesn't make sense for this flag to have a minimum CodeQL version

src/feature-flags.ts
```diff
@@ -26,6 +27,14 @@ export const featureConfig: Record<
...
+  },
+  [Feature.BypassToolcacheKotlinSwiftEnabled]: {
+    envVar: "CODEQL_BYPASS_TOOLCACHE_KOTLIN_SWIFT",
+    // Cannot specify a minimum version because this flag is checked before we have
+    // access to the CodeQL instance.
     minimumVersion: undefined,
```
henrymercer, line 38: Agree, this one ideally would have a minimum version, but it's not practical to figure out what CodeQL version is in the toolcache first.

src/util.test.ts
```diff
@@ -449,3 +456,105 @@ test("withTimeout doesn't call callback if promise resolves", async (t) => {
...
+    name: "bypass swift language from repository",
+    features: [Feature.BypassToolcacheKotlinSwiftEnabled],
+    hasCustomCodeQL: false,
+    languagesInput: "",
+    languagesInRepository: ["SwiFt", "other"],
+    expected: true,
+    expectedApiCall: true,
+  },
+].forEach((args) => {
+  test(`shouldBypassToolcache: ${args.name}`, async (t) => {
```
henrymercer, line 545: Nice.  I think it might also be useful to have tests where we have Java in the repository / input but not Kotlin, and both Kotlin and Java in the repository / input.

src/util.test.ts
```diff
@@ -449,3 +456,105 @@ test("withTimeout doesn't call callback if promise resolves", async (t) => {
...
+    const mockRequest = mockLanguagesInRepo(args.languagesInRepository);
+    const mockLogger = getRecordingLogger([]);
```
henrymercer, line 547: Minor, non-blocking: since we're recording the logs, we should either check the log messages, or switch this to `getRunnerLogger(true)`.

src/util.ts
```diff
@@ -832,3 +834,51 @@ export function isHostedRunner() {
...
+    );
+    const bypass = rawLanguages.some((lang) =>
+      KOTLIN_SWIFT_BYPASS.includes(lang)
+    );
+    if (bypass) {
+      logger.info(
+        `Bypassing toolcache for kotlin or swift. Languages: ${rawLanguages}`
+      );
+    }
+    return bypass;
```
henrymercer, line 881: Nit, non-blocking: push this into `if (bypass)` since if we added future cases below the Kotlin / Swift bypass, we might be surprised by returning false here rather than falling through.

henrymercer (commented):
This generally looks great.

Some questions:
1. Should we revert this change once the next release is fully rolled out and we no longer need to bypass the toolcache for Kotlin / Swift?  It does look like there's some complexity here that we wouldn't want to keep around.
2. Do we want to add a PR check that runs some integration tests, or are we happy with the unit tests here?  Personally I'd be ok with unit tests here plus some manual testing.

src/config-utils.ts
aeisenberg: I added it below, next to where we throw the error for unknown languages.

src/feature-flags.ts
aeisenberg: Well, if we know we are using 2.11.3 or lower, then this feature flag is not useful since we are not going to be affected by the wobble.

src/util.ts
aeisenberg: I am refactoring this area anyway. I need to add an extra condition to always make an API check if java is a non-autodetected language. 

I will make sure there is only one return statement (except for the short-circuiting checks above).

henrymercer (approved):

PR #1400, opened 2022-11-28
Fixes spurious error messages in tests

aeisenberg:
Previously, `isAnalyzingDefaultBranch` was failing because there are some missing env vars: `GITHUB_SHA`, `GITHUB_REF`, and `GITHUB_EVENT_PATH`. Also, `checkout_path` is missing as an input.

Rather than trying to set them to mock values, which would require setting the paths to existing paths in the file system, I chose to stub the entire function. I think this is fine since the point of the test is to check the ram and threads values, not testing the `isAnalyzingDefaultBranch` function.


angelapwen (approved):
Thanks for finding this!!

Just curious, were the tests passing in CI because `GITHUB_SHA`, `GITHUB_REF`, and `GITHUB_EVENT_PATH` _were_ set by default in CI, just not locally?

henrymercer (commented):
Very nice!  I checked this out locally and `src/analyze-action-input.test.ts` is still failing.  Thankfully we can just apply the same fix to get that passing.  Once we've done that all the tests succeed locally.

aeisenberg:
> Just curious, were the tests passing in CI because `GITHUB_SHA`, `GITHUB_REF`, and `GITHUB_EVENT_PATH` _were_ set by default in CI, just not locally?

The tests are still "passing" locally (ie- the test process doesn't fail). It's just that there is a subprocess command that gets invoked and fails. The error is swallowed by the action (I think that's the behaviour we want).

aeisenberg:
> Just curious, were the tests passing in CI because `GITHUB_SHA`, `GITHUB_REF`, and `GITHUB_EVENT_PATH` _were_ set by default in CI, just not locally?

The tests are still "passing" locally (ie- the test process doesn't fail). It's just that there is a subprocess command that gets invoked and fails. The error is swallowed by the action (I think that's the behaviour we want).

aeisenberg:
> Very nice! I checked this out locally and `src/analyze-action-input.test.ts` is still failing. Thankfully we can just apply the same fix to get that passing. Once we've done that all the tests succeed locally.

Right...forgot about the second one. Let me get that in, too.

angelapwen (approved):
Nice, I pulled down the branch and ran to confirm this time as well 👍

henrymercer (approved):

PR #1419, opened 2022-12-06
python-setup: Don't allow Poetry to make venv in project

RasmusWL:
As highlighted in the code comments, Projects that specify `in-project = true` in their poetry.toml would get the venv created inside the repo directory, which would cause CodeQL to consider it as user-written code. We don't want this to happen.

I have not updated changelog or readme since I thought this was a small enough change, but let me know if you think that woudl be needed.

### DIFF
--- a/python-setup/auto_install_packages.py
+++ b/python-setup/auto_install_packages.py
@@ -33,10 +33,16 @@ def _check_output(command, extra_env={}):
 
 def install_packages_with_poetry():
 
-    # To handle poetry 1.2, which started to use keyring interaction MUCH more, we need
-    # add a workaround. See
-    # https://github.com/python-poetry/poetry/issues/2692#issuecomment-1235683370
-    extra_poetry_env = {"PYTHON_KEYRING_BACKEND": "keyring.backends.null.Keyring"}
+    extra_poetry_env = {
+        # To handle poetry 1.2, which started to use keyring interaction MUCH more, we need
+        # add a workaround. See
+        # https://github.com/python-poetry/poetry/issues/2692#issuecomment-1235683370
+        "PYTHON_KEYRING_BACKEND": "keyring.backends.null.Keyring",
+        # Projects that specify `in-project = true` in their poetry.toml would get the
+        # venv created inside the repo directory, which would cause CodeQL to consider
+        # it as user-written code. We don't want this to happen.
+        "POETRY_VIRTUALENVS_IN_PROJECT": "False",
+    }
 
     command = [sys.executable, '-m', 'poetry']
     if sys.platform.startswith('win32'):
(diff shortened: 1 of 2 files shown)


python-setup/auto_install_packages.py
```diff
@@ -33,10 +33,16 @@ def _check_output(command, extra_env={}):
...
-    extra_poetry_env = {"PYTHON_KEYRING_BACKEND": "keyring.backends.null.Keyring"}
+    extra_poetry_env = {
+        # To handle poetry 1.2, which started to use keyring interaction MUCH more, we need
+        # add a workaround. See
+        # https://github.com/python-poetry/poetry/issues/2692#issuecomment-1235683370
+        "PYTHON_KEYRING_BACKEND": "keyring.backends.null.Keyring",
+        # Projects that specify `in-project = true` in their poetry.toml would get the
+        # venv created inside the repo directory, which would cause CodeQL to consider
+        # it as user-written code. We don't want this to happen.
+        "POETRY_VIRTUALENVS_IN_PROJECT": "False",
```
angelapwen, line 44: I couldn't easily find documentation/links for this environment variable; do you have one handy that we could add to the code comment or the PR?

angelapwen (commented):
I don't think we need to update the README for this change 😄 

As for the CHANGELOG — do we expect the change to affect many users? If the venv was previously considered user-written code, and now isn't, it shouldn't affect any user's results (unless the venv itself had alerts), right?

RasmusWL:
> As for the CHANGELOG — do we expect the change to affect many users?

Many? that's not my expectation, but it's a gut feeling and not based on hard data.

> If the venv was previously considered user-written code, and now isn't, it shouldn't affect any user's results (unless the venv itself had alerts), right?

spot on. I think there's a fair chance there would have been spurious alerts found within the in-project venv though.

Altogether, I have time defending not making adding anything to the changelog, so let me just do that :+1:

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Add a step that tries to upload a SARIF file for the workflow run when that workflow run fails. This will help better surface failed code scanning workflow runs. [#1393](https://github.com/github/codeql-action/pull/1393)
+- Python automatic dependency installation will no longer consider dependecy code installed in venv as user-written, for projects using Poetry that specify `virtualenvs.in-project = true` in their `poetry.toml`. [#1419](https://github.com/github/codeql-action/pull/1419).
```
angelapwen, line 6:
```suggestion
- Python automatic dependency installation will no longer consider dependency code installed in `venv` as user-written, for projects using Poetry that specify `virtualenvs.in-project = true` in their `poetry.toml`. [#1419](https://github.com/github/codeql-action/pull/1419).
```

Quick typo suggestion

angelapwen (approved):
Thanks for updating!

PR #1421, opened 2022-12-07
update-required-checks.sh: fix argument handling

cklin:


henrymercer (approved):
Wow, I guess we haven't passed a SHA to this script for a while... thanks!

henrymercer:
While we're here, perhaps we should also add `check-expected-release-files` to the ignore list on line 26.  I noticed that this is appearing as an expected check here, but it only runs when certain paths are changed.

henrymercer (approved):

PR #1422, opened 2022-12-08
Set up the Swift version the extractor declares

angelapwen:
This PR changes the `setup-swift` Action to set up the version of Swift the extractor declares (except on Windows). It also adds the `nightly-latest` PR check back to a couple of PR checks, which were removed in https://github.com/github/codeql-action/pull/1412 to unblock the release.

Note that after this PR merges, we should remember to update the corresponding integration test within the CLI, and include the bumped `setup-swift` SHA (https://github.com/github/codeql-action/pull/1415) as well. 

When the newest version of the CLI (that supports 5.7.1) makes it to `latest` and `cached`, we should be able to remove this logic and unconditionally setup 5.7.1, as future versions of the extractor should maintain backwards compatibility with prior Swift versions. 

### DIFF
--- a/.github/setup-swift/action.yml
+++ b/.github/setup-swift/action.yml
@@ -1,0 +1,32 @@
+name: "Setup Swift"
+description: Performs necessary steps to set up appropriate Swift version.
+inputs:
+  codeql-path:
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Get Swift version
+      id: get_swift_version
+      # We don't support Swift on Windows or prior versions of CLI.
+      if: "(runner.os != 'Windows') && (matrix.version == 'cached' || matrix.version == 'latest' || matrix.version == 'nightly-latest')"
+      shell: bash
+      env: 
+        CODEQL_PATH: ${{inputs.codeql-path}}
+      run: |
+        if [ $RUNNER_OS = "macOS" ]; then
+          PLATFORM="osx64"
+        else # We do not run this step on Windows.
+          PLATFORM="linux64"
+        fi 
+        SWIFT_EXTRACTOR_DIR=$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')
+        VERSION=$($SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor --version | awk '/version/ { print $3 }')
+        # Specify 5.7.0, otherwise setup Action will default to latest minor version. 
+        if [ $VERSION = "5.7" ]; then
+          VERSION="5.7.0"
+        fi
+        echo "version=$VERSION" | tee -a $GITHUB_OUTPUT
+    - uses: swift-actions/setup-swift@194625b58a582570f61cc707c3b558086c26b723
+      if: "(runner.os != 'Windows') && (matrix.version == 'cached' || matrix.version == 'latest' || matrix.version == 'nightly-latest')"
+      with:
+        swift-version: "${{steps.get_swift_version.outputs.version}}"
--- a/.github/setup-swift/action.yml
+++ b/.github/setup-swift/action.yml
@@ -30,3 +30,7 @@ runs:
       if: "(runner.os != 'Windows') && (matrix.version == 'cached' || matrix.version == 'latest' || matrix.version == 'nightly-latest')"
       with:
         swift-version: "${{steps.get_swift_version.outputs.version}}"
+     # We run prepare-test again to return to the appropriate directory before building.
+    - uses: ./.github/prepare-test
+      with:
+        version: ${{ matrix.version }}
--- a/.github/setup-swift/action.yml
+++ b/.github/setup-swift/action.yml
@@ -19,8 +19,8 @@ runs:
         else # We do not run this step on Windows.
           PLATFORM="linux64"
         fi 
-        SWIFT_EXTRACTOR_DIR=$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')
-        VERSION=$($SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor --version | awk '/version/ { print $3 }')
+        SWIFT_EXTRACTOR_DIR="$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')"
+        VERSION="$("$SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor" --version | awk '/version/ { print $3 }')"
         # Specify 5.7.0, otherwise setup Action will default to latest minor version. 
         if [ $VERSION = "5.7" ]; then
           VERSION="5.7.0"
--- a/.github/setup-swift/action.yml
+++ b/.github/setup-swift/action.yml
@@ -1,4 +1,4 @@
-name: "Setup Swift"
+name: "Set up Swift"
 description: Performs necessary steps to set up appropriate Swift version.
 inputs:
   codeql-path:
@@ -26,8 +26,6 @@ runs:
           VERSION="5.7.0"
         fi
         echo "version=$VERSION" | tee -a $GITHUB_OUTPUT
-        # We return to the appropriate directory for the rest of the workflow.
-        cd ../action
     - uses: swift-actions/setup-swift@194625b58a582570f61cc707c3b558086c26b723
       if: "(runner.os != 'Windows') && (matrix.version == 'cached' || matrix.version == 'latest' || matrix.version == 'nightly-latest')"
       with:
--- a/.github/setup-swift/action.yml
+++ b/.github/setup-swift/action.yml
@@ -26,11 +26,9 @@ runs:
           VERSION="5.7.0"
         fi
         echo "version=$VERSION" | tee -a $GITHUB_OUTPUT
+        # We return to the appropriate directory for the rest of the workflow.
+        cd ../action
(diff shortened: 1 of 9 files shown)


angelapwen:
`cached` and `latest` are failing (expected as they only support 5.7.0 and not 5.7.1).

I can conditionally setup 5.7.0 for `latest`, or we could wait to merge this PR until `latest` and `cached` reflect the last release which should support 5.7.1 🤔

henrymercer:
For `cached` to reflect the latest release, we'd need to wait another 2 weeks or so.

I think perhaps while Swift is in beta, we should just test that the Swift extractor is compatible with the Swift version it declares:

```
codeql/experimental/swift/tools/*/extractor --version | awk '/version/ { print $3 }'
```

This mimics the [internal tests](https://github.com/github/codeql/tree/main/swift/actions/run-integration-tests), and while it's further away from customer setups, it avoids us having to update these tests every release while Swift is in beta.

henrymercer:
Ah, that approach isn't very easy because we need to setup Swift before running `codeql-action/init`, but at that point CodeQL hasn't been downloaded yet, so we can't get the Swift version.

In that case, we might have to stop running this test on `cached` due to the bumpiness of which CodeQL version the cache contains, and manually specify which Swift versions correspond to each bundle.

Optionally, we could define a top-level `setup-swift` boolean in each generated workflow definition, and update `sync.py` to add the appropriate setup step to the Action.

angelapwen:
I've updated all 4 Swift-related PR checks to use 5.7.1, and stopped testing `cached` where I could but running into a couple issues:
- `Swift analysis using a custom build command` on `latest` is failing because the latest version of CLI still is the version that requires 5.7.0 and not 5.7.1. This should cease to be an issue once our newest version goes into `latest` (and afterwards as well, as long as the extractor continues to support 5.7.1).
- The multi-language autodetect check is failing on `cached` (and presumably `latest`) — even though I am no longer checking for the Swift database, it's still attempting to extract Swift because it's been automatically detected in the repository. I can't specify the languages to check because that defeats the purpose of this PR check (the autodetect). 

Is it a viable solution to just wait until the newest CLI version that supports 5.7.1 makes it into `latest` and `cached`, and then merge this PR? It doesn't solve the longer-term problem of extractors not providing backwards compatibility with prior Swift versions, but hopefully this is not a common occurrence even in beta, and ideally should be tested in the CLI.

angelapwen:
Aditya suggested checking the Swift version from the extractor after the `init` step as a stopgap until the newest CLI makes it to `latest`. Will give that a try now.

angelapwen:
I think the above approach should work, but we are still getting this error — 

```
/home/runner/work/_temp/4bebb1c6-be8f-43cf-9dfd-75dc640e9229.sh: line 1: codeql/experimental/swift/tools/*/extractor: No such file or directory
```

Is this perhaps the wrong path, or has the extractor been cleaned up after the `init` step?

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -69,7 +86,7 @@ steps:
       fi
 
   - name: Check language autodetect for Swift
-    if: "!startsWith(matrix.os, 'windows') && (matrix.version == 'cached' || matrix.version == 'latest' || matrix.version == 'nightly-latest')"
```
adityasharad, line 72: We can drop the Windows check here because we haven't put any Windows versions in the matrix?

pr-checks/checks/swift-autobuild.yml
```diff
@@ -4,17 +4,28 @@ versions: ["latest", "cached", "nightly-latest"]
...
+  - name: Get Swift version
+    id: get_swift_version
+    shell: bash
+    env: 
+      CODEQL_PATH: ${{steps.init.outputs.codeql-path}}
+    run: |
+      SWIFT_EXTRACTOR_DIR=$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')
+      VERSION=$($SWIFT_EXTRACTOR_DIR/tools/osx64/extractor --version | awk '/version/ { print $3 }') # Update platform if more operating systems are added to this test.
+      if [ $VERSION = "5.7" ]; then
+        VERSION="5.7.0"
```
adityasharad, line 23: Worth commenting why we need this.

pr-checks/checks/export-file-baseline-information.yml
```diff
@@ -1,19 +1,39 @@
...
+      SWIFT_EXTRACTOR_DIR=$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')
+      VERSION=$($SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor --version | awk '/version/ { print $3 }')
+      if [ $VERSION = "5.7" ]; then
+        VERSION="5.7.0"
+      fi
+      echo "version=$VERSION" | tee -a $GITHUB_OUTPUT
+  - uses: swift-actions/setup-swift@194625b58a582570f61cc707c3b558086c26b723
+    if: runner.os != 'Windows'
+    with:
+      swift-version: "${{steps.get_swift_version.outputs.version}}"
```
adityasharad, line 36: Optional, but could save some code duplication: Could we pull these two steps out into a `setup-swift` composite action within this repo (like `.github/prepare-test`)? Then we only have to change one place in future for Swift setup.

adityasharad (dismissed):
Looks good as is, but some ideas for further clarity.

pr-checks/checks/swift-autobuild.yml
angelapwen: Ah thanks, I commented in the custom build check but forgot to copy it over!

pr-checks/checks/multi-language-autodetect.yml
angelapwen: Yes — but I've actually added it back in the composite action as that might be run from a Windows workflow!

pr-checks/checks/export-file-baseline-information.yml
angelapwen: That's a good idea! Just updated.

angelapwen:
😮‍💨 I had an error that the new `setup-swift` step couldn't be found without running `checkout`, so I ran it again just before `setup-swift` but this overwrote the whole directory (I think). Reverted and trying to figure out why `setup-swift` can't be found when the `checkout` Action is run at the beginning of each check. But hit API rate limits again 🕐

henrymercer:
> 😮‍💨 I had an error that the new `setup-swift` step couldn't be found without running `checkout`, so I ran it again just before `setup-swift` but this overwrote the whole directory (I think). Reverted and trying to figure out why `setup-swift` can't be found when the `checkout` Action is run at the beginning of each check. But hit API rate limits again 🕐

Hmm, this is a little annoying.  I think the problem is that the `prepare-test` step moves everything in the CodeQL Action repo apart from the multi-language test repository into `action`:

```
Run mkdir ../action
  mkdir ../action
  mv * .github ../action/
  mv ../action/tests/multi-language-repo/{*,.github} .
  mv ../action/.github/workflows .github
```

I expect that if we replace the call by `./action/.github/setup-swift` then that should work.

angelapwen:
> Hmm, this is a little annoying. I think the problem is that the `prepare-test` step moves everything in the CodeQL Action repo apart from the multi-language test repository into `action`:
> 
> ```
> Run mkdir ../action
>   mkdir ../action
>   mv * .github ../action/
>   mv ../action/tests/multi-language-repo/{*,.github} .
>   mv ../action/.github/workflows .github
> ```
> 
> I expect that if we replace the call by `../action/.github/setup-swift` then that should work.

Thank you — that did it! That makes sense 😸

.github/setup-swift/action.yml
```diff
@@ -0,0 +1,32 @@
...
+      env: 
+        CODEQL_PATH: ${{inputs.codeql-path}}
+      run: |
+        if [ $RUNNER_OS = "macOS" ]; then
+          PLATFORM="osx64"
+        else # We do not run this step on Windows.
+          PLATFORM="linux64"
+        fi 
+        SWIFT_EXTRACTOR_DIR=$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')
+        VERSION=$($SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor --version | awk '/version/ { print $3 }')
```
adityasharad, line 23: Does not change the meaning in this case, but this is a good practice to adopt, since it will save you a lot of annoyance if the command output ever contains spaces.
```suggestion
        SWIFT_EXTRACTOR_DIR="$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')"
        VERSION="$("$SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor" --version | awk '/version/ { print $3 }')"
```

adityasharad (dismissed):
Looks sensible, good work! Minor Bash suggestions.

.github/setup-swift/action.yml
angelapwen: Makes sense, thank you!

adityasharad (approved):

angelapwen:
Hm... the PR checks I changed don't seem to have run anymore. I'll close and reopen to see if this triggers them.

angelapwen:
Ok.. that seems to have done it 😆 re-requesting review (again), sorry!

adityasharad (approved):

PR #1432, opened 2022-12-09
Add telemetry for uploading failed runs

henrymercer:
#1393 started uploading failed runs to the code scanning API.  In preparation for rolling this out, it would be useful to gather some telemetry, particularly since things are set up to be unlikely to hard-fail, so we probably won't get any customer feedback.

- [Example run with a skipped reason](https://github.com/github/codeql-action/actions/runs/3651268514/jobs/6168362274)
- [Example run with an error](https://github.com/github/codeql-action/actions/runs/3651420972/jobs/6168671829)


src/init-action-post-helper.ts
```diff
@@ -69,31 +100,32 @@ export async function uploadFailedSarif(
...
-      await uploadFailedSarif(config, repositoryNwo, featureEnablement, logger);
+      return await uploadFailedSarif(
+        config,
+        repositoryNwo,
+        featureEnablement,
+        logger
+      );
     } catch (e) {
       if (expectFailedSarifUpload) {
         throw new Error(
```
aeisenberg, line 131: This only happens during testing, right? Also, below.

src/init-action-post-helper.ts
```diff
@@ -69,31 +100,32 @@ export async function uploadFailedSarif(
     logger,
     { isUnsuccessfulExecution: true }
   );
+  return {
+    upload_failed_run_raw_upload_size_bytes:
+      uploadResult?.statusReport?.raw_upload_size_bytes,
+    upload_failed_run_zipped_upload_size_bytes:
+      uploadResult?.statusReport?.zipped_upload_size_bytes,
+  };
```
aeisenberg, line 108: So, with this change, there will be two status reports in this case? One for the upload and one for logging the upload failure, I think. If so, that might lead to some confusion.

src/init-action-post-helper.ts
henrymercer: Correct — `expectFailedSarifUpload` is just for testing.  There's a comment about that in the code, but you have to expand the diff to see it.

src/init-action-post-helper.ts
henrymercer: There's a bit of name overloading going on here — the upload process doesn't submit a status report to the API.  `uploadResult?.statusReport` is rather a list of fields that:

- Are included in the `finish` status report sent at the end of `analyze`
- Are included in the status report sent at the end of `upload-sarif`
- (As of this PR) are included in the status report sent at the end of the post `init` Action.

So we only ever upload one status report about the failed SARIF file.

Note that if `analyze` produces a SARIF file, tries to upload that SARIF file, but uploading that SARIF fails catastrophically, we will get a `finish` status report with info about that SARIF upload attempt, and also a `init-post` status report with info about the failed SARIF file upload.

henrymercer: There's a bit of name overloading going on here — the upload process doesn't submit a status report to the API.  `uploadResult?.statusReport` is rather a list of fields that:

- Are included in the `finish` status report sent at the end of `analyze`
- Are included in the status report sent at the end of `upload-sarif`
- (As of this PR) are included in the status report sent at the end of the post `init` Action.

So we only ever upload one status report about the failed SARIF file.

Note that if `analyze` produces a SARIF file, tries to upload that SARIF file, but uploading that SARIF fails catastrophically, we will get a `finish` status report with info about that SARIF upload attempt, and also a `init-post` status report with info about the failed SARIF file upload.

aeisenberg: Ah...ok. That makes sense, but this leads to another question: why change the name from `raw_upload_size_bytes` to `upload_failed_run_raw_upload_size_bytes` and the other field too? When we upload the failed status report, we already know that the report is for a failure. Is there any need to make these fields differently named?

henrymercer: Introducing a new field is a more conservative option that reduces the risk of breaking internal queries that look at `raw_upload_size_bytes` without filtering to `action_name in { "finish", "upload-sarif" }`.  I have looked at some common places and can't find any queries that don't look at all the fields, but I don't believe there's a way to check all uses of telemetry.

I'm ambivalent about which way we go here — let me know how convinced you are here.

angelapwen: Just as another data point, I also found the rename a bit confusing (but understand the conservative approach)

src/init-action-post-helper.ts
```diff
@@ -104,12 +136,41 @@ export async function run(
...
+  featureEnablement: FeatureEnablement,
+  logger: Logger
+) {
+  const config = await getConfig(actionsUtil.getTemporaryDirectory(), logger);
+  if (config === undefined) {
+    logger.warning(
+      "Debugging artifacts are unavailable since the 'init' Action failed before it could produce any."
+    );
+    return;
+  }
```
angelapwen, line 166: 👍 thanks!

src/init-action-post.ts
```diff
@@ -27,22 +42,45 @@ async function runWrapper() {
...
+    "init-post",
+    "success",
+    startedAt
+  );
+  const statusReport: InitPostStatusReport = {
+    ...statusReportBase,
+    ...uploadFailedSarifResult,
+  };
+  console.log("status report");
+  console.log(statusReport);
```
angelapwen, line 82: Are these just for debug purposes? I don't think we're logging the full status report in other instances.

src/init-action-post-helper.ts
henrymercer: Thanks, that's helpful.  I've pushed a commit to re-use the existing fields.

src/init-action-post.ts
henrymercer: Yeah, just for debugging.  I think I removed this in a subsequent commit.

henrymercer: Oops, these should go, thanks!

src/init-action-post-helper.ts
```diff
@@ -16,17 +16,37 @@ import {
   getWorkflow,
 } from "./workflow";
 
+export interface UploadFailedSarifResult extends uploadLib.UploadStatusReport {
```
angelapwen, line 19: Non-blocking: just wondering why we're now extending this `UploadStatusReport`? Is it just best practice to do so when we're submitting results related to the status of an upload?

angelapwen (approved):
Thanks!!

src/init-action-post-helper.ts
henrymercer: This is just to indicate that this result may include the `raw_upload_size_bytes` and `zipped_upload_size_bytes` fields from the `UploadStatusReport`

angelapwen: A ha, I see 👍

PR #1434, opened 2022-12-12
remove use of query-string package

nickfyson:


src/codeql.ts
```diff
@@ -499,7 +497,7 @@ export async function setupCodeQL(
         // We also don't want to send an authorization header if there's already a token provided in the URL.
         if (
           codeqlURL.startsWith(`${apiDetails.url}/`) &&
-          parsedQueryString["token"] === undefined
+          new URLSearchParams(parsedCodeQLURL.search).get("token") === undefined
```
henrymercer, line 500: I believe `URLSearchParams.get` returns `null` not `undefined` when the parameter's missing.  Could we swap this to `.has` instead?

henrymercer (approved):
Nice!

PR #1441, opened 2022-12-20
Remove tests with old certifi dependency

henrymercer:
See https://github.com/advisories/GHSA-43fp-rhv2-5gv8.  It is not possible to upgrade the Python 2 tests to use the patched certifi, as this requires Python >=3.6, so for now we're removing them.


angelapwen (approved):
Looks fine on the Action-side, assume we are waiting for @github/codeql-python to review before merge 👍

henrymercer:
We can merge with a review from either team here :)

PR #1444, opened 2022-12-21
Improve reporting failed runs via SARIF

henrymercer:
This PR makes some minor improvements to the way we report failed runs via SARIF.  The first commit adjusts the pre-check to check whether the `analyze` Action has completed, to allow a quicker short circuit exit when `upload: false` is specified but the workflow run was successful.  The remaining commits improve error messages.


src/init-action-post-helper.ts
```diff
@@ -98,11 +98,11 @@ export async function uploadSarifIfRunFailed(
   featureEnablement: FeatureEnablement,
   logger: Logger
 ): Promise<UploadFailedSarifResult> {
-  // Environment variable used to integration test uploading a SARIF file for failed runs
+  // This environment variable is used to integration test uploading a SARIF file for failed runs
   const expectFailedSarifUpload =
     process.env["CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF"] === "true";
 
-  if (process.env[CODEQL_ACTION_ANALYZE_DID_UPLOAD_SARIF] !== "true") {
+  if (process.env[CODEQL_ACTION_ANALYZE_DID_COMPLETE_SUCCESSFULLY] !== "true") {
```
angelapwen, line 105: I'm a little confused about this change: doesn't it change the original behavior? It looks like previously we attempted to upload the failed SARIF here only if `upload: false`, which means the env variable wasn't set to `true`; but now we attempt to upload it regardless of the the value of `upload`.

I think the change makes more sense to me and I like the `Analyze Action completed successfully` reason below, but I'm not clear on how it preserves original behavior and causes a short-circuit 🤔

henrymercer: The `uploadFailedSarif` inner function checks the `upload` and `category` inputs before actually doing the upload — perhaps I could rename that function 🤔

The short-circuit is the circumstance in which the run succeeded but the user specified `upload: false`.  Before, `CODEQL_ACTION_ANALYZE_DID_UPLOAD_SARIF` wouldn't be `true`, so we'd run `uploadFailedSarif`, introspect the workflow, and only then discover that `upload` was set to `false`.  Now, `CODEQL_ACTION_ANALYZE_DID_COMPLETE_SUCCESSFULLY` will be true, so we can skip uploading the failed SARIF without having to introspect the workflow.

angelapwen: A ha... I see, I didn't click into the `uploadFailedSarif` inner function because I figured it just uploaded 😄

angelapwen (approved):
✅, optionally could rename `uploadFailedSarif` function 😸

angelapwen (approved):
Looks good, thanks for renaming + adding extra tests! 😄

PR #1462, opened 2023-01-05
Refactor CodeQL setup

henrymercer:
This PR refactors how the Action sets up the CodeQL tools.  The aim is to make the code more readable and to set the stage for improving how we roll out new CodeQL versions.


angelapwen (dismissed):
Thank you for this!!!! It's so much clearer now. And, I learned that we have `AssertNever` now

src/codeql.ts
```diff
@@ -403,21 +403,162 @@ async function getCodeQLBundleDownloadURL(
...
+    };
+
+async function getCodeQLSource(
+  toolsInput: string | undefined,
+  bypassToolcache: boolean,
+  apiDetails: api.GitHubApiDetails,
+  variant: util.GitHubVariant,
+  logger: Logger
+): Promise<CodeQLToolsSource> {
+  if (toolsInput && toolsInput !== "latest" && !toolsInput.startsWith("http")) {
```
aeisenberg, line 427: Minor:

```suggestion
  if (toolsInput !== "latest" && !toolsInput?.startsWith("http")) {
```

src/codeql.ts
```diff
@@ -403,21 +403,162 @@ async function getCodeQLBundleDownloadURL(
...
+  // This avoids leaking Enterprise tokens to dotcom.
+  // We also don't want to send an authorization header if there's already a token provided in the URL.
+  if (
+    codeqlURL.startsWith(`${apiDetails.url}/`) &&
+    !searchParams.has("token")
+  ) {
+    logger.debug("Downloading CodeQL bundle with token.");
+    headers.authorization = `token ${apiDetails.auth}`;
+  } else {
+    logger.debug("Downloading CodeQL bundle without token.");
```
aeisenberg, line 524: In this case, the URL may still have a `token` search param, Should we explicitly remove it?

aeisenberg: Looks like this was always here, so maybe not a problem?

src/codeql.ts
henrymercer: I think that would start accepting `toolsInput === undefined`, which we don't want here.

src/codeql.ts
henrymercer: "without adding token" would be more accurate — the goal here is to log that we aren't using the `token` input to `init` here.

aeisenberg (approved):

PR #1464, opened 2023-01-06
Send the external repository token to the CLI

aeisenberg:


aeisenberg:
Failing test is because the cached version of the cli is too old to be using cli config parsing now, so the cli config files are not generated.

aeisenberg:
@henrymercer, I think you are the one with the most background to review. Can you take a look?

src/codeql.ts
```diff
@@ -885,24 +884,35 @@ async function getCodeQLForCmd(
         }
       }
 
+      // A config file is only generated if the CliConfigFileEnabled feature flag is enabled.
+      // Only pass external repository token if a config file is
```
henrymercer, line 888: Minor: Trailing sentence here.

src/codeql.ts
```diff
@@ -1354,6 +1364,7 @@ async function runTool(cmd: string, args: string[] = []) {
       },
     },
     ignoreReturnCode: true,
+    input: Buffer.from(stdin || ""),
```
henrymercer, line 1367: Minor: I'd feel more comfortable only providing an `input` property when we actually have an input (e.g. `...(stdin ? { input: Buffer.from(stdin) } : {})`), but this is probably fine.

src/codeql.ts
```diff
@@ -885,24 +884,35 @@ async function getCodeQLForCmd(
...
+        [
+          "database",
+          "init",
+          "--db-cluster",
+          config.dbLocation,
+          `--source-root=${sourceRoot}`,
+          ...extraArgs,
+          ...getExtraOptionsFromEnv(["database", "init"]),
+        ],
+        externalRepositoryToken
```
henrymercer, line 914: Nit: it's a little unclear what this is doing.  Consider having `runTool` destructure its input like the `toolrunner.ToolRunner` constructor, so we can make it very clear it's input.

henrymercer (dismissed):
A few minor comments, but this looks generally good to me 👍

src/codeql.ts
aeisenberg: I added an `opts` parameter that has an optional `stdin` property. I think this is what you are asking for.

henrymercer: Thanks!

aeisenberg:
Failures due to using a kotlin version that is too new. Updating the base branch should fix it.

henrymercer (approved):

PR #1466, opened 2023-01-06
Update bundle to 2.12.0

dbartol:
### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #1474, opened 2023-01-10
Refactor CodeQL setup tests and fix GHAE test

henrymercer:
This PR refactors the CodeQL setup tests and fixes a GHAE setup test.

### Refactoring of CodeQL setup tests

- I've split up `mockApiAndSetupCodeQL` into `mockDownloadApi`, `installIntoToolcache`, and `setupCodeQL` so it's a easier to understand the intent of the tests.
- I've renamed the tests to include assertions about expected behavior.
- I've also renamed the bundle tag name from `version` to `tagName` to prepare for the controlled rollout changes, which will involve keeping track of both a CLI version and a bundle tag name.

### Fix to GHAE setup test

- Previously we weren't actually testing GHAE setup.
  - The GHAE API URL was missing, so calling the GHAE endpoint failed.
  - However the test didn't fail because a mock for the fallback Dotcom endpoint was hanging around from a previous test.
- I have added the missing GHAE API URL and added a call to the test infrastructure to clear all `nock` network request mocks after each test, just as we clear all `sinon` mocks.


henrymercer:
Checks are failing on main — merging in main once #1466 is merged should fix.

src/codeql.test.ts
```diff
@@ -69,21 +67,15 @@ test.beforeEach(() => {
   };
 });
 
-async function mockApiAndSetupCodeQL({
-  apiDetails,
-  bypassToolcache,
+async function mockDownloadApi({
```
angelapwen, line 70: Nit: it wasn't super clear to me what the `Promise` this function returned was, could we add a comment indicating it's returning the URL or rename the function?

src/codeql.test.ts
```diff
@@ -93,8 +85,8 @@ async function mockApiAndSetupCodeQL({
 
   const baseUrl = apiDetails?.url ?? "https://example.com";
   const relativeUrl = apiDetails
-    ? `/github/codeql-action/releases/download/${version}/codeql-bundle-${platform}.tar.gz`
-    : `/download/codeql-bundle-${version}/codeql-bundle.tar.gz`;
+    ? `/github/codeql-action/releases/download/${tagName}/codeql-bundle-${platform}.tar.gz`
```
angelapwen, line 88: To be clear — this is where you're renaming the bundle tag name in the tests from `version` to tagName` that includes the `codeql-bundle-` prefix, right?

angelapwen (dismissed):
A couple of small notes and there's a merge conflict with one of the source map files, but overall it looks great and clean! Thanks for doing all this refactoring ✨

src/codeql.test.ts
henrymercer: Good call, I've added some JSDoc.

src/codeql.test.ts
henrymercer: Correct!

angelapwen (approved):
Thank you!!

PR #1475, opened 2023-01-10
Enable a controlled switchover between CodeQL releases

henrymercer:
This PR adds functionality that we will use to enable a controlled switchover between CodeQL releases.  Specifically, we would like to ensure that each repository running CodeQL transitions from the previous CodeQL release to the new CodeQL release once.  We want to avoid repositories flip-flopping between CodeQL releases during the rollout of a new CodeQL release, specifically during the Actions runner image update process, as this creates alert churn.

To do this, we decide the default version of CodeQL on Dotcom using feature flags rather than by using what's in the toolcache.  Specifically, we look at all the `default_codeql_version_x_enabled` feature flags and pick the latest enabled version.

One niggle with this approach is that if the CLI version we need isn't in the toolcache, there isn't yet a nice way to find the CodeQL Bundle release corresponding to that CLI version.  In the medium term, we'll consider tagging each CodeQL bundle release with a CLI version number, for instance `codeql-bundle-v2.12.0`.  In the meantime, we add a special asset to the release `cli-version-2.12.0.txt` specifying the version number.  The contents of this asset don't matter, since we don't want to have to download anything.

GHES behaviour is unchanged: we account for the new toolcache format, but we continue to use the CodeQL bundle release specified within the Action (in `defaults.json`) by default.  We also continue to allow CodeQL bundles that have been baked into Actions runner images to override the version in `defaults.json`.

Commit-by-commit-review recommended.

### TODO before merging

- [x] Add a changelog note for the impact on the layout of CodeQL within the Actions runner images.
- [x] Adapt the functionality that enables using the toolcache for manually specified bundle URLs to the new runner image structure.  This will require a search through toolcache versions or a bundle tag name to CodeQL CLI version lookup.
- [x] Add some more unit tests for setting up CodeQL.  Also require an owner and a repo in the GitHub Releases API mocking as a regression test.

- [ ] Confirm this change is backwards compatible with existing workflows. — see changelog notes


src/init.ts
```diff
@@ -8,7 +8,7 @@ import * as analysisPaths from "./analysis-paths";
 import { GitHubApiCombinedDetails, GitHubApiDetails } from "./api-client";
 import { CodeQL, CODEQL_VERSION_NEW_TRACING, setupCodeQL } from "./codeql";
```
angelapwen, line 9: I think the test errors are coming from `setupCodeQL` being imported from `./codeql` rather than `./setup-codeql`. (And likely the version in `./codeql` can be removed?)

henrymercer: It was pretty confusing having two `setupCodeQL` functions, so I've renamed the one in `setupCodeQL` to `setupCodeQLBundle`.  I think the test errors were coming from forgetting to specify the owner and repo in the releases API call.  We didn't catch this in tests because we weren't requiring a owner and repo in the mock.

angelapwen (commented):
Looked through commit-by-commit and everything makes sense to me. Just a few comments for now, will take another 👀 tomorrow!

src/codeql.ts
```diff
@@ -314,6 +314,45 @@ export function getCodeQLActionRepository(logger: Logger): string {
...
+  const codeQLActionRepository = getCodeQLActionRepository(logger);
+  const releases = await apiClient.paginate(apiClient.repos.listReleases, {
+    owner: codeQLActionRepository.split("/")[0],
+    repo: codeQLActionRepository.split("/")[1],
+  });
+  logger.debug(`Found ${releases.length} releases.`);
+
+  for (const release of releases) {
+    const cliVersionFileVersions = release.assets
+      .map((asset) => asset.name.match(/cli-version-(.*)\.txt/)?.[1])
```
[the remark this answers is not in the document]
henrymercer: Good call, done!

src/feature-flags.test.ts
```diff
@@ -371,7 +372,61 @@ test("Environment variable can override feature flag cache", async (t) => {
...
+      await featureEnablement.getDefaultCliVersion(GitHubVariant.DOTCOM),
+      {
+        cliVersion: "2.12.1",
+        variant: GitHubVariant.DOTCOM,
+      }
+    );
+  });
+});
+
+test(`selects CLI v2.11.6 on Dotcom when no default version feature flags are enabled`, async (t) => {
```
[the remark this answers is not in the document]
henrymercer: Good question, this is something that's up for discussion.  My preference is to remove this fallback once 2.12.0 is the default CLI version and instead fail.  The benefit of this is that we avoid alert churn from selecting a different default CLI version if we mess up the feature flags and set them all to disabled.

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The default version of the CodeQL tools when running the CodeQL Action on github.com will now be more stable during the release of new GitHub Actions runner images. This will prevent fluctuations in code scanning alerts while a new runner image release rolls out to GitHub-hosted Actions runners. [#1475](https://github.com/github/codeql-action/pull/1475).
+- Upcoming versions of the GitHub Actions runner images will include a change to the layout of the CodeQL tools within the Actions toolcache. Specifically, the Actions toolcache will be pre-populated with the latest two versions of the CodeQL tools, rather than just the latest CodeQL tools. The version number of the CodeQL tools within the toolcache will also change to incorporate the version number of the corresponding release of the CodeQL CLI. Users should continue to avoid depending on the layout of the CodeQL tools within the toolcache, as this may change without warning in the future.
```
aeisenberg, line 6: Can you include a link to the actions documentation about the toolcache? Some users may not be familiar with it.

Also, "pre-populated with the latest two versions", can you mention somewhere that it was previously only the latest version?

"The version number of the CodeQL tools within the toolcache will also change to incorporate the version number of the corresponding release of the CodeQL CLI."

It took me a second to realize what this meant. Previously, you had to choose a cli version by release date, but now we are accepting release date _or_ semver? I don't think this is clear to someone who already isn't familiar with this.

aeisenberg:
> One niggle with this approach is that if the CLI version we need isn't in the toolcache, there isn't yet a nice way to find the CodeQL Bundle release corresponding to that CLI version. In the medium term, we'll consider tagging each CodeQL bundle release with a CLI version number, for instance `codeql-bundle-v2.12.0`. In the meantime, we add a special asset to the release `cli-version-2.12.0.txt` specifying the version number. The contents of this asset don't matter, since we don't want to have to download anything.

Could we add a json file in this repo that has mappings from bundle release number to version number or URL? This would be helpful in general for our users since there's no easy way for them to figure out this mapping right now without downloading and running `codeql version`.

src/codeql.ts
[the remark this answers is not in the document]
angelapwen: Thanks, super clear now!

src/feature-flags.test.ts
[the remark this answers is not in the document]
angelapwen: If we mess up the feature flags and set them all to disabled, then is it correct that we would start failing all workflows using the Action if we don't have a fallback? That would allow us to avoid alert churn and set off our monitors very quickly, but it seems like a catastrophic failure 👀

angelapwen:
👍 on new additions. CI failure is due to known Swift autobuild hang and I've rerun.

CHANGELOG.md
AlonaHlobina: I agree with @aeisenberg, it also took me a good time to understand the second point. What is not clear:
- Do CodeQL users familiar with the concept of Actions toolcache? Do we need to disclose such details? 
- Who exactly is affected by the change? 
- What action should users take?
- What exactly should they not do starting now?

src/feature-flags.test.ts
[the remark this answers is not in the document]
henrymercer: Let's get some product input on this.  My intuition is that failed workflows might be better than alert churn, but I think we should discuss this further.  I'll create an internal issue.

henrymercer:
> Could we add a json file in this repo that has mappings from bundle release number to version number or URL? This would be helpful in general for our users since there's no easy way for them to figure out this mapping right now without downloading and running `codeql version`.
> 
> This could go in `defaults.json`, which we're already updating for each release.

I have a slight preference for keeping the GitHub Releases the source of truth:
- This aligns better with our medium term goal of changing the version numbering system of bundle releases to include the CLI version number.  Once we've done that a mapping in `defaults.json` would be redundant.
- We can automatically add the CLI version marker file to the bundle release as part of the publishing process, saving a bit of manual work.

src/feature-flags.test.ts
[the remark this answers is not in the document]
henrymercer: Hmm, while writing up that issue I changed my mind.  See the backlinked internal issue.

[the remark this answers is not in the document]
angelapwen: I actually don't see an update in the backlinked issue on this PR 👀

CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
...
+- Code scanning alerts will now be more stable during the release of new GitHub Actions runner images for customers running the CodeQL Action on github.com. [#1475](https://github.com/github/codeql-action/pull/1475).
+- To accomplish this, the internal layout of the GitHub Actions tool cache has changed. This will affect a minority of customers as follows.
+  - Customers with workflows on github.com that are pinned to specific old versions of the CodeQL Action (e.g. `v2.1.32`) will no longer obtain the newest version of CodeQL from the tool cache, and will instead download a compatible older version of CodeQL from GitHub Releases. To continue using the newest version of CodeQL, please update your workflows to reference the latest version of the CodeQL Action (`v2`).
+  - Customers directly interacting with the GitHub Actions tool cache directly, for example via the `@actions/tool-cache` npm package or directly on the filesystem of a GitHub Actions runner, may need to update their workflows to take into account the following changes:
```
aeisenberg, line 8:
```suggestion
  - Customers interacting with the GitHub Actions tool cache directly, for example via the `@actions/tool-cache` npm package or directly on the filesystem of a GitHub Actions runner, may need to update their workflows to take into account the following changes:
```

aeisenberg: I think you had something before where you mentioned that referencing codeql directly in the file system is not recommended since that is internal. I think that is an important point.

src/setup-codeql.ts
```diff
@@ -153,7 +153,7 @@ export async function tryFindCliVersionDotcomOnly(
     });
     return tryGetCodeQLCliVersionForRelease(release.data, logger);
   } catch (e) {
-    logger.error(
+    logger.debug(
       `Failed to find the CLI version for the CodeQL bundle tagged ${tagName}. Error: ${
```
aeisenberg, line 157: Minor: Since this is not an "error":

```suggestion
      `Failed to find the CLI version for the CodeQL bundle tagged ${tagName}. Reason: ${
```

aeisenberg (commented):
Given the complexity of this change and the fact that the actions changelog is not very visible to users, are you considering adding this to a changelog post?

src/feature-flags.test.ts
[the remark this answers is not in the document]
henrymercer: Let's continue discussion on the internal issue

CHANGELOG.md
henrymercer: Thanks both for your comments, PTAL at the latest update if you haven't seen it already!

henrymercer:
> Given the complexity of this change and the fact that the actions changelog is not very visible to users, are you considering adding this to a changelog post?

I think a changelog post would be a good idea.  Will discuss with Alona once we've agreed on a good Action changelog note.

CHANGELOG.md
henrymercer: I've added a sentence recommending users obtain the CLI through `gh codeql` or GitHub Releases instead.

angelapwen (dismissed):
✅ from me on the code changes, but I believe there is still ongoing discussion about the inclusion of the section for advanced users in the changelog note.

angelapwen (approved):
Approving but looks like merge conflicts need to be resolved 😄

henrymercer:
Hmm since we already have merge conflicts here, it might make sense to get https://github.com/github/codeql-action/pull/1492 in first and then rebase on that, as I expect the TypeScript update will probably create some more conflicts.

henrymercer:
TypeScript PR is failing on Windows, so let's do this PR first instead.

angelapwen (approved):
Looks like the PR checks are failing with parsing SARIF on Windows but it should be unrelated to this PR 🤔

PR #1485, opened 2023-01-17
Hide error message and stack for non-error

aeisenberg:
Avoid printing the error message and stack when we fail to find the commit. This will happen in several non-error states (e.g., when there is no repository checked out at the CWD). Move the error message to a debug message so that it is still available if someone really wants to see it.


src/actions-util.ts
```diff
@@ -88,9 +88,10 @@ export const getCommitOid = async function (
     return commitOid.trim();
   } catch (e) {
     core.info(
-      `Failed to call git to get current commit. Continuing with data from environment or input: ${e}`
+      "Could not determine current commit SHA. Continuing with data from the environment."
```
angelapwen, line 91: The change to the error message seems to lose a bit of information to me — does getting the SHA from the input count as data from the environment? How about:

```suggestion
      "Could not determine current commit SHA using git. Continuing with data from user input or environment."
```

aeisenberg: Sure...that reads better, even though it's a little more complicated since in most situations, by the time it hits here, we will have already checked user input. It actually looks like the logic is a bit overly complicated and it would be nice to simplify, but that's out of the scope of this change.

aeisenberg:
@angelapwen Thanks for your review. Can you re-approve?

angelapwen (approved):
These changes look unrelated to the CI failures to me — are we failing CI on `main`?

aeisenberg:
Yep. Failing on main. Checking on this now.

PR #1487, opened 2023-01-18
Fix a bug in cli config parsing

aeisenberg:
When the config specifies queries in external repositories and no other queries and cli config parsing is enabled, the analyze job is not able to find any queries to run and it fails. This PR fixes the problem.

Also add some logging statements when writing a copy of the user config file.


aeisenberg:
Closing and re-opening with config parsing enabled in this repository.

src/codeql.ts
```diff
@@ -1448,6 +1451,10 @@ async function generateCodeScanningConfig(
       augmentedConfig.packs["javascript"].push(packString);
     }
   }
+  logger.info(`Writing augmented user configuration file to ${configLocation}`);
+  logger.startGroup("Augmented user configuration file contents");
+  logger.info(yaml.dump(augmentedConfig));
+  logger.endGroup();
```
henrymercer, line 1457: I think this makes sense as info while we're rolling out the flag, but let's consider making this debug afterwards.

henrymercer (approved):

src/codeql.ts
aeisenberg: Sure.

PR #1492, opened 2023-01-18
Upgrade to TypeScript 9.2.0

henrymercer:
Now that we've dropped Node 12 support, we can upgrade to the latest version of TypeScript!

- I've brought a bunch of other packages up to date, since we need to update our linter and test suite to be compatible with the new TypeScript version.
- I've added type annotations to account for the stricter type checking that's now enabled.
- I've bumped our compilation target from ES2019 to ES2021 (see https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping). 


henrymercer:
@github/codeql-action-reviewers This is ready for review.  The two checks that are still yellow have been removed, and we'll need to update the required checks prior to merging.

src/codeql.ts
```diff
@@ -525,7 +525,7 @@ async function downloadCodeQL(
     `Downloading CodeQL tools from ${codeqlURL}. This may take a while.`
   );
 
-  const dest = path.join(tempDir, uuidV4());
+  const dest = path.join(tempDir, uuidV4() as string);
```
aeisenberg: Is this because we don't have the typings for the uuid library?

src/codeql.ts
```diff
@@ -987,7 +987,7 @@ async function getCodeQLForCmd(
       // Set trace command
       const ext = process.platform === "win32" ? ".cmd" : ".sh";
       const traceCommand = path.resolve(
-        JSON.parse(extractorPath),
+        JSON.parse(extractorPath) as string,
```
aeisenberg, line 679: So `extractorPath` is a JSON string that when parsed is a javascript string? That's surprising, but trying it out, I see that's exactly what is returned.

src/autobuild-action.ts
```diff
@@ -36,7 +36,7 @@ async function sendCompletedStatusReport(
   failingLanguage?: string,
   cause?: Error
 ) {
-  initializeEnvironment(pkg.version);
+  initializeEnvironment(pkg.version as string);
```
aeisenberg, line 39: Minor, we can change line 24 to add a type for pkg:

```
interface PackageJson { ... }
const pkg = require("../package.json") as PackageJson;
```

src/init-action.ts
```diff
@@ -138,8 +138,8 @@ async function sendSuccessStatusReport(
 async function run() {
   const startedAt = new Date();
   const logger = getActionsLogger();
-  initializeEnvironment(pkg.version);
-  await checkActionVersion(pkg.version);
+  initializeEnvironment(pkg.version as string);
+  await checkActionVersion(pkg.version as string);
```
aeisenberg, line 142: See comment above about a type for pacakge.json

src/upload-sarif-action.ts
```diff
@@ -36,8 +36,8 @@ async function sendSuccessStatusReport(
 
 async function run() {
   const startedAt = new Date();
-  initializeEnvironment(pkg.version);
-  await checkActionVersion(pkg.version);
+  initializeEnvironment(pkg.version as string);
+  await checkActionVersion(pkg.version as string);
```
aeisenberg, line 40: See above

aeisenberg (commented):
Nice! 

This was really hard to review since the PR tools in the browser just wasn't loading the code for me. Or I had to wait 10-15 seconds for rendering.

aeisenberg (commented):
A few minor comments for typing, but I think it's basically good.

src/codeql.ts
henrymercer: I'm not sure why we request a JSON format here.  `format=text` sounds potentially more sensible really when we're just asking for a path.  I think in the absence of a reason not to, let's change this so we can eliminate a typecast.

src/codeql.ts
henrymercer: Yes, let's add these via `@types/uuid`

src/autobuild-action.ts
henrymercer: 👍 Let's use the JSON schema exposed by `@schemastore/package`

src/codeql.ts
henrymercer: I've added a comment to explain why I think we do this — it eliminates the need to remove the trailing new line we get with `--format=text`.

src/codeql.ts
```diff
@@ -650,6 +650,9 @@ export async function getCodeQLForCmd(
     async extractScannedLanguage(config: Config, language: Language) {
       const databasePath = util.getCodeQLDatabasePath(config, language);
       // Get extractor location
+      //
+      // Request it using `format=json` so we don't need to strip the trailing new line generated by
+      // the CLI.
```
aeisenberg, line 655: We could just call `.trim()` on the result and strip the trailing newline anyway, but it's fine the way it is. I guess it's ever so slightly safer parsing the JSON in case the path ends with whitespace.

aeisenberg (approved):
LGTM.

src/codeql.ts
henrymercer: That was exactly my thinking 👍

PR #1494, opened 2023-01-19
Fix a bug that forced the `init` Action to run for at least two minutes on JavaScript

henrymercer:
The bug is that we wait for timeouts to elapse at the end of the `init` Action, even if the operation we were applying the timeout to already finished.

### Context

Some of the Actions caching APIs we call don't have reliable timeouts, so we implement our own timeout on top of this.  Once this timeout elapses, the Action continues executing without downloading / uploading the cache.  However, once the top level of the Action has finished executing, the Node runtime will continue waiting on the background tasks to finish executing.  We get around this by adding a call to `util.checkForTimeout()` to the end of the top level of the Action.  This kills the Node process after 30 seconds if there was a timeout at some point during the Action's execution (see https://github.com/github/codeql-action/pull/1354 for more).

However it turns out we're liable to the same waiting problem with the timeouts themselves.  If the cache operation finishes before the timeout, we'll still wait for the timeout itself to finish executing before the Action exits.  This impacts JavaScript runs, which use TRAP caching.  The TRAP caching timeout is currently two minutes, so the `init` Action will now wait for up to 2 minutes to finish on any workflows that run a language that supports TRAP caching.

### Fix

To fix this, we immediately `unref` the timer so that Node won't wait for it to finish.


src/util.ts
```diff
@@ -552,7 +552,9 @@ export async function bundleDb(
 }
 
 export async function delay(milliseconds: number) {
-  return new Promise((resolve) => setTimeout(resolve, milliseconds));
+  // Immediately `unref` the timer such that it doesn't hold up the event loop
```
adityasharad, line 555: Looks good. Would it be possible to extract this unref pattern into a common helper so that we can easily remember to do this elsewhere, or is that not worth the trouble?

henrymercer: It's a good question.  My initial thought was that if we created our own `setTimeout` function, it's fairly possible that someone would overlook it and just call Node's `setTimeout` directly.  That said, perhaps it's worth it if only to avoid writing that comment twice.

adityasharad: Then we could easily enforce it with a custom query! But certainly none of this should block the PR.

henrymercer: With my latest commit we implement `withTimeout` using `delay` so we only need to use Node's `setTimeout` once.  I think I'll defer the custom query to future work.

adityasharad (approved):

aeisenberg (dismissed):

adityasharad (approved):

PR #1496, opened 2023-01-20
Increase strictness of tool caching threshold to avoid caching nightlies

henrymercer:
If the CLI version number of a CodeQL Bundle is a pre-release, then we want to avoid versioning the bundle in the toolcache as `<cliVersion>-<bundleVersion>`, so we don't interpret this toolcache entry as containing a stable CLI release.

However, we currently version CodeQL nightlies as `x.y.z+<timestamp>`.  Even though these are not pre-releases, we do not want bundles containing them to override CodeQL bundles containing stable versions of the CodeQL CLI.

Therefore this PR ensures we only version bundles as `<cliVersion>-<bundleVersion>` in the toolcache when `cliVersion` is of the form `x.y.z`.  All other bundles will be versioned as `0.0.0-<bundleVersion>` within the toolcache.


angelapwen (approved):
Looks good to me~

PR #1497, opened 2023-01-21
Send tools telemetry to `init` status report

angelapwen:
This PR sends the following fields to the `init` status report on success and failure:

- `tools_source`
- if `tools_source` was `DOWNLOAD`, `tools_download_duration_ms` and `tools_feature_flags_valid`

Note that previously we only sent additional fields on `init` success, but because we would like to support the tools fields on failure as well, the PR changes the status report behavior so that the `init` report is sent on failure too, with as many fields populated as possible.


angelapwen:
Looks like a bit of an inscrutable error, `init action failed: TypeError: (0 , actions_util_1.getActionVersion) is not a function` when it is correctly defined in `actions-util` and I haven't changed anything about it. 

Also, oddly enough the error fails in the `analyze` step even though it originates from `init`

henrymercer:
I think this could be as boring as just needing to merge in main and recompile the Action — some of the compiled files seem to be missing at the moment.

src/init.ts
```diff
@@ -9,6 +9,7 @@ import { GitHubApiCombinedDetails, GitHubApiDetails } from "./api-client";
 import { CodeQL, CODEQL_VERSION_NEW_TRACING, setupCodeQL } from "./codeql";
 import * as configUtils from "./config-utils";
 import { CodeQLDefaultVersionInfo, FeatureEnablement } from "./feature-flags";
+import { ToolsSource } from "./init-action";
```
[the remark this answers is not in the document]
angelapwen: Oh, that's interesting. I'll move the declaration to `init` and see how it goes.

angelapwen:
I have an issue with `node` in VSCode so the automated linting isn't working very well.. trying to resolve now.

src/codeql.test.ts
```diff
@@ -205,6 +206,8 @@ test("downloads and caches explicitly requested bundles that aren't in the toolc
       );
       t.assert(toolcache.find("CodeQL", `0.0.0-${version}`));
       t.is(result.toolsVersion, `0.0.0-${version}`);
+      t.is(result.toolsSource, ToolsSource.Download);
+      t.not(result.toolsDownloadDurationMs, undefined);
```
henrymercer, line 210: Is it possible to mock the `performance` API?  If not, then we could slightly strengthen this to checking something like `typeof result.toolsDownloadDurationMs === "number"`

src/init-action.ts
```diff
@@ -111,25 +127,43 @@ async function sendSuccessStatusReport(
     queries.push(...queriesInput.split(","));
   }
 
-  const statusReport: InitSuccessStatusReport = {
+  const statusReport: InitStatusReport = {
     ...statusReportBase,
     disable_default_queries: disableDefaultQueries,
-    languages,
+    languages: languages || "",
```
henrymercer, line 133: Minor: There's a few places here, for example the `languages` property where before we could use `x` but now we're using `x || ""`.  This means there'll be a few bugs that the typechecker used to catch but won't any more.  For example, beforehand this line made sure that `languages` was not `undefined`, but now with the `||` this will succeed even if `languages` is `undefined`.  Could you explain why we need the `|| ""`?

src/init-action.ts
```diff
@@ -111,25 +127,43 @@ async function sendSuccessStatusReport(
...
+    tools_resolved_version: toolsVersion || "",
     workflow_languages: workflowLanguages || "",
-    trap_cache_languages: Object.keys(config.trapCaches).join(","),
-    trap_cache_download_size_bytes: Math.round(
-      await getTotalCacheSize(config.trapCaches, logger)
-    ),
-    trap_cache_download_duration_ms: Math.round(config.trapCacheDownloadTime),
+    // TODO(angelapwen): I think the following will crash if config.trapCaches is undefined because we might
+    // be sending a failure report before we initialize config.
+    trap_cache_languages: Object.keys(config.trapCaches).join(",") || "",
```
henrymercer, line 143: I think accessing any property on `undefined` is a type error, so we'll probably error quite early on if the config doesn't exist.  It might be simpler to keep a separate paths for when the `config` is defined / undefined.

Another pattern that could be useful is defining a separate `ToolsStatusReport` for the new fields, and having a function produce just that.  You can then union on the `ToolsStatusReport` object right before sending off the status report.

src/init-action.ts
```diff
@@ -111,25 +127,43 @@ async function sendSuccessStatusReport(
...
+    trap_cache_download_size_bytes:
+      Math.round(await getTotalCacheSize(config.trapCaches, logger)) || -1, // Placeholder value if cache size is not determined yet.
+    trap_cache_download_duration_ms:
+      Math.round(config.trapCacheDownloadTime) || -1, // Placeholder value if cache size is not determined yet.,
```
henrymercer, line 147: I think we should probably omit this field if we don't have TRAP caching information.  That still allows us to indicate that we don't have the information, but lessens the risk of making mistakes when querying the downstream data e.g. not excluding `-1`, or casting the data into an unsigned int.

henrymercer (commented):
I like how you've used this additional telemetry information to improve the tests for setting up the CodeQL tools!  Here's a couple of initial comments:

src/init-action.ts
angelapwen: Ahh, I had thought because we're erroring out early, we would want to still send the status report even if the `config` was undefined (and thus stop erroring out before we send the status report).

It's related to your comment below. I'll try out keeping separate paths for when `config` is defined/undefined as that seems to be the main new place that we might send an init status report now with a failure, compared to before.

src/codeql.test.ts
angelapwen: I'm not really clear with testing terminology but I was able to set a `spy` on the `performance.now()` function and make sure it's called 0 or 2 times. I'm not sure if mocking the `performance` API would have worked because theoretetically want `now()` to return  different values each time (even if we're not really checking that the duration field ends up as a positive number).

I also added the suggested `number` assertions to the times that we expect this result.

src/init-action.ts
angelapwen: No longer need to do this as I'm only populating if we have `config`.

src/init-action.ts
angelapwen: I did kind of a mix of both suggestions: we have 3 paths and types of reports for init now: 

1. without config or tools
2. with config and without tools
3. without config and with tools (so we don't error when we can't access some properties)
4. with config and tools

I kept the path for when `config` is defined so we don't error when we can't access some properties, and then unioned on the 2 tool download fields when the tool source was download. I think that should cover all our paths for now.

src/codeql.test.ts
angelapwen: As discussed during 🍐 ing, `performance` is used with trap caching timing so we decided not to mock it.

angelapwen:
I think I've addressed all comments and PR is ready for review again!

src/feature-flags.ts
```diff
@@ -27,9 +27,10 @@ export type CodeQLDefaultVersionInfo =
 
 export interface FeatureEnablement {
   /** Gets the default version of the CodeQL tools. */
-  getDefaultCliVersion(
-    variant: util.GitHubVariant
-  ): Promise<CodeQLDefaultVersionInfo>;
+  getDefaultCliVersion(variant: util.GitHubVariant): Promise<{
+    codeQLDefaultVersionInfo: CodeQLDefaultVersionInfo;
+    toolsFeatureFlagsValid?: boolean;
+  }>;
```
henrymercer, line 33: Optional: We could also embed `toolsFeatureFlagsValid` into `CodeQLDefaultVersionInfo`, since you could consider whether we understood the feature flags or not part of the version info.

src/init-action.ts
```diff
@@ -44,7 +46,20 @@ import {
 } from "./util";
 import { validateWorkflow } from "./workflow";
 
-interface InitSuccessStatusReport extends StatusReportBase {
+// Fields of the init status report that can be sent before `config` is populated.
```
henrymercer, line 49: Nit: Use Javadoc `/** */` to make this comment show as doc when hovering over a reference to `InitStatusReport` in an IDE.

src/init-action.ts
```diff
@@ -73,63 +82,104 @@ interface InitSuccessStatusReport extends StatusReportBase {
   trap_cache_download_duration_ms: number;
 }
 
-async function sendSuccessStatusReport(
+// Fields that we only send to the status report if the tools source was download.
+interface InitToolsDownloadFields {
+  /** Time taken to download the bundle, in milliseconds. */
+  tools_download_duration_ms: number;
+  /** Whether the relevant tools dotcom feature flags have been misconfigured. Only sent if we attempt to download based off dotcom flags. */
```
henrymercer, line 89:
```suggestion
  /** Whether the relevant tools dotcom feature flags have been misconfigured. Only populated if we attempt to determine the default version based on the dotcom feature flags. */
```

src/init-action.ts
```diff
@@ -73,63 +82,104 @@ interface InitSuccessStatusReport extends StatusReportBase {
...
+  tools_feature_flags_valid: boolean;
+}
+
+async function sendInitStatusReport(
+  actionStatus: ActionStatus,
   startedAt: Date,
   config: configUtils.Config,
```
henrymercer, line 96: Adjust this type to indicate that `config` might be undefined.

src/init-action.ts
henrymercer: 👍 I just had a look at this and I think this is a nice solution.

src/feature-flags.ts
angelapwen: Not sure why I didn't think of that 😄 done!

src/init-action.ts
angelapwen: Nice catch, thank you!

src/init-action.ts
```diff
@@ -73,63 +82,104 @@ interface InitSuccessStatusReport extends StatusReportBase {
...
-      await getTotalCacheSize(config.trapCaches, logger)
-    ),
-    trap_cache_download_duration_ms: Math.round(config.trapCacheDownloadTime),
   };
 
-  await sendStatusReport(statusReport);
+  let initToolsDownloadFields: InitToolsDownloadFields = {
+    tools_download_duration_ms: -1, // Placeholder value in case field is undefined.
+    tools_feature_flags_valid: false, // Report invalid in case field is undefined.
+  };
```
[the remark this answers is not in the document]
angelapwen: Ah, I added these lines because I thought I had to to be of type `InitToolsDownloadFields`, but I worked around it now (it can be of inferred type here, and we can only add the fields/assign type later).

src/init-action.ts
angelapwen: Done!

src/feature-flags.ts
```diff
@@ -17,11 +17,13 @@ const MINIMUM_ENABLED_CODEQL_VERSION = "2.11.6";
 export type CodeQLDefaultVersionInfo =
   | {
       cliVersion: string;
+      toolsFeatureFlagsValid?: boolean;
```
henrymercer, line 20: We know we're on Dotcom here, so `toolsFeatureFlagsValid` should always be present

```suggestion
      toolsFeatureFlagsValid: boolean;
```

src/feature-flags.ts
```diff
@@ -17,11 +17,13 @@ const MINIMUM_ENABLED_CODEQL_VERSION = "2.11.6";
...
       variant: util.GitHubVariant.DOTCOM;
     }
   | {
       cliVersion: string;
       tagName: string;
+      toolsFeatureFlagsValid?: boolean;
```
henrymercer, line 26: Similarly, we know we're on GHAE / GHES here, so `toolsFeatureFlagsValid` will never be present
```suggestion
```

henrymercer (approved):
Great stuff!  A couple of suggestions we can address as followup, but I'll merge this now to get this into the release that I'm doing as part of the CodeQL release process.

src/feature-flags.ts
angelapwen: Done 👍

src/feature-flags.ts
angelapwen: This ended up being somewhat difficult to implement without adding some ugly casts, so I've left it optional.

henrymercer (approved):
🎉

PR #1505, opened 2023-01-23
More cleanup as a result of dropping Node 12

henrymercer:
Clean up a few things now that we no longer need to support Node 12.


aeisenberg:
> `Promise.race(x)` will wait for either (a) one of the promises in x to resolve, or (b) all the promises in x to reject.

Do you have this reversed? That sounds like the behaviour of `Promise.any`.

[Promise.any](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any):

> This returned promise fulfills when any of the input's promises fulfills, with this first fulfillment value. It rejects when all of the input's promises reject (including when an empty iterable is passed)

[Promise.race](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race):

> This returned promise settles with the eventual state of the first promise that settles.

henrymercer:
You're completely right 🤦 Thanks for the catch, I'll drop that commit.

aeisenberg:
It's subtle. I had to read the docs a few times just to make sure.

aeisenberg (approved):

PR #1506, opened 2023-01-23
Fail `prepare-test` early when `gh release list` fails

henrymercer:
We've been seeing some tests fail recently due to the `gh release list` command failing.  Because Bash doesn't fail on error by default, those errors weren't manifesting themselves until a later step.  This PR adds `set -e` so the `prepare-test` Action fails straight away.  This makes it slightly quicker to triage the error and saves a bit of runner time. 


angelapwen (approved):
Thanks for the drive by fixes ✨

PR #1507, opened 2023-01-23
Limit Swift autobuild runtime in PR check to 10 minutes

henrymercer:
There's a known issue that causes the Swift autobuilder to hang.  By setting a timeout, we'll fail earlier and we can rerun the check earlier.


adityasharad:
https://github.com/github/codeql/pull/11900 should fix the timeout but this is useful while we're still testing older versions.

adityasharad (approved):

PR #1508, opened 2023-01-24
Fall back to the `defaults.json` CLI version if feature flags misconfigured

henrymercer:
We expect at least one default CLI version to be enabled on Dotcom at any time. However if the feature flags are misconfigured, rather than crashing, we fall back to the CLI version shipped with the Action in defaults.json. This has the effect of immediately rolling out new CLI versions to all users running the latest Action.

A drawback of this approach relates to the small number of users that run old versions of the Action on Dotcom. As a result of this approach, if we misconfigure the feature flags then these users will experience some alert churn. This is because the CLI version in the defaults.json shipped with an old version of the Action is likely older than the CLI version that would have been specified by the feature flags before they were misconfigured.


angelapwen (approved):
🙇

PR #1510, opened 2023-01-26
Fix computation of fallback version number

henrymercer:
This PR addresses a bug where we would compute a fallback version number like `0.0.0-codeql-bundle-20230105` rather than one like `0.0.0-20230105`.  This happened because we used the tag name `0.0.0-codeql-bundle-20230105` rather than the bundle version `20230105` to compute the fallback version number.

This fallback version number is crucial at the moment, because the controlled switchover changes haven't yet rolled out to the GitHub-hosted runner images, and therefore at the moment, the GitHub-hosted runner images version the CodeQL tools as `0.0.0-20230105` rather than `2.0.0-20230105` (or `2.0.0`).  So if we get the fallback version number wrong, we end up having to download the bundle from the GitHub Release in most workflows.

The fix is pretty simple — I've split out the tag -> bundle version logic from `getBundleVersionFromUrl` into a new function `getBundleVersionFromTagName` and called this in `getCodeQLSource` when we compute the fallback version.  Another approach would have been to call ```getBundleVersionFromUrl(`/${tagName}/`)``` but that's a little hacky.

I think part of the reason why this bug happened is that some of our unit tests do too much stuff.  So I've rewritten the ones that test lookup to mock the toolcache, rather than using our own logic to install things into the toolcache.  This allows us to test fallback versions like `0.0.0-20230105` more easily and readably by separating out concerns.  I've removed one of the cases where the bundle is not cached because this is already tested in the "downloads and caches explicitly requested bundles that aren't in the toolcache" test.

- [Example run before](https://github.com/dsp-testing/henrymercer-test/actions/runs/4014436795/jobs/6894988836) — note the following logs:
    ```
  ##[debug]Computed a fallback toolcache version number of 0.0.0-codeql-bundle-20230105 for CodeQL tools version 2.12.0.
  ##[debug]isExplicit: 0.0.0-codeql-bundle-20230105
  ##[debug]explicit? true
  ##[debug]checking cache: /opt/hostedtoolcache/CodeQL/0.0.0-codeql-bundle-20230105/x64
  ##[debug]not found
  ##[debug]Did not find CodeQL tools version 2.12.0 in the toolcache.
   ```
- [Example run after](https://github.com/dsp-testing/henrymercer-test/actions/runs/4014877912/jobs/6895932765) — note that the bundle is now found in the toolcache:
    ````
  ##[debug]Computed a fallback toolcache version number of 0.0.0-20230105 for CodeQL tools version 2.12.0.
  ##[debug]isExplicit: 0.0.0-20230105
  ##[debug]explicit? true
  ##[debug]checking cache: /opt/hostedtoolcache/CodeQL/0.0.0-20230105/x64
  ##[debug]Found tool in cache CodeQL 0.0.0-20230105 x64
  ##[debug]CodeQL found in cache /opt/hostedtoolcache/CodeQL/0.0.0-20230105/x64
    ````


nickfyson (approved):
LGTM

Bug fix itself makes sense, though I have lower confidence in my assessment of the test improvements (all rather unfamiliar) 😅

PR #1516, opened 2023-02-03
Supply authorization parameter to toolcache.downloadTool()

aibaars:
Previously we supplied the authorization information via the 'headers' parameter and left the `auth` parameter `undefined`. This works fine, except in some cases when the request is retried.

We observed difference in behaviour, when using `undefined` for the `auth` parameter we got a 504 error on the first attempt and a 401 error on the second attempt. When the `auth` parameter was set we observed a sequences of 504 errors until the number of retries was exceeded. I'm not really sure where the `Authorization` header got lost, but I suspect the HTTP client drops it when following redirects and the retry-helper doesn't restore it before trying again.

In case someone else has similar problems: the underlying cause of the 504 errors were due to proxy settings. When downloading a release artifact from a GitHub Enterprise server (say `github.company.com`) you get redirected to `media.github.company.com`. The `no_proxy` variable did not contain the `media` subdomain which caused a gateway timeout. Explicitly adding the subdomain to the `no_proxy` variable solved the issue. Alternatively the following pull request https://github.com/actions/toolkit/pull/1223 would have solved the issue too, because it updates the no_proxy handling to automatically include subdomains.


angelapwen (approved):
Thanks, read the associated context and this makes sense. Re-ran the autobuild test that is known to time out.

PR #1519, opened 2023-02-03
Support `security-experimental` as a well-known suite

angelapwen:
This change adds `security-experimental` as a well-known query suite that can be specified under the `queries` input. Note that this change is intended for internal/field users and will not be publicized in our changelog or documentation.

The suite will be supported for CLI versions 2.11.1+, and an error will be thrown if the suite was specified but the CLI version is not up to date.

As `security-experimental` includes the ML-powered queries, we test via the existing ML-powered query tests. 


angelapwen:
Skipped/failing test is unrelated (observed on other SHAs as well) so marking ready for review.

src/config-utils.ts
```diff
@@ -401,6 +406,17 @@ async function addBuiltinSuiteQueries(
...
   }
+  if (
+    suiteName === "security-experimental" &&
+    !(await codeQlVersionAbove(
+      codeQL,
+      CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE
+    ))
+  ) {
+    throw new Error(
+      `'security-experimental' suite is not supported on CodeQL versions less than ${CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE}.`
```
henrymercer, line 417:
```suggestion
      `'security-experimental' suite is not supported on CodeQL versions less than ${CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE}.`
```
```suggestion
      `The 'security-experimental' suite is not supported on CodeQL CLI versions earlier than ${CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE}. Please upgrade to CodeQL CLI version ${CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE} or later.`
```

src/codeql.ts
```diff
@@ -278,6 +278,11 @@ export const CODEQL_VERSION_ML_POWERED_QUERIES_WINDOWS = "2.9.0";
  */
 export const CODEQL_VERSION_BETTER_RESOLVE_LANGUAGES = "2.10.3";
 
+/**
+ * Versions 2.11.1+ of the CodeQL CLI introduces `security-experimental` query suites for all languages.
```
henrymercer, line 282:
```suggestion
 * Versions 2.11.1+ of the CodeQL Bundle include a `security-experimental` built-in query suite for each language.
```

src/config-utils.test.ts
```diff
@@ -1993,7 +1993,7 @@ test(
   process.platform === "win32" ? undefined : "~0.1.0"
 );
 // Test that ML-powered queries aren't run when the user hasn't specified that we should run the
-// `security-extended` or `security-and-quality` query suite.
+// `security-experimental`, `security-extended`, or `security-and-quality` query suite.
```
henrymercer, line 1996: As a nit, consider listing `security-experimental` last, since it's not going to be advertised to external users in docs etc, and therefore should have the least usage.

henrymercer (commented):
Looks good!  Another thing to consider is adding the `security-experimental` suite to `.github/codeql/codeql-config.yml`.  This would give us a little more assurance that this suite will continue working.

src/config-utils.ts
angelapwen: Done!

src/codeql.ts
angelapwen: Done!

angelapwen:
Have added the suite into `.github/codeql/codeql-config.yml` — thank you, I had meant to do that but it slipped my mind!

src/config-utils.ts
```diff
@@ -401,6 +406,17 @@ async function addBuiltinSuiteQueries(
...
   }
+  if (
+    suiteName === "security-experimental" &&
+    !(await codeQlVersionAbove(
+      codeQL,
+      CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE
+    ))
+  ) {
+    throw new Error(
+      `The 'security-experimental' suite is not supported on CodeQL CLI versions earlier than ${CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE}. Please upgrade to CodeQL CLI version ${CODEQL_VERSION_SECURITY_EXPERIMENTAL_SUITE} or later.`
```
henrymercer, line 417: Nit: break this line to 120 characters max (outside of this PR, we should probably update the linter to enforce this!)

henrymercer (approved):
Looks good, though I realise now that we run the CodeQL check using both the latest and the cached tools, so we'll probably have to wait until 2.12.2 is released before we can update `.github/codeql/codeql-config.yml`.

src/config-utils.ts
angelapwen: Sounds good!

angelapwen:
Ah yeah that makes sense. I've made the changes and will make a note for the two follow-up changes (adding the suite to `codeql-config.yml` after the 2.12.2 release, and improving the linter)

henrymercer (approved):

PR #1524, opened 2023-02-06
Add max line length of 120 to linter

angelapwen:
This change adds the `max-len` property to the linter, which will enforce a maximum of 120 characters per line. It also ignores URLs, strings, and template literals in case these cases exceed 120 characters. 

We also have the option to ignore all comments, or just trailing comments (https://eslint.org/docs/latest/rules/max-len). I have left those out for now but am open to adding them! The PR also fixes some comments that have exceeded the 120 character limit. 


aeisenberg (approved):
Nice. My understanding is that the linting rules here are copied from gh/gh. I think it's generally ok to become stricter than gh/gh, but not more lenient. So, this PR LGTM.

henrymercer (approved):
LGTM.  Perhaps we can look to remove the `ignore*` settings as we go forwards, but we should prioritize this alongside re-enabling the rules we disabled while transitioning to eslint at the bottom of the file.

angelapwen:
Had to rebase `main` in so requesting re-review!

src/init-action.ts
```diff
@@ -87,7 +87,8 @@ interface InitWithConfigStatusReport extends InitStatusReport {
 interface InitToolsDownloadFields {
   /** Time taken to download the bundle, in milliseconds. */
   tools_download_duration_ms?: number;
-  /** Whether the relevant tools dotcom feature flags have been misconfigured. Only populated if we attempt to determine the default version based on the dotcom feature flags. */
+  /** Whether the relevant tools dotcom feature flags have been misconfigured.
+   * Only populated if we attempt to determine the default version based on the dotcom feature flags. */
```
henrymercer, line 91: For next time (I don't think this is worth a recompiling and rerunning the checks for)

```suggestion
  /**
   * Whether the relevant tools dotcom feature flags have been misconfigured.
   * Only populated if we attempt to determine the default version based on the dotcom feature flags. */
```

henrymercer (approved):

src/init-action.ts
angelapwen: Made the change in this PR: https://github.com/github/codeql-action/pull/1532

PR #1532, opened 2023-02-09
Tools telemetry: accurately report when feature flags were inaccessible 

angelapwen:
Previously, we sent a `false` value of `tools_feature_flags_valid` in our status reports when we were actually unable to access our feature flags, for example in cases where the repository does not have the appropriate security permissions to access feature flags. This change should allow us to catch the error where we are unable to access feature flags, and not send the `tools_feature_flags_valid` field at all to the status report.

This will enable us to create a monitor on the `tools_feature_flags_valid` telemetry field, as we will always expect it to be 0. 


src/feature-flags.ts
```diff
@@ -153,13 +153,17 @@ export class Features implements FeatureEnablement {
 class GitHubFeatureFlags implements FeatureEnablement {
   private cachedApiResponse: GitHubFeatureFlagsApiResponse | undefined;
 
+  // We cache whether the feature flags were accessed or not in order to accurately report whether flags were
+  // incorrectly configured vs. inaccessible in our telemetry.
+  private areRemoteFeatureFlagsAccessible: boolean;
```
henrymercer, line 158: Minor: There are two cases where we didn't get the feature flags: one, a transient failure happened, and we failed the run to avoid alert churn; two, our privileges are restricted e.g. from being on a Dependabot PR.  I feel "accessible" is a bit ambiguous between these two cases.  Maybe `hasAccessedRemoteFeatureFlags` or something similar would be clearer?

src/feature-flags.ts
```diff
@@ -325,13 +336,19 @@ class GitHubFeatureFlags implements FeatureEnablement {
     }
   }
 
-  private async loadApiResponse(): Promise<GitHubFeatureFlagsApiResponse> {
+  private async loadApiResponse(): Promise<{
```
henrymercer, line 339: Minor: I'd consider having this write directly to `hasAccessedRemoteFeatureFlags` to simplify things a tad.

henrymercer (commented):
Looks good, just a couple of minor comments.

src/feature-flags.ts
angelapwen: Aren't we throwing the `Encountered an error while trying to determine feature enablement` error in the former case? I'm only setting the flag to `false` if the privileges are restricted 🤔 

I don't mind renaming to the suggested name though, it works regardless!

src/feature-flags.ts
angelapwen: Makes sense — done!

henrymercer (approved):

PR #1533, opened 2023-02-10
Remove TRAP caching feature flag

cklin:
This PR makes the CodeQL action assume that the `trap_caching_enabled` is true without actually checking it.  The feature flag has been set to true globally for months, so this change should have no practical effect.


angelapwen (approved):
Looks good to me, reran the PR check timeout.

henrymercer (approved):
Typically, removing the feature flag check in the Action would enable the feature on GHES, since features are off by default in GHES.  However since `getTrapCachingEnabled()` returns false for self-hosted runners, TRAP caching is effectively disabled for GHES.  So I agree this PR doesn't change behaviour in practice.

cklin:
> Typically, removing the feature flag check in the Action would enable the feature on GHES, since features are off by default in GHES.

Thanks for pointing that out!  For future reference, what do we usually do for GHES in feature flag removals?  If we want the feature to remain disabled in GHES, we could use `util.getGitHubVersion()` and specifically exclude GHES.  What about features that we do want to enable on GHES?

henrymercer:
Most of the time, we want to enable the feature on GHES, which we do by dropping the feature flag check, and making sure there aren't any code paths to specifically disable GHES, like you mention.

PR #1538, opened 2023-02-13
Optionally send tools download telemetry

angelapwen:
This PR sends the two optional telemetry fields `tools_download_duration_ms` and `tools_feature_flags_valid` conditionally: if either field is undefined, the field is not sent.

This should not change anything in our telemetry because the fields should have been ignored regardless, but we add them conditionally just to be sure.


src/init-action.ts
```diff
@@ -118,13 +118,17 @@ async function sendInitStatusReport(
...
-      tools_feature_flags_valid: toolsFeatureFlagsValid,
-    };
+    if (toolsDownloadDurationMs !== undefined) {
+      initToolsDownloadFields.tools_download_duration_ms =
+        toolsDownloadDurationMs;
+    }
+    if (toolsFeatureFlagsValid !== undefined) {
+      initToolsDownloadFields.tools_feature_flags_valid =
+        toolsFeatureFlagsValid;
+    }
```
henrymercer, line 131: Minor: You could probably delete the surrounding `if (toolsSource === ToolsSource.Download)`, as I assum these fields would only be defined when we want to send them.

henrymercer (approved):

src/init-action.ts
angelapwen: Yeah, you're right 👍 that would be cleaner

henrymercer (approved):

PR #1539, opened 2023-02-13
Avoid unref-ing timer while awaiting status upload

aeisenberg:
We had a problem where `waitForProcessing` was not completing before the node process ends. This is because using `unref` would allow the node process to end without having the `delay` function complete.


aeisenberg:
This was not a user-facing bug, so no change note required.

src/util.ts
```diff
@@ -548,10 +548,19 @@ export async function bundleDb(
...
 
-export async function delay(milliseconds: number) {
-  // Immediately `unref` the timer such that it only prevents the process from exiting if the
-  // surrounding promise is being awaited.
-  return new Promise((resolve) => setTimeout(resolve, milliseconds).unref());
+/**
+ * @param milliseconds time to delay
+ * @param opts.unref if true, the timer will not prevent the process from exiting
+ */
+export async function delay(milliseconds: number, opts: { unref: boolean }) {
```
henrymercer, line 555: Idea you can take or leave: it might make our code slightly more readable to rename `unref` to something like `allowProcessExitDuringDelay`.

henrymercer (approved):

src/util.ts
```diff
@@ -548,10 +548,22 @@ export async function bundleDb(
   return databaseBundlePath;
 }
 
-export async function delay(milliseconds: number) {
-  // Immediately `unref` the timer such that it only prevents the process from exiting if the
-  // surrounding promise is being awaited.
-  return new Promise((resolve) => setTimeout(resolve, milliseconds).unref());
+/**
+ * @param milliseconds time to delay
+ * @param unref if true, the timer will not prevent the process from exiting
```
henrymercer, line 553: Nit: update JSDoc.  The spec recommends writing down two `@param`s for a destructured parameter, which seems a little verbose (https://jsdoc.app/tags-param.html#parameters-with-properties).

henrymercer (approved):

src/util.ts
aeisenberg: Hmmm...that's a little verbose, but I'll go with the spec.

henrymercer (approved):

PR #1540, opened 2023-02-14
Set --expect-discarded-cache option

cklin:
This PR changes the action to pass `--expect-discarded-cache` to the last `codeql database run-queries` invocation to improve performance.


src/analyze.ts
```diff
@@ -369,6 +383,16 @@ export async function runQueries(
     return await codeql.databasePrintBaseline(databasePath);
   }
 
+  /**
+   * Update queryFlags for the last `codeql database run-queries` invocation.
+   * This function should be called prior to the very last `run-queries` call.
+   */
+  async function configureForLastQueryRun(): Promise<void> {
+    if (await util.supportExpectDiscardedCache(codeql)) {
+      queryFlags.push("--expect-discarded-cache");
```
aeisenberg, line 392: I find this a little confusing that this is changing a variable that is declared outside of the scope of this function. Also, it's not clear from this function (other than by the name) that it can only be called once, otherwise it will create invalid arguments.

Could you try a different approach?

Maybe add a new parameter to `runQueryGroup` called something like `isLastRun`. And then `runQueryGroup` will pass `--expect-discarded-cache` in this case only if the CLI version is new enough to accept it. I feel like this is a bit more of an explicit way to handle it.

src/analyze.ts
```diff
@@ -267,6 +277,13 @@ export async function runQueries(
           );
         }
 
+        const customQueryIndices: number[] = [];
+        for (let i = 0; i < queries["custom"].length; ++i) {
+          if (queries["custom"][i].queries.length > 0) {
```
aeisenberg, line 282: Minor:

```suggestion
        for (let i = 0; i < queries.custom.length; ++i) {
          if (queries.custom[i].queries.length > 0) {
```

src/analyze.ts
```diff
@@ -276,37 +293,42 @@ export async function runQueries(
               language,
               "builtin",
               createQuerySuiteContents(queries["builtin"], queryFilters),
```
aeisenberg, line 295: Minor:

```suggestion
              createQuerySuiteContents(queries.builtin, queryFilters),
```

src/analyze.ts
```diff
@@ -222,6 +223,9 @@ export async function runQueries(
     );
     const packsWithVersion = config.packs[language] || [];
 
+    const isLastLanguage =
+      language === config.languages[config.languages.length - 1];
```
aeisenberg, line 227: Not sure how `--expect-discarded-cache` actually works, but does it affect all languages in a cluster at the same time? Ie- if you have a clustered db with java and javascript, will running with `--expect-discarded-cache` on java queries affect the cache for javascript?

aeisenberg (commented):
A question inline. 

Also, can you think of any way of testing this? Probably a unit test or two would be fine. We are testing the `runQueries` command in this test: https://github.com/aeisenberg/codeql-action/blob/main/src/analyze.test.ts#L25 This test could be expanded, or a new test that focuses explicitly on this flag could be written.

We can pair on this if you like.

src/analyze.ts
cklin: Based on discussions in the tracking issue, I think we should set `--expect-discarded-cache` for the last `run-queries` command of each language (as opposed to for all languages).

cklin:
I made the suggested changes and added unit tests.  PTAL.  Thank you!

src/analyze.test.ts
```diff
@@ -229,6 +232,175 @@ test("status report fields and search path setting", async (t) => {
...
+
+    await runQueriesWithConfig(config, [Feature.CliConfigFileEnabled]);
+    t.deepEqual(
+      getDatabaseRunQueriesCalls(codeql).map((c) => c.args[4]),
+      [true]
+    );
+  });
+});
+
+test("optimizeForLastQueryRun for two languages, CliConfigFileEnabled", async (t) => {
```
aeisenberg, line 389: Going forward, `Feature.CliConfigFileEnabled` should be enabled for all users (except some specific DCA repos).

aeisenberg (approved):
Nice work.

PR #1556, opened 2023-03-03
Upload per-database diagnostic SARIFs on green and red runs

angelapwen:
On failed runs we now call `codeql database diagnostic-export` to create and upload a SARIF file for per-database diagnostics. If this file was not created, we fall back on the original behavior of uploading the result of `codeql export diagnostics`. 

On successful runs, we add the `--sarif-include-diagnostics` option to the `codeql database interpret-results` command. 

Both the addition of the option to the `interpret-results` call, and the `codeql database diagnostic-export` call, are gated behind the `ExportDiagnosticsEnabled` feature flag. 

We test the green run scenario with an integration test that adds a diagnostic prior to the `analyze` step, with the feature flag environment variable override to enable the feature. We check the SARIF file generated to make sure it includes the manually added diagnostic. 

[GitHub internal only]: we test the red run scenario by writing a similar workflow as the one in the integration test, but failing the workflow after adding the diagnostic. See [run](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/actions/runs/4452610920/jobs/7820394828) and [corresponding tool status page](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/security/code-scanning/tool-status/CodeQL/configurations/actions-FZTWS5DIOVRC653POJVWM3DPO5ZS6Y3PMRSXC3BNMFXGC3DZONUXGLTZNVWA/2cd0eaa2455a0fd762610e5d9f96e1ffb3c81f02274d16db17f95b37b00d265a) for the surfaced diagnostic. 


src/init-action-post-helper.ts
```diff
@@ -65,36 +73,70 @@ async function maybeUploadFailedSarif(
...
+        databasePath,
+        maybeIsDatabaseCluster === undefined ||
+          maybeIsDatabaseCluster.length === 0
+          ? false
+          : true,
+        databaseSarifFile,
+        category
+      );
+      core.info(`Uploading failed SARIF file ${databaseSarifFile}`);
+      const uploadResult = await uploadLib.uploadFromActions(
```
angelapwen, line 116: I just saw that `uploadFromActions` can take a directory, so I can refactor this to only upload once once the per-database SARIFs are generated.

henrymercer (commented):
In the green runs path, we call `codeql database interpret-results` once per database to produce a SARIF file per DB, combine all these SARIF files together, and then submit a single SARIF file.  This has different behaviour to uploading each SARIF file separately, since code scanning will overwrite the existing data for that category each time you upload a new SARIF file.

I think for red runs we can do something simpler — `config.dbLocation` should always be a DB cluster (at least for all CodeQL versions that support diagnostics).  Therefore we should be able to run `codeql database export-diagnostics` once on that DB cluster and submit the resulting SARIF file.

angelapwen:
Unrelated CI failures that I'll fix in a separate PR!

src/init-action-post-helper.ts
```diff
@@ -69,9 +71,21 @@ async function maybeUploadFailedSarif(
   }
   const category = getCategoryInputOrThrow(workflow, jobName, matrix);
   const checkoutPath = getCheckoutPathInputOrThrow(workflow, jobName, matrix);
+  const databasePath = config.dbLocation;
 
+  // We call 'database export-diagnostics' to find any per-database diagnostics.
   const sarifFile = "../codeql-failed-run.sarif";
-  await codeql.diagnosticsExport(sarifFile, category);
+  await codeql.databaseExportDiagnostics(
+    databasePath,
```
henrymercer, line 79: Can this be undefined or null at this point?

src/init-action-post-helper.ts
```diff
@@ -69,9 +71,21 @@ async function maybeUploadFailedSarif(
...
+    true, // Database is always a cluster for CodeQL versions that support diagnostics.
+    sarifFile,
+    category
+  );
+
+  // If there was no SARIF file produced, then we fall back on 'export diagnostics'.
+  if (!fs.existsSync(sarifFile)) {
```
henrymercer, line 86: When would this happen?  It seems like if we don't get a SARIF file then something badly went wrong and `databaseExportDiagnostics` probably threw an exception.

src/codeql.ts
```diff
@@ -970,6 +983,28 @@ export async function getCodeQLForCmd(
       ];
       await new toolrunner.ToolRunner(cmd, args).exec();
     },
+    async databaseExportDiagnostics(
+      databasePath: string,
+      isCluster: boolean,
+      sarifFile: string,
+      automationDetailsId: string | undefined
+    ): Promise<void> {
+      const args = [
```
henrymercer, line 992: To export the diagnostic information, we'll need to pass `--sarif-add-diagnostics`.  We should do this behind a feature flag so that if CodeQL's diagnostics are problematic in any way we can easily turn them off.

henrymercer (commented):
Looking good so far.  An integration test might be tricky to write due to most of the work happening in `post-init`.  We could work around that by creating a separate Action for the test (like we did for the query filters tests), but I'd be happy to merge this with detailed unit tests instead.

src/init-action-post-helper.ts
angelapwen: Hm, I had thought that this would catch the case where no database has been created yet (in which case we should call `codeql export diagnostics`.

src/init-action-post-helper.ts
angelapwen: I traced through where `config.dbLocation` is set, and it's set during the `initConfig` call in `init-action`(https://github.com/github/codeql-action/blob/f13b180fb88fab31693634cb19e73cb3ed1cb7d8/src/init-action.ts#L252). By the time this function returns, it returns either the user input database location, or the default location. 

So it seems pretty unlikely that we crash out of the init action before this point, but perhaps to be safe we should check if it's undefined and run `diagnosticExport`?

henrymercer: > So it seems pretty unlikely that we crash out of the init action before this point, but perhaps to be safe we should check if it's undefined and run `diagnosticExport`?

That would be my preference 👍 If a particular step timeouts, or the run is cancelled, we can exit the code at unexpected points.

src/codeql.ts
angelapwen: Added behind a feature flag 👍

src/init-action-post-helper.ts
angelapwen: Done!

angelapwen:
I've addressed the existing comments other than testing. I need to update some of the existing unit tests to reflect the new behavior, as well as write new ones to confirm that the appropriate export command is called as expected ✍️

src/init-action-post-helper.ts
```diff
@@ -64,9 +66,27 @@ async function maybeUploadFailedSarif(
...
+      databasePath,
+      sarifFile,
+      category,
+      features
+    );
+
+    // If there was no SARIF file produced, then we fall back on 'export diagnostics'.
+    if (!fs.existsSync(sarifFile)) {
+      await codeql.diagnosticsExport(sarifFile, category, config, features);
+    }
```
henrymercer, line 88: I like the idea of reducing risk here if there's a problem with `database export-diagnostics`.  Since an issue could cause that command to crash completely, I'd recommend we instead gate the call to `databaseExportDiagnostics` behind the export diagnostics feature flag.

angelapwen: Oh, I see, so rather than only append the `--sarif-include-diagnostics` option behind the export diagnostics feature flag, we'll only call the `database export-diagnostics` (always with the `--sarif-include-diagnostics` option) if the flag is set. That does sound safer. 👍

angelapwen:
Hrm... I am testing the integration test and getting

```
A fatal error occurred: Could not save the diagnostic to disk.
(eventual cause: FileNotFoundException "/home/runner/work/_temp/javascript/diagnostic/cli-diagnostics-add-20[23](https://github.com/github/codeql-action/actions/runs/4442028134/jobs/7797833947?pr=1556#step:6:24)0316T22065...")
/home/runner/work/_temp/8ab836c8-289d-4b7c-8200-8c2ebd265dae.sh: line 1: : command not found
```

I thought it might be a permissions issue so tried to create the `/home/runner/work/_temp/javascript/diagnostic/` directory beforehand, but now not seeing any relevant error from the CLI.

pr-checks/checks/diagnostics-export.yml
```diff
@@ -0,0 +1,43 @@
...
+      output: "${{ runner.temp }}/results"
+      upload-database: false
+  - name: Run `diagnostics add`
+    shell: bash
+    env:
+      CODEQL_PATH: ${{ steps.init.outputs.codeql-path }}
+    run: |
+      sudo mkdir -p ${{ runner.temp }}/javascript/diagnostic
+      sudo chmod -R 777 ${{ runner.temp }}/javascript/diagnostic
+      "$("$CODEQL_PATH" diagnostics add --diagnostic-dir=${{ runner.temp }}/javascript/diagnostic --plaintext-message="Plaintext message" --source-id="lang/diagnostics/example" --source-name="Diagnostic name")"
```
henrymercer, line 24: You might want to use the `dbLocation` input to place the JS database in a specific place.  I think by default they go in `${{ runner.temp }}/codeql_databases`.  You could also use `codeql database add-diagnostic` to avoid having to specify the diagnostic directory here.

henrymercer: I think we may also want to remove the `$(` wrapper, so we don't try to execute the result of this shell command as another command.

src/codeql.ts
tied2token: 1🫣

tied2token (approved):

karbonnanroid (approved):

karbonnanroid (approved):

src/codeql.ts
```diff
@@ -982,6 +997,27 @@ export async function getCodeQLForCmd(
...
+        "database",
+        "export-diagnostics",
+        `${databasePath}`,
+        "--db-cluster", // Database is always a cluster for CodeQL versions that support diagnostics.
+        "--format=sarif-latest",
+        `--output=${sarifFile}`,
+        "--sarif-include-diagnostics", // ExportDiagnosticsEnabled is always true if this command is run.
```
henrymercer, line 1012: Add this to the JSDoc for `databaseExportDiagnostics` to increase visibility.

henrymercer (approved):
Nice!

pr-checks/checks/diagnostics-export.yml
```diff
@@ -0,0 +1,52 @@
+name: "Diagnostic export"
+description: "Tests that a manually added diagnostic is exported to SARIF."
+versions: ["latest"]
```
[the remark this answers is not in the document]
angelapwen: Done

pr-checks/checks/diagnostics-export.yml
```diff
@@ -0,0 +1,52 @@
...
+        const run = sarif.runs[0];
+
+        const toolExecutionNotifications = run.invocations[0].toolExecutionNotifications;
+        const diagnosticToolExecutionNotification = toolExecutionNotifications[toolExecutionNotifications.length - 1];
+        if (diagnosticToolExecutionNotification.descriptor.id !== 'lang/diagnostics/example' || diagnosticToolExecutionNotification.message.text !== 'Plaintext message') {
+          core.setFailed('`toolExecutionNotification` related to this diagnostic was not found in SARIF');
+        }
+
+        const notifications = run.tool.driver.notifications;
+        const diagnosticNotification = notifications[notifications.length - 1];
```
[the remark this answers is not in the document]
angelapwen: Done

src/init-action-post-helper.test.ts
```diff
@@ -288,15 +358,25 @@ async function testFailedSarifUpload(
...
-      diagnosticsExportStub.calledOnceWith(
-        sinon.match.string,
-        category,
-        sinon.match.any,
-        sinon.match.any
-      ),
-      `Actual args were: ${diagnosticsExportStub.args}`
-    );
+    if (databaseExists && exportDiagnosticsEnabled) {
+      t.true(
```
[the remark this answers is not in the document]
angelapwen: Yep! Thanks for catching ✨

src/codeql.ts
angelapwen: Done!

pr-checks/checks/diagnostics-export.yml
```diff
@@ -0,0 +1,52 @@
...
+        const fs = require('fs');
+
+        const sarif = JSON.parse(fs.readFileSync(process.env['SARIF_PATH'], 'utf8'));
+        const run = sarif.runs[0];
+
+        const toolExecutionNotifications = run.invocations[0].toolExecutionNotifications;
+        const diagnosticToolExecutionNotification = toolExecutionNotifications[toolExecutionNotifications.length - 1];
+        if (diagnosticToolExecutionNotification.descriptor.id !== 'lang/diagnostics/example' || diagnosticToolExecutionNotification.message.text !== 'Plaintext message') {
+          core.setFailed('`toolExecutionNotification` related to this diagnostic was not found in SARIF');
+        }
```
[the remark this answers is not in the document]
henrymercer: Filtering might be slightly neater.  You could then also check we have exactly one diagnostic with that ID.

henrymercer (approved):
I'd recommend using some functional list methods to simplify the tests, otherwise LGTM

PR #1560, opened 2023-03-03
Bump `setup-swift` and re-configure dependabot

angelapwen:
Our Swift extractor is now using Swift v5.7.3, but our `setup-swift` version was pinned to a prior version that doesn't support v5.7.3, so our CI is failing.

This change bumps `setup-swift` to its most recent release commit SHA (https://github.com/swift-actions/setup-swift/releases/tag/v1.22.0).

It also adds a new entry to the `dependabot.yml` file because Dependabot should have bumped this version — `setup-swift` v1.22.0 was released two weeks ago and we run Dependabot checks weekly. Per the dependabot maintainers, the [filepath needed is the root of the repo](https://github.com/dependabot/dependabot-core/issues/5137#issuecomment-1441799333) (`/.github/workflows` is specially cased under `/`). 

We also add a comment reminding the next person who approves a Dependabot update to manually update the SHA in the CLI's CodeQL Action Integration Test. (Note that we should also replicate the logic of finding the appropriate Swift version in the CLI's integration test).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,7 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+  - package-ecosystem: github-actions
+    directory: "/.github/setup-swift/"
+    schedule:
+      interval: weekly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,6 @@ updates:
     schedule:
       interval: weekly
   - package-ecosystem: github-actions
-    directory: "/.github/setup-swift/"
+    directory: "/.github/setup-swift/" # All subdirectories outside of "/.github/workflows" must be explicitly included.
     schedule:
       interval: weekly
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -15,3 +15,7 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+  - package-ecosystem: github-actions
```
adityasharad, line 18: Perhaps comment here why we need this separately from the `/` pattern above. I understand `/` to mean "repo root and .github/workflows` and any other subdir needs explicit inclusion?

adityasharad (dismissed):

.github/dependabot.yml
angelapwen: Yes, that's correct 👍 will do

henrymercer (approved):

PR #1578, opened 2023-03-14
Fix a circular dependency that led a minimum version to be `undefined`

henrymercer:
A circular dependency between `codeql.ts` and `feature-flags.ts` meant that `featureConfig[Feature.ExportCodeScanningConfigEnabled].minimumVersion` was undefined.  This in turn meant that we tried to pass some CLI options even if we were running with an early version of the CLI where these options didn't exist.

The fact that we didn't catch this in PR checks highlights the need for more comprehensive testing across the full range of minor versions we support.


src/feature-flags.ts
```diff
@@ -94,7 +94,7 @@ export class Features implements FeatureEnablement {
     gitHubVersion: util.GitHubVersion,
     repositoryNwo: RepositoryNwo,
     tempDir: string,
-    logger: Logger
+    private readonly logger: Logger
```
angelapwen, line 97: Why do we make this `private readonly`?

aeisenberg (approved):

angelapwen (approved):
Thanks for investigation & extra debug statements 🕵️

src/feature-flags.ts
henrymercer: This makes it a field on the `Features` class so we can access it within `getValue`.

henrymercer: We could also add a `logger` argument to `getValue`, but that would cause a reasonable amount of churn, and sharing the logger makes sense.

tied2token (approved):
i dont see any change

src/feature-flags.ts
angelapwen: Ah, got it, thank you

PR #1598, opened 2023-03-21
Update `upload` input values and logic

angelapwen:
- The `upload` input to the `analyze` Action now accepts the following values:
    - `always` is the default value, which uploads the SARIF file to Code Scanning for successful and failed runs.
    - `failure-only` is recommended for customers post-processing the SARIF file before uploading it to Code Scanning. This option uploads debugging information to Code Scanning for failed runs to improve the debugging experience.
    - `never` avoids uploading the SARIF file to Code Scanning even if the code scanning run fails. This is not recommended for external users since it complicates debugging.
    - The legacy `true` and `false` options will be interpreted as `always` and `failure-only` respectively.


henrymercer (commented):
Nice!

src/init-action-post-helper.test.ts
```diff
@@ -161,7 +161,7 @@ test("uploads failed SARIF run with database export-diagnostics if the database
   });
 });
 
-test("doesn't upload failed SARIF for workflow with upload: false", async (t) => {
+test("uploads failed SARIF run for workflow with upload: failure-only", async (t) => {
```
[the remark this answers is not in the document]
angelapwen: Makes sense — added for `true` and `false` as well.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Update default CodeQL bundle version to 2.12.5. [#1585](https://github.com/github/codeql-action/pull/1585)
+- Customers using `upload: false` to post-process the SARIF output of the `analyze` Action before uploading it to Code Scanning will benefit from an improved debugging experience. The CodeQL Action will now upload a SARIF file with debugging information to Code Scanning on failed runs for these customers. Previously, this was only available for customers using the default value of the `upload` input. You can disable this behavior by passing an `upload: never` input to the `analyze` Action. [#1598](https://github.com/github/codeql-action/pull/1598)
```
henrymercer, line 6:
```suggestion
- Customers post-processing the SARIF output of the `analyze` Action before uploading it to Code Scanning will benefit from an improved debugging experience. [#1598](https://github.com/github/codeql-action/pull/1598)
  - The CodeQL Action will now upload a SARIF file with debugging information to Code Scanning on failed runs for customers using `upload: false`. Previously, this was only available for customers using the default value of the `upload` input.
  - The `upload` input to the `analyze` Action now accepts the following values:
    - `always` is the default value, which uploads the SARIF file to Code Scanning for successful and failed runs.
    - `failure-only` is recommended for customers post-processing the SARIF file before uploading it to Code Scanning. This option uploads debugging information to Code Scanning for failed runs to improve the debugging experience.
    - `never` avoids uploading the SARIF file to Code Scanning even if the code scanning run fails. This is not recommended for external users since it complicates debugging.
    - The legacy `true` and `false` options will be interpreted as `always` and `failure-only` respectively.
```

analyze/action.yml
```diff
@@ -10,10 +10,15 @@ inputs:
     required: false
     default: "../results"
   upload:
-    description: Upload the SARIF file to Code Scanning
+    description: >-
+      Upload the SARIF file to Code Scanning.
+      This defaults to 'always' which uploads the SARIF file to Code Scanning for successful and failed runs.
+      You can specify 'failure-only' to only upload the SARIF file to Code Scanning for failed runs, for instance if you are post-processing the SARIF file before uploading it to Code Scanning.
+      Setting this input to 'never' avoids uploading the SARIF file to Code Scanning even if the code scanning run fails.
+      This is not recommended for external users since it complicates debugging.
```
henrymercer, line 18:
```suggestion
    description: >-
      Upload the SARIF file to Code Scanning.
      This defaults to 'always' which uploads the SARIF file to Code Scanning for successful and failed runs.
      You can specify 'failure-only' to only upload debugging information to Code Scanning if the workflow run fails, for instance if you are post-processing the SARIF file before uploading it to Code Scanning.
      Setting this input to 'never' avoids uploading the SARIF file to Code Scanning even if the code scanning run fails.
      This is not recommended for external users since it complicates debugging.
```

henrymercer (commented):
Some further changelog improvement suggestions!

src/actions-util.ts
```diff
@@ -674,3 +674,8 @@ export async function printDebugLogs(config: Config) {
     walkLogFiles(logsDirectory);
   }
 }
+
+export async function isValidUploadInput(input: string | undefined) {
```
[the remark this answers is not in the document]
angelapwen: Makes sense — done!

src/init-action-post-helper.test.ts
```diff
@@ -179,7 +179,61 @@ test("doesn't upload failed SARIF for workflow with upload: false", async (t) =>
...
+        upload: "unrecognized string",
+      },
+    },
+  ]);
+  await testFailedSarifUpload(t, actionsWorkflow, {
+    category: "my-category",
+  });
+});
+
+test("uploading failed SARIF run fails for workflow with upload: never", async (t) => {
```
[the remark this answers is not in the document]
angelapwen: Done

src/workflow.ts
```diff
@@ -424,25 +424,34 @@ export function getCategoryInputOrThrow(
...
+    getAnalyzeActionName(),
+    "upload",
+    matrixVars
   );
+  if (userInput === undefined || userInput === "true") {
+    return "always";
+  }
+  if (userInput === "false") {
+    return "failure-only";
+  }
```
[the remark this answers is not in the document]
angelapwen: That makes sense, so I don't have to implement some of this logic twice. I moved it in and allowed this function to return `undefined` since I am converting that to `always` in the `actions-util` function anyway.

src/actions-util.ts
```diff
@@ -674,3 +674,24 @@ export async function printDebugLogs(config: Config) {
     walkLogFiles(logsDirectory);
   }
 }
+
+export type ValidUploadValues = "always" | "failure-only" | "never";
```
henrymercer, line 678:
```suggestion
export type UploadKind = "always" | "failure-only" | "never";
```

src/actions-util.ts
```diff
@@ -674,3 +674,24 @@ export async function printDebugLogs(config: Config) {
...
+  if (input === "false") {
+    return "failure-only";
+  }
+  if (!["always", "failure-only", "never", "true", "false"].includes(input)) {
+    core.warning(
+      `Unrecognized 'upload' input to 'analyze' Action: ${input}. Defaulting to 'always'.`
+    );
+    return "always";
+  }
+  return <ValidUploadValues>input;
```
henrymercer, line 696: Minor: I'd be curious to see whether we can remove this cast so that the type checker can help us verify that we've implemented this correctly.

What about including `input == "always"` in the first clause, `input == "failure-only"` in the second clause, and adding another clause for `input === "never"`?

src/init-action-post-helper.test.ts
```diff
@@ -179,7 +179,142 @@ test("doesn't upload failed SARIF for workflow with upload: false", async (t) =>
...
+        upload: "always",
+      },
+    },
+  ]);
+  await testFailedSarifUpload(t, actionsWorkflow, {
+    category: "my-category",
+  });
+});
+
+test("uploads failed SARIF run for workflow with upload: true", async (t) => {
```
henrymercer, line 218: Thanks for adding these tests!  As a minor point, you could probably move these inside a loop to save some duplication.

src/actions-util.ts
angelapwen: Yep, that seems to work 👍

src/actions-util.ts
angelapwen: Done!

src/init-action-post-helper.test.ts
angelapwen: Yeah, that made it a lot cleaner 😄

src/init-action-post-helper.test.ts
```diff
@@ -161,33 +161,70 @@ test("uploads failed SARIF run with database export-diagnostics if the database
...
+    uploadInput: "never",
+    shouldUpload: false,
+  },
+  {
+    uploadInput: "unrecognized-value",
+    shouldUpload: true,
+  },
+];
+
+for (const { uploadInput, shouldUpload } of UPLOAD_INPUT_TEST_CASES) {
```
henrymercer, line 191: Nice!

src/actions-util.ts
```diff
@@ -674,3 +674,24 @@ export async function printDebugLogs(config: Config) {
     walkLogFiles(logsDirectory);
   }
 }
+
+export type UploadKind = "always" | "failure-only" | "never";
+
+// Returns a ValidUploadValue, converting unspecified and deprecated upload inputs appropriately.
+export async function getUploadValue(
```
henrymercer, line 681: We could probably drop the `async` here right?

angelapwen: Yep, good catch!

src/actions-util.ts
```diff
@@ -674,3 +674,22 @@ export async function printDebugLogs(config: Config) {
     walkLogFiles(logsDirectory);
   }
 }
+
+export type UploadKind = "always" | "failure-only" | "never";
+
+// Parses the `upload` input into an `UploadKind`, converting unspecified and deprecated upload inputs appropriately.
+export function getUploadValue(input: string | undefined): UploadKind {
+  if (input === undefined || input === "true" || input === "always") {
```
henrymercer, line 682: It just occured to me that this could be more readably written as a `switch` statement, but no need to change this.

henrymercer (dismissed):
Very nice!

src/actions-util.ts
angelapwen: Just did this for fun 😆 thanks for suggestion!

henrymercer (dismissed):
:shipit:

henrymercer (approved):

PR #1604, opened 2023-03-24
update actions/checkout version

GrantBirki:
This pull request updates the `actions/checkout` versions


henrymercer (commented):
I looked at [the tags on `actions/checkout`](https://github.com/actions/checkout/tags) and v3 and v3.5.0 point to the same commits.  So merging this would only hold this workflow back from future releases of `actions/checkout`.

GrantBirki:
Oh interesting! That is good to know, I'll close this out

PR #1605, opened 2023-03-24
Work around duplicate locations bug in diagnostics export

henrymercer:
This PR works around a bug in diagnostics export that could lead to an invalid SARIF file being produced by `database interpret-results` or `database export-diagnostics`.

The bug lies in the CLI functionality that groups diagnostics together.  When grouping together multiple diagnostics that have the same location `a`, the CLI produced a single diagnostic with multiple identical locations `[a, ..., a]`.  This is not valid SARIF, as the spec mandates that the locations for each `notification` object are distinct.

To allow us to roll out diagnostics export without waiting for a new CLI release, we workaround this bug in the CodeQL Action.  To do this, when diagnostics export is enabled, we go through the SARIF file produced by `database interpret-results` and `database export-diagnostics` and remove any duplicate locations from the `runs[].invocations[].toolExecutionNotifications` property.

We add unit tests and integration tests, as well as an environment variable `CODEQL_ACTION_DISABLE_DUPLICATE_LOCATION_FIX` to disable this functionality if unexpected problems arise.


aeisenberg:
I haven't had a chance to review yet, but a general question. Are we fairly sure the fix will be available in the next CLI? If so, should we be gating this functionality on the CLI version as well? I.e., use the original behaviour of version is greater than 2.12.x?

angelapwen:
> Are we fairly sure the fix will be available in the next CLI? If so, should we be gating this functionality on the CLI version as well? I.e., use the original behaviour of version is greater than 2.12.x?

The fix has already been merged and patched in the CLI (I backlinked it as it's an internal PR), so I don't believe we need to add an additional gate.

src/util.ts
```diff
@@ -81,6 +89,18 @@ export interface SarifResult {
   };
 }
 
+export interface SarifNotification {
+  locations?: SarifLocation[];
```
angelapwen, line 94: Is there a reason we don't want this to be a `Set<SarifLocation>` to enforce that each member of the set is distinct? is it so that we want to be able to disable this functionality via the environment variable?

src/codeql.ts
```diff
@@ -1040,6 +1048,13 @@ export async function getCodeQLForCmd(
         args.push("--sarif-category", automationDetailsId);
       }
       await new toolrunner.ToolRunner(cmd, args).exec();
+
+      // Fix invalid notifications in the SARIF file output by CodeQL.
+      let sarif = JSON.parse(
+        fs.readFileSync(intermediateSarifFile, "utf8")
+      ) as util.SarifFile;
+      sarif = util.fixInvalidNotifications(sarif, logger);
+      fs.writeFileSync(sarifFile, JSON.stringify(sarif));
```
angelapwen, line 1057: Would it be worthwhile to pull this out into a separate function to make sure that the logic remains the same between here and in `diagnostics export`?

angelapwen (approved):
Looked through and makes sense to me. Approving but will wait for Andrew's review too!

src/util.ts
aeisenberg: I think it is because this interface is part of an object being parsed from a JSON string. `Set` is not available in JSON.

src/util.ts
```diff
@@ -828,3 +799,82 @@ export function parseMatrixInput(
...
+  logger: Logger
+): SarifFile {
+  if (process.env[CODEQL_ACTION_DISABLE_DUPLICATE_LOCATION_FIX] === "true") {
+    logger.info(
+      "SARIF notification object duplicate location fix disabled by the " +
+        `${CODEQL_ACTION_DISABLE_DUPLICATE_LOCATION_FIX} environment variable.`
+    );
+    return sarif;
+  }
+  if (!(sarif.runs instanceof Array)) {
```
aeisenberg, line 826: Minor: using `Array.isArray` is generally better. However, in this particular case `instanceof` is fine. It's only really necessary when using iframes in DOM.

```suggestion
  if (!Array.isArray(sarif.runs)) {
```

https://stackoverflow.com/questions/22289727/difference-between-using-array-isarray-and-instanceof-array

src/util.ts
```diff
@@ -828,3 +799,82 @@ export function parseMatrixInput(
...
+  // This is a workaround for a bug in the CodeQL CLI that causes duplicate locations to be
+  // emitted in some cases.
+  let numDuplicateLocationsRemoved = 0;
+
+  const newSarif = {
+    ...sarif,
+    runs: sarif.runs.map((run) => {
+      if (
+        run.tool?.driver?.name !== "CodeQL" ||
+        !(run.invocations instanceof Array)
```
aeisenberg, line 840: minor: as above

```suggestion
        !(Array.isArray(run.invocations))
```

src/util.ts
```diff
@@ -828,3 +799,82 @@ export function parseMatrixInput(
...
+      ) {
+        return run;
+      }
+      return {
+        ...run,
+        invocations: run.invocations.map((invocation) => {
+          if (!(invocation.toolExecutionNotifications instanceof Array)) {
```
aeisenberg, line 847:
```suggestion
          if (!Array.isArray(invocation.toolExecutionNotifications)) {
```

src/util.ts
```diff
@@ -828,3 +799,82 @@ export function parseMatrixInput(
...
+            return invocation;
+          }
+          return {
+            ...invocation,
+            toolExecutionNotifications:
+              invocation.toolExecutionNotifications.map((notification) => {
+                if (!(notification.locations instanceof Array)) {
```
aeisenberg, line 854:
```suggestion
                if (!Array.isArray(notification.locations)) {
```

aeisenberg (approved):
I have a handful of questions and a stylistic suggestion. Though, I think this generally looks good assuming we don't want a full feature flag.

src/util.ts
henrymercer: What Andrew said — since we're just using `JSON.parse` here, we can't define clever things like in Jackson where we say deserialize this field to a set.

henrymercer:
Thanks folks for the review.  I'll address these improvements as followup to avoid blocking this on another review cycle.

PR #1615, opened 2023-03-28
Use apiURL to determine whether to use the token

lcartey:
The action includes this check to determine whether to include the token in the request:
```typescript
  } else if (codeqlURL.startsWith(`${apiDetails.url}/`)) {
    logger.debug("Providing an authorization token to download CodeQL tools.");
    authorization = `token ${apiDetails.auth}`;
  }
```
The intention is to only include the token if we are fetching from the same instance, to avoid leaking tokens.

However, the `codeqlURL` in this check will refer to the API `url` to use to download the release. For example:

https://api.github.com/repos/myrepo/codeql-action/releases/assets/<id>

However, the `apiDetails.url` refers to the instance URL, e.g. `https://github.com`, which means the check will always fail. We therefore need to use the `apiDetails.apiURL` to get the appropriate comparison.


src/setup-codeql.ts
```diff
@@ -608,7 +608,7 @@ export async function downloadCodeQL(
   let authorization: string | undefined = undefined;
   if (searchParams.has("token")) {
     logger.debug("CodeQL tools URL contains an authorization token.");
-  } else if (codeqlURL.startsWith(`${apiDetails.url}/`)) {
+  } else if (codeqlURL.startsWith(`${apiDetails.apiURL}/`)) {
```
henrymercer, line 611: If a user provides a tools URL explicitly like `tools: https://github.com/github/codeql-action/releases/download/codeql-bundle-20230317/codeql-bundle-linux64.tar.gz`, I think we'll want to authenticate this too.  Therefore I think we might want to loosen this to ```codeqlURL.startsWith(`${apiDetails.url}/`) || codeqlURL.startsWith(`${apiDetails.apiURL}/`)```.

henrymercer (commented):
Seems reasonable, but I think we want to preserve the old behaviour.

henrymercer:
Closing in favor of https://github.com/github/codeql-action/pull/1945.

PR #1619, opened 2023-03-28
Allow workflow to be passed via an environment variable for default setup

henrymercer:
The default setup workflow is not checked into the repository, but the Action relies on the workflow file to make features like reporting failed runs work.  We'd like to move to a more robust method of making features like reporting failed runs work, but this is hard to do without making a breaking change to the workflow file, and we aren't currently in a good position to do that.

Therefore as a medium term solution, default setup will pass its workflow file to the Action via the `CODE_SCANNING_WORKFLOW_FILE` environment variable.

GitHub staff: see [an example run](https://github.com/dsp-testing/henrymercer-default-setup-test/actions/runs/4678604021/jobs/8287855026) and [the resulting status page](https://github.com/dsp-testing/henrymercer-default-setup-test/security/code-scanning/tools/CodeQL/status/configurations/automatic/820e8a4e1e3d03257373e351d381b0157bb0795f835e6f57cfb81db8cb074aac).


src/workflow.ts
```diff
@@ -237,24 +240,46 @@ export function formatWorkflowCause(errors: CodedError[]): undefined | string {
   return errors.map((e) => e.code).join(",");
 }
 
-export async function getWorkflow(): Promise<Workflow> {
+async function getWorkflowFilePath(logger: Logger): Promise<string> {
+  const maybePath = process.env["CODE_SCANNING_WORKFLOW_FILE"];
```
marcogario, line 244: The variable contains the base64 encoded workflow, not the path.

henrymercer: Thanks for the catch!

angelapwen (approved):
Looks good and found the appropriate debug log in the internal run (2nd attempt) linked, but assume you want @Daverlo's 👀 too!

A couple of questions:
- will we add integration tests for default setup at a later point?
- do we want to add a changenote since this is user facing?

henrymercer:
> * will we add integration tests for default setup at a later point?

Yes!  Tracking this internally.

> * do we want to add a changenote since this is user facing?

Good point, let's add a changenote for this.

angelapwen (approved):
Not sure why the PR check is failing though. Re-ran with debug mode. 

Edit: looks like we might have a test flake, but because the re-run succeeded I'm not sure what it was!

src/workflow.ts
```diff
@@ -237,30 +241,52 @@ export function formatWorkflowCause(errors: CodedError[]): undefined | string {
...
+      "This can happen if the currently running workflow checks out a branch that doesn't contain " +
+      "the corresponding workflow file."
+  );
 }
 
 /**
- * Get the path of the currently executing workflow.
+ * Get the path of the currently executing workflow relative to the repository root.
  */
-export async function getWorkflowPath(): Promise<string> {
```
Daverlo, line 263: What is this this returning when using default setup?

I don't think it really matters as I only see it being used for getting the analysis key, and for default setup we are always passing a category, so this value should be overwritten. 

But I'm wondering if this should highlight that there is no workflow file somehow?

Daverlo (approved):
LGTM!

src/workflow.ts
henrymercer: For default setup, this will return `dynamic/github-code-scanning/codeql`, but as you say we get the analysis key directly from `CODEQL_ACTION_ANALYSIS_KEY`.  I think this is a good thing to highlight though, as this is part of some work I'd like us to do around nailing down in the code what context the CodeQL Action expects from default setup, in terms of the `github` job context, environment variables, etc.

henrymercer: At the moment, I don't think we want to fail if the workflow file isn't found — we want to preserve the way we've been generating the analysis key to avoid breaking advanced setups where the workflow file doesn't exist, e.g. reusable workflows or workflows executing in other repositories.

PR #1625, opened 2023-03-29
Remove checks for triggering on specific paths

rneatherway:
These are no longer necessary with the new approach to selecting alerts to show on pull requests.


henrymercer (approved):
Looks reasonable.  Is the `MismatchedBranches` warning also obsolete?

rneatherway:
> Is the `MismatchedBranches` warning also obsolete?

That one needs adjusting, but I need to think about how to rephrase it a bit more, so thought I would just these done first.

PR #1631, opened 2023-04-04
Skip the SARIF notification object workaround for CLIs that have fixed this bug

henrymercer:
The latest version of the CodeQL CLI (2.12.6) contains a fix for the duplicate SARIF notification object bug in 2.12.5.

This PR updates the Action so it no longer applies the workaround to CodeQL CLI version 2.12.6 and later.  We extend the integration tests to make sure we keep checking the workaround on 2.12.5.


src/codeql.ts
```diff
@@ -878,7 +884,13 @@ export async function getCodeQLForCmd(
         Feature.ExportDiagnosticsEnabled,
         this
       );
-      const codeqlOutputFile = shouldExportDiagnostics
+      const shouldWorkaroundInvalidNotifications =
+        shouldExportDiagnostics &&
+        !(await util.codeQlVersionAbove(
+          this,
+          CODEQL_VERSION_DUPLICATE_NOTIFICATIONS_FIXED
+        ));
```
aeisenberg, line 892: It would be nice to extract this to a function so that the logic can be reused below.

aeisenberg (commented):
Small suggestion that I think would make the code easier to understand.

src/codeql.ts
henrymercer: If I've understood your comment correctly, only:

```
!(await util.codeQlVersionAbove(
          this,
          CODEQL_VERSION_DUPLICATE_NOTIFICATIONS_FIXED
        )
```

is reused.  This doesn't seem worth extracting into a function.

aeisenberg (approved):

PR #1636, opened 2023-04-06
Re-enable duplicate SARIF notification location workaround

henrymercer:
We thought we'd fixed this in version 2.12.6 of the CLI, and our relevant test cases all passed.  However now that 2.12.6 has fully rolled out, we've found another instance of this in the wild, so we need to re-enable the workaround in the Action.


henrymercer:
Investigation will follow into what caused this problem to reoccur, but I'd like to re-enable the workaround as a fast fix.

angelapwen (approved):

PR #1640, opened 2023-04-06
Add `workflow_run_attempt` data to status report

angelapwen:
With this change, we begin reporting the `workflow_run_attempt` field in the status report base, using the `GITHUB_RUN_ATTEMPT` default environment variable ([docs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)) 

Similar to the `workflow_run_id`, we report -1 if the environment variable isn't found for whatever reason. 


src/actions-util.ts
```diff
@@ -410,6 +412,11 @@ export async function createStatusReportBase(
   if (workflowRunIDStr) {
     workflowRunID = parseInt(workflowRunIDStr, 10);
   }
+  const workflowRunAttemptStr = process.env["GITHUB_RUN_ATTEMPT"];
+  let workflowRunAttempt = -1;
```
adityasharad, line 416: I was going to suggest `undefined` if not set, but if this is the convention we use for the workflow run ID, stick with it.

src/actions-util.ts
```diff
@@ -437,6 +444,7 @@ export async function createStatusReportBase(
 
   const statusReport: StatusReportBase = {
     workflow_run_id: workflowRunID,
+    workflow_run_attempt: workflowRunAttempt,
```
adityasharad, line 447: Do we have tests that check this is set in the status report?

angelapwen: I took a look and there are no existing tests for any of the status report fields except for specific cases, such as [if the duration fields for each language are set ](https://github.com/github/codeql-action/blob/988e1bc94181d8b9c03964b72e28974e2dbca322/src/analyze.test.ts#L28)

I actually noticed that we don't have any tests for `init` or `init-action` themselves, just `init-action-post-helper` as well. 

I was planning to validate this field manually after merging and releasing, which is what we've done while adding fields previously, and I can file a follow-up issue for testing.

adityasharad: I would like to see some validation runs (e.g. referencing the action from this branch), and we can inspect the status reports internally before releasing.

angelapwen: I've created some 3 runs at https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/actions/runs/4660282650/jobs/8248393316 (GitHub internal only) and validated internally that the `workflow_run_attempt` field is populating appropriately. Will DM ✉️

adityasharad (dismissed):

adityasharad (approved):

PR #1648, opened 2023-04-14
Fix pre-release trigger for update-bundle action

cklin:
This PR switches the update-bundle release trigger from `prereleased` to `published` because the former has been documented not to work.

From https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release:

> Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

### DIFF
--- a/.github/workflows/update-bundle.yml
+++ b/.github/workflows/update-bundle.yml
@@ -2,11 +2,20 @@ name: Update default CodeQL bundle
 
 on:
   release:
-    types: [prereleased]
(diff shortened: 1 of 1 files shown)


angelapwen (approved):
Wow, that's not intuitive at all! Thank you for catching this 😄

cklin:
cc @henrymercer

PR #1590, opened 2023-03-18
Add new configuration Parameter

tgrall:
- Write test to check it is read from configuration
- Update documentation

### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.
- [ ] I may have to work on some exception management when the `configuration` parameter is not well formatted. 

----

close #1589 1589

PR #1653, opened 2023-04-19
Add `threat-models` as a property to config file and inputs

aeisenberg:
There's a lot of changes here, but it's pretty formulaic. It follows the approach used by the `queries` input and config property. `threat-models` can appear as an input or in the config file. If it appears in the input, then we need to either merge it with the threat-models in the config (if prefixed with `+`) or overwrite it.

There's no danger if someone uses `threat-models` with an older CLI since the CLI can handle configs with extra properties.


henrymercer (commented):
Another approach here would be to merge the work on allowing users to pass config properties via a `config` input (https://github.com/github/codeql-action/pull/1590), then letting users pass threat models via something like:

```yml
uses: github/codeql-action/init@v2
with:
  config: |
    threat-models: <threat models>
```

An advantage of this is consistency — we have a single way to pass in threat models.  Another is less special-purpose code.  A disadvantage is that users couldn't use a config file but then also override or add some threat models.  I'm not sure how important that is.

aeisenberg:
Thanks good point. It keeps the code significantly simpler. 

My only concern about this is the inconsistency. We allow users to specify `packs` and `queries` via inputs. Would we not want to do the same with threat models?

Perhaps the best approach would be to explicitly deprecate the packs and queries inputs (and any others that are duplicated in the config file).

PR #1658, opened 2023-04-25
Add `workload_run_attempt` to analysis upload

angelapwen:
Previously we began sending the `workflow_run_attempt` field to our status reports. We also want to send it to the analysis upload API 😄 

This PR does a bit of refactoring so that `workflow_run_id` and `workflow_run_attempt` are populated the same across the status report and analysis payloads. 


src/workflow.ts
```diff
@@ -319,6 +319,22 @@ export function getWorkflowRunID(): number {
...
+ * Get the workflow run attempt number.
+ */
+export function getWorkflowRunAttempt(): number {
+  const workflowRunAttempt = parseInt(
+    getRequiredEnvParam("GITHUB_RUN_ATTEMPT"),
+    10
+  );
+  if (Number.isNaN(workflowRunAttempt)) {
+    throw new Error(
+      "GITHUB_RUN_ATTEMPT must define a non NaN workflow run attempt"
```
aeisenberg, line 332: Maybe include the original value here so we can see what the error was.

aeisenberg: Also, (minor) maybe make sure that the value is `> 0`?

angelapwen: Done. It wasn't clear to me if `GITHUB_RUN_ID` could possibly be 0, so I only checked for negative values 🤔 `GITHUB_RUN_ATTEMPT` clearly states that the value should begin at 1: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

aeisenberg (approved):

PR #1675, opened 2023-05-09
Remove MismatchedBranches check from code scanning workflow validation

shaikhul:
Remove `MismatchedBranches` check from code scanning workflow validation as these two have misleading wording.


src/workflow.test.ts
```diff
@@ -303,27 +242,16 @@ test("formatWorkflowErrors() when there is one error", (t) => {
   t.true(message.startsWith("1 issue was detected with this workflow:"));
 });
 
-test("formatWorkflowErrors() when there are multiple errors", (t) => {
```
henrymercer, line 306: Nit: ideally we'd keep this test since `formatWorkflowErrors()` takes an array of errors, and we could reasonably add another `WorkflowErrors` member in the future.  Given that `WorkflowErrors.MismatchedBranches` no longer exists, we could just use something like `{ message: "Something else went wrong", code: 42 }`.

henrymercer (approved):

henrymercer:
My understanding is that the main change of this PR is to remove the checks for `push` triggers in code scanning workflows, since with the simplified diff approach this is no longer required.  If that's right, could we update the PR title / description accordingly?  We may also want to add a brief changelog note in `CHANGELOG.md`.  This will make it easier to find this PR in the future.

src/workflow.test.ts
shaikhul: isn't this covered by previous test at line 240? https://github.com/github/codeql-action/pull/1675/files#diff-f559966d12ec92f40e27ca272f4e68e70a9f284724c90b45f2a7c02319dbe0abR240

henrymercer: We're covering the case where there's only a single error there, but not the multiple errors case.

henrymercer (approved):

henrymercer:
@shaikhul Would you mind recompiling the Action and pushing again? 🙏 Unfortunately for now we need to check in the compiled code.

shaikhul:
@henrymercer done 903cb27

henrymercer (approved):
Thanks for doing this!  One fix to the changelog format I didn't notice before, otherwise this is good to go!

henrymercer (approved):

PR #1678, opened 2023-05-12
Flag up functionality that may not exist in default setup workflows

henrymercer:
This PR adds two internal CodeQL queries to identify environment variables and Actions context variables that may not work with default setup.

While we're here, we switch `libraryPathDependencies` to `dependencies` and modify the "Inconsistent Action inputs" query to ignore internal Actions.


queries/codeql-pack.yml
```diff
@@ -1,4 +1,4 @@
 name: codeql-action-custom-queries-javascript
 version: 0.0.0
-libraryPathDependencies: codeql-javascript
-
+dependencies:
+  codeql/javascript-all: 0.6.1
```
aeisenberg, line 4: Wow...surprised we haven't updated this yet.

queries/default-setup-event-context.ql
```diff
@@ -0,0 +1,62 @@
...
+      DataFlow::SharedFlowStep::step(src, trg, _, _)
+    ) and
+    inlbl = outlbl
+  }
+}
+
+from EventContextAccessConfiguration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
+where
+  cfg.hasFlowPath(source, sink) and
+  not sink.getNode().asExpr().getFile().getBaseName().matches("%.test.ts")
```
aeisenberg, line 57: Minor: This could be moved to within the `isSink` predicate.

queries/default-setup-event-context.ql
```diff
@@ -0,0 +1,62 @@
...
+select sink.getNode(), source, sink,
+  "This context property may not exist in default setup workflows. If all uses are safe, add it to the list of "
+    + "context properties that are known to be safe in " +
+    "'queries/default-setup-event-context.ql'. If this use is safe but others are not, " +
```
aeisenberg, line 61: I don't see a list of properties yet.

aeisenberg (commented):
Nice. I think I understand the QL.

queries/default-setup-event-context.ql
henrymercer: So far we don't have any.  There aren't any event variables we want to rely on currently, so I've updated the alert message to be a little less prescriptive.

aeisenberg (approved):
NIce.

PR #1682, opened 2023-05-16
Extract semantic CLI version from URL when requesting specific tools

henrymercer:
This PR prepares the Action for migrating the CodeQL Bundle to be versioned using a semantic version (e.g. `codeql-bundle-v1.2.3`).  Most of the work is already done, but this PR makes a small improvement to ensure we can look up CLIs requested via `tools` in the tool cache before downloading them.  Specifically, we populate `cliVersion` when we can recognize that the requested CodeQL Bundle tag uses a semantic version.  The rest of the changes here are pulling out shared test utilities so we can write more specific CodeQL setup tests in `setup-codeql.test.ts`.

There's scope for some cleanup and refactoring here now we're migrating the bundle version to semver, but this will be easier to do once this change is fully rolled out.

We also add a changelog note to notify customers about the change to the way that bundles will be tagged for future CodeQL versions.


angelapwen (approved):
Nice, thanks for doing some of the testing refactoring too 😸

angelapwen (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 ## [UNRELEASED]
 
 - We are rolling out a feature in May 2023 that will disable Python dependency installation for new users of the CodeQL Action. This improves the speed of analysis while having only a very minor impact on results. [#1676](https://github.com/github/codeql-action/pull/1676)
+- We are improving the way that the CodeQL bundle is tagged to streamline the process of obtaining the CodeQL tools.
```
adityasharad, line 6: A few suggestions:
- Link to https://github.com/github/codeql-action/releases so that readers know what we mean by bundles.
- Clarify the reason: this allows us to easily identify bundles by their CodeQL semantic version.
- Clarify what a user should change with an example, e.g. if they are downloading specific artifacts from a `codeql-bundle-<timestamp>` release, they should change it to either the latest release or a `codeql-bundle-<version>` tag.

adityasharad (approved):

PR #1617, opened 2023-03-28
Autobuild using CodeQL CLI

henrymercer:
This PR implements autobuilding using the hidden `database autobuild` CLI subcommand.  This simplifies the implementation of autobuild, opens the door to sharing more code between the CodeQL Action and third party CI for CodeQL, and enable us to more efficiently implement new features like improved diagnostics for autobuilders.

We will gradually roll this out for CodeQL CLI 2.12.6 and later using the `code_scanning_cli_autobuild_enabled` feature flag.  While the CLI and Action implementation of autobuild is very similar, we will not enable this by default on GHES 3.9 to reduce risk.

### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #1684, opened 2023-05-18
Add `resolve-environment` action

mbg:


resolve-environment/action.yml
```diff
@@ -0,0 +1,27 @@
+name: 'CodeQL: Resolve Build Environment'
+description: 'Attempt to infer a suitable environment configuration for the autobuilder'
```
henrymercer, line 2:
```suggestion
description: '[Experimental] Attempt to infer a suitable environment configuration for the autobuilder'
```

resolve-environment/action.yml
```diff
@@ -0,0 +1,27 @@
...
+  matrix:
+    default: ${{ toJson(matrix) }}
+  language:
+    description: The language to infer the build environment configuration for.
+    required: true
+  working-directory:
+    description: >-
+      Run the autobuilder using this path (relative to $GITHUB_WORKSPACE) as
+      working directory. If this input is not set, the autobuilder runs with
+      $GITHUB_WORKSPACE as its working directory.
```
henrymercer, line 20: Nit: let's update this before we merge.

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,121 @@
...
+        `Changing autobuilder working directory to ${workingDirectory}`
+      );
+      process.chdir(workingDirectory);
+    }
+
+    const result = await runResolveBuildEnvironment(
+      initCodeQLResult.codeql.getPath(),
+      logger,
+      language
+    );
```
henrymercer, line 94: One thing that's missing here is tolerance to older CLIs.  We should discuss what we want to do here, whether that's returning some kind of empty result, returning nothing, or raising a better error including the minimum version number.

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,121 @@
...
+import { runResolveBuildEnvironment } from "./resolve-environment";
+import {
+  checkForTimeout,
+  checkGitHubVersionInRange,
+  getRequiredEnvParam,
+  wrapError,
+} from "./util";
+import { validateWorkflow } from "./workflow";
+
+const actionName = "resolve-environment";
```
henrymercer, line 25: Nit: rename this to `ACTION_NAME` since it's static.

src/resolve-environment.ts
```diff
@@ -0,0 +1,15 @@
+import { getCodeQL } from "./codeql";
+import { Language } from "./languages";
+import { Logger } from "./logging";
+
+export async function runResolveBuildEnvironment(
+  cmd: string,
+  logger: Logger,
+  language: Language
+) {
+  logger.startGroup(`Attempting to resolve build environment`);
```
henrymercer, line 10:
```suggestion
  logger.startGroup(`Attempting to resolve build environment for ${language}`);
```

src/codeql.ts
```diff
@@ -229,6 +235,20 @@ export interface ResolveQueriesOutput {
...
+      os?: BuildEnvironmentOS;
+      [key: string]: unknown;
+    };
+  };
+}
+
+export interface BuildEnvironmentOS {
+  name?: string;
+  version?: string;
+}
```
henrymercer, line 250: At the moment the CodeQL Action doesn't do anything with the output of `resolve build-environment` beyond just passing it on.  I think for now we should consider having the CodeQL Action treat the output  as opaque — an empty interface or `unknown` would be fine.  That way we avoid taking a dependency while this command is still experimental.

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,121 @@
...
+      logger
+    );
+
+    const workingDirectory = getOptionalInput("working-directory");
+    if (workingDirectory) {
+      logger.info(
+        `Changing autobuilder working directory to ${workingDirectory}`
+      );
+      process.chdir(workingDirectory);
+    }
```
henrymercer, line 88: We should pass this to the CLI via the `--working-dir` argument to `resolve build-environment`.  That saves us a step in the Action :)

henrymercer (commented):
This is looking in good shape!

- I think it's worth writing a CHANGELOG note for observant customers who are watching the repo to say that this Action is an internal experiment, is subject to change and shouldn't be used in production.
- It'd be good to have a basic integration test of this once we have a CLI to test it against, just to make sure everything is wired up correctly.
- We will need to make some internal changes to start capturing telemetry properly — for instance we need to add `resolve-environment` to an allowlist in the monolith.  Though there's no rush to make this change.

src/resolve-environment-action.ts
mbg: I copied this from the `autobuild` action, so the behaviour of the two actions is equivalent in that regard. If we pass this to the CLI here, would it make sense to do that for the `autobuild` action as well?

src/codeql.ts
mbg: Ah this is actually related to the one thing I mentioned that's still missing here: we discussed over in https://github.com/github/codeql/pull/13211#discussion_r1201759876 that the action could add in a `runnerImage` key based on the OS name and version with a suitable GHA image name that matches those.

src/resolve-environment-action.ts
henrymercer: Yes — I actually have [a PR](https://github.com/github/codeql-action/pull/1617) up for that!  At the moment, the `autobuild` Action works by finding autobuild scripts and calling them directly, but I would like to have the Action instead call a CLI command.

src/codeql.ts
henrymercer: Ah thanks, I'll comment over there.

mbg: As discussed over in the other PR, this is not essential right now, so I will modify this to treat the output as opaque.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Experimental: add a new `resolve-environment` action which invokes a corresponding, new CLI command that allows CodeQL extractors to try and infer a configuration for the build environment that is required to build a given project. Do not use this in production as it is part of an internal experiment and subject to change at any time.
```
aeisenberg, line 5: Minor: 

```suggestion
- Experimental: add a new `resolve-environment` action which invokes a CodeQL extractor to attempt to infer a configuration for the build environment that is required to build a given project. Do not use this in production as it is part of an internal experiment and subject to change at any time.
```

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,84 @@
...
+    const workingDirectory = getOptionalInput("working-directory");
+    const result = await runResolveBuildEnvironment(
+      config.codeQLCmd,
+      logger,
+      workingDirectory,
+      language
+    );
+    core.setOutput("environment", result);
+  } catch (unwrappedError) {
+    const error = wrapError(unwrappedError);
```
aeisenberg, line 60: Are there any known user errors that we should ignore? Eg- trying to resolve an unknown language? Trying to avoid spurious triggers of our SLOs.

CHANGELOG.md
henrymercer: Alternatively, we could omit the implementation altogether

```suggestion
- Experimental: add a new `resolve-environment` action which attempts to infer a configuration for the build environment that is required to build a given project. Do not use this in production as it is part of an internal experiment and subject to change at any time.
```

pr-checks/checks/resolve-environment-action.yml
```diff
@@ -0,0 +1,17 @@
+name: "resolve-environment-action"
```
henrymercer, line 1: Minor, for consistency:

```suggestion
name: "Resolve environment"
```

resolve-environment/action.yml
```diff
@@ -0,0 +1,23 @@
+name: 'CodeQL: Resolve Build Environment'
+description: '[Experimental] Attempt to infer a suitable environment configuration for the autobuilder'
```
henrymercer, line 2:
```suggestion
description: '[Experimental] Attempt to infer a build environment suitable for automatically building the repo'
```

src/codeql.ts
```diff
@@ -229,6 +236,15 @@ export interface ResolveQueriesOutput {
   };
 }
 
+export interface ResolveBuildEnvironmentOutput {
+  configuration?: {
+    [language: string]: {
+      os?: unknown;
```
henrymercer, line 242: We're not using this, so let's avoid introducing a dependency on the format here.

```suggestion
```

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const ACTION_NAME = "resolve-environment";
+
+async function run() {
+  const startedAt = new Date();
+  const logger = getActionsLogger();
+  const language: Language = resolveAlias(getRequiredInput("language"));
+
+  try {
+    const workflowErrors = await validateWorkflow(logger);
```
henrymercer, line 26: We'll have already called `init` here, which will have already validated the workflow, so I think we can remove this.

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,88 @@
...
+      language
+    );
+    core.setOutput("environment", result);
+  } catch (unwrappedError) {
+    const error = wrapError(unwrappedError);
+    core.setFailed(error.message);
+    await sendStatusReport(
+      await createStatusReportBase(
+        ACTION_NAME,
+        "aborted",
```
henrymercer, line 65: We use `aborted` for cases where we didn't manage to download the CodeQL CLI, and `getActionsStatus` elsewhere.  This lets us distinguish between user errors and non-user errors in telemetry.

```suggestion
        getActionsStatus(error),
```

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,88 @@
...
+    const result = await runResolveBuildEnvironment(
+      config.codeQLCmd,
+      logger,
+      workingDirectory,
+      language
+    );
+    core.setOutput("environment", result);
+  } catch (unwrappedError) {
+    const error = wrapError(unwrappedError);
+    core.setFailed(error.message);
```
henrymercer, line 61:
```suggestion
    core.setFailed(`Failed to resolve a build environment suitable for automatically building your code. ${error.message}`);
```

src/resolve-environment.ts
```diff
@@ -0,0 +1,18 @@
...
+  cmd: string,
+  logger: Logger,
+  workingDir: string | undefined,
+  language: Language
+) {
+  logger.startGroup(
+    `Attempting to resolve build environment for ${language} in ${workingDir}`
+  );
+  const codeQL = await getCodeQL(cmd);
+  const result = await codeQL.resolveBuildEnvironment(workingDir, language);
```
henrymercer, line 15: Nit: we already have some inconsistency in the repo, but `codeql` is more common.

```suggestion
  const codeql = await getCodeQL(cmd);
  const result = await codeql.resolveBuildEnvironment(workingDir, language);
```

src/resolve-environment.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Logger } from "./logging";
+
+export async function runResolveBuildEnvironment(
+  cmd: string,
+  logger: Logger,
+  workingDir: string | undefined,
+  language: Language
+) {
+  logger.startGroup(
+    `Attempting to resolve build environment for ${language} in ${workingDir}`
```
henrymercer, line 12: Handle the case where `workingDir` is `undefined` here.

henrymercer (commented):
Nice!  Some minor suggestions to improve this before we merge.

mbg:
In addition to the basic integration test included in this PR, we also have an internal repo with a more comprehensive test matrix which covers most cases that the Go autobuilder distinguishes between when resolving build environments. That forms an end-to-end test for all the new components related to this work.

src/resolve-environment-action.ts
mbg: The CLI returns the following JSON if it cannot find the `identify-environment` script for a language pack:
```json
{ "configuration": {} }
```
The lack of a language key signals that no language pack was successfully queried for the build environment. 

I'd advocate that we just return an empty JSON result here if we can't run the CLI:
```json
{}
```
With the meaning of: no `"configuration"` key means we didn't run the CLI. We can also emit a warning to the log. My reasoning is as follows:

- If we are running with an older CLI version that doesn't support the command, then we want to get as close to the old workflow behaviour as possible. 
- We do not want the new Action to stop a workflow run.
- Returning an empty JSON object provides no instructions for how to configure the build environment.
- All our new workflow steps are guarded by checks for specific keys in the configuration. If no key is present, no additional steps are run. 
- Not returning a JSON value means we have to deal with the special case where trying to parse what is expected to be a JSON output may fail / guard against this.

src/resolve-environment-action.ts
mbg: I think this will probably get addressed by the proposed handling of CLI errors in general: https://github.com/github/codeql-action/pull/1684#discussion_r1229453530

src/resolve-environment-action.ts
henrymercer: Makes sense 👍 It's probably more convenient for callers to account for the output being `{}` rather than needing to version check the CodeQL CLI before calling the Action.

mbg: I have pushed a commit which implements this and that also addresses @aeisenberg's remaining comment, I think.

src/resolve-environment-action.ts
```diff
@@ -0,0 +1,98 @@
...
+  } catch (unwrappedError) {
+    const error = wrapError(unwrappedError);
+
+    if (error instanceof CommandInvocationError) {
+      // If the CLI failed to run successfully for whatever reason,
+      // we just return an empty JSON object and proceed with the workflow.
+      core.setOutput(ENVIRONMENT_OUTPUT_NAME, {});
+      logger.warning(
+        `Failed to resolve a build environment suitable for automatically building your code. ${error.message}`
+      );
```
henrymercer, line 63: A common case here might be running with a CLI that doesn't support the `codeql resolve build-environment` command.  To produce a better error message, I'd recommend checking the version of the CLI and tailoring the error accordingly (preferably avoiding calling the CLI command at all if the CLI is too old).  Check out the section of code near `CODEQL_VERSION_INIT_WITH_QLCONFIG` in `codeql.ts` for examples of how we do this for other commands.

henrymercer (commented):
One improvement around better CLI error handling for old CLIs, otherwise this LGTM.

henrymercer (dismissed):
Checks are currently failing since `nightly-latest` doesn't have a version number of 2.13.4+ yet.  I suggest we point to the new 2.13.4 bundle directly (`stable-v2.13.4`) or alternatively wait until https://github.com/github/codeql-action/pull/1721 is merged and use `latest` in the checks.

henrymercer (approved):
Thanks for addressing the feedback on the naming of the check step!  This LGTM :shipit:

PR #1685, opened 2023-05-18
Send `job_run_uuid` to status report telemetry

angelapwen:
This change generates a `job_run_uuid` in the `init` Action, and then exports it as an environment variable to propagate it across steps. When any status report is created, the value is read from the environment variable and then sent to our internal telemetry API. This will allow us to join status reports from various steps to one another without using multiple fields.

The PR also tests the telemetry fields in `statusReportBase` are assigned appropriately and the proper type (eg. `string` for the `job_run_uuid`). 

[GitHub internal only]: I validated against this branch, with a logging statement, on this [job](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/actions/runs/5017824320/jobs/8996644547). Note that in a single attempt, the `job_run_uuid` logged is the same across steps, and in each attempt it is unique. 


src/actions-util.ts
```diff
@@ -411,6 +413,7 @@ export async function createStatusReportBase(
 ): Promise<StatusReportBase> {
   const commitOid = getOptionalInput("sha") || process.env["GITHUB_SHA"] || "";
   const ref = await getRef();
+  const jobRunUUID = process.env[sharedEnv.JOB_RUN_UUID] || "";
```
angelapwen, line 416: The other environment variable constants use a blank string if the variable is not found so I have followed that pattern, but this variable is a bit different in that it is not an Actions default environment variable or user input. 

I wonder if it makes sense to use a more descriptive string such as `job_run_uuid_not_found` or something along those lines.

adityasharad: Blank is okay IMO, it is easy to notice when looking at the data that the field is blank/missing in rows where we expect it to be present.

src/init-action.ts
```diff
@@ -212,6 +214,8 @@ async function run() {
     logger
   );
 
+  core.exportVariable(sharedEnv.JOB_RUN_UUID, uuidV4());
```
adityasharad, line 217: For this PR: I think an explanatory comment (here or in the status report field) would be useful. Include:
- why we generate an ID rather than obtaining it from the Actions context/env variables
- what its intended purpose is (identifying reports from different steps in the same workflow job without having to infer from workflow run IDs and configuration parameters)

Not critical for this PR but I'd like to know (and record in a comment somewhere): What do you prefer we do if Actions does begin to expose a job ID? Would we stick with this UUID (keeping all data consistent in format, useful if we want proper distributed traces) or switch to the numeric job ID (more intuitive but one-time switch in data format)?

adityasharad (commented):
Nice, I was pleasantly surprised by how small a change this was. Minor suggestions only.

src/actions-util.ts
adityasharad: Also do we have tests that look at the base fields?

src/init-action.ts
angelapwen: I've added a longer comment on the status report field, as I think it makes more sense there. Regarding if Actions does begin to expose a job ID:
```
   * If and when Actions does expose a unique job ID, we plan to populate
   * this field with the Actions-generated identifier, as it will allow us
   * to more easily join our telemetry data with Actions telemetry tables.
```

I think that having the ability to unify our data across GitHub/Actions telemetry would make it worthwhile to switch over. We could always add a UUID back in the future, if we want to create something closer to distributed traces.

src/actions-util.ts
angelapwen: We do not — I'm working on a follow-up PR that adds a test for `createStatusReportBase` that will also test this field. For now I've just tested this field in the testing repo I linked in the PR description.

angelapwen: Nevermind, I actually finished up the basics of the test and pushed it to this branch 😸

src/init-action.ts
esbena: Sorry for catching this late after the hydro schema merge.

On this topic, the "UUID" status field name can be confusing for both syntactic and semantic reasons.

> If and when Actions does expose a unique job ID

When that happens, the ID is no longer a UUID, but rather some numeric (syntax) ID that is unique within the domain of GitHub Actions workflow job ids (semantics).

Ideally, `job_run_uuid` should be renamed to the less informative `job_run_id`.

angelapwen: That's okay! Thank you for your feedback ✨ 

> When that happens, the ID is no longer a UUID, but rather some numeric (syntax) ID that is unique within the domain of GitHub Actions workflow job ids (semantics).

> Ideally, `job_run_uuid` should be renamed to the less informative `job_run_id`. 

That's a good point that it would be numeric, and the UUID we're emitting now is a string. I'd like to keep this as a string UUID for now, and when GitHub Actions exposes a unique job ID, we can use `job_run_id` as a bigint in the data warehouse, and start querying based off the `job_run_id` rather than `job_run_uuid` at that point. This would keep the two clearly separate. 

Let me know if any concerns on this approach 😄

angelapwen: That's okay! Thank you for your feedback ✨ 

> When that happens, the ID is no longer a UUID, but rather some numeric (syntax) ID that is unique within the domain of GitHub Actions workflow job ids (semantics).

> Ideally, `job_run_uuid` should be renamed to the less informative `job_run_id`. 

That's a good point that it would be numeric, and the UUID we're emitting now is a string. I'd like to keep this as a string UUID for now, and when GitHub Actions exposes a unique job ID, we can use `job_run_id` as a bigint in the data warehouse, and start querying based off the `job_run_id` rather than `job_run_uuid` at that point. This would keep the two clearly separate. 

Let me know if any concerns on this approach 😄

esbena: Agreed. Having two different fields would also make the eventual migration less painful on the Kusto side.

src/actions-util.test.ts
```diff
@@ -265,3 +266,54 @@
...
+    t.assert(typeof statusReport.job_run_uuid === "string");
+    t.assert(statusReport.workflow_run_id === 100);
+    t.assert(statusReport.workflow_run_attempt === 2);
+    t.assert(
+      statusReport.workflow_name === (process.env["GITHUB_WORKFLOW"] || "")
+    );
+    t.assert(statusReport.job_name === (process.env["GITHUB_JOB"] || ""));
+    t.assert(statusReport.analysis_key === "analysis-key");
+    t.assert(statusReport.commit_oid === process.env["GITHUB_SHA"]);
+    t.assert(statusReport.ref === process.env["GITHUB_REF"]);
```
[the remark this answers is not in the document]
angelapwen: Closed alert because we are only using the environment variable in a test.

angelapwen:
I realized we hadn't gotten back to this PR, and resolved a merge conflict with `main`. @adityasharad could you please re-review? Thank you!

src/actions-util.test.ts
```diff
@@ -265,3 +266,54 @@ test("isAnalyzingDefaultBranch()", async (t) => {
...
+
+test("createStatusReportBase", async (t) => {
+  await withTmpDir(async (tmpDir: string) => {
+    setupActionsVars(tmpDir, tmpDir);
+
+    process.env["GITHUB_REF"] = "refs/heads/main";
+    process.env["GITHUB_SHA"] = "a".repeat(40);
+    process.env["GITHUB_RUN_ID"] = "100";
+    process.env["GITHUB_RUN_ATTEMPT"] = "2";
+    process.env["GITHUB_REPOSITORY"] = "octocat/HelloWorld";
```
adityasharad, line 278: These are supposed to be protected environment variables in the Actions runtime.

So two questions:
- Do these assignments actually affect the environment variable?
- If you change the assertions below to compare `statusReport.ref` etc with the constant values, not with `process.env["GITHUB_REF"]`, do the tests still pass?

adityasharad (commented):
Code changes look fine, some questions around comments and tests.

src/actions-util.test.ts
angelapwen: Good point! It seems they actually do affect the environment variable. I changed the assertions to `t.assert(statusReport.ref === "refs/heads/main");` and `t.assert(statusReport.commit_oid === "a".repeat(40));
`for the SHA, and the tests still passed. 

If I don't populate these variables manually, I get 
```
  Rejected promise returned by test. Reason:

  Error {
    message: 'GITHUB_SHA environment variable must be set',
  }
```

errors. I assume that these just aren't populated at all in the unit tests.

src/actions-util.ts
```diff
@@ -315,6 +315,16 @@ export type ActionStatus =
...
 
 export interface StatusReportBase {
+  /** UUID representing the job run that this status report belongs to.
+   * We generate our own UUID here because Actions currently does not expose
+   * a unique job run identifier. This UUID will allow us to more easily
+   * match reports from different steps in the same workflow job.
+   *
+   * If and when Actions does expose a unique job ID, we plan to populate
+   * this field with the Actions-generated identifier, as it will allow us
+   * to more easily join our telemetry data with Actions telemetry tables.
```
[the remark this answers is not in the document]
angelapwen: Done 👍

src/actions-util.test.ts
angelapwen: Following up on this after our 1:1: these tests failed locally because I wasn't in the Actions runtime. When run in Actions, the protected Actions environment variables are able to be overridden for the purpose of the checks, but are not populated across steps and tests. 

@henrymercer and/or @aeisenberg, does this match your understanding of how this unit test works?

henrymercer: That's my understanding.  I would go further and say that we should be running our unit tests starting from an empty environment — our tests shouldn't depend on the environment we run them in.

adityasharad: Ok great, I'm happy enough with this test checking the status report is populated from `process.env`, and we can leave further test cleanup for a future PR:
- starting from clean environment
- passing locally when not running in an Actions environment

adityasharad (approved):

PR #1695, opened 2023-05-23
PR checks: Update requests to 2.31.0

henrymercer:
Update `requests` to 2.31.0 to address https://github.com/advisories/GHSA-j8r2-6x86-q33q.  We don't use this in the Action itself, but instead in PR checks that test Python dependency installation.

@github/codeql-python I'd appreciate it if one of you could take a careful look at this PR to make sure I've preserved test coverage.  Thanks!


henrymercer:
Debug artifact upload failures fixed by #1696.

yoff (approved):
LGTM

(nit: It would have been nice to split the first commit in two so the rewriting of the files was clear from the diff and the moving happened separately.)

henrymercer:
Thanks, that's good feedback.  I had hoped that the diff would recognize the change as a rename!

PR #1696, opened 2023-05-23
Fix Swift PR Checks on `nightly-latest` CLI

angelapwen:
Now that Swift analysis is on by default in the latest nightly CLI version, we are autodetecting Swift, but Swift is not set up in this workflow. This change fixes the Swift setup workflow to unblock CI by:

- adding `setup-swift` workflow, and `CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT` for all versions except `nightly-latest` to the debug artifacts check
- no longer setting the `CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT` flag for the `nightly-latest` CLI version, for any other check. This allows us to test the case where the flag is not set for `nightly-latest` as well. 
- updating the `setup-swift` workflow to only set up Swift if the `codeql resolve languages` includes Swift. 

Once CLI v2.13.3 is released, we can stop setting the experimental flag for `latest`.

### DIFF
--- a/.github/actions/setup-swift/action.yml
+++ b/.github/actions/setup-swift/action.yml
@@ -1,5 +1,5 @@
 name: "Set up Swift"
-description: Sets up an appropriate Swift version if Swift is enabled via CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT.
+description: Sets up an appropriate Swift version if supported.
 inputs:
   codeql-path:
     description: Path to the CodeQL CLI executable.
@@ -9,24 +9,28 @@ runs:
   steps:
     - name: Get Swift version
       id: get_swift_version
-      if: env.CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT == 'true'
       shell: bash
       env: 
         CODEQL_PATH: ${{ inputs.codeql-path }}
       run: |
-        if [ $RUNNER_OS = "macOS" ]; then
+        if [[ $RUNNER_OS = "macOS" ]]; then
           PLATFORM="osx64"
         else # We do not run this step on Windows.
           PLATFORM="linux64"
         fi 
         SWIFT_EXTRACTOR_DIR="$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')"
-        VERSION="$("$SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor" --version | awk '/version/ { print $3 }')"
-        # Specify 5.7.0, otherwise setup Action will default to latest minor version. 
-        if [ $VERSION = "5.7" ]; then
-          VERSION="5.7.0"
+        if [ $SWIFT_EXTRACTOR_DIR = "null" ]; then
+          VERSION="null"
+        else
+          VERSION="$("$SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor" --version | awk '/version/ { print $3 }')"
+          # Specify 5.7.0, otherwise setup Action will default to latest minor version. 
+          if [ $VERSION = "5.7" ]; then
+            VERSION="5.7.0"
+          fi
         fi
         echo "version=$VERSION" | tee -a $GITHUB_OUTPUT
+
     - uses: swift-actions/setup-swift@65540b95f51493d65f5e59e97dcef9629ddf11bf # Please update the corresponding SHA in the CLI's CodeQL Action Integration Test.
-      if: env.CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT == 'true'
+      if: steps.get_swift_version.outputs.version != null
       with:
         swift-version: "${{ steps.get_swift_version.outputs.version }}"
--- a/.github/actions/setup-swift/action.yml
+++ b/.github/actions/setup-swift/action.yml
@@ -9,6 +9,7 @@ runs:
   steps:
     - name: Get Swift version
       id: get_swift_version
+      if: runner.os != 'Windows'
       shell: bash
       env: 
         CODEQL_PATH: ${{ inputs.codeql-path }}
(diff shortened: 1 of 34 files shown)


henrymercer (dismissed):

angelapwen:
Oh no, the step is now failing on past CLI versions. I'll simply pass the environment variable to the failing test then, and we can do The Right Thing when we remove the experimental flag for new versions.

henrymercer (commented):
How about:

- Updating the autogenerated "Set environment variable for Swift enablement" step to stop setting `CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT=true` on `nightly-latest`
- Updating `.github/actions/setup-swift` to run Swift setup if we're running on macOS/Linux and `swift` is included in `codeql resolve languages` (Note that based on [this](https://github.com/github/codeql/blob/8daa8d78ff7f1769749d21c6215b37678aaaea99/docs/codeql/reusables/supported-versions-compilers.rst#id10), Linux analysis requires 5.7.3)

This should handle the case where Swift is enabled by default in `nightly-latest`.  Once 2.13.3 is released, we can remove `latest` from the list in "Set environment variable for Swift enablement".

.github/actions/setup-swift/action.yml
```diff
@@ -1,5 +1,5 @@
 name: "Set up Swift"
-description: Sets up an appropriate Swift version if Swift is enabled via CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT.
+description: Sets up an appropriate Swift version if supported.
```
aeisenberg, line 2:
```suggestion
description: Sets up an appropriate Swift version if supported on this platform.
```

pr-checks/sync.py
```diff
@@ -83,8 +83,7 @@ def writeHeader(checkStream):
                     matrix.version == '20220908' ||
                     matrix.version == '20221211' ||
                     matrix.version == 'cached' ||
-                    matrix.version == 'latest' ||
-                    matrix.version == 'nightly-latest'
+                    matrix.version == 'latest'
```
aeisenberg, line 86: Is there a reason that we are no longer running on nightly-latest? Is this temporary? It seems to me that running on nightly-latest is a good early warning sign.

.github/workflows/debug-artifacts.yml
```diff
@@ -54,6 +54,10 @@ jobs:
           debug: true
           debug-artifact-name: my-debug-artifacts
           debug-database-name: my-db
+      - uses: ./../action/.github/actions/setup-swift
+        if: matrix.version == 'nightly-latest'
+        with:
+          codeql-path: ${{ steps.init.outputs.codeql-path }}
```
henrymercer, line 60: Nice, I think we just need to add `id: init` to the init step to make this work.

henrymercer (commented):
Thanks for making the changes, I think it made sense to do the work of updating the PR checks now.  I think the only remaining piece is to update `multi-language-autodetect.yml` to check language autodetection for Swift on `nightly-latest` too.  This is the only remaining read of `CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT` I found in the codebase.

pr-checks/sync.py
aeisenberg: Oh...right. I didn't expand this to see the previous lines. You're entirely correct.

.github/workflows/debug-artifacts.yml
angelapwen: Ah... I was wondering what the failure was coming from. Thanks.

angelapwen:
> I think the only remaining piece is to update multi-language-autodetect.yml to check language autodetection for Swift on nightly-latest too. This is the only remaining read of CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT I found in the codebase. 

Nice find — I've modified it to check on `nightly-latest` when the platform isn't Windows, in addition to the experimental flag.

.github/actions/setup-swift/action.yml
```diff
@@ -9,24 +9,29 @@ runs:
...
           PLATFORM="osx64"
         else # We do not run this step on Windows.
           PLATFORM="linux64"
         fi 
         SWIFT_EXTRACTOR_DIR="$("$CODEQL_PATH" resolve languages --format json | jq -r '.swift[0]')"
-        VERSION="$("$SWIFT_EXTRACTOR_DIR/tools/$PLATFORM/extractor" --version | awk '/version/ { print $3 }')"
-        # Specify 5.7.0, otherwise setup Action will default to latest minor version. 
-        if [ $VERSION = "5.7" ]; then
-          VERSION="5.7.0"
+        if [ $SWIFT_EXTRACTOR_DIR = "null" ]; then
```
henrymercer, line 23: Not critical here, but for the future we should quote variables to prevent problems with word splitting.

.github/actions/setup-swift/action.yml
```diff
@@ -9,24 +9,29 @@ runs:
...
     - name: Get Swift version
       id: get_swift_version
-      if: env.CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT == 'true'
+      if: runner.os != 'Windows'
       shell: bash
       env: 
         CODEQL_PATH: ${{ inputs.codeql-path }}
       run: |
-        if [ $RUNNER_OS = "macOS" ]; then
+        if [[ $RUNNER_OS = "macOS" ]]; then
```
henrymercer, line 17: Nit: we should probably pick one of (single brace, double brace) and (single equals, double equals) and run with it.  I think for our purposes here they will behave the same though.

henrymercer (approved):

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -66,7 +66,9 @@ steps:
       fi
 
   - name: Check language autodetect for Swift
-    if: env.CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT == 'true'
+    if: >-
```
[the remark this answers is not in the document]
angelapwen: Oh yes, I noticed this as well. I wonder if it's because we had previously set the `CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT` variable globally, but then moved it individually to specific steps. I'll open a follow-up PR.

.github/actions/setup-swift/action.yml
angelapwen: I agree! I actually changed it to double brace here to match the other instances of `$RUNNER_OS`. 

I think the bash I've seen so far in the workflow uses single equals for equality, but the Actions syntax outside of the bash scripts use double equals. That's what I've been following at least 😅

pr-checks/checks/multi-language-autodetect.yml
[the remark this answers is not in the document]
angelapwen: PR at https://github.com/github/codeql-action/pull/1699

PR #1712, opened 2023-05-31
Remove unused `CODESCANNING_EVENT_NAME` environment variable

henrymercer:
This environment variable is not used by default setup, so here we remove the Action's support for it to avoid potential confusion around what `workflowEventName()` would produce in default setup workflow runs. 


src/actions-util.ts
```diff
@@ -576,21 +576,18 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
+ * This will be "dynamic" for default setup workflow runs.
+ */
+export function getWorkflowEventName() {
+  return getRequiredEnvParam("GITHUB_EVENT_NAME");
 }
 
 // Was the workflow run triggered by a `push` event, for example as opposed to a `pull_request` event.
 function workflowIsTriggeredByPushEvent() {
-  return workflowEventName() === "push";
+  return getWorkflowEventName() === "push";
```
aeisenberg, line 590: Aren't default setup workflows fun on `push` and `schedule` implicitly? If so, how do we determine when this is happening?

aeisenberg: Thanks for the explanation. Maybe you could update the comment on line 588 since the name of this function is no longer 100% correct. Or are you suggesting removing this function and inlining all uses of it?

henrymercer: I was thinking of removing it and inlining its one use — as you say the name's a bit confusing and I'd rather not rely  on the JSDoc.

aeisenberg (approved):

henrymercer:
I'll wait until we've done the Swift release before merging this out of an abundance of caution.

adityasharad:
Just to be sure: this was never used in a shipped GHES version?

marcogario (approved):

PR #1717, opened 2023-06-05
Fix changelog for 2.3.6

henrymercer:
One of the changelog entries for 2.3.6 ended up incorrectly being associated with 2.3.4 due to a bad merge conflict resolution.  As a long term fix, we could store unreleased changelog entries in a separate directory and only edit `CHANGELOG.md` with automation.  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ No user facing changes.
 
 ## 2.3.6 - 01 Jun 2023
 
-No user facing changes.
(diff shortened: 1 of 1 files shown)


adityasharad (approved):
Separately tracking unreleased entries is a good idea, though perhaps confusing for OSS contributors. Let's figure out how to make that work and minimise conflicts.

angelapwen (approved):
Thanks for catching. I'd noticed the same happening in the 2.3.5 release as well (a change from 2.3.5 was added to 2.3.3 as a result of a merge conflict resolution).

PR #1718, opened 2023-06-05
PR checks: stop setting experimental Swift var for new CLI versions

angelapwen:
Now that `latest` and `cached` are both 2.13.3, which is the version in which we GA'ed Swift, we should stop setting this experimental variable when we test these CLI versions so we can test the case where the variable is unset.


henrymercer (approved):
Thanks!  Separately, we might want to rename `cached` to something like `ghes-default` — on Dotcom with controlled switchover we will follow the feature flags rather than what's in the toolcache.

PR #1741, opened 2023-06-28
Send new per-query alert count event reports for QA telemetry

angelapwen:
This change sends a list of event reports representing the per-query alert counts from a run, appended as a field to the `analyze` status report, when the `qa_telemetry` feature flag is set. The `event_report` schema in the data warehouse is mimicked in the `EventReport` interface here.

We generate an event report just after `codeql database-interpret results` is run for each language, so the event reports are language-specific. The `properties` field of the event report is an object representing the per-query alert counts processed from a single SARIF file. 

I have manually tested this by setting the environment variable `CODEQL_ACTION_QA_TELEMETRY: true` in a test run ([GitHub internal only link](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/actions/runs/5402843513/jobs/9814910935#step:4:3663)) to confirm that the event report and alert counts are being generated appropriate. I have also modified the existing `runQueries` test to confirm that the appropriate `event_report` status report field is set with event property `codeql database interpret-results`.


src/analyze.ts
```diff
@@ -392,6 +416,38 @@ export async function runQueries(
...
 
+  /** Get an object with all queries and their counts parsed from a SARIF file path. */
+  function getPerQueryAlertCounts(
+    sarifPath: string,
+    log: Logger
+  ): Record<string, number> {
+    validateSarifFileSchema(sarifPath, log);
+    const sarifObject = JSON.parse(
+      fs.readFileSync(sarifPath, "utf8")
+    ) as util.SarifFile;
```
henrymercer, line 427: Trying to load the whole SARIF file in memory can cause us to exceed the max length of a string in Node.js on certain repositories.  Even if we only enable the feature flag for the QA org, this might cause some repositories to start failing in QA.  I'd recommend looking into doing this step with a streaming JSON parser.

src/analyze.ts
```diff
@@ -392,6 +416,38 @@ export async function runQueries(
...
+    // All rules (queries), from all results, from all runs
+    for (const sarifRun of sarifObject.runs) {
+      if (sarifRun.results) {
+        for (const result of sarifRun.results) {
+          if (result.ruleId) {
+            if (result.ruleId in perQueryAlertCounts) {
+              perQueryAlertCounts[result.ruleId] =
+                perQueryAlertCounts[result.ruleId] + 1;
+            } else {
+              perQueryAlertCounts[result.ruleId] = 1;
```
henrymercer, line 442:
```suggestion
            perQueryAlertCounts[result.ruleId] = (perQueryAlertCounts[result.ruleId] || 0) + 1;
```

src/analyze.ts
```diff
@@ -392,6 +416,38 @@ export async function runQueries(
...
+    // We do not need to compute fingerprints because we are not sending data based off of locations.
+
+    // Generate the query: alert count object
+    const perQueryAlertCounts: Record<string, number> = {};
+
+    // All rules (queries), from all results, from all runs
+    for (const sarifRun of sarifObject.runs) {
+      if (sarifRun.results) {
+        for (const result of sarifRun.results) {
+          if (result.ruleId) {
```
henrymercer, line 437: Eventually we may want to stop populating this field in favour of a reporting descriptor reference object `result.rule`, so I suggest we switch to `result.rule?.id` ([more info](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317645)).

src/analyze.ts
angelapwen: I'd thought this as well. It turns out that we already read the entire SARIF file into memory in `uploadLib`, even several times for a single run:

Validating the SARIF — https://github.com/github/codeql-action/blob/cff3d9e3c9a51dfcf1ea2692cd1cd876dda6d71e/src/upload-lib.ts#L227

Combining SARIFs into 1 — https://github.com/github/codeql-action/blob/cff3d9e3c9a51dfcf1ea2692cd1cd876dda6d71e/src/upload-lib.ts#L30

Counting the number of results — https://github.com/github/codeql-action/blob/cff3d9e3c9a51dfcf1ea2692cd1cd876dda6d71e/src/upload-lib.ts#L205

Since it seems we already read SARIF into memory for uploading, I thought I could address this as a follow-up. It's true that currently DCA/QA does not hit the upload steps and therefore SARIF parsing at all, though, so we don't necessarily have evidence that QA will _not_ fail on any repositories. If it does, though, we can always opt out those repos on a per-repo level from the feature flag until the parsing is implemented. What do you think?

henrymercer: Gotcha, so we'll hit this in the `upload: true` path anyway.  In that case, I think we're good to go for QA, since QA is intended to replicate the code scanning experience, and that experience would fail in this case for typical runs using `upload: true`. FYI @esbena

src/analyze.ts
angelapwen: Thanks, done! (though the [latest commit](https://github.com/github/codeql-action/compare/main...angelapwen:codeql-action:per-query-alert-counts) seems to be delayed in showing up on this PR..)

src/analyze.ts
angelapwen: That makes sense 👍 For now I've changed it to check `result.rule?id` and if that's undefined, `result.ruleId` based off of this line in the link you sent: 

```If rule.id (§3.52.4) is absent, it SHALL default to thisObject.ruleId. If rule.id and thisObject.ruleId are both present, they SHALL be equal.```

If we stop populating the field, we'll want to in any case scrub through the codebase and tests to change all existing references of `ruleId`

**Review by henrymercer — APPROVED** [review]

**Comment by angelapwen** [comment]
I've set the feature to `true` in the existing `runQueries` test, and confirmed that the appropriate `event_report` status report field is set with event property `codeql database interpret-results`.

**Review by henrymercer — APPROVED** [review]

**Comment by esbena** [comment]
> and confirmed that the appropriate event_report status report field

I can see that in your test run. But it does not look like the data enters the database. Is that intentional?

```
database("hydro").code_scanning_v0_status_report
| where timestamp between (now(-30d) .. now())
| summarize count() by array_length(event_reports)
```
Results in 34035006 reports with length 0 event_reports, and nothing else.

angelapwen:
@esbena Thank you for pointing this out. I looked at the debug logs of the run more closely and it looks like the request returned an error because I had sent the date/times for the `event_report` with the incorrect format. I'll put up a new PR, start a test run, and check the data warehouse for the report.

PR #1757, opened 2023-07-06
Add a deprecation warning for CodeQL CLIs < 2.9.4

henrymercer:
Add a deprecation warning for customers using the now deprecated 2.8.6–2.9.3 range of CLIs that these CLIs will not be supported by the next version of the Action:

image

In the future, we can integrate this into our GHES compatibility automation to automatically add deprecation warnings once a CLI is deprecated.

Commit-by-commit review recommended, as the second commit refactors how we represent environment variables in the Action.


src/codeql.ts
```diff
@@ -1032,6 +1038,22 @@ export async function getCodeQLForCmd(
...
       `Expected a CodeQL CLI with version at least ${CODEQL_MINIMUM_VERSION} but got version ${await codeql.getVersion()}`
     );
+  } else if (
+    checkVersion &&
+    !(await util.codeQlVersionAbove(codeql, CODEQL_NEXT_MINIMUM_VERSION))
+  ) {
+    core.warning(
+      `CodeQL CLI version ${await codeql.getVersion()} was deprecated on 2023-06-20 alongside ` +
+        "GitHub Enterprise Server 3.5 and will not be supported by the next release of the " +
+        "CodeQL Action. Please update to a newer version of the CodeQL CLI " +
```
angelapwen, line 1050: Maybe "Please update to a version of the CodeQL CLI newer than..." as it can be any version greater than this one?

src/environment.ts
```diff
@@ -0,0 +1,60 @@
+export enum EnvVar {
```
angelapwen, line 1: I added some of these myself, but now that I'm looking at them all together I'm having a hard time figuring out why some are prepended with `CODEQL_ACTION_` and others aren't. I guess it would be difficult to change them as customers might be using those environment variables though.

src/codeql.ts
```diff
@@ -1054,6 +1056,8 @@ export async function getCodeQLForCmd(
         "'github/codeql-action/*@v2.20.4' in your code scanning workflow to ensure you continue " +
         "using this version of the CodeQL Action."
     );
+    process.env[EnvVar.SUPPRESS_DEPRECATED_SOON_WARNING] = "true";
+    core.exportVariable(EnvVar.SUPPRESS_DEPRECATED_SOON_WARNING, "true");
```
angelapwen, line 1060: Do we need both of these lines? I thought the `core.exportVariable()` one would suffice.

angelapwen (approved):

src/codeql.ts
henrymercer: Ah nice, I thought `core.exportVariable` was a direct analogy to `echo "a=b" >> $GITHUB_ENV` but it turns out it sets the variable in the current step too.

src/environment.ts
henrymercer: Some are used in other parts of CodeQL too, but others are just inconsistent.  You're right that it's unlikely to be worth changing these.

src/codeql.ts
henrymercer: I think it's clearer to show the minimum version (>= rather than >), but I agree we could phrase this better.

angelapwen: Ah yes, agree that >= is clearer!

angelapwen (approved):

angelapwen (approved):

PR #1758, opened 2023-07-06
Skip an API call when downloading a CodeQL version with a semver bundle

henrymercer:
Previously, we made an API call to look up the CLI version to release mapping when the default CLI version was requested on Dotcom and that CLI wasn't in the toolcache.  Now we compute the tag name directly for semver bundles.


src/codeql.ts
```diff
@@ -315,6 +315,11 @@ export const CODEQL_VERSION_INIT_WITH_QLCONFIG = "2.12.4";
  */
 export const CODEQL_VERSION_RESOLVE_ENVIRONMENT = "2.13.4";
 
+/**
+ * Versions 2.13.4+ of the CodeQL CLI had a semantically versioned CodeQL Bundle.
```
angelapwen, line 319:
```suggestion
 * Versions 2.13.4+ of the CodeQL CLI have a semantically versioned CodeQL Bundle.
```

angelapwen: Super nit 😸

angelapwen (dismissed):
✨ great!

angelapwen (approved):

PR #1768, opened 2023-07-11
Telemetry: Mark configuration errors as user errors

henrymercer:
This will help us better identify CodeQL Action bugs by enabling us to create more precise monitors.


src/config-utils.test.ts
```diff
@@ -230,11 +230,11 @@ test("load input outside of workspace", async (t) => {
         createFeatures([]),
         getRunnerLogger(true)
       );
-      throw new Error("initConfig did not throw error");
+      throw new UserError("initConfig did not throw error");
```
angelapwen, line 233: Hm... should this test simply throw a normal `Error`? I'm not clear on why the tests should throw a `UserError`.

(Same comment on all instances of `initConfig did not throw error` below)

angelapwen (approved):
Thanks! Just a quick question.

src/config-utils.test.ts
henrymercer: Good catch — these should indeed be `Error`s!

angelapwen (approved):

PR #1770, opened 2023-07-11
Telemetry: Nest `alertCounts` object in `event_report.properties`

angelapwen:
Previously, for QA runs, we were sending the map of alerts to counts as a top-level object in the event report `properties` bag. Because we eventually would like to send other (different) properties attached to events, let's nest them under an `alertCounts` object in `properties`.

This change will take effect before the next QA run, for v2.14.1. 

[Internal only]: [Test run with the new format](https://github.com/dsp-testing/angelapwen-codeql-test-juice-shop/actions/runs/5520260560/jobs/10066777295#step:4:3615). I have also confirmed by querying the data warehouse, which I will link in backlinked issue. 


angelapwen:
I'll hold off on merging until the current QA run is confirmed to have completed!

PR #1771, opened 2023-07-11
Bump semver

henrymercer:
Bumps semver to fixed versions (see https://github.com/advisories/GHSA-c2qf-rxjj-qqgw).  Use overrides to address dependencies on semver that come from packages that haven't been updated yet.  This is a precaution as the attack vector here (ReDoS) is likely not relevant to the Action.


package.json
```diff
@@ -75,6 +75,18 @@
...
+    },
+    "@actions/tool-cache": {
+      "semver": ">=6.3.1"
+    },
+    "eslint-plugin-import": {
+      "semver": ">=6.3.1"
+    },
+    "eslint-plugin-jsx-a11y": {
+      "semver": ">=6.3.1"
+    },
```
aeisenberg, line 89: Can you explain why this isn't sufficient? 

```suggestion
    "semver": "^7.5.4",
```

henrymercer: It's not possible to specify an override for a direct dependency.  Furthermore, I'd prefer to force upgrade dependencies to the same major version, to reduce the chances of breaking something.

henrymercer: I'd prefer to force upgrade dependencies to the same major version, to reduce the chances of breaking something.

aeisenberg: thanks.

aeisenberg (approved):
Deleting 16k lines is not bad either.

PR #1795, opened 2023-07-21
Improve handling of fatal CLI errors

henrymercer:
Currently we are losing information about CLI errors in `database {finalize,interpret-results,run-queries}`.  When the CLI command includes "A fatal error occurred" in its stderr, we lose other context about the error (such as whether it was an OOM, or out of disk, or something else) in the Actions run summary and in telemetry.

Furthermore, the Action is suppressing the improved errors for no code found shipped in CodeQL v2.12.4.

This PR stops suppressing these errors by displaying fatal errors in full, and using the CLI error message for `database finalize` no code found errors for CodeQL CLI v2.12.4 and later.  Most of the commits are tweaking the error message to try to improve its utility.  In particular, we're quite keen to get it on a single line if we can, as this lets us include all the useful detail we can on the Actions run summary page.


src/codeql.ts
```diff
@@ -1138,11 +1153,47 @@ async function runTool(
...
   }).exec();
-  if (exitCode !== 0)
+  if (exitCode !== 0) {
+    error = extractFatalErrors(error) || error;
     throw new CommandInvocationError(cmd, args, exitCode, error, output);
+  }
   return output;
 }
 
+function extractFatalErrors(error: string): string | undefined {
```
aeisenberg, line 1163: I'm finding it a little hard to understand how this method transforms error messages. Can you add a test that shows some sort of reasonable error message and the output that it should be?

henrymercer: Thanks, that's a good point.  The tests I've added to `src/codeql.test.ts` test this functionality through `runTool`, but I'll add some JSDoc too.

aeisenberg: Sorry...I missed the tests. They put things into context. JSDoc would be helpful, too.

aeisenberg (approved):

PR #1796, opened 2023-07-21
Scale the amount of reserved RAM on large runners only

henrymercer:
Initially we planned to roll this out to all runners, however we don't currently have the right data to make a decision on the tradeoff between the slowdown and the reduced frequency of OOMs.  Therefore, let's roll this out to large runners only where we know the tradeoff is worth it.

This PR also improves the unit tests to check combinations of platform and total RAM without needing to run on different machines.  Commit-by-commit review recommended to see examples of the scaling reserved RAM change in the tests.


nickrolfe (approved):
This looks sensible.

PR #1809, opened 2023-07-27
Add support for the code_scanning_codeql_java_lombok feature flag

igfoo:
I think the `minimumVersion` could be older, but I don't think it's worth the time to work out exactly how old.


src/feature-flags.ts
```diff
@@ -58,6 +59,11 @@ export const featureConfig: Record<
   Feature,
   { envVar: string; minimumVersion: string | undefined; defaultValue: boolean }
 > = {
+  [Feature.CodeqlJavaLombokEnabled]: {
+    envVar: "CODEQL_JAVA_LOMBOK_ENABLED",
```
angelapwen, line 63: For consistency with the others environment variable overrides for the feature flag! 
```suggestion
    envVar: "CODEQL_JAVA_LOMBOK",
```

angelapwen (dismissed):
:shipit:

angelapwen (dismissed):
That should do it 😸

angelapwen (approved):

src/init-action.ts
```diff
@@ -342,6 +342,20 @@ async function run() {
       core.exportVariable("CODEQL_EXTRACTOR_JAVA_AGENT_DISABLE_KOTLIN", "true");
     }
 
+    if (await features.getValue(Feature.CodeqlJavaLombokEnabled, codeql)) {
+      logger.info("Enabling CodeQL Java Lombok support");
```
henrymercer, line 346: Currently we'll print this even if the user isn't analyzing Java, which might be a bit confusing.  Could we run this later in the job, or alternatively change the message?

angelapwen: Oh yes, good point. We could also check the feature flag only if one of the languages to be analyzed is Java, similar to what's done for the python dependencies check: https://github.com/github/codeql-action/blob/de6681ceb7967227f8760bfad7bde8540662f894/src/init-action.ts#L276

henrymercer (dismissed):

henrymercer (approved):

PR #1815, opened 2023-07-28
Simplify README to recommend default setup and refer to docs

henrymercer:
- Recommend default setup by default :)
- Refer to the docs for setting up advanced setup, rather than recommending a multi-language DB build.  Customers will get a better experience setting up a matrixed build for repos with multiple languages.
- Configuration information re the config file and the config input is now included in the docs.  Refer to the docs rather than duplicating that information here.


adityasharad (approved):

angelapwen (approved):
Also looks good to me ✨

felicitymay (approved):
The changes look good to me. Thanks for the ping ✨ 

We're going to be changing some of these docs URLs, but we'll have redirects in place as usual for the short term.

I'll create an issue for us to update the URLs once the next stage of the docs refactoring is complete.

PR #1816, opened 2023-07-28
Categorize more user errors correctly in telemetry

henrymercer:
Categorizing user errors correctly in telemetry helps us detect and respond to incidents faster.  This PR starts reporting two user errors as user errors in telemetry:

- Not specifying a required input for an Action
- Submitting an advanced CodeQL analysis for a repo that has default setup enabled


src/actions-util.ts
```diff
@@ -24,7 +24,11 @@ const pkg = require("../package.json") as JSONSchemaForNPMPackageJsonFiles;
  * This allows us to get stronger type checking of required/optional inputs.
  */
 export const getRequiredInput = function (name: string): string {
-  return core.getInput(name, { required: true });
+  const value = core.getInput(name);
+  if (!value) {
```
adityasharad, line 28: Does this throw on a provided input with a falsy value? (I think that problem also exists currently, looking at the implementation of `getInput`.)

src/upload-lib.ts
```diff
@@ -488,6 +489,14 @@ export async function waitForProcessing(
   }
 }
 
+function areProcessingErrorsUserError(processingErrors: string[]): boolean {
```
adityasharad, line 492: Confusing name but hard to find a better one. How about `shouldConsiderAsUserError`?

src/upload-lib.ts
```diff
@@ -488,6 +489,14 @@ export async function waitForProcessing(
...
+  return (
+    processingErrors.length === 1 &&
+    processingErrors[0] ===
+      "CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled"
```
adityasharad, line 496: I am a bit worried about this breaking if the upstream text changes. Do we get any richer data (eg response code) from the API response?

src/actions-util.ts
henrymercer: Here's the current implementation:

https://github.com/actions/toolkit/blob/91d3933eb52b351f437151400a88ba7d57442a9b/packages/core/src/core.ts#L127

My understanding is:

- Action inputs are stored as environment variables
- `process.env[x]` is a string or undefined
- For strings, everything is truthy apart from `""`, which counts as not specifying the input

src/upload-lib.ts
henrymercer: I think that's better.  It's not clear it's about processing errors from the function name, but I think it's sufficient to provide that context in the JSDoc.

src/upload-lib.ts
henrymercer: I don't believe so — they're marked as unrecoverable delivery errors, but some unrecoverable delivery errors are user errors and others aren't.  Given the impact of this breaking is a potential monitor false alarm, I'm happy with relying on this.

adityasharad: Good enough. Let's ask our colleagues to let us know if they change it.

adityasharad (approved):

PR #1817, opened 2023-07-31
Use feature flag to enable evaluator intra-layer parallelism

nickrolfe:


angelapwen (approved):
Looks good to me 🎏

src/codeql.ts
```diff
@@ -778,6 +780,14 @@ export async function getCodeQLForCmd(
...
         codeqlArgs.push(querySuitePath);
       }
+      if (
+        await features.getValue(
+          Feature.EvaluatorIntraLayerParallelismEnabled,
+          this,
+        )
+      ) {
+        codeqlArgs.push("--intra-layer-parallelism");
+      }
```
henrymercer, line 790: Tip for the future: sometimes we may want to disable a feature flag after it's already been enabled by default in the CLI.  To do that, we can use a pattern like this where we pass the `--no-` flag for sufficiently new CLIs when the feature is disabled: https://github.com/github/codeql-action/blob/2f8540dd316af4b6d54b847dba3e67bea2b870e0/src/codeql.ts#L844C1-L851C1.

nickrolfe: Ah that's nice, thanks. I'll make a PR to do that.

PR #1820, opened 2023-08-01
Clean up the bundle archive after extracting it

henrymercer:
Make some more disk space available to future steps by deleting the CodeQL bundle archive once we've extracted it.  We tolerate this step failing since deleting files doesn't always work, particularly on Windows where an antivirus program might lock the file.

Edit: There's another saving to be had — when we cache the bundle we copy it from a temporary directory to a toolcache directory, so we can also delete the extracted CodeQL bundle from this temporary directory.

In total, we make an additional 2.3 GB available to the `autobuild` and `analyze` steps on a Linux run using a single-platform dist.  The savings are better for multi-platform dists, but these are less common. 


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 ## [UNRELEASED]
 
 - We are rolling out a feature in August 2023 that will improve multi-threaded performance on larger runners. [#1817](https://github.com/github/codeql-action/pull/1817)
+- Reduce disk space usage when downloading the CodeQL bundle. [#1820](https://github.com/github/codeql-action/pull/1820)
```
adityasharad, line 8:
```suggestion
- Reduce disk space usage just after downloading the CodeQL bundle. [#1820](https://github.com/github/codeql-action/pull/1820)
```

adityasharad (dismissed):

CHANGELOG.md
henrymercer: This change now reduces the disk space required to download the CodeQL bundle too, so I think I'll stick with the original.

src/setup-codeql.ts
```diff
@@ -612,13 +629,30 @@ export async function downloadCodeQL(
   const toolcacheVersion = maybeCliVersion?.match(/^[0-9]+\.[0-9]+\.[0-9]+$/)
     ? `${maybeCliVersion}-${bundleVersion}`
     : convertToSemVer(bundleVersion, logger);
+
+  logger.debug("Caching CodeQL bundle.");
+  const codeqlFolder = await toolcache.cacheDir(
```
adityasharad, line 634: Minor: could we name these variables something like `codeqlDownloadedPath`, `codeqlExtractedPath`, `codeqlToolcachedPath`? I am already getting confused about the differences.

src/setup-codeql.ts
```diff
@@ -612,13 +629,30 @@ export async function downloadCodeQL(
...
+    codeqlExtracted,
+    "CodeQL",
+    toolcacheVersion,
+  );
+
+  // Safety check to make sure that we don't delete the bundle we're about to use.
```
adityasharad, line 640: Why are we concerned about deleting that bundle, if we've untarred that bundle to another location anyway?

henrymercer: This is quite defensive and is just guarding against the hypothetical case where `cacheDir` returns the input path.

src/setup-codeql.ts
henrymercer: Sure, I can do a bit of drive-by cleanup :)

src/setup-codeql.ts
```diff
@@ -612,13 +629,30 @@ export async function downloadCodeQL(
...
+    extractedBundlePath,
+    "CodeQL",
+    toolcacheVersion,
+  );
+
+  // Defensive check: we expect `cacheDir` to copy the bundle to a new location.
+  if (toolcachedBundlePath !== extractedBundlePath) {
+    logger.debug("Cleaning up downloaded CodeQL bundle.");
+    try {
+      await del(archivedBundlePath, { force: true });
```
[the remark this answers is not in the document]
henrymercer: Thanks for the catch, I've added a warning log if we don't end up deleting anything in the `del` call too.

adityasharad (approved):

henrymercer:
I reran https://github.com/github/codeql-action/actions/runs/5730681885/job/15530162792?pr=1820 in debug.  Logs are looking good.

angelapwen:
Just curious: how exactly did you figure out the disk space savings here?

henrymercer:
I did a fairly basic calculation: I added a bunch of print statements to dump the available disk space, and summed the amount of disk space made available by deleting the archived bundle and deleting the extracted bundle from the temporary directory.

PR #1822, opened 2023-08-03
PR Checks: Test `codeql-bundle.tar.gz`

angelapwen:
This PR check makes sure that we are testing the case where we use the non-platform specific `codeql-bundle.tar.gz` file for analysis. To do so, we've added a `use-all-platform-bundle` input to the `prepare-test` action that we set to true in the new PR check.

### DIFF
--- a/.github/actions/prepare-test/action.yml
+++ b/.github/actions/prepare-test/action.yml
@@ -28,7 +28,7 @@ runs:
       run: |
         set -e # Fail this Action if `gh release list` fails.
 
-        if [[ ${{ inputs.require-codeql-bundle-all }} == "true" ]]; then
+        if [[ ${{ inputs.use-all-platform-bundle }} == "true" ]]; then
           artifact_name="codeql-bundle.tar.gz"
         elif [[ "$RUNNER_OS" == "Linux" ]]; then
           artifact_name="codeql-bundle-linux64.tar.gz"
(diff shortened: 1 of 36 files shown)


.github/actions/prepare-test/action.yml
```diff
@@ -4,6 +4,10 @@ inputs:
   version:
     description: "The version of the CodeQL CLI to use. Can be 'latest', 'default', 'nightly-latest', 'nightly-YYYY-MM-DD', or 'stable-YYYY-MM-DD'."
     required: true
+  require-codeql-bundle-all:
```
henrymercer, line 7: Minor: How about `use-all-platform-bundle`, to make it more clear what the "all" means?

.github/workflows/test-codeql-bundle-all.yml
```diff
@@ -0,0 +1,56 @@
+name: 'PR Check - CodeQL Bundle All'
```
henrymercer, line 1: Optional: We could consider autogenerating this by adding a top-level property to the workflow specifications and forwarding this to `prepare-test`

.github/workflows/test-codeql-bundle-all.yml
```diff
@@ -0,0 +1,56 @@
...
+    - reopened
+    - ready_for_review
+  workflow_dispatch: {}
+jobs:
+  test-codeql-bundle-all:
+    strategy:
+      matrix:
+        include:
+        - os: ubuntu-latest
+          version: nightly-latest
```
henrymercer, line 26: Perhaps matrix over `default, latest, nightly-latest` to test the three most relevant bundles.

henrymercer (commented):
Looks good, just a couple of comments.

.github/actions/prepare-test/action.yml
angelapwen: Done!

.github/workflows/test-codeql-bundle-all.yml
angelapwen: Done 😄

PR #1825, opened 2023-08-07
Add runner disk info and image version to status report

henrymercer:
This PR adds three new fields to the status report:

- Available disk space on the runner
- Total disk space on the runner
- Image version of the runner, for GitHub-hosted runners

We also log a warning if there is less than 2 GB of disk space available at various points in the run.  For example, we log a warning if there is less than 2 GB of disk available at the start of `init`, but not at the end of `analyze` as the user is unlikely to need much space for the rest of the job.  For neatness, we only log the warning once per run, but we continue logging the amount of free disk space in the debug logs.

Commit-by-commit review recommended since we do some refactoring, e.g. pulling out status reporting into its own file.


.vscode/settings.json
```diff
@@ -9,5 +9,8 @@
   },
   // Installing a new Node package often triggers VS Code's git limit warnings as there is typically
   // an intermediate stage where many files are modified. This setting suppresses these warnings.
-  "git.ignoreLimitWarning": true
+  "git.ignoreLimitWarning": true,
+  // Use the vendored TypeScript version to have a consistent development experience across
+  // machines.
+  "typescript.tsdk": "node_modules/typescript/lib"
```
angelapwen, line 15: Just curious: how did you know to add this?

src/status-report.ts
```diff
@@ -240,6 +244,10 @@ export async function createStatusReportBase(
   if (codeQlCliVersion !== undefined) {
     statusReport.codeql_version = codeQlCliVersion;
   }
+  const imageVersion = process.env["ImageVersion"];
```
angelapwen, line 247: Also curious here: where did you find that this was under `ImageVersion`? I wasn't able to find it in the GitHub docs.

angelapwen (dismissed):
Amazing, thank you for all the refactoring and sorting ✨ a couple of questions for my own curiosity!

.vscode/settings.json
henrymercer: I believe this is documented [here](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript) but I just ran the "TypeScript: Select TypeScript Version" command in VS Code to point it to the vendored TypeScript version.

src/status-report.ts
henrymercer: This isn't officially documented, but has been in place for some time.  There are a few references to it in `actions/runner-images` issues.  Since it's not in the docs, I wouldn't want to rely on it, but I think as an optional thing we add to telemetry it's fine.

henrymercer: Yeah, I'm not sure on the reason behind that.

angelapwen (approved):
Good call on the changenote!

PR #1832, opened 2023-08-10
Simplify toolcache version number for semantically versioned bundles

henrymercer:
When we download a semantically versioned bundle, version it in the toolcache using just the CLI version number, `2.14.0`, since for semantically versioned bundles that's sufficient information to look them up in the toolcache from a `tools` URL.


angelapwen (approved):
Nice!!

PR #1811, opened 2023-07-28
Only print the analysis summary once

henrymercer:
Currently we print the analysis summary twice, once within the "interpreting results" log group, and once at the top-level of the analyze Action logs.  This PR changes this so we print the analysis summary once at the top-level.

### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #1833, opened 2023-08-10
Log commands manually before executing them

henrymercer:
This PR fixes a bad interaction between https://github.com/github/codeql-action/pull/1795 and https://github.com/github/codeql-action/pull/1811 which we didn't pick up since the problem only appears when looking at the combination of these two PRs.

In https://github.com/github/codeql-action/pull/1811, to make sure we only print the analysis summary once, we invoked the tool runner with `silent: true` to capture CodeQL's stdout without logging it.  We then had our stdout listener forward CodeQL's stdout to the Action's stdout.  However we missed another implication of setting `silent: true`: this disables the log line which prints the command being invoked.  This is because we only started running most of our CodeQL invocations through the tool runner in https://github.com/github/codeql-action/pull/1795.

With the combination of these two PRs, we stop logging most CodeQL invocations, which reduces the utility of the logs.

This PR starts printing the commands manually to make sure we log them in all cases, including ones where we want to capture the stdout.


angelapwen (approved):
Agh.. nice catch.

Issue #1162, opened 2022-08-02
Avoid printing an error message when `upload-sarif` location is not a git directory


crenshaw-dev:
I have a workflow which has to check out the repo to a specific (non-default) directory.

The upload-sarif action produces this error:

```
fatal: not a git repository (or any of the parent directories): .git
Failed to call git to get current commit. Continuing with data from environment or input: Error: The process '/usr/bin/git' failed with exit code 128
Error: The process '/usr/bin/git' failed with exit code 128
    at ExecState._setResult (/home/runner/work/_actions/github/codeql-action/v2/node_modules/@actions/exec/lib/toolrunner.js:592:25)
    at ExecState.CheckComplete (/home/runner/work/_actions/github/codeql-action/v2/node_modules/@actions/exec/lib/toolrunner.js:575:18)
    at ChildProcess.<anonymous> (/home/runner/work/_actions/github/codeql-action/v2/node_modules/@actions/exec/lib/toolrunner.js:469:27)
    at ChildProcess.emit (node:events:390:28)
    at maybeClose (node:internal/child_process:1064:16)
    at Socket.<anonymous> (node:internal/child_process:450:11)
    at Socket.emit (node:events:390:[28](https://github.com/argoproj/argo-cd/runs/7634411066?check_suite_focus=true#step:5:29))
    at Pipe.<anonymous> (node:net:687:12)
```

I believe the action expects the working directory to be a git repo.

aibaars:
Did the upload action actually fail? The phrase `Continuing with data from environment or input:` suggests that is should carry on regardless.

crenshaw-dev:
It did not fail. Maybe the appropriate fix is to change the error message? It's really unclear based on the current message whether I should be concerned, i.e. why it would want to be run in a git repo.

hmakholm:
Since there's a fallback with a fairly good chance of working (namely falling back to the environment variable set by the Actions runner), the message arguably at least shouldn't be presented with a stack trace -- that makes it look more alarming than it ought to.

aeisenberg:
I changed the issue title to better reflect the actual problem.

PR #1860, opened 2023-08-28
Add better error messages when determining merge-base

aeisenberg:
Avoid printing scary error messages to console when the current directory is not a git repo. Instead provide a better reason for the git failure and continue on.

Fixes https://github.com/github/codeql-action/issues/1162.


src/actions-util.test.ts
```diff
@@ -267,3 +269,50 @@ test("isAnalyzingDefaultBranch()", async (t) => {
     getAdditionalInputStub.restore();
   });
 });
+
+test("determineMergeBaseCommitOid non-pullrequest", async (t) => {
+  const stub = sinon.stub(core, "info");
```
henrymercer, line 274: Nit, optional: renaming this to `infoStub` makes the assertion a little more readable.

src/actions-util.ts
```diff
@@ -63,6 +63,7 @@ export const getCommitOid = async function (
   // the merge commit, which must mean that git is available.
   // Even if this does go wrong, it's not a huge problem for the alerts to
   // reported on the merge commit.
+  let output = "";
```
henrymercer, line 66: Perhaps rename to `stderr` so we don't confuse this with the standard out, i.e. the git sha?

src/actions-util.ts
```diff
@@ -146,10 +154,15 @@ export const determineMergeBaseCommitOid = async function (): Promise<
...
-    core.info((e as Error).stack || "NO STACK");
+    if (output.includes("not a git repository")) {
+      core.info(
+        "The checkout path provided to the action does not appear to be a git repository. Continuing with data from environment.",
+      );
+    } else {
+      core.info(
+        `Failed to call git to determine merge base. Continuing with data from environment: ${output}`,
+      );
+    }
```
henrymercer, line 165: I think the message here isn't quite accurate as we'll just continue without a merge base.  This shouldn't be a problem since we compute a merge base server side.

henrymercer (commented):
Looks good overall, some minor suggestions.  Also I've been suggesting we add changelog notes to more of our PRs and I think this one would benefit from one too.

src/actions-util.ts
```diff
@@ -75,19 +76,24 @@ export const getCommitOid = async function (
...
     return commitOid.trim();
   } catch (e) {
-    core.info(
-      "Could not determine current commit SHA using git. Continuing with data from user input or environment.",
-    );
-    core.debug(`Reason: ${(e as Error).message}`);
-    core.debug((e as Error).stack || "NO STACK");
+    if (output.includes("not a git repository")) {
+      core.info(
+        "Could not determine current commit SHA using git. Continuing with data from user input or environment. The checkout path provided to the action does not appear to be a git repository.",
```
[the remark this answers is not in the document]
aeisenberg: I don't know if eslint will break up a single line string into multiline. I can look, though.

src/actions-util.test.ts
```diff
@@ -267,3 +268,50 @@ test("isAnalyzingDefaultBranch()", async (t) => {
...
+  process.env["GITHUB_SHA"] = "100912429fab4cb230e66ffb11e738ac5194e73a";
+  const result = await actionsUtil.determineMergeBaseCommitOid(__dirname);
+  t.deepEqual(result, undefined);
+  t.deepEqual(0, infoStub.callCount);
+
+  infoStub.restore();
+});
+
+test("determineMergeBaseCommitOid no error", async (t) => {
+  const stub = sinon.stub(core, "info");
```
henrymercer, line 285: `s/stub/infoStub` here too?

src/actions-util.test.ts
```diff
@@ -267,3 +268,50 @@ test("isAnalyzingDefaultBranch()", async (t) => {
...
+    stub.firstCall.args[0].startsWith(
+      "The checkout path provided to the action does not appear to be a git repository.",
+    ),
+  );
+
+  stub.restore();
+});
+
+test("determineMergeBaseCommitOid other error", async (t) => {
+  const stub = sinon.stub(core, "info");
```
henrymercer, line 301: `s/stub/infoStub` here too?

src/actions-util.ts
```diff
@@ -75,19 +76,26 @@ export const getCommitOid = async function (
...
-    core.debug((e as Error).stack || "NO STACK");
+    if (stderr.includes("not a git repository")) {
+      core.info(
+        "Could not determine current commit SHA using git. Continuing with data from user input or environment. " +
+          "The checkout path provided to the action does not appear to be a git repository.",
+      );
+    } else {
+      core.info(
+        `Could not determine current commit SHA using git. Continuing with data from user input or environment. ` +
+          `Failed to call git to determine merge base. Continuing with data from environment: ${stderr}`,
```
henrymercer, line 95:
```suggestion
        `Could not determine current commit SHA using git. Continuing with data from user input or environment. ${stderr}`,
```

henrymercer (dismissed):

src/actions-util.ts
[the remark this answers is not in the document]
aeisenberg: [This is the setting.](https://eslint.org/docs/latest/rules/max-len) I can set `ignoreStrings` to `false`. The problem is that there is no autofix for this. Maybe that's ok. I'll create an issue for this.

henrymercer (approved):
Nice, thanks for addressing this bit of customer feedback!

PR #1862, opened 2023-08-31
Improve logging for Go autobuild when invoked in `analyze` step

angelapwen:
We want to make it more explicit when we run the Go autobuilder in `analyze` when the user has not specified the autobuild step. I have used the `debug` log level (which will show up when re-run on debug mode is activated) rather than `info` because we have customers on legacy workflows where this is the expected workflow.


MathiasVP (approved):
LGTM!

PR #1888, opened 2023-09-15
Respect RAM constraints imposed by Linux cgroups

henrymercer:
This updates the Action's RAM calculation to incorporate RAM constraints imposed by Linux cgroups, for example when running in a Docker container on Linux.

We support constraints from both cgroups v1 and v2.


angelapwen (approved):
NIce! Could we add a comment indicating that once the relevant NodeJS fix is in we'd be able to remove this workaround?

henrymercer:
> NIce! Could we add a comment indicating that once the relevant NodeJS fix is in we'd be able to remove this workaround?

Yes, I'll do so as follow up so I can get this out.

PR #1889, opened 2023-09-18
C++: introduce automatic installation of dependencies in the autobuilder

redsun82:
This introduces the possibility to automatically install dependencies when running the C++ autobuilder on an Ubuntu runner, that will be available with upcoming version 2.15.0. This behaviour can be explicitly requested by setting `CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES` in the `autobuild` action `env`, or is otherwise driven by the `code_scanning_cpp_dependency_installation` feature flag.


autobuild/action.yml
```diff
@@ -12,6 +12,17 @@ inputs:
...
+  cpp-autoinstall-dependencies:
+    description: >-
+      Experimental input, the API may change in the future.
+      Set to true to enable trying to automatically detect and install
+      dependencies when running the C/C++ autobuilder, set to anything else
+      to explicitly disable it. This only works when running on Ubuntu and
+      is automatically disabled otherwise. If this input is not set the
+      default is currently driven by the cpp_dependency_installation_enabled
+      feature flag or the CODEQL_CPP_DEPENDENCY_INSTALLATION environment
+      variable.
```
henrymercer, line 24: Would it be simpler to expose a single way of configuring this, i.e. just via the `CODEQL_CPP_DEPENDENCY_INSTALLATION` environment variable?  If we want to make it more obvious that this feature is experimental, we could also rename it to `CODEQL_CPP_EXPERIMENTAL_DEPENDENCY_INSTALLATION` or similar.

src/autobuild.ts
```diff
@@ -91,13 +98,60 @@
...
+      core.exportVariable(envVar, "false");
+    } else {
+      logger.info(
+        `${
+          actionInput === "true" ? "Enabling" : "Disabling"
+        } ${featureName} explicitly requested`,
+      );
+      core.exportVariable(envVar, actionInput);
+    }
+  } else if (process.env["RUNNER_ENVIRONMENT"] === "self-hosted") {
```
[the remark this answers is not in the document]
henrymercer: Let's double check this environment variable exists in default setup and then add it to `queries/default-setup-environment-variables.ql`.

src/autobuild.ts
```diff
@@ -91,13 +98,60 @@ export async function determineAutobuildLanguages(
   return languages;
 }
 
+async function setupCppAutobuild(codeql: CodeQL, logger: Logger) {
+  const envVar = "CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES";
+  const actionInput = getOptionalInput("cpp-autoinstall-dependencies");
+  const featureName = "C++ automatic installation of dependencies";
+  if (actionInput === "true") {
+    if (!(await codeQlVersionAbove(codeql, "2.15.0"))) {
```
henrymercer, line 106: Nit: Flip the condition for readability

src/feature-flags.ts
```diff
@@ -74,6 +75,11 @@ export const featureConfig: Record<
     minimumVersion: "2.14.0",
     defaultValue: false,
   },
+  [Feature.CppDependencyInstallation]: {
+    envVar: "CODEQL_CPP_DEPENDENCY_INSTALLATION",
+    minimumVersion: "2.15.0",
```
henrymercer, line 80: Move this into a static final constant so we can reuse it in `autobuild.ts` too.

src/autobuild.ts
```diff
@@ -91,13 +98,60 @@ export async function determineAutobuildLanguages(
   return languages;
 }
 
+async function setupCppAutobuild(codeql: CodeQL, logger: Logger) {
+  const envVar = "CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES";
```
henrymercer, line 102: Should this be different to `CODEQL_CPP_DEPENDENCY_INSTALLATION`?  It could be confusing if we need to use a different environment variable to enable deptrace in Actions vs in third-party CI where customers are using the CodeQL CLI directly.

henrymercer (commented):
This looks like it will do the right thing, though I have a couple of questions about whether we can simplify the experience by reducing the amount of configuration options.  Let me know what you think!

src/autobuild.ts
redsun82: That is a good point, I was a bit torn on it myself. I wanted to decouple the action configuration from the internal extractor flag, but maybe it's just simpler to just go with `CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES` everywhere?

autobuild/action.yml
redsun82: I guess it's acceptable to do this kind of configuration via environment variables... what I think is important is that even once the rollout is complete and this becomes the default, there should be a clear opt out for users. I guess this would mean (outside the scope of this PR):
* clear documentation about this option in [here](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages#autobuild-for-cc)
* possibly a commented out opt-out in the default workflow template created when C/C++ analysis is involved?

src/autobuild.ts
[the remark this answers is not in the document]
redsun82: This would seem to be something going set pretty much at the very core of actions:

https://github.com/github/actions-dotnet/blob/6f69e9eaf146862329edbc8a87a1e2aec9a9c17a/Actions/Runtime/Sdk/Server/TaskHub.cs#L3636

how can we double check what exists in default setup, just to be sure?

autobuild/action.yml
redsun82: Here's the PR that introduced that by the way:

https://github.com/github/actions-dotnet/pull/15224

src/autobuild.ts
redsun82: Here's the PR that introduced that by the way:

https://github.com/github/actions-dotnet/pull/15224

redsun82: Here's the PR that introduced that by the way:

https://github.com/github/actions-dotnet/pull/15224

src/autobuild.ts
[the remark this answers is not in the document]
redsun82: Here's the PR that introduced that by the way:

https://github.com/github/actions-dotnet/pull/15224

src/autobuild.ts
henrymercer: I've had the same dilemma before and I think overall it was best to minimize the number of environment variables — it can be confusing to explain to folks that there's one environment variable for Actions and another for the CLI.  I'd recommend we go with `CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES`.

autobuild/action.yml
henrymercer: Good ideas.  Would it also be practical to recommend that users try adding `CODEQL_CPP_DEPENDENCY_INSTALLATION=false` to their workflow in the diagnostic message we produce when the autobuilder fails?

src/autobuild.ts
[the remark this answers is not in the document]
henrymercer: I've messaged you some internal resources we have about testing on default setup.

autobuild/action.yml
redsun82: hmm, that would probably not be helpful most of the time. The latest QA experiment has shown that out of around 5000 repos there were still 3751 failures, of which only 11 would have succeeded with deptrace off. So the vast majority of the time the suggestion would be useless. I need to think what could actually constitute a sign that this feature should be turned off.

src/feature-flags.ts
redsun82: `featureConfig` is actually exported, so I can share this information without defining another constant by using `featureConfig[Feature.CppDependencyInstallation].envVar` in `autobuild.ts`

src/autobuild.ts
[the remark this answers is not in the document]
redsun82: [I can confirm](https://github.com/dsp-testing/test-cpp-default-setup/actions/runs/6246955844/job/16958607196#step:4:47) that that variable is indeed available in default setup, I will update the query then :+1:

src/autobuild.ts
```diff
@@ -91,13 +98,48 @@ export async function determineAutobuildLanguages(
...
+    ) {
+      logger.info(`Disabling ${featureName} as we are on a self-hosted runner`);
+      logger.info(`This can be enabled by setting ${envVar}: true in the env`);
+      core.exportVariable(envVar, "false");
+    } else {
+      logger.info(`Enabling ${featureName}`);
+      core.exportVariable(envVar, "true");
+    }
+  } else {
+    logger.info(`Disabling ${featureName}`);
```
henrymercer, line 128:
```suggestion
    logger.info(`Disabling ${featureName}.`);
```

src/autobuild.ts
```diff
@@ -91,13 +98,48 @@ export async function determineAutobuildLanguages(
...
+    // disable autoinstall on self-hosted runners unless explicitly requested
+    if (
+      process.env["RUNNER_ENVIRONMENT"] === "self-hosted" &&
+      process.env[envVar] !== "true"
+    ) {
+      logger.info(`Disabling ${featureName} as we are on a self-hosted runner`);
+      logger.info(`This can be enabled by setting ${envVar}: true in the env`);
+      core.exportVariable(envVar, "false");
+    } else {
+      logger.info(`Enabling ${featureName}`);
```
henrymercer, line 124:
```suggestion
      logger.info(`Enabling ${featureName}. This can be disabled by setting the ${envVar} environment variable to 'false' (see https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow).`);
```

src/autobuild.ts
```diff
@@ -91,13 +98,48 @@ export async function determineAutobuildLanguages(
...
+    logger,
+  );
+  if (await features.getValue(Feature.CppDependencyInstallation, codeql)) {
+    // disable autoinstall on self-hosted runners unless explicitly requested
+    if (
+      process.env["RUNNER_ENVIRONMENT"] === "self-hosted" &&
+      process.env[envVar] !== "true"
+    ) {
+      logger.info(`Disabling ${featureName} as we are on a self-hosted runner`);
+      logger.info(`This can be enabled by setting ${envVar}: true in the env`);
```
henrymercer, line 121:
```suggestion
      logger.info(`Disabling ${featureName} as we are on a self-hosted runner.` + getWorkflowEventName() !== "dynamic" ? ` To override this, set the ${envVar} environment variable to 'true' in your workflow (see https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow).` : "");
```

henrymercer (dismissed):
Some minor suggestions around logging, but LGTM otherwise!

henrymercer (approved):
Excellent!

PR #1891, opened 2023-09-18
Fix reading `cgroup` limits on `ubuntu-20.04`

henrymercer:
Fix an issue resolving RAM on the `ubuntu-20.04` Actions runner image.  When running on Ubuntu 20.04, `/sys/fs/cgroup/memory/memory.limit_in_bytes` sometimes returns MAX_INT to indicate a lack of constraints, which we weren't handling properly.

This PR starts running CodeQL against all supported Actions runner images to ensure that we will catch issues like this in CI in the future.


src/util.ts
```diff
@@ -225,17 +226,31 @@ function getTotalMemoryAvailable(logger: Logger): number {
...
+          const displayLimit = `${Math.floor(limit / (1024 * 1024))} MiB`;
+          if (limit < osTotalMemory) {
+            logger.info(
+              `While resolving RAM, found cgroup limit of ${displayLimit} in ${limitFile}.`,
+            );
+            return limit;
+          } else {
+            logger.debug(
+              `While resolving RAM, ignoring cgroup file ${limitFile} since the limit (${displayLimit}) ` +
+                "is greater than the total RAM reported by the operating system.",
```
cklin, line 238: (nit) The debug message is technically incorrect, since we also enter the `else` branch when the cgroup limit is the same as `osTotalMemory`.

cklin (approved):

src/util.ts
henrymercer: Good point.  I've fixed in the last commit.

cklin (approved):
👍

PR #1900, opened 2023-09-22
Bump min. version for evaluator parallelism feature flag

nickrolfe:
Since we suspect fine-grained parallelism may make the OOM issues worse, we don't want to enable it on 2.14.0-2.14.5, even though they support it.


aeisenberg (approved):

henrymercer:
Note that when we ramp up the feature flag, customers running old versions of the CodeQL Action may have fine-grained parallelism enabled despite using an older version of the CLI.  This is probably an acceptable risk, given that most customers will be running the up-to-date Action, and running on the older CLI versions will only rarely cause breakage.

nickrolfe:
Good point, but I agree the risk is low. Fine-grained parallelism _should_ work just fine if it gets enabled on 2.14.0-5.

PR #1901, opened 2023-09-22
Warn when workflow analyzes the same language twice

henrymercer:
With the introduction of new aliases for C/C++, Java/Kotlin, and JavaScript/TypeScript, we want to avoid customers analyzing the same language more than once.  The most common case of this will likely be customers who have started with a starter workflow and have added a duplicate language to the `language` matrix parameter.

This PR introduces workflow validation for this common case, so that customers who are analyzing the same language twice will get an Actions warning when they execute that workflow.


src/init-action.ts
```diff
@@ -217,16 +217,13 @@ async function run() {
...
-    const workflowErrors = await validateWorkflow(logger);
-
     if (
       !(await sendStatusReport(
         await createStatusReportBase(
           "init",
           "starting",
           startedAt,
           await checkDiskUsage(logger),
-          workflowErrors,
```
adityasharad, line 229: This change means all workflow errors will be reported as Actions errors, but not part of the `starting` status report, right?

src/workflow.ts
```diff
@@ -104,14 +106,56 @@ export const WorkflowErrors = toCodedErrors({
...
+          const matrixLanguagesByExtractor: {
+            [extractorName: string]: string[];
+          } = {};
+          for (const language of matrixLanguages) {
+            const extractorName = aliases[language] || language;
+            if (!matrixLanguagesByExtractor[extractorName]) {
+              matrixLanguagesByExtractor[extractorName] = [];
+            }
+            matrixLanguagesByExtractor[extractorName].push(language);
+          }
```
adityasharad, line 137: I think this could usefully be pulled out into a `groupLanguagesByExtractorName` helper.

src/workflow.ts
```diff
@@ -104,14 +106,56 @@ export const WorkflowErrors = toCodedErrors({
...
+            matrixLanguagesByExtractor,
+          )) {
+            if (languages.length > 1) {
+              errors.push({
+                message:
+                  `CodeQL language '${extractor}' is referenced by more than one entry in the ` +
+                  `'language' matrix parameter for job '${jobName}'. This may result in duplicate alerts. ` +
+                  `Please edit the 'language' matrix parameter to keep only one of the following: ${languages
+                    .map((language) => `'${language}'`)
+                    .join(", ")}.`,
```
adityasharad, line 150: Perhaps add a phrase like "these language names are considered identical by CodeQL". Do we want to recommend they use the combined alias key in this situation?

src/workflow.test.ts
```diff
@@ -22,227 +24,387 @@ function errorCodes(
...
+      on: ["push", "pull_request"],
+      jobs: { test: { steps: [{ run: "git checkout HEAD^2" }] } },
+    },
+    await getCodeQLForTesting(),
+  );
 
   t.deepEqual(...errorCodes(errors, [WorkflowErrors.CheckoutWrongHead]));
 });
 
+test("getWorkflowErrors() for workflow with language name and its alias", async (t) => {
```
adityasharad, line 315: Just to be sure, have we got a test for two unrelated languages?

adityasharad (commented):
Good idea. A few minor questions.

src/init-action.ts
henrymercer: That's right.  This improves consistency since we're using the `error` and `cause` fields of the status report to indicate errors, rather than sometimes errors and sometimes warnings.  I think if we want to collect workflow validation warnings, it would be more appropriate to add a new field.

src/workflow.test.ts
henrymercer: Good idea, let's add one.

src/workflow.ts
henrymercer: The tricky thing is that the CLI doesn't really have a notion of the preferred name, since to reduce risk we did not change the `name`s of the CodeQL extractors.  I think we're going to hit this warning rarely enough that it's not worth hardcoding the language-specific information into the Action, but I don't feel strongly about this — feel free to convince me otherwise!

adityasharad: Makes sense. We can introduce later a concept of "recommended name" if we want to do this, agree it's not urgent.

adityasharad (dismissed):
Optionally consider a changenote.

adityasharad (approved):

PR #1907, opened 2023-09-27
Bump minimum CodeQL version to 2.10.5

henrymercer:
Now that the 2.9.* series is deprecated, we can remove tech debt in the Action and update our testing matrix to improve our test coverage against the latest versions of CodeQL.


aeisenberg (approved):
This is great. 

Can you also take a look at the individual checks to see if any of them can start testing against more versions. For example, `pr-checks/checks/packaging-config-inputs-js.yml` says that it works against recent CLIs, but now, I think it is available in all CLIs that are still supported.

This could happen separately.

And, only if you think it would be helpful, you could add a section in the `CONTRIBUTING.md` on the different steps to take when removing support for an old CLI.

henrymercer:
> Can you also take a look at the individual checks to see if any of them can start testing against more versions. For example, `pr-checks/checks/packaging-config-inputs-js.yml` says that it works against recent CLIs, but now, I think it is available in all CLIs that are still supported.

I have a separate plan for this — we are currently running too many CI jobs on each PR, which causes friction both internally when jobs fail for transient reasons and externally, because API rate limits can prevent contributors from being able to run the checks.  Specifically, I'd like to start testing particularly old CLIs on a nightly basis and on the release branches instead of on `main`.

> And, only if you think it would be helpful, you could add a section in the `CONTRIBUTING.md` on the different steps to take when removing support for an old CLI.

Good idea, I'll do this.

aeisenberg (approved):
Nice.

PR #1909, opened 2023-09-28
Install wrapper script for Go on Linux to support tracing for Go 1.21 and above

mbg:
This PR modifies the `init` Action to add a `go` wrapper script to the `PATH` if we are running on Linux in order to better cope with tracing builds for Go 1.21 and above. This works as follows:

- If we are initialising CodeQL for Go and we are running on Linux, then we obtain the current path to the `go` binary.
- We query the type of executable with the `file` command.
- If this contains the string `"statically linked"`, we install the wrapper script to a `bin` directory in the CodeQL temporary directory.
- We add that `bin` directory to the system `PATH`
- We store the location of the wrapper script in a new environment variable, `CODEQL_ACTION_GO_BINARY` (we could probably omit this since the path can be reconstructed in the `analyze` step)

Additionally, we verify that the result of `which go` points to our wrapper script in the `analyze` step by checking it against the value of `CODEQL_ACTION_GO_BINARY` to make sure that users did not e.g. use a `setup-go` step after the `init` action.


src/actions-util.ts
```diff
@@ -425,3 +425,33 @@ export function getWorkflowRunAttempt(): number {
...
+          },
+          stderr: (data) => {
+            stderr += data.toString();
+          },
+        },
+      },
+    ).exec();
+    return fileOut;
+  } catch (e) {
+    core.info(`Could not determine type of ${fp}. ${stderr}`);
```
smowton, line 453: Show stdout as well?

src/actions-util.ts
```diff
@@ -425,3 +425,33 @@ export function getWorkflowRunAttempt(): number {
...
+
+/**
+ * Tries to obtain the output of the `file` command for the file at the specified path.
+ */
+export const getFileType = async (fp: string): Promise<string> => {
+  let stderr = "";
+  try {
+    let fileOut = "";
+    await new toolrunner.ToolRunner(
+      await safeWhich.safeWhich("file"),
```
smowton, line 437: Do we need to `which` it? Can we just run it by basename?

src/analyze-action.ts
```diff
@@ -236,6 +237,19 @@ async function run() {
...
 
+    // Check that the Go wrapper script still exists, if set
+    const goWrapperPath = process.env[EnvVar.GO_BINARY_LOCATION];
+
+    if (goWrapperPath !== undefined) {
+      const goBinaryPath = await safeWhich("go");
+
+      if (goWrapperPath !== goBinaryPath) {
+        core.warning(
+          "Unexpected result for `which go`: please ensure that the correct version of Go is installed before the `codeql-action/init` Action is used.",
```
smowton, line 248: Say what we're comparing it to -- it's changed since `init`

src/actions-util.ts
mbg: I largely followed the example of other invocations of `ToolRunner`, which all use `safeWhich` to find the executable to run.

criemen:
Can we already add a feature gate/version flag to that, so that we automatically stop this wrapper script installation at some point in the future? Or a turn-off switch from the CLI?
Once we fix this in the CLI, we need to be able to retrospectively turn off this hack from the CLI, otherwise we'll extract go files with older codeql-action versions double.

src/actions-util.ts
smowton: ^^ ?

src/analyze-action.ts
smowton: To consider: TSP warning?

smowton (commented):
This broadly looks good but still needs some sort of version gating mechanism so that we don't start double-tracing if and when the CLI starts handling this situation for itself.

src/analyze-action.ts
mbg: I don't think the Action currently generates any diagnostics of its own (@henrymercer can confirm?) so I am not entirely sure how much work it would be for us to start adding code to do that.

src/init-action.ts
```diff
@@ -321,6 +328,44 @@ async function run() {
...
+      !isSupportedToolsFeature(
+        versionInfo,
+        ToolsFeature.IndirectTracingSupportsStaticBinaries,
+      )
+    ) {
+      try {
+        const goBinaryPath = await safeWhich("go");
+        const fileOutput = await getFileType(goBinaryPath);
+
+        if (fileOutput.includes("statically linked")) {
```
henrymercer, line 344: Add a debug log line that we're applying a tracing workaround.

src/init-action.ts
```diff
@@ -321,6 +328,44 @@ async function run() {
...
+          // Create a directory that we can add to the system PATH.
+          const tempBinPath = path.resolve(getTemporaryDirectory(), "bin");
```
henrymercer, line 346: Currently this uses the temporary directory for the entire workflow by default.  To reduce the chances of colliding with other Actions, consider making this a bit more specific, e.g. adding a middle `codeql-action-go-tracing` part.

src/init-action.ts
```diff
@@ -321,6 +328,44 @@ async function run() {
...
+          );
+          fs.chmodSync(goWrapperPath, "755");
+
+          // Store the original location of our wrapper script somewhere where we can
+          // later retrieve it from and cross-check that it hasn't been changed.
+          core.exportVariable(EnvVar.GO_BINARY_LOCATION, goWrapperPath);
+        }
+      } catch (e) {
+        core.warning(
+          `Analyzing Go on Linux, but failed to install wrapper script: ${e}`,
```
henrymercer, line 364: Can we say something about how this might affect the analysis here?  Also, use `logger` instead of `core` here.

src/resolve-environment.test.ts.ignore
```diff
@@ -0,0 +1,35 @@
+import test from "ava";
```
henrymercer, line 1: Did you intend to check this in?

src/tools-features.ts
```diff
@@ -0,0 +1,20 @@
+import { VersionInfo } from "./codeql";
+
+export enum ToolsFeature {
+  FeaturesInVersionResult = "featuresInVersionResult",
```
henrymercer, line 4: Nit: drop this since we're not using it.

henrymercer (commented):
As a general comment, are we happy with the test coverage here?  Specifically, do we want to add some unit tests or some specific assertions to the PR checks?  Or do the existing checks passing on Go 1.21.1 give us enough confidence?

pr-checks/checks/go-tracing-autobuilder.yml
```diff
@@ -4,6 +4,9 @@ operatingSystems: ["ubuntu", "macos"]
 env:
   DOTNET_GENERATE_ASPNET_CERTIFICATE: "false"
 steps:
+  - uses: actions/setup-go@v4
+    with:
+      go-version: "1.21.1"
```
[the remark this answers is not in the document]
mbg: We will update it as necessary with future version upgrades. I can modify this to always use the latest patch version, similar to what we have in other workflows now.

src/actions-util.ts
```diff
@@ -425,3 +425,33 @@ export function getWorkflowRunAttempt(): number {
...
+
+/**
+ * Tries to obtain the output of the `file` command for the file at the specified path.
+ */
+export const getFileType = async (fp: string): Promise<string> => {
+  let stderr = "";
+  let stdout = "";
+  try {
+    await new toolrunner.ToolRunner(
+      await safeWhich.safeWhich("file"),
```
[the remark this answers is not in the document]
mbg: Hmm. I'd generally expect unix-like operating systems to have it installed by default, although e.g. alpine does not. I think we should handle this at the call-site of `getFileType` though depending on what not being able to obtain the result from `file` has on the caller.

src/analyze-action.ts
```diff
@@ -231,6 +232,19 @@ async function run() {
       logger,
     );
 
+    // Check that the Go wrapper script still exists, if set
```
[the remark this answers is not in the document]
mbg: The comment is wrong, sorry. I will fix that. We are actually checking that `which go` still points at the wrapper script.

src/actions-util.ts
```diff
@@ -425,3 +425,33 @@ export function getWorkflowRunAttempt(): number {
...
+          stdout: (data) => {
+            stdout += data.toString();
+          },
+          stderr: (data) => {
+            stderr += data.toString();
+          },
+        },
+      },
+    ).exec();
+    return stdout;
```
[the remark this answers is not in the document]
mbg: We could. I have no preference. Our one application at the moment doesn't need it, since we are just checking for a substring.

src/actions-util.ts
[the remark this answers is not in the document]
mbg: In the case of our one use of `getFileType`, we have the whole procedure in a `try`/`catch` block which prints a warning to the log if something goes wrong. I think that is probably OK. 

The alternative would be to install the wrapper script even if we can't detect that `go` is a static binary. There'd probably not be any harm in doing that here, so that would also be OK.

I have no strong preference.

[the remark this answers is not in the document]
mbg: @smowton any thoughts?

src/init-action.ts
```diff
@@ -321,6 +328,44 @@ async function run() {
...
 
+    // https://github.com/github/codeql-team/issues/2411
+    if (
+      config.languages.includes(Language.go) &&
+      process.platform === "linux" &&
+      !isSupportedToolsFeature(
+        versionInfo,
+        ToolsFeature.IndirectTracingSupportsStaticBinaries,
+      )
+    ) {
```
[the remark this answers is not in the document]
mbg: There is very little risk with this workaround. We are simply wrapping the actual `go` binary in a script that just invokes the actual `go` binary with all arguments passed to the script. 

Without the workaround we have broken tracing for Go 1.21+ on Linux.

src/actions-util.ts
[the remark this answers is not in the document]
henrymercer: Let's do it.  This has caught us out in the past.

henrymercer (commented):
Thanks, I'd like to hear what Chris thinks about how we should behave in the absence of `file`, but otherwise this looks good to me.

src/actions-util.ts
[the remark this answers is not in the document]
smowton: Installing the wrapper when `go` is a dynamically-linked binary will cause double-extraction, so I'd prefer that we make sure we have hard evidence that it's necessary before installing. The circumstance where we don't have this basic utility on an actions runner sounds reasonably far-fetched, and Alpine in particular we can't run on anyhow due to its not using glibc.

src/analyze-action.ts
smowton: @henrymercer could you comment on this? If possible it would be great to flag clearly to the user when they're running `setup-go` after `codeql-init` and are thereby defeating this workaround.

henrymercer: I like this idea.  The Action doesn't yet generate any diagnostics of its own, but it should be fairly simple — we just need to call `codeql database add-diagnostic`, or drop a JSON diagnostic in the database directory.

src/actions-util.ts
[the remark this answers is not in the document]
mbg: Perfect, then we don't need to change anything since the current implementation catches the exception, emits a warning, and then carries on as is without installing the wrapper.

henrymercer (approved):

src/analyze-action.ts
```diff
@@ -245,6 +246,25 @@ async function run() {
...
         );
+
+        addDiagnostic(
+          config,
+          Language.go,
+          makeDiagnostic(
+            "go/workflow/go-installed-after-codeql-init",
+            "Go was installed after the `codeql-action/init` Action was run",
+            {
+              plaintextMessage:
```
henrymercer, line 257: Use `markdownMessage` since we're using the inline code block :)

src/analyze-action.ts
```diff
@@ -245,6 +246,25 @@ async function run() {
...
+                "To avoid interfering with the CodeQL analysis after the `codeql-action/init` Action is run, you should perform all installation steps beforehand.",
```
henrymercer, line 258:
```suggestion
                "To avoid interfering with the CodeQL analysis, perform all installation steps before calling the `github/codeql-action/init` Action.",
```

henrymercer (approved):
Very nice!

henrymercer (approved):

PR #1918, opened 2023-10-04
Support `features` key in `version` JSON output

mbg:
Add general support for extracting the `features` key in the output of the `version` command. 


src/codeql.ts
```diff
@@ -212,6 +212,19 @@ export interface CodeQL {
...
+  productName: string;
+  vendor: string;
+  version: string;
+  sha: string;
+  branches: string[];
+  copyright: string;
+  unpackedLocation: string;
+  configFileLocation: string;
+  configFileFound: boolean;
+  features?: { [name: string]: boolean };
```
henrymercer, line 225: We can simplify this down to just the entries the CodeQL Action cares about:

```suggestion
  version: string;
  features?: { [name: string]: boolean };
```

src/codeql.ts
```diff
@@ -430,6 +443,21 @@ function resolveFunction<T>(
...
+  vendor: "GitHub",
+  sha: "",
+  branches: [],
+  copyright: "",
+  unpackedLocation: "",
+  configFileLocation: "",
+  configFileFound: false,
+  version,
+});
+
```
henrymercer, line 460: Minor: Move this to `testing-utils.ts`

src/codeql.ts
```diff
@@ -212,6 +212,19 @@ export interface CodeQL {
   resolveExtractor(language: Language): Promise<string>;
 }
 
+export interface VersionOutput {
```
henrymercer, line 215: Nit: rename to `VersionInfo`?

src/codeql.ts
```diff
@@ -528,7 +556,9 @@ export async function getCodeQLForCmd(
     async getVersion() {
       let result = util.getCachedCodeQlVersion();
       if (result === undefined) {
-        result = (await runTool(cmd, ["version", "--format=terse"])).trim();
+        result = JSON.parse(
+          await runTool(cmd, ["version", "--format=json"]),
+        ) as VersionOutput;
```
henrymercer, line 561: Probably worth wrapping this in a `try/catch` to handle unexpected output from `codeql version --format=json`.

henrymercer (commented):
Nice!  A couple of minor comments.

src/codeql.ts
mbg: I went for `VersionOutput` to be consistent with the other interface names, but agree that `VersionInfo` is nicer. Consistency or nice name?

henrymercer: Ah good point — I missed that while looking at just the diff.  I mildly prefer `VersionInfo`, but either is fine.

src/testing-utils.ts
```diff
@@ -208,10 +208,17 @@ export function mockLanguagesInRepo(languages: string[]) {
   return listLanguages;
 }
 
-export function mockCodeQLVersion(version) {
+/**
+ * Constructs a `VersionInfo` object for testing purposes only.
+ */
+export const makeVersionInfo = (version: string): CodeQL.VersionInfo => ({
```
henrymercer, line 214: Non-blocking: Out of curiosity, why convert to `export const`?

henrymercer (approved):

src/testing-utils.ts
mbg: (It was `const` from the start.)

Mostly a style preference coming from functional programming languages and I find the semantics end up causing me to shoot myself in the foot less often, but not sure there's much of a concrete benefit in this particular instance. Happy to change it to `function` if you have some good reasons (debugging?) for having that.

PR #1923, opened 2023-10-05
Fix using the `resolve-environment` Action with language aliases

henrymercer:
This PR fixes a bug that impacted the rollout of language aliases to default setup configurations that analyzed Go alongside at least one of the following languages:

- C/C++
- Java/Kotlin
- JavaScript/TypeScript

The problem was that while we passed the CLI option to enable language aliasing to `resolve languages`, we also needed to pass it to `resolve build-environment`.

We have mitigated this by disabling language aliases in default setup until this fix is out.


src/resolve-environment.ts
```diff
@@ -1,17 +1,36 @@
...
-  logger.startGroup(`Attempting to resolve build environment for ${language}`);
+  logger.startGroup(
+    `Attempting to resolve build environment for ${languageInput}`,
+  );
 
   const codeql = await getCodeQL(cmd);
+
+  let language = languageInput;
+  if (await util.codeQlVersionAbove(codeql, CODEQL_VERSION_LANGUAGE_ALIASING)) {
+    // Delegate to the CodeQL CLI to handle aliasing.
```
mbg, line 24: Is there something meant to be here?

henrymercer: No, in the case that the CLI supports the language aliasing we give it the raw language input.  Perhaps it's clearest to just check the negation.

aeisenberg (approved):

PR #1928, opened 2023-10-06
Add a fix for python 3.12

aeisenberg:
The python extractor does not yet support 3.12. Check for this and instead make sure we run python 3.11. Only need to check on windows since we are extremely unlikely to be running 3.12 on linux or macos.


python-setup/check_python12.ps1
```diff
@@ -0,0 +1,19 @@
...
+Write-Host "Checking python version"
+if ((py -3 -c "import sys; print(0 if sys.version_info >= (3, 12) else 1)") -eq "0") {
+  Write-Host "Python 3.12+ detected, setting PY_PYTHON3=3.11"
+  # First make sure we have python 3.11 installed
+  py -3.11 -c "import imp"
+  if ($LASTEXITCODE -eq 0) {
+    Write-Host "Python 3.11 detected, using this version instead of 3.12+."
+    Write-Output "PY_PYTHON3=3.11" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+  } else {
+    Write-Host "FAILURE: Python 3.12+ is not supported. Please install Python 3.11."
```
adityasharad, line 14: Optional.
```suggestion
    Write-Host "FAILURE: Python 3.12+ is not supported, and Python 3.11 could not be detected on the system. Please install Python 3.11."
```

adityasharad (commented):
Clever. Looks reasonable.
Can we have an integration test on Windows that calls `setup-python` with 3.12 to exercise this path?

python-setup/check_python12.ps1
```diff
@@ -0,0 +1,19 @@
+
+#! /usr/bin/pwsh
+
+# If we are running greater than or equal to python 3.12, change py to run version 3.11
+Write-Host "Checking python version"
```
angelapwen, line 5:
```suggestion
Write-Host "Checking Python version"
```

Super-nit

aeisenberg:
Should be simple enough to create a test.

python-setup/check_python12.ps1
aeisenberg: Actually, I think all `Python` should be `python`.

criemen:
I think we should query a feature flag from the CLI so that we can turn off this workaround from within the CLI once we've a release out that fixes the python extractor.

aeisenberg:
Good call. I can check for less than 2.15.0.

criemen:
I think the new tool-features feature (see https://github.com/github/codeql-action/pull/1909/files#diff-a2b4f46aa09bd648db18893b59331e2e037dfb60954c80383d820703b09d2519) would be more appropriate to use.

aeisenberg:
That requires adding the feature to the CLI directly, no?

aeisenberg:
Ah...I see what you're saying. I choose a feature name now and that will be the feature name in the version of codeql that has the fix. Good idea.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add a workaround python 3.12, which is not supported in CodeQL CLI version 2.14.6 or earlier. If you are running an analysis on windows and using python 3.12 or later, the CodeQL Action will switch to running python 3.11.
```
angelapwen, line 7:
```suggestion
- Add a workaround python 3.12, which is not supported in CodeQL CLI version 2.14.6 or earlier. If you are running an analysis on Windows and using python 3.12 or later, the CodeQL Action will switch to running python 3.11.
```

Also, should we include a note about if 3.11 isn't installed, the workflow will fail and the user should install it?

angelapwen (approved):
Looks good modulo question on the changelog file!

CHANGELOG.md
aeisenberg: Yes...

aeisenberg:
```suggestion
- Add a workaround python 3.12, which is not supported in CodeQL CLI version 2.14.6 or earlier. If you are running an analysis on Windows and using python 3.12 or later, the CodeQL Action will switch to running python 3.11. If python 3.11 is not found, then the workflow will fail.
```

angelapwen (approved):

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add a workaround python 3.12, which is not supported in CodeQL CLI version 2.14.6 or earlier. If you are running an analysis on Windows and using python 3.12 or later, the CodeQL Action will switch to running python 3.11. If python 3.11 is not found, then the workflow will fail.
```
angelapwen, line 7:
```suggestion
- Add a workaround for python 3.12, which is not supported in CodeQL CLI version 2.14.6 or earlier. If you are running an analysis on Windows and using python 3.12 or later, the CodeQL Action will switch to running python 3.11. If python 3.11 is not found in this case, then the workflow will fail.
```
Just realized it's missing a word..

aeisenberg: Thanks I'll put this in a follow-up.

brumboy59 (approved):

PR #1473, opened 2023-01-10
Temporarily disable Kotlin analysis in PR checks

henrymercer:
Kotlin analysis is incompatible with Kotlin 1.8.0, which is now rolling out to the Actions runner images.

While we work on a more permanent fix to our PR checks, this will prevent us losing other test coverage.

### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #1939, opened 2023-10-10
Re-enable kotlin tests

aeisenberg:
Reverts https://github.com/github/codeql-action/pull/1473


aeisenberg:
Hmmm...now it's failing, but because there's an even newer version of kotlin. We are going to need to choose a specific version of kotlin based on the version of the CLI.

aeisenberg:
Converting to draft mode since I won't be able to work on this. If someone else wants to pick it up, please do so.

iamannalynleynes1206 (approved):

aeisenberg:
Not going to be able to get to this.

aeisenberg:
This should now be working properly. The version of kotlin is fixed to 1.8.21. We can consider upgrading the version in the future, or even running against multiple versions if we like.

angelapwen (approved):
Thanks for getting back to this!! ✨

PR #1940, opened 2023-10-11
Rename feature flag for evaluator fine-grained parallelism

nickrolfe:
There was a bug in the implementation of fine-grained parallelism in 2.14.x, making it unsafe to enable the feature flag. By renaming the feature flag, we can safely enable it without turning the feature on for people using 2.14.x and old versions of the Action.


henrymercer (dismissed):

nickrolfe:
Marking ready for review, since we've now reached sufficient confidence in the feature to begin rolling it out for 2.15.1 onwards.

henrymercer (approved):

Issue #1959, opened 2023-10-18
📢 Node 16 deprecation, upcoming CodeQL Action v3 📢


angelapwen:
[Node.js 16 reached end of life](https://nodejs.org/en/blog/announcements/nodejs16-eol) last month, September 2023, half a year before its original anticipated EOL date. GitHub Actions will begin to display a warning to users notifying them about the upcoming migration starting October 23, 2023 ([GitHub Changelog post](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)). 

Users of the CodeQL Action and code scanning workflows on GitHub.com, please know that:
- You will begin to see these warnings in your Actions logs on code scanning runs starting October 23, 2023.
- All code scanning workflows should continue to succeed regardless of the warning.
- The team at GitHub maintaining the CodeQL Action is aware of the deprecation timeline and actively working on creating another version of the CodeQL Action, v3, that will bump us to Node 20. 

This note will be linked to from the Action CHANGELOG as well as the repository README.

Please let us know in comments on this issue if there are any questions or concerns. Thank you!

TWiStErRob:
Is this fixed by #2006? Can you please create a GitHub release for the new tag? (and maybe explain why it's 3.22 and not 3.0)

Background: I got a Renovate PR, but don't understand the new release.

aeisenberg:
Yes, v3 is now available and supports node 20. We're working through the details of exactly how to upgrade existing users from v2 to v3. but #2006 is a major part of deprecating node 16. 

We have released 3.22 since v3 is identical to v2 except for the node version. This is an easy way for us (and for users) to track exactly which features you are getting.

Can you explain why you would like a new release for this? We generally only use releases for new codeql CLI versions, which are largely independent of the codeql action version.

TWiStErRob:
Ah, CLI, that makes sense, that's why I didn't get it. I thought it was missing the release for 3.22.

So the action "release notes" are only in CHANGELOG? If so, adding the middle paragraph of your above reply to that file might help people understand the version number, rather than implying/inferring it. (It makes total sense, but unconventional.)

aeisenberg:
Thanks for the feedback. I'll let the team know.

mydea:
Hey there, just cross-posting from https://github.com/github/codeql-action/issues/2059:

[...]

PR #1960, opened 2023-10-18
Add announcement on Node 16 deprecation

angelapwen:
Make the announcement in the [issue](https://github.com/github/codeql-action/issues/1959) visible via the Action changelog and the README. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,10 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Users will begin to see warnings on Node.js 16 deprecation in their Actions logs on code scanning runs starting October 23, 2023.
+  - All code scanning workflows should continue to succeed regardless of the warning.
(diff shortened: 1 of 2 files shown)


aeisenberg (approved):
Looks good.

One more suggestion. Can we detect when this warning is emitted and emit a second warning pointing to the issue for more information (only if the original warning is also emitted)?

angelapwen:
Thanks!

> One more suggestion. Can we detect when this warning is emitted and emit a second warning pointing to the issue for more information (only if the original warning is also emitted)?

Hm.. that's an interesting idea. I'm not sure actually, I think it will be more clear once the warning is actually emitted. But so far I don't think we've used a mechanism to catch underlying Actions warnings.

aeisenberg:
After the warning starts being emitted, it will be emitted for _every_ workflow run? If so, we can wait until it's emitted and then include a custom warning that links to the issue.

angelapwen:
> After the warning starts being emitted, it will be emitted for _every_ workflow run? If so, we can wait until it's emitted and then include a custom warning that links to the issue.

Yes, I think so — at least that's what I understood from the Changelog post: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

PR #1963, opened 2023-10-20
Use the correct action input parameter

aeisenberg:
Fixes an integration test that isn't quite testing what we think it is.

Also, ensures that all macos tests run on python 3.11. See separate commits.

### DIFF
--- a/pr-checks/checks/with-checkout-path.yml
+++ b/pr-checks/checks/with-checkout-path.yml
@@ -1,29 +1,43 @@
 name: "Use a custom `checkout_path`"
 description: "Checks that a custom `checkout_path` will find the proper commit_oid"
 steps:
+  # This ensures we don't accidentally use the original checkout for any part of the test.
+  - name: Delete original checkout
+    shell: bash
+    run: |
+      # delete the original checkout so we don't accidentally use it.
+      # Actions does not support deleting the current working directory, so we
+      # delete the contents of the directory instead.
+      rm -rf ./* .github .git
   # Check out the actions repo again, but at a different location.
   # choose an arbitrary SHA so that we can later test that the commit_oid is not from main
   - uses: actions/checkout@v4
     with:
       ref: 474bbf07f9247ffe1856c6a0f94aeeb10e7afee6
       path: x/y/z/some-path
+
   - uses: ./../action/init
     with:
       tools: ${{ steps.prepare-test.outputs.tools-url }}
       # it's enough to test one compiled language and one interpreted language
       languages: csharp,javascript
-      source-path: x/y/z/some-path/tests/multi-language-repo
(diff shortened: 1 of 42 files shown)


angelapwen:
Looks like we're not allowed to remove the current directory in Actions.

aeisenberg:
When I get a chance, I'll selectively remove a few key directories to ensure we're not accidentally using them.

aeisenberg:
@angelapwen can you take another look? I think this should work now.

angelapwen (dismissed):
Fix looks good, but we'll need to fix the Python 3.12 issue causing the tests to fail separately.

angelapwen:
Re-ran and we got an older version of the runner so PR checks have passed 👍

angelapwen (approved):
Thanks!! I guess at some point this runner image will stop being `macos-latest` but looks good for now.

aeisenberg:
Updated to only apply python 3.11 for older CLIs. At some point, we can purge references to CLIs we no longer support, but that can happen at some point in the future.

pr-checks/checks/with-checkout-path.yml
```diff
@@ -1,29 +1,43 @@
...
+      source-root: x/y/z/some-path/tests/multi-language-repo
       debug: true
+
   - name: Build code (non-windows)
     shell: bash
     if: ${{ runner.os != 'Windows' }}
+    working-directory: x/y/z/some-path/tests/multi-language-repo
     run: |
-      $CODEQL_RUNNER x/y/z/some-path/tests/multi-language-repo/build.sh
+      $CODEQL_RUNNER ./build.sh
```
henrymercer, line 32: I don't think `$CODEQL_RUNNER` is needed, but we can address that separately.

pr-checks/sync.py
```diff
@@ -74,6 +74,23 @@ def writeHeader(checkStream):
             useAllPlatformBundle = checkSpecification['useAllPlatformBundle']
 
     steps = [
+        {
+            'name': 'Setup Python on MacOS',
+            'uses': 'actions/setup-python@v2',
```
henrymercer, line 79: Nit: use v4

henrymercer (approved):
Thanks!  Couple of non-blocking suggestions.

pr-checks/sync.py
aeisenberg: Changed this as well as the other references to `setup-python`.

pr-checks/checks/with-checkout-path.yml
henrymercer: I think per the paragraph below what your linked, the tracer should inject the `$CODEQL_RUNNER` prefix for us.  If this doesn't work, we should update the docs so users are aware when this is needed.

PR #1967, opened 2023-10-25
Enable analysis summary v2 and sub-language file coverage information on GHES

henrymercer:
This PR:

- Enables analysis summary v2 on GHES 3.9 and later (since the new summary currently links to the status page unconditionally).
- Enables sub-language file coverage information on the upcoming GHES 3.12 and later.
- Cleans up the feature flag for exporting failed SARIF.


src/codeql.ts
```diff
@@ -317,6 +316,12 @@ const GHES_MOST_RECENT_DEPRECATION_DATE = "2023-09-12";
  * flag is older than the oldest supported version above, it may be removed.
  */
 
+/**
+ * Versions 2.13.3+ of the CodeQL CLI support exporting a failed SARIF file via
+ * `codeql database export-diagnostics` or `codeql diagnostics export`.
+ */
+export const CODEQL_VERSION_EXPORT_FAILED_SARIF = "2.11.3";
```
aeisenberg, line 323: Is this what you meant? At least that's what the comment above says.

```suggestion
export const CODEQL_VERSION_EXPORT_FAILED_SARIF = "2.13.3";
```

src/feature-flags.ts
```diff
@@ -24,16 +24,6 @@ export const CODEQL_VERSION_BUNDLE_SEMANTICALLY_VERSIONED = "2.13.4";
  */
 export const CODEQL_VERSION_INTRA_LAYER_PARALLELISM = "2.14.6";
 
-/**
```
aeisenberg, line 27: Should we be moving all version constraints outside of this file and keep them all in `codeql.ts`?

src/codeql.ts
henrymercer: Thanks for the catch, the comment is incorrect.  I moved this over from `UploadFailedSarifEnabled` in `features.ts`.

src/feature-flags.ts
henrymercer: We have to keep any version constraints that the feature flags depend on in this file to avoid circular dependencies between `feature-flags.ts` and `codeql.ts`.  We could add another file just for version constraints, but it feels a little overkill.

angelapwen (dismissed):
Looks good, thanks for the 🧹 cleanup on failed SARIF too!

henrymercer:
Now I get an opportunity to use #1971 :)

henrymercer:
Well, I'm not too surprised that didn't work.  Pushing a fix.

angelapwen:
Ah.. I'd also thought that it'd check out the current branch by default, but it does make sense that the checkout action isn't aware of the current branch ref.

aeisenberg (approved):

henrymercer:
> Ah.. I'd also thought that it'd check out the current branch by default, but it does make sense that the checkout action isn't aware of the current branch ref.

I think it checks out the synthetic merge commit vs the head of the branch, and the latter is what we want to push to.

PR #1970, opened 2023-10-26
Validate workflow within new log group

henrymercer:
This cleans up the logs for `init`.


angelapwen (dismissed):
🚢

angelapwen:
Although this seems to come out empty — I wonder if that's confusing to users? Does it need to run in debug mode to actually show something in the group? 

Screenshot 2023-10-26 at 10 17 11 AM

henrymercer:
> Although this seems to come out empty — I wonder if that's confusing to users? Does it need to run in debug mode to actually show something in the group?

Our test workflows are a little unusual since they don't have the typical `languages` matrix configuration.  Therefore we're not seeing the CodeQL CLI logs for `codeql resolve languages`, since that validation isn't being called.

I think the best way to address this is to add a log that no issues were detected.

angelapwen (approved):
Looks good 👍

PR #1971, opened 2023-10-26
Add workflow to rebuild the Action on a label

henrymercer:
Reduce friction when applying suggestions or handling merge conflicts via the GitHub UI.  Previously we needed to go back into the developer environment and rebuild.  Now we can apply a label and have Actions do it for us.

### DIFF
--- a/.github/workflows/rebuild.yml
+++ b/.github/workflows/rebuild.yml
@@ -1,0 +1,58 @@
+name: Rebuild Action
+
+on:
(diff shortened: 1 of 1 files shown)


angelapwen (approved):
Wow very nice! Will makes things easier when we have external contributions as well. 

Tested the label on this PR and has worked at least in the case where there's nothing to rebuild 😄

PR #1981, opened 2023-11-04
Delete analysis after uploading

aeisenberg:
The analysis is purposefully failing. We don't want a failed analysis sitting in the security center since this can cause some internal checks to erroneously fail.


aeisenberg:
I think this is working now, but I am confused by something. The call to `/repos/github/codeql-action/code-scanning/analyses?sarif_id=XXX` seems to be returning multiple analyses when I call it in the workflow. Some of the results are overlapping. 

For example, you can see here https://github.com/github/codeql-action/actions/runs/6791589351/job/18463369144?pr=1981#step:22:44 that there are a number of analyses that we attempt to delete, but they don't exist. They have already been deleted by another job in the workflow, ie- https://github.com/github/codeql-action/actions/runs/6791589351/job/18463373448?pr=1981#step:18:53.

Two questions:

1. Why is the single SARIF upload generating so many analyses?
2. Why are two calls to retrieve analyses from different SARIF uploads retrieving a partially overlapping set of analyses?

aeisenberg:
This PR is now ready for review. I separated out the changes for the debug-artifacts-upload workflows, since we can and should merge that first in order to get main passing again.

src/init-action-post-helper.ts
```diff
@@ -193,3 +203,74 @@ export async function run(
...
 }
+
+async function removeUploadedSarif(
+  uploadFailedSarifResult: UploadFailedSarifResult,
+) {
+  const sarifID = uploadFailedSarifResult.sarifID;
+  if (sarifID) {
+    core.startGroup("Deleting failed SARIF upload");
+    core.info(
+      `Uploaded failed SARIF file with ID ${sarifID}. Because this is a test, the analysis associated with it will now be deleted.`,
```
henrymercer, line 214:
```suggestion
      `In test mode, so deleting the failed analysis to avoid impacting tool status for the Action repository. SARIF ID to delete: ${sarifID}.`,
```

src/init-action-post-helper.ts
```diff
@@ -193,3 +203,74 @@ export async function run(
 
   return uploadFailedSarifResult;
 }
+
+async function removeUploadedSarif(
+  uploadFailedSarifResult: UploadFailedSarifResult,
+) {
```
henrymercer, line 209: Minor style comment: We tend to pass a `logger` object around explicitly rather than using `core` directly.  This makes it easier to test logging calls, though was mainly an artifact from the CodeQL runner, which couldn't depend on Actions functionality. Perhaps we should consider using `core` directly and testing via spies instead.

src/init-action-post-helper.ts
```diff
@@ -193,3 +203,74 @@ export async function run(
...
+    const client = getApiClient();
+
+    try {
+      const repositoryNwo = parseRepositoryNwo(
+        getRequiredEnvParam("GITHUB_REPOSITORY"),
+      );
+
+      // Get the analysis associated with the uploaded sarif
+      const analysisInfo = await client.request(
+        "GET /repos/:owner/:repo/code-scanning/analyses?sarif_id=:sarif_id",
```
henrymercer, line 225: Should we perhaps wait a few seconds before calling this endpoint, just in case there is some lag between processing the SARIF file and it being available via this API?

henrymercer (commented):
Nice, a bunch of suggestions, but this looks good.

src/init-action-post-helper.ts
```diff
@@ -193,3 +203,74 @@ export async function run(
...
+        core.info(`Deleting analysis ${analysis.id}`);
+        try {
+          await client.request(
+            "DELETE /repos/:owner/:repo/code-scanning/analyses/:analysis_id?confirm_delete",
+            {
+              owner: repositoryNwo.owner,
+              repo: repositoryNwo.repo,
+              analysis_id: analysis.id,
+            },
+          );
```
henrymercer, line 245: Minor: Log that we successfully deleted the analysis so this is obvious when we read the logs.

.github/workflows/python312-windows.yml
```diff
@@ -38,4 +38,5 @@ jobs:
     - name: Analyze
       uses: ./../action/analyze
       with:
+        upload: false
```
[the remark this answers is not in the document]
aeisenberg: I'll create a tech debt issue for this.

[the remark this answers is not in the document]
aeisenberg: https://github.com/github/codeql-core/issues/3988

src/init-action-post-helper.ts
```diff
@@ -29,6 +30,9 @@ export interface UploadFailedSarifResult extends uploadLib.UploadStatusReport {
   upload_failed_run_stack_trace?: string;
   /** Reason why we did not upload a SARIF payload with `executionSuccessful: false`. */
   upload_failed_run_skipped_because?: string;
+
+  /** The internal ID of SARIF analysis. */
+  sarifID?: string;
```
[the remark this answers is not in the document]
aeisenberg: I'm not sure about that. Are you thinking about the `analysis_id`? That is a number. The `sarid_id` is a guid.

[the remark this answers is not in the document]
henrymercer: Sorry, I mixed these up, thanks for the correction.

src/init-action-post-helper.ts
aeisenberg: I'll try it and see if it gets too messy to pass around the logger

src/init-action-post-helper.ts
aeisenberg: I don't know if that is necessary? I guess there's no harm in waiting 5 seconds.

src/init-action-post-helper.ts
```diff
@@ -193,3 +203,74 @@ export async function run(
...
+            },
+          );
+        } catch (e) {
+          if (
+            e instanceof Error &&
+            e.message.includes("No analysis found for analysis ID")
+          ) {
+            core.info(
+              `Analysis ${analysis.id} does not exist. It was likely already deleted.`,
+            );
```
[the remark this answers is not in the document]
aeisenberg: I can make it a warning now. Earlier, there were race conditions, but that was because I had an incorrect url that was not filtering by sarifID.

aeisenberg:
Looks like we can't use the `cause` option in `Error` yet.

src/util.ts
```diff
@@ -909,6 +909,14 @@ export function wrapError(error: unknown): Error {
...
 }
 
+export function getErrorMessage(error: unknown): string {
+  return error instanceof Error ? error.toString() : String(error);
+}
+
+export async function wait(ms: number) {
+  await new Promise((resolve) => setTimeout(resolve, ms));
+}
+
```
henrymercer, line 919: I think we can reuse `delay` instead of implementing a new function.

henrymercer (dismissed):
Looks good, just one minor comment about the changes.

src/util.ts
aeisenberg: Right...forgot about that function.

henrymercer (approved):
Nice

PR #2013, opened 2023-12-01
Extract scanned languages using `--index-traceless-dbs`

henrymercer:
This is a more elegant way to extract scanned languages that avoids the need to compute the path to the autobuilder in the Action.  It was introduced in CodeQL v2.5.6 so is supported by all versions of CodeQL we need to support.


angelapwen (approved):
Nice ✨

PR #2041, opened 2023-12-15
improve handling of changelog processing for backports

nickfyson:
This PR fixes handling of backporting the changelog when the only listed change for a version is not applicable for the older targeted release.


.github/update-release-branch.py
```diff
@@ -174,6 +175,58 @@ def get_today_string():
   today = datetime.datetime.today()
   return '{:%d %b %Y}'.format(today)
 
+def process_changelog_for_backports(source_branch_major_version, target_branch_major_version):
+
+  # changelog entries use a speficic format to indicate
```
henrymercer, line 180:
```suggestion
  # changelog entries can use the following format to indicate
```

.github/update-release-branch.py
```diff
@@ -174,6 +175,58 @@ def get_today_string():
...
+  # that they only apply to newer versions
+  regex = re.compile(r'\[v(\d+)\+ only\]')
```
henrymercer, line 182: Rename to `some_versions_only_regex`?

.github/update-release-branch.py
```diff
@@ -174,6 +175,58 @@ def get_today_string():
...
+        if int(target_branch_major_version) < int(match.group(1)):
+          continue
+
+      if line.startswith('## '):
+        line = line.replace(f'## {source_branch_major_version}', f'## {target_branch_major_version}')
+        if found_content == False:
+          # we have found two headings in a row, so we need to add the placeholder message.
+          output += 'No user facing changes.\n'
+        found_content = False
+        output += '\n' +line + '\n\n'
```
henrymercer, line 219:
```suggestion
        output += f'\n{line}\n\n'
```

henrymercer (dismissed):
Looks good.  Another approach would be parsing the document in full into a version -> changelog notes map, performing our tweaks on the map, then rewriting to Markdown.  I think that might be more readable, but it's probably not better enough to justify changing it — what do you think?

.github/update-release-branch.py
nickfyson: 👍

henrymercer (dismissed):

nickfyson:
Yeah, I started doing that and then decided it was overkill. But then had to work pretty hard to make this approach even vaguely grokkable. 😅

I think I'll leave it with this approach, and if/when it gets more complicated then refactor it 'properly'. 😄

.github/update-release-branch.py
```diff
@@ -174,6 +175,58 @@ def get_today_string():
...
+
+  output = ''
+
+  with open('CHANGELOG.md', 'r') as f:
+
+    # until we find the first section, just duplicate all lines
+    while True:
+      line = f.readline().rstrip()
+
+      output += line + '\n'
```
aeisenberg, line 192: This is minor and non-blocking, but why `rstrip` and then add `\n`? If the original changelog has trailing whitespace (even though it shouldn't), then would it be best to keep the new one as close to the original as possible?

.github/update-release-branch.py
```diff
@@ -174,6 +175,58 @@ def get_today_string():
...
+      if line.startswith('## '):
```
aeisenberg, line 193: We _shouldn't_ have to handle this, but this script will fail if there is no `## ` line. Maybe just to be safe include a fallback to handle this case (I could imagine this _might_ happen if we're testing something).

aeisenberg (approved):
Looks reasonable to me. Two non-blocking suggestions.

.github/update-release-branch.py
nickfyson: I've added an explicit error message in this situation, to try and avoid confusion should it arise. 👍

aeisenberg (approved):

PR #2048, opened 2023-12-20
backport checks changes to v2 branch

nickfyson:
In order to keep alignment of check names across all the maintained branches, we backport the matrix job to the `releases/v2` branch.

The check runs on node 16 and skips for the current version (node 20).

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -15,6 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 45
 
+    strategy:
+      matrix:
+        node-types-version: [16.11, current] # we backport this matrix job in order to maintain the same check names
+
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


adityasharad (approved):

.github/workflows/pr-checks.yml
```diff
@@ -15,14 +15,41 @@ jobs:
...
       - name: Lint
         run: npm run-script lint
 
+      - name: Update version of @types/node
+        if: matrix.node-types-version != 'current'
+        env:
+          NODE_TYPES_VERSION: ${{ matrix.node-types-version }}
+        run: |
+          # Export `NODE_TYPES_VERSION` so it's available to jq
+          export NODE_TYPES_VERSION="${NODE_TYPES_VERSION}"
```
aeisenberg, line 35: I didn't know this was necessary.

Pretty sure this works the same:

```suggestion
          export NODE_TYPES_VERSION
```

.github/workflows/pr-checks.yml
```diff
@@ -15,14 +15,41 @@ jobs:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Lint
         run: npm run-script lint
 
+      - name: Update version of @types/node
+        if: matrix.node-types-version != 'current'
+        env:
```
aeisenberg, line 31:
```suggestion
        shell: bash
        env:
```

aeisenberg (approved):
I think this works.  Though, is there a benefit to running the checks on a different node version on v2? I was imagining that when the matrix value is not `current`, then the job can just trivially end in success. If there's a reason to run on both versions, then let's do it.

PR #2051, opened 2023-12-21
Fix type error in `Update release branch` workflow

angelapwen:
Simply casts a string to an int 😸 

### DIFF
--- a/.github/actions/release-branches/release-branches.py
+++ b/.github/actions/release-branches/release-branches.py
@@ -13,7 +13,7 @@ config = configparser.ConfigParser()
 with open(os.path.join(grandparent_dir, 'releases.ini')) as stream:
     config.read_string('[default]\n' + stream.read())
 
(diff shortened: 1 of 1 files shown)


.github/actions/release-branches/release-branches.py
```diff
@@ -46,7 +46,7 @@ def main():
     if consider_backports:
       for i in range(int(major_version.strip("v"))-1, 0, -1):
         branch_name = f"releases/v{i}"
-        if i >= OLDEST_SUPPORTED_MAJOR_VERSION:
+        if i >= int(OLDEST_SUPPORTED_MAJOR_VERSION):
```
adityasharad, line 49: This is the only use of this variable so it works; the more durable option would be to typecast at the definition of `OLDEST_SUPPORTED_MAJOR_VERSION`.

angelapwen: Ah, that's fair — I had thought it might be easier to explicitly cast when used but I guess we'd always be using it as an int. I'll update.

adityasharad (approved):

PR #2061, opened 2024-01-04
Improve `paths` config handling for JS/Python when globs are used

RasmusWL:


RasmusWL:
I have only tested this change against the Python extractor implementation, where I know the _ordering_ of LGTM_INDEX_FILTER is respected. I assume it would be by the other languages as well, but better be safe :blush: 

So I want to make an integration test to ensure that with a directory structure as below, a config with `paths: ['include/foo*']`, should only extract ONE file for Python (and likewise for JS/Ruby)

```
.
├── include
│   ├── bar.py
│   └── foo.py
└── top.py
```

or do you rather think that should be part of the tests for the extractors? (let me know what you think)

henrymercer:
Thanks for looking into this, this fix looks good.  However I think we want to implement it in the CLI rather than the Action.

Context: We moved the logic for computing `LGTM_INDEX_FILTERS` into the CLI as of v2.11.6, the motivation being that we want this functionality to work for customers using third-party CI systems.  Since then, we have had to maintain both the CLI implementation and a duplicate implementation in the Action so we can continue supporting older CLIs.

As of today, GHES 3.7 was deprecated, and therefore we've been able to drop support for CodeQL v2.11.5 and earlier in the Action.  This means we no longer need to maintain the implementation of this functionality in the Action, and I've put up a PR to remove it here: https://github.com/github/codeql-action/pull/2062

Would you mind porting your implementation to the CLI?  If not, I'm happy to take this on since it's a small change — let me know!

PR #2062, opened 2024-01-04
Remove config parsing logic that is now implemented in the CLI

henrymercer:
In CodeQL CLI v2.11.6, we finished implementing key parts of parsing the code scanning configuration within the CLI.  This enabled this functionality to be used by folks running code scanning using a third-party CI system.

However due to the need to support earlier versions of the CLI that do not support config parsing, we have needed to duplicate config parsing code between the Action and the CLI.

Now that CodeQL CLI v2.11.5 and earlier have been deprecated, we can remove this legacy config parsing code from the Action.


angelapwen (dismissed):
Whew, I think I 👀 d all the changes — looks so much cleaner now! I assume the status report changes will be a follow-up?

henrymercer:
There are two aspects to the status report changes:

1. Renaming `builtin` to `all` — I think this is fairly low priority and can happen as follow up
2. Removing the `custom` fields on the server-side — I think it might make sense to keep these around a little longer so we still receive the full telemetry from earlier Actions, so we'll do this as follow up too.

angelapwen (approved):

PR #2065, opened 2024-01-05
Clean up running queries some more

henrymercer:
Now that we have deprecated parsing the configuration in the Action, we can simplify running queries to reflect that:

- We will run all the queries for a language in one group, so we can always pass `--expect-discarded-cache`.
- We no longer specify query suites when running queries, so we can delete the logic for doing this.  Instead the query suite is stored within the database.


angelapwen (approved):
👍 looks good to me!

PR #2077, opened 2024-01-10
Dependabot alerts for shield only

aeisenberg:


angelapwen (approved):
Great!

henrymercer:
I suspect `@github/codeql-action-reviewers` will still be notified as CODEOWNERS, but this still makes sense as we want to notify the production shield even if they don't happen to be in `@github/codeql-action-reviewers`.

henrymercer (commented):
I suspect `@github/codeql-action-reviewers` will still be notified as CODEOWNERS, but this still makes sense as we want to notify the production shield even if they don't happen to be in `@github/codeql-action-reviewers`.

angelapwen:
Yep! https://github.com/github/codeql-action/pull/2078

PR #2080, opened 2024-01-11
Fix `paths`/`paths-ignore` warning that would appear unconditionally

henrymercer:
This warning should only appear on traced languages if the user has configured `paths` or `paths-ignore`, however it was appearing unconditionally.  This is because `undefined?.length !== 0` evaluates to `true`.


tamasvajk (dismissed):
LGTM

PR #2083, opened 2024-01-12
Respect Linux `cgroup` CPU number for `--threads` value

angelapwen:
We want to respect `cgroup` constraints so that when we run in a container, the default max threads value is appropriately set to the maximum number of cores in this container rather than the host OS's.

We check both `/sys/fs/cgroup/cpuset.cpus.effective` (`cgroup v2`) and `/sys/fs/cgroup/cpuset.cpus` (`cgroup v1`) to find the number of cores.


src/util.ts
```diff
@@ -380,6 +395,73 @@ export function getThreadsFlagValue(
...
+): number | undefined {
+  if (!fs.existsSync(cpuMaxFile)) {
+    logger.debug(
+      `While resolving threads, did not find a cgroup CPU file at ${cpuMaxFile}.`,
+    );
+    return undefined;
+  }
+
+  const cpuMaxString = fs.readFileSync(cpuMaxFile, "utf-8");
+  const cpuLimit = cpuMaxString.split(" ")[0];
```
henrymercer, line 415: Optional: log a warning and return `undefined` if there are more than two values

src/util.ts
```diff
@@ -380,6 +395,73 @@ export function getThreadsFlagValue(
...
+    );
+    return undefined;
+  }
+
+  let cpuCount = 0;
+  // Comma-separated numbers and ranges, for eg. 0-1,3
+  const cpusString = fs.readFileSync(cpusFile, "utf-8");
+  for (const token of cpusString.split(",")) {
+    // If it's a single character
+    if (token.length === 1) {
```
henrymercer, line 448: What happens if we have something like `10`?

henrymercer (commented):
Nice!  A couple of comments:

src/util.ts
angelapwen: Ah... you're right 😆 nice catch

angelapwen: Checked for if it includes the `-` character instead.

src/util.ts
```diff
@@ -380,6 +395,79 @@ export function getThreadsFlagValue(
...
+  let cpuCount = 0;
+  // Comma-separated numbers and ranges, for eg. 0-1,3
+  const cpusString = fs.readFileSync(cpusFile, "utf-8");
+  for (const token of cpusString.split(",")) {
+    if (!token.includes("-")) {
+      // Not a range
+      ++cpuCount;
+    } else {
+      const cpuStartIndex = parseInt(token.charAt(0));
+      const cpuEndIndex = parseInt(token.charAt(2));
```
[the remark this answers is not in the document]
angelapwen: 👍 done! Thanks!

henrymercer (approved):
Looks good, thanks!

PR #2084, opened 2024-01-12
Only delete SARIF in PR check if not running on a fork 

angelapwen:
The `Submit SARIF after failure` PR Check was failing when opened on a fork because of a permissions problem when deleting the uploaded SARIF. This change should fix this by only deleting the SARIF when the head of the pull request is explicitly not a fork.

Tested because this PR was opened on a fork and the PR check is passing, and the SARIF deletion step is not entered 😸 I also pushed the same commits to a branch on this repo and the deletion behavior is preserved: https://github.com/github/codeql-action/pull/2085 (draft PR)

Also fixed some linting errors in the `analyze-action.ts` file as a drive-by second commit.


src/init-action-post-helper.ts
```diff
@@ -182,7 +182,12 @@ export async function run(
     );
   }
 
-  if (process.env["CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF"] === "true") {
+  // We do not delete uploaded SARIFs if we're on a fork, as we're missing the
+  // appropriate permissions.
+  if (
+    process.env["CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF"] === "true" &&
+    repositoryNwo.owner !== "github"
```
aeisenberg, line 189: Is this the correct thing to do?

```suggestion
    repositoryNwo.owner === "github"
```

angelapwen: You're right.. I think there's some other logical bug here given that the checks passed with this conditional incorrect. Will do some print debugging.

angelapwen:
Okay, I've finally worked out how to differentiate the behavior between fork and non-fork. The Actions runs linked to the PR are a bit confused because the branch in my fork is named the same as the branch in this repo, but here are the correct associated runs:

Run from a fork (this PR), where the `Deleting failed SARIF upload` step does not run: https://github.com/github/codeql-action/actions/runs/7547884719/job/20548780850 

Run from not-a-fork, where the `Deleting failed SARIF upload` step does run: https://github.com/github/codeql-action/actions/runs/7547885210/job/20548798454

src/init-action-post-helper.ts
```diff
@@ -186,7 +188,7 @@ export async function run(
   // appropriate permissions.
   if (
     process.env["CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF"] === "true" &&
-    repositoryNwo.owner !== "github"
+    github.context.payload.pull_request?.head.repo.fork === false
```
aeisenberg, line 191: Nice. That's easy. Though, this will also fail to run for `push` and `workflow_dispatch` triggers. I think that neither of these would ever run in a fork.

So, this should be more generic since it also captures non-pr triggered runs:

```suggestion
    !github.context.payload.pull_request?.head.repo.fork
```

aeisenberg: And now that I think about it a little more, can you restructure this so that there is a log message emitted explaining that we don't delete the SARIF on forks when `CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF` is true?

```javascript
if (process.env["CODEQL_ACTION_EXPECT_UPLOAD_FAILED_SARIF"] === "true") {    
  if (github.context.payload.pull_request?.head.repo.fork === false) {
    // do delete
  } else {
    // log
}
```

angelapwen: That makes sense 👍 the log message emitted also means I can remove the code comment ✨ 

I've pushed the changes here and to the non-fork PR. I'll re-request review once the checks pass as expected on both.

aeisenberg:
Interesting...the check run for the SARIF failure CI job on this PR is matching to the CI job for #2085. 

See https://github.com/github/codeql-action/actions/runs/7548705896/job/20551327490

The code looks fine. I just clicked to re-run all the jobs.

aeisenberg (approved):
Anyway, I think this is good to go in.

angelapwen:
> Interesting...the check run for the SARIF failure CI job on this PR is matching to the CI job for #2085.
> 
> See https://github.com/github/codeql-action/actions/runs/7548705896/job/20551327490

Yes, Henry and I picked that up earlier 🤔 we think it's because I named the branches the same on both the fork and this repo, and they have the exact same commits...

PR #2087, opened 2024-01-15
C++: Add feature flag for TRAP caching.

sashabu:

 - Closes https://github.com/github/codeql-c-team/issues/2146


src/feature-flags.ts
```diff
@@ -68,6 +69,11 @@ export const featureConfig: Record<
     minimumVersion: "2.15.0",
     defaultValue: false,
   },
+  [Feature.CppTrapCachingEnabled]: {
+    envVar: "CODEQL_CPP_TRAP_CACHING",
```
henrymercer, line 73: Was your intention to set this environment variable to `true` when the feature is enabled?  The current design only uses `envVar` as an input (we will consider the feature flag enabled if the env var is true, but we won't set the env var to true if the feature flag is enabled).

To set the environment variable to true when the feature flag is on you can do something similar to what we do for Lombok in `init-action.ts`.

It's definitely a valid question whether feature flags _should_ work this way — I'm open to reconsidering this.

sashabu: > Was your intention to set this environment variable to true when the feature is enabled?

It was, thanks for catching this. Pushed a commit to address this.

sashabu:
> we probably want to let the `trap-caching` input override the value of the feature flag.

I don't think we do: the feature flag is intended for a gradual rollout rather than long-lived, so I think we do want the feature flag to override it. (And if I understand correctly, `CODEQL_CPP_TRAP_CACHING` will set the feature flag, so customers would have some control if they need to.)

> * We may want to set `minimumVersion` on the feature flag to avoid printing a log message like "Enabling CodeQL C++ TRAP caching support" when running against an old CLI that doesn't support TRAP caching.

We're not using the environment variable in production yet (or turning the feature flag on). What's the right way to go about `minimumVersion` in this case? Last release, next release, leave undefined for now and fix later, hold off merging this PR?

henrymercer:
> I don't think we do: the feature flag is intended for a gradual rollout rather than long-lived, so I think we do want the feature flag to override it. (And if I understand correctly, `CODEQL_CPP_TRAP_CACHING` will set the feature flag, so customers would have some control if they need to.)

That's fair — the expectation for `trap-caching: true` could be that TRAP caching is enabled for all languages for which it's GA.  However I do think `trap-caching: false` should override the feature flag, as this is a signal that the customer does not want us to populate Actions cache (typically because they are using it for something else and don't want their other caches to be evicted).

> We're not using the environment variable in production yet (or turning the feature flag on). What's the right way to go about `minimumVersion` in this case? Last release, next release, leave undefined for now and fix later, hold off merging this PR?

I think we should probably hold off on merging this PR until we know what the minimum CLI version is that we'd want to enable this for.  The alternative is to add a CLI flag to the `features` property of `codeql version` and check for that instead of using a version number.

sashabu:
> However I do think `trap-caching: false` should override the feature flag, as this is a signal that the customer does not want us to populate Actions cache (typically because they are using it for something else and don't want their other caches to be evicted).

Agreed. My last commit should do that.

> I think we should probably hold off on merging this PR until we know what the minimum CLI version is that we'd want to enable this for.

👍

henrymercer (dismissed):

henrymercer (approved):

PR #2088, opened 2024-01-17
Remove the Lombok feature flag

igfoo:
This essentially reverts https://github.com/github/codeql-action/pull/1809


henrymercer (commented):
This PR would disable Lombok for CLI v2.14.0 to 2.14.3, where it is enabled by the feature flag currently but it is not enabled by default in the CLI.  Should we change this to set `CODEQL_EXTRACTOR_JAVA_RUN_ANNOTATION_PROCESSORS` to `true` for CLIs >= v2.14.0?

igfoo:
Good point, thanks! I've made it check for 2.14.0-2.14.3, so we know when it's safe to remove it in the future.

I've amended the commits, so that the history is simpler (rather than removing code in one commit, then partially reinstating it in a later one).

src/init-action.ts
```diff
@@ -423,20 +423,19 @@ async function run() {
       core.exportVariable(kotlinLimitVar, "1.9.20");
     }
 
-    if (config.languages.includes(Language.java)) {
+    if (
+      config.languages.includes(Language.java) &&
+      (await codeQlVersionAbove(codeql, "2.14.0")) &&
+      !(await codeQlVersionAbove(codeql, "2.14.4"))
+    ) {
```
henrymercer, line 430: Optional: add a comment that Lombok is on by default for 2.14.4 and later.

henrymercer (approved):

henrymercer (approved):

PR #2092, opened 2024-01-18
add v2 deprecation warning

nickfyson:
This PR adds a deprecation warning when `v2` of any CodeQL Action is found to be running on a GitHub instance that is capable of upgrading to `v3`.

The following three workflow runs demonstrate these changes...

- [Run using `v2` of the action](https://github.com/github/codeql-action/actions/runs/7585306499)
- [Run with f72cffc cherry-picked, adding the warnings](https://github.com/github/codeql-action/actions/runs/7585325872)
- [Run with a854253 also cherry-picked, demonstrating that now only a single warning is shown.](https://github.com/github/codeql-action/actions/runs/7585505149)


aeisenberg (approved):

src/util.ts
```diff
@@ -944,3 +944,38 @@ export async function checkDiskUsage(logger?: Logger): Promise<DiskUsage> {
...
+        semver.satisfies(
+          semver.coerce(githubVersion.version) ?? "0.0.0",
+          ">=3.11",
+        ))
+    ) {
+      core.warning(
+        "CodeQL Action v2 will be deprecated on December 5th, 2024. Please upgrade to v3. For " +
+          "more information, see " +
+          "https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/",
+      );
```
henrymercer, line 978: We could optionally set an environment variable such that we only show this warning for the first Action you run, so it only appears once in the workflow logs (at least per job).

henrymercer (approved):

src/util.ts
nickfyson: (though, we're linking to the changelog where the individual actions are listed, so probably is better to just limit ourselves to one warning...)

henrymercer: We could change the phrasing to something like "Please update all occurrences of the CodeQL Action in your workflow files to v3"?

nickfyson: 👍 I've pushed updates, and included links in the description that demonstrate that the limit to one warning is working.

henrymercer (approved):

PR #2097, opened 2024-01-24
Send overall job status in init-post status report

angelapwen:


henrymercer (commented):
Looks good, a couple of thoughts:

henrymercer:
> * We also explicitly set the environment variable to `Success` if `ANALYZE_DID_COMPLETE_SUCCESSFULLY` is true. Note that in the case where analyze was successful, but some step between `analyze` and the `init-post` action fails, we'll incorrectly mark the job status as `Success`. Presumably this is a very uncommon scenario though 🤔

Just to check my understanding here, if one of the Actions we maintain is called after we successfully uploaded a SARIF file and it fails, then this will set the job status environment variable, so we'll have a failure.  If a third-party Action is called after we successfully uploaded a SARIF file, then we won't know about that so we'll mark this as a success.  This seems fine as code scanning has succeeded in that situation at analyzing the repo and uploading the analysis results.

src/environment.ts
```diff
@@ -35,6 +35,9 @@ export enum EnvVar {
   /** UUID representing the current job run. */
   JOB_RUN_UUID = "JOB_RUN_UUID",
 
+  /** Status for the entire job, submitted to the status report in `init-post` */
+  JOB_STATUS = "JOB_STATUS",
```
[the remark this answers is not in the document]
angelapwen: Done, and added a comment on top of the enum to remind myself and others 😆

src/status-report.ts
```diff
@@ -122,12 +130,27 @@ export interface DatabaseCreationTimings {
   trap_import_duration_ms?: number;
 }
 
+/** Returns the current ActionStatus and sets the JOB_STATUS environment variable
+ * to ConfigurationError or Failure, if not already set.
+ */
 export function getActionsStatus(
```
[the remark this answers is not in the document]
angelapwen: Yeah that makes sense! I added the extra function and returned this function to its original state

src/init-action-post.ts
```diff
@@ -83,6 +86,8 @@ async function runWrapper() {
   const statusReport: InitPostStatusReport = {
     ...statusReportBase,
     ...uploadFailedSarifResult,
+    job_status:
+      (process.env[EnvVar.JOB_STATUS] as JobStatus) ?? JobStatus.Unknown,
```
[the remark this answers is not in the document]
angelapwen: I actually only meant to check whether the value was `undefined` and only added the `as JobStatus` because of the compiler complaining 😆 but it's a good idea to be extra safe and make sure we're not setting some other value without updating the enum. I added the extra function to return `Unknown` if the value is `undefined` or if it's not recognized.

angelapwen:
> Just to check my understanding here, if one of the Actions we maintain is called after we successfully uploaded a SARIF file and it fails, then this will set the job status environment variable, so we'll have a failure. If a third-party Action is called after we successfully uploaded a SARIF file, then we won't know about that so we'll mark this as a success. This seems fine as code scanning has succeeded in that situation at analyzing the repo and uploading the analysis results.

Yes!

src/environment.ts
```diff
@@ -1,3 +1,4 @@
+// NB: prefix these variables with `CODEQL_ACTION` to reduce risk of their being overwritten by other steps. */
```
henrymercer, line 1:
```suggestion
/**
 * Environment variables used by the CodeQL Action.
 *
 * We recommend prefixing environment variables with `CODEQL_ACTION_` to reduce the risk that they are overwritten by other steps.
 */
```

src/init-action-post-helper.ts
```diff
@@ -282,3 +301,22 @@ async function removeUploadedSarif(
...
+export function getFinalJobStatus(): JobStatus {
+  const jobStatusFromEnvironment = process.env[EnvVar.JOB_STATUS];
+  if (!jobStatusFromEnvironment) {
+    return JobStatus.Unknown;
+  }
+  if (
+    !Object.values(JobStatus).includes(jobStatusFromEnvironment as JobStatus)
+  ) {
+    return JobStatus.Unknown;
+  }
```
henrymercer, line 320: Nit: group these conditions together

src/status-report.ts
```diff
@@ -133,6 +141,25 @@ export function getActionsStatus(
   }
 }
 
+/**
+ * Sets the overall job status environment variable to configuration error
+ * or failure, unless it's already been set to one of these values in a
+ * previous step.
+ */
+function setJobStatusIfUnsuccessful(actionStatus: ActionStatus) {
```
henrymercer, line 149: Should we produce failure for `aborted` too?  Not related to your code, but we should probably have a comment explaining when we use `aborted` instead of `failure` (or consider removing it entirely).

henrymercer (dismissed):
Nice!

src/status-report.ts
angelapwen: Hm yeah that's fair, we should as otherwise we'll fall through and set it to `Success` if nothing was set already.

src/init-action-post-helper.ts
angelapwen: Done!

src/status-report.ts
angelapwen: I added a comment next to when `aborted` is declared as a possible status. I would be in favor of changing it to `failure` in a future PR though.

angelapwen:
Will hold on merging this until the monolith PR is approved and merged.

src/status-report.ts
```diff
@@ -32,12 +32,20 @@ export type ActionName =
   | "upload-sarif";
 
 export type ActionStatus =
-  | "aborted"
+  | "aborted" // Only used in the init Action, if init failed and it wasn't a configuration error.
```
henrymercer, line 35:
```suggestion
  | "aborted" // Only used in the init Action, if init failed before initializing the tracer due to something other than a configuration error.
```

henrymercer (approved):

henrymercer:
Ah, rebuild doesn't have permission to write to forks.

angelapwen:
The failures are actually unrelated upstream API failures. The comment change doesn't affect the transpilation files (I think)

angelapwen:
Monolith PR is merged so I'll merge this now and keep an eye on our telemetry to make sure the change is looking okay!

PR #2100, opened 2024-01-26
Add a README section about supported versions

henrymercer:
This information is already included in the changelog, but let's surface it in the README to so that it's more visible.


jsoref:
Thanks. This text is very helpful. 

It's probably worth mentioning that workflows running outside GHES on v2 will trigger node16 depreciation warnings on each and every run.

It's a bit surprising. (I have a customer that filed a ticket asking me to release an upgrade so that the warning goes away. - thankfully I had already been working on fixing it because I also find the warning very annoying, but it took me quite a bit to develop and validate a fix for one of the other breaking action updates.)

aeisenberg (approved):
Thanks for writing this up.

henrymercer:
I think I'll merge this as is, since we [produce our own warning](https://github.com/github/codeql-action/blob/738d232550ddb0cd6a20e566b3142456af0a6bcc/src/util.ts#L1065) as well about upgrading from v2 to v3 (when the user's running on a GitHub instance that supports Node 20 Actions).

PR #2105, opened 2024-01-30
Refactor configuration errors

angelapwen:
As we add more configuration errors to improve our uptime measurements, I thought it would be nice to refactor the existing errors into their own file. I took inspiration from the `FeatureConfig` `Record` that we use for feature flags.

Also, I thought it would be nice to consolidate the place we check for configuration errors into `codeql.ts`, where the actual command invocations happen.

The PR also renames the `UserError` type to `ConfigurationError` to standardize on a single term.


src/codeql.ts
```diff
@@ -616,20 +617,40 @@ export async function getCodeQLForCmd(
...
+          }
+          if (
+            isCliConfigurationError(
+              CliError.IncompatibleWithActionVersion,
+              e.message,
+            ) ||
+            isCliConfigurationError(CliError.InvalidSourceRoot, e.message)
+          ) {
+            throw new util.UserError(e.message);
+          }
```
nickfyson, line 650: D'you think it would be possible to define a function such that a block like this would just be replaced by (eg) `throw processCliConfigurationError(e)`, which in the general case would be a no-op that returns `e`, but matching cases will return the altered/improved error? With that it ought to be possible to handle a new class of error just by adding a new definition to `src/cli-config-errors.ts`, which might be cleaner/easier to grok...? 🤷

nickfyson (commented):
I really like the direction, I think it makes a lot of sense to consolidate this class of errors and their handling into a single place. I've added a couple of comments/questions...

src/codeql.ts
angelapwen: Oh that might work well! Let me try that. I had been thinking about also defining some kind of function closure in `cli-config-errors` for each error type, but that seemed kind of complicated/unneeded. I think it would make sense to have it just return the wrapped error string though.

src/cli-config-errors.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * All of our caught CLI error messages that we handle specially: ie. if we
+ * would like to categorize an error as a configuration error or not. Optionally
+ * associated with a CLI error code as well. Note that either of the conditions
+ * is enough to be considered a match: if the exit code is a match, or the error
+ * messages match.
+ */
+export const cliErrorsConfig: Record<
```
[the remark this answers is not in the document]
henrymercer: Looks good to me! :)

src/codeql.ts
angelapwen: Okay, I gave this a try and it still seems a bit messy but I agree that in the future it'll be nicer as we add more errors. 

Initially I tried it with a second configuration `Record`, as it seems to be a different function (the original one included data that helps the Action figure out whether the error is a known CLI error or not, whereas the new data is for which Action error message to return once the CLI error type is known). ([commit](https://github.com/github/codeql-action/pull/2105/commits/253c29f8cc940dd38cd7066620e403bb3be7d242))

Then I tried combining them because it seems a bit odd to have two map with the exact same keys... I think the code is cleaner but maybe harder to understand how each field in the configuration `Record` is being used. ([commit](https://github.com/github/codeql-action/pull/2105/commits/c5a7e3e3f35d89fb02b381a5c87979b991c1e2d2))

Would be interested in your thoughts as a likely maintainer/user of this piece of the code 😸

src/codeql.ts
```diff
@@ -616,20 +621,52 @@ export async function getCodeQLForCmd(
...
+          if (isCliConfigurationError(CliError.InvalidSourceRoot, e.message)) {
+            throw new util.UserError(
+              processCliConfigurationError(
+                CliError.InvalidSourceRoot,
+                e.message,
+              ),
+            );
+          }
+        }
+        throw e;
```
nickfyson, line 668: Sorry, I definitely wasn't sufficiently clear in my original (late night) comment, but I was hoping we could end up with something like this...
```suggestion
throw processCliConfigurationError(error);
```
So we'd end up not having the export the error types at all, just the processing function that would both identify and then enhance/change the error as required. Does that make sense? (quite possibly not... 😅)

angelapwen: Actually, it was clear! But I ended up not really being able to do that because of the `NoJavaScriptTypeScriptCodeFound` case, which also requires a check to the CLI version 🤔 it seemed a little extraneous to _also_ add that case and logic to the configuration file.

It seems like that specific error can be removed once we stop supporting CLI v2.11.6, which is our oldest supported version now, so maybe we can make that refactoring change once it's deprecated.

For now — do you think it was cleaner before? 😆

nickfyson: Ah, I see... 🤔

I think you might be right that it was clearer beforehand, apologies for derailing this PR somewhat! :laugh

src/cli-config-errors.ts
```diff
@@ -0,0 +1,108 @@
...
+      "No code found during the build. Please see: " +
+      "https://gh.io/troubleshooting-code-scanning/no-source-code-seen-during-build",
+    appendCliErrorToActionError: false,
+  },
+};
+
+/**
+ * Checks whether or not the error message received from the CLI is a config
+ * error. Returns true as long as either of the conditions holds: the exit
+ * codes are a match, or the error message matches the expected message snippets.
```
aeisenberg, line 61: Small clarification:

```suggestion
 * codes are a match, or all of the error message matches the expected message snippets.
```

src/cli-config-errors.ts
```diff
@@ -0,0 +1,108 @@
...
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
+ * Returns the error message that the Action should return in case of this CLI error. If no
+ * `actionErrorMessage` was defined, return the original CLI error. Otherwise, return the
+ * `actionErrorMessage` with the CLI error message appended, depending on the value of
```
aeisenberg, line 89: Is this what you meant?

```suggestion
 * `cliErrorMessage` was defined, return the original CLI error. Otherwise, return the
 * `cliErrorMessage` with the CLI error message appended, depending on the value of
```

src/cli-config-errors.ts
```diff
@@ -0,0 +1,108 @@
...
+ * `appendCliErrorToActionError` in `cliErrorsConfig`.
+ */
+export function processCliConfigurationError(
+  cliError: CliError,
+  cliErrorMessage: string,
+): string {
+  const cliErrorConfig = cliErrorsConfig[cliError];
+  if (
+    cliErrorConfig === undefined ||
+    cliErrorConfig.actionErrorMessage === undefined
```
aeisenberg, line 99: Nit: (very minor)

```suggestion
    cliErrorConfig?.actionErrorMessage === undefined
```

src/cli-config-errors.ts
```diff
@@ -0,0 +1,108 @@
...
+  ) {
+    return cliErrorMessage;
+  }
+
+  // Append the CLI error message by default.
+  return cliErrorConfig.appendCliErrorToActionError === false
```
aeisenberg, line 105: nit:

```suggestion
  return !cliErrorConfig.appendCliErrorToActionError
```

src/cli-config-errors.ts
```diff
@@ -0,0 +1,108 @@
+/** Error messages from the CLI that we handle specially. */
+export enum CliError {
```
aeisenberg, line 2: The convention in JS and TS is usually that any class name or type name ending in `...Error` is  subtype of a javascript error. In this case, it's an enum, which was a little confusing to me below.

Can you rename this to something else? Maybe `CliErrorKind`, `CliErrorCategory`, or `CliErrorNames`?

src/cli-config-errors.ts
angelapwen: Hm, I actually wanted to catch the case where `appendCliErrorToActionError` wasn't set, in which case it would evaluate to `undefined` — in that situation we'd want to default to appending. But, I think this will also be covered in the refactor.

src/cli-config-errors.ts
angelapwen: Ah, that's good to know — thank you! I'll do that.

src/cli-config-errors.ts
angelapwen: Updated clarification in new method comment.

src/codeql.ts
```diff
@@ -630,20 +606,26 @@ export async function getCodeQLForCmd(
...
+            config.dbLocation,
+            `--source-root=${sourceRoot}`,
+            ...(await getLanguageAliasingArguments(this)),
+            ...extraArgs,
+            ...getExtraOptionsFromEnv(["database", "init"]),
+          ],
+          { stdin: externalRepositoryToken },
+        );
+      } catch (e) {
+        if (e instanceof Error) {
```
nickfyson, line 625: I think we need a fallback `throw e` just incase we've caught something that is not an Error...?

src/codeql.ts
```diff
@@ -708,17 +690,13 @@ export async function getCodeQLForCmd(
         await runTool(cmd, args);
       } catch (e) {
         if (
-          e instanceof CommandInvocationError &&
+          e instanceof Error &&
           !(await util.codeQlVersionAbove(
             this,
             CODEQL_VERSION_BETTER_NO_CODE_ERROR_MESSAGE,
```
nickfyson, line 696: Is the behaviour here what we want with these changes? Do we not want to try and wrap the error at all if the CodeQL version isn't sufficiently old? I'm not clear which types of error are possible here...

nickfyson: if this check is moved into `src/cli-errors.ts` somewhere do things get egregiously messy...?

src/status-report.ts
```diff
@@ -43,7 +43,7 @@ export enum JobStatus {
   Unknown = "JOB_STATUS_UNKNOWN",
   Success = "JOB_STATUS_SUCCESS",
   Failure = "JOB_STATUS_FAILURE",
-  ConfigurationError = "JOB_STATUS_CONFIGURATION_ERROR",
+  ConfigError = "JOB_STATUS_CONFIGURATION_ERROR",
```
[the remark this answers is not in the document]
angelapwen: 🤔 I think I've already encoded this enum in the upstream data warehouse schema, so I'm not sure it's worth changing now. That said, if we were to change it this would be the time 😆

src/codeql.ts
angelapwen: Yes, nice catch!

src/codeql.ts
angelapwen: I think if the version is newer, we no longer need this check (and we can delete the whole wrapping check too). This one in particularly is pretty well-tested:

https://github.com/github/codeql-action/blob/cf7e9f23492505046de9a37830c3711dd0f25bb3/src/codeql.test.ts#L862

https://github.com/github/codeql-action/blob/cf7e9f23492505046de9a37830c3711dd0f25bb3/src/codeql.test.ts#L885

https://github.com/github/codeql-action/blob/cf7e9f23492505046de9a37830c3711dd0f25bb3/src/codeql.test.ts#L902

so I'm pretty certain that there aren't any logical changes in the PR. 

I tried moving the check but it does get a bit messy, because then the configuration would need to know which CLI version is necessary for it to be wrapped.. I'm not sure if it's worth moving the check as we're almost about to deprecate this case (the next time we stop supporting a CLI version, it'll be this one).

nickfyson: 👍 that makes sense, particularly if this special case is going to be so short-lived

src/status-report.ts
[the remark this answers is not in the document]
nickfyson: The string `JOB_STATUS_CONFIGURATION_ERROR` is what has been committed to elsewhere, right? So within the action code we could use (eg) `ConfigErrorStatus` instead, perhaps along with `UnknownStatus`, `SuccessStatus` etc for symmetry?

My concern is just that calling it `ConfigError` makes it seem like an actual Error object, when it isn't. But this is far from a blocking concern!

nickfyson (approved):
🚀

src/status-report.ts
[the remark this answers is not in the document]
angelapwen: Oh — yes, I see what you mean! Yep, I can do that 😸

nickfyson (approved):
🚀

nickfyson (approved):

PR #2107, opened 2024-01-30
Refactoring: Simplify adding new configuration properties

henrymercer:
This PR refactors the way we initialize and test configurations to make it easier to add new configuration properties.  It shouldn't have any user visible effects.  Specifically, adding a new configuration property should not require updating irrelevant tests.


src/config-utils.ts
```diff
@@ -391,39 +391,69 @@ export async function getRawLanguages(
...
+  repository: RepositoryNwo;
+  tempDir: string;
+  codeql: CodeQL;
+  workspacePath: string;
+  githubVersion: GitHubVersion;
+  apiDetails: api.GitHubApiCombinedDetails;
+  logger: Logger;
+}
+
+type GetDefaultConfigInputs = Omit<
```
angelapwen, line 415: TIL about the `Omit` type 🤯

src/config-utils.ts
```diff
@@ -391,39 +391,69 @@ export async function getRawLanguages(
...
+  InitConfigInputs,
+  "configFile" | "configInput"
+>;
+
+type LoadConfigInputs = Omit<InitConfigInputs, "configInput"> & {
```
angelapwen, line 420: What does this `&` operator do here? Thought it represented bitwise AND but I can't really make sense of it in this context.

src/config-utils.ts
angelapwen: We could use `Omit`/`Pick` to refactor the `StatusReportBase` and all those other interfaces later, perhaps...

angelapwen (approved):
Just a quick question 😄

src/config-utils.ts
aeisenberg: These are [intersections](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#intersections). Though I wonder if this should be a union type (using `|`).

henrymercer: I believe we want intersections, i.e. we want the type to include the properties from both the left hand side and the right hand side.  For instance if we have (1) `{ a: 1 }` and (2) `{ a: 1 }`, then (1) and (2) would satisfy `{ a: int } | { b: string }` but only (2) would satisfy `{ a: int } & { b: string }`.

src/config-utils.ts
henrymercer: There are quite a few of these nice types :) https://www.typescriptlang.org/docs/handbook/utility-types.html

PR #2108, opened 2024-01-31
Add experimental `build-mode` input

henrymercer:
This PR introduces an experimental `build-mode` input to the `init` Action.  This is currently for internal testing only.

We forward this input to `codeql database init` and add a test to check that the build mode is forwarded to the database as expected, and Java buildless extraction completes successfully.

This PR makes use of the config refactoring implemented in #2107.


init/action.yml
```diff
@@ -10,6 +10,20 @@ inputs:
...
+  build-mode:
+    description: >-
+      [Experimental, for internal testing only] The build mode that will be used to analyze the language.
+      This input is only available in single-language analyses.
+
+      Available build modes will differ based on the language being analyzed. One of: 
+
+      - none:      The database will be created without building the source code.
+                   Available for all interpreted languages and some compiled languages.
+      - autobuild: The database will be created by attempting to automatically build the source code.
```
aeisenberg, line 22: It's unclear here if this still requires an explicit autobuild step.

aeisenberg (dismissed):
I think the action iput description can be clarified, but it's not a blocking change.

init/action.yml
henrymercer: Good point thanks!  I'll update this.  Right now we do need the explicit autobuild step, but I'm working on a change that will mean no autobuild step is required.

angelapwen (approved):
🆒 😄

PR #2110, opened 2024-02-01
Adjust Tokens

jsoref:
I can't validate this, at the moment my jobs are stuck waiting for runners for macOS. I'm assuming that's because of the PRs I've been working on for github/codeql. I filed https://support.github.com/ticket/personal/0/2568638.

Goals:
* remove some trailing whitespace
* resolve https://github.com/github/codeql/issues/15462#issuecomment-1919186317 by converting `sendStatusReport` into a function that always returns `true` and uses `core.warning` instead of `core.error`
* add an optional `secrets.CODEQL_TOKEN` to allow workflows in forks of this repository to _try_ to work around github limits which result in failures due to limits (presumably on the default github-actions application) -- I can't tell if this works (see preamble)
* add `input.token.description` to `action.yml` files to explain the purpose of the token in each case (more or less -- I'm not remotely certain about any of these)
* hopefully I've correctly adjusted the pr-check & workflow things -- this sync stuff is a PITA to deal with -- it'd be nice if there was autogenerated output to update pr-check, but afaict there isn't


analyze/action.yml
```diff
@@ -69,6 +69,8 @@ inputs:
     required: true
     default: "true"
   token:
+    description: "Token for accessing /code-scanning/analysis endpoints (used by `upload-database` and `upload`)"
+    required: false
```
angelapwen, line 73: We think this provides a bit too much detail on internals so would like to suggest:
```suggestion
    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
    required: false
```

src/status-report.ts
```diff
@@ -334,19 +341,19 @@ export async function sendStatusReport<S extends StatusReportBase>(
...
-            core.setFailed(
+            core.warning(
               'Workflows triggered by Dependabot on the "push" event run with read-only access. ' +
                 "Uploading Code Scanning results requires write access. " +
                 'To use Code Scanning with Dependabot, please ensure you are using the "pull_request" event for this workflow and avoid triggering on the "push" event for Dependabot branches. ' +
                 "See https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning#scanning-on-push for more information on how to configure these events.",
             );
           } else {
-            core.setFailed(e.message || GENERIC_403_MSG);
+            core.warning(e.message || GENERIC_403_MSG);
```
angelapwen, line 351: Because we're no longer failing the entire workflow at this stage, we'd like to provide a better error when we upload SARIF if that API call returns 403/404. So perhaps we can simply do `core.warning(e.message)` in this file (and below, in the 404 case, and move the `GENERIC_403_MSG` and `GENERIC_404_MSG` to catch these response codes after this call: https://github.com/github/codeql-action/blob/cf7e9f23492505046de9a37830c3711dd0f25bb3/src/upload-lib.ts#L110?

autobuild/action.yml
```diff
@@ -3,6 +3,7 @@ description: 'Attempt to automatically build code'
 author: 'GitHub'
 inputs:
   token:
+    description: "Token for accessing /code-scanning/analysis endpoints (used to for status reporting)"
```
angelapwen, line 6: Same as in `analyze/action.yml`, and let's explicitly add the `required: false` block? 
```suggestion
    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
    required: false
```

resolve-environment/action.yml
```diff
@@ -3,6 +3,7 @@ description: '[Experimental] Attempt to infer a build environment suitable for a
 author: 'GitHub'
 inputs:
   token:
+    description: "Token for accessing /code-scanning/analysis endpoints (used for ???)"
```
angelapwen, line 6: Same as other comments!
```suggestion
    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
    required: false
```

upload-sarif/action.yml
```diff
@@ -20,6 +20,7 @@ inputs:
     description: "The sha of the HEAD of the ref where results will be uploaded. If not provided, the Action will use the GITHUB_SHA environment variable. If provided, the ref input must be provided as well. This input is ignored for pull requests from forks."
     required: false
   token:
+    description: "Token for accessing /code-scanning/analysis endpoints (used to publish sarif -- generally requires `security-events: write`)"
```
angelapwen, line 23: Same as other comments!
```suggestion
    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
    required: false
```

angelapwen (changes requested):
Thank you for your contribution 💖 and your patience as we reviewed the PR! I wrote in some suggestions and specific code-related comments, but here are some other higher-level comments:

- Could we drop the `node_modules` changes here, as they seem to be unrelated to the changes? We'd also like to make sure we `Squash merge` rather than `Merge commit` when the PR is in so that the change isn't in the commit history of `main`.
- We'd also like to separate out the last change (adding `${{ secrets.CODEQL_TOKEN }}` to all workflows owned by us, to keep the PR simpler. We may look into a better way of auto-generating this line for steps using actions that we own. 
- If you could rename the PR title to be more descriptive of the changes, that would be great!

Let us know if we can clarify further!! Thanks again 😸

angelapwen:
Thank you for splitting these up ✨

PR #2111, opened 2024-02-01
add first_party_analysis boolean to all status reports

nickfyson:
In order to define more granular metrics and define more appropriate SLOs we add a new field to the status reports uploaded by the CodeQL Action.

This field `first_party_analysis` is based on whether the init action has been used, which is only used for first party analysis. When a SARIF file has been generated by other means and submitted using the upload action, this is considered to be a third party analysis and will be treated differently when calculating SLOs. To ensure misconfigured workflows are not treated as third party, only the `upload-sarif` action can submit status reports that are not first-party.

See back-linked internal issue for example status reports, showing these changes in action.


src/environment.ts
```diff
@@ -41,6 +41,9 @@ export enum EnvVar {
   /** Whether the CodeQL Action has already warned the user about low disk space. */
   HAS_WARNED_ABOUT_DISK_SPACE = "CODEQL_ACTION_HAS_WARNED_ABOUT_DISK_SPACE",
 
+  /** Whether the init action has been run. */
+  INIT_ACTION_HAS_RUN = "CODEQL_INIT_ACTION_HAS_RUN",
```
henrymercer, line 45: Minor: Doesn't read as well but it's nice to have everything the Action defines prefixed by `CODEQL_ACTION_`

```suggestion
  INIT_ACTION_HAS_RUN = "CODEQL_ACTION_INIT_HAS_RUN",
```

src/status-report.test.ts
```diff
@@ -58,3 +58,108 @@ test("createStatusReportBase", async (t) => {
...
+    process.env["GITHUB_REPOSITORY"] = "octocat/HelloWorld";
+    process.env["GITHUB_RUN_ATTEMPT"] = "2";
+    process.env["GITHUB_RUN_ID"] = "100";
+    process.env["GITHUB_SHA"] = "a".repeat(40);
+    process.env["ImageVersion"] = "2023.05.19.1";
+    process.env["RUNNER_OS"] = "macOS";
+    process.env["RUNNER_TEMP"] = tmpDir;
+
+    const getRequiredInput = sinon.stub(actionsUtil, "getRequiredInput");
+    getRequiredInput.withArgs("matrix").resolves("input/matrix");
```
henrymercer, line 77: Minor: consider creating a function for all of this as it's shared across a few tests

henrymercer (approved):
Nice!

angelapwen (approved):
Looks great to me ✨ thanks for the care in thinking about misconfigured workflows!

PR #2112, opened 2024-02-02
Telemetry: Handle failures determining disk usage

henrymercer:
We've noticed some situations where `df` fails due to the tracer not initializing correctly, and therefore we are unable to determine available disk space.  Let's instead fail gracefully when we can't determine available disk space.


src/status-report.ts
```diff
@@ -241,6 +239,11 @@ export async function createStatusReportBase(
     workflow_run_id: workflowRunID,
   };
 
+  if (diskInfo) {
+    statusReport.runner_available_disk_space_bytes = diskInfo.numAvailableBytes;
+    statusReport.runner_total_disk_space_bytes = diskInfo.numTotalBytes;
```
aeisenberg, line 244: Do we want to capture the reason for the `df` failure and send it as part of the status report?

aeisenberg (approved):
My comment shouldn't block having this merged, but something to add later if we want.

src/status-report.ts
aeisenberg: The message wouldn't make it into kusto, but at least it would be available in splunk.

henrymercer: We could probably add an internal error diagnostic for this.  Let's address as potential follow up, as you suggest.

PR #2120, opened 2024-02-07
Allow omitting the autobuild Action when build mode specified 

henrymercer:
This PR enables users to omit the `autobuild` Action when the build mode is specified and the CLI supports the `traceCommandUseBuildMode` feature (available in CodeQL CLI 2.16.2 and later).  This will remove redundancy and reduce the possibility for errors.


src/analyze.ts
```diff
@@ -166,29 +168,49 @@ async function setupPythonExtractor(
   process.env["LGTM_PYTHON_SETUP_VERSION"] = output;
 }
 
-export async function createdDBForScannedLanguages(
+export async function runExtraction(
```
angelapwen, line 171: Oh great, this naming is a lot clearer!

angelapwen (dismissed):
Cool! I like how simple that test was 💖

src/analyze.ts
henrymercer: This is in some sense the core of the change — when the build mode is specified, the `analyze` Action will now do what's necessary to extract the database, rather than only extracting scanned languages.

henrymercer:
@angelapwen I added a couple more tests for the other build modes.  Would you mind taking another look?

angelapwen (approved):
The new tests look good to me! Should this have a changelog note? Or will we wait for a general larger-scale announcement of the new build modes?

henrymercer:
Right now, this is just for internal testing.  Once we've finished testing, we can remove all the internal labels and add a changelog note with all the changes.

PR #2121, opened 2024-02-08
Treat status reports as non-critical

jsoref:
Resolve https://github.com/github/codeql/issues/15462#issuecomment-1919186317 by converting `sendStatusReport` into a function that always returns `void` and uses `core.warning` instead of `core.error`

Split from:
* #2110


src/upload-lib.ts
```diff
@@ -107,19 +112,34 @@ async function uploadPayload(
...
+      switch (e.status) {
+        case 403:
+          core.warning(GENERIC_403_MSG);
+          break;
+        case 404:
+          core.warning(GENERIC_404_MSG);
+          break;
+        default:
+          core.warning(e.message);
+          break;
```
angelapwen, line 139: 🤔 Correct me if I'm wrong, but I _think_ we still want to `throw e` unconditionally here, on top of the custom warnings we're emitting. 

Otherwise, the error will be uncaught and the workflow will potentially succeed, even though in this case we do want to fail the calling workflow because the upload endpoint is in our critical path. 

Alternatively, we could use `core.setFailed(...)` rather than `core.warning(...)` here as we did in the status report catch clause.

src/upload-lib.ts
```diff
@@ -107,19 +112,34 @@ async function uploadPayload(
...
 
-  return response.data.id;
+    logger.debug(`response status: ${response.status}`);
+    logger.info("Successfully uploaded results");
+    return response.data.id;
+  } catch (e) {
+    if (util.isHTTPError(e)) {
+      switch (e.status) {
+        case 403:
+          core.warning(GENERIC_403_MSG);
```
angelapwen, line 132:
```suggestion
          core.warning(e.message || GENERIC_403_MSG);
```
for parity with the prior message emitted in the status report file.

angelapwen (commented):
Thanks again for this — our telemetry reporting should indeed be non-critical! I've just added a few comments. 

I think this may warrant a CHANGELOG entry 🌞 users who may have seen this failure in the past should be happy to see the change.

src/upload-lib.ts
jsoref: Hmm.

I'm in the process of trying to get one of the GitHub actions to optionally *not* unconditionally fail on failure:

* https://github.com/actions/download-artifact/pull/286

It's incredibly frustrating having "reusable" components that force fatal behaviors on their calling components.

Note that I'm not really trying to change this specific behavior in this PR, so if the behavior before this PR was that failure to upload was fatal, then we can retain that here and I can make yet another PR to add a flag similar to the one above to let callers say "I know, but I don't want *my* action to die just because of this potential failure".

An example of this disaster is me trying to run CI for this stuff -- the Ruby/ Swift macOS runners appear to run the github-actions application out of its request limit which results in the workflows dying.

Maybe someone writing an action or workflow doesn't think that should be fatal, or maybe they have a way of handling it. For instance, they could stash some artifacts and then later upload the sarif items after the limits reset. -- check-spelling uploads a sarif artifact in addition to using upload-sarif, so it really wouldn't be inconceivable to me that someone might upload an artifact on upload sarif failure and have a scheduled job responsible for collecting all of the queued artifacts, downloading them, and calling upload-sarif (possibly with a different token that is likely to have more calls available).

I actually think implementing the above would be a lot prettier than the other commit that is no longer in the PR series -- instead of each workflow taking `secrets.codeql_token` or `github.token`, it could use an additional step `if: steps.upload-sarif.outputs.failed` `uses: upload-artifact` / `jobs.call-retry-upload.uses: ./.github/workflows/upload-sarif-from-artifact.yml` `with:` (artifact information).

One positive is that the expensive calculation side (a bunch of these runs take hours especially on limited macOS runners) would be split from the cheap to rerun upload step.

Now, to prove that this repository can work on macOS, you might not want to make that workflow split, but if I had a real workload similar to this repository's workload, and I ran into the same problems as I run into here, I'd definitely *want* to be able to do so.

Anyway, this flexibility is probably out of scope for this PR.

---

FWIW, this comment was written using GitHub Mobile for Android beta and Gboard -- if there are typos, I'll try to correct them later.

angelapwen: That's totally fair and an interesting use of our `upload-sarif` Action that I hadn't considered (though in retrospect it seems obvious with your use case 😸) 

I would agree that it's out of scope for this PR as I'd like to "preserve" the original behavior of the 403/404s here, just in a different spot.

jsoref: Fwiw, I've opened #2125 which relates to this.

It's a piece that I'm _already_ papering over:
https://github.com/check-spelling/check-spelling/blob/7b1e024bbfaf52adfa7111a09a9bfbd25c4b3349/action.yml#L672

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "codeql",
-  "version": "3.24.1",
+  "version": "3.25.0",
```
aeisenberg, line 3: The version is bumped during the release process. It shouldn't be changed manually.

```suggestion
  "version": "3.24.1",
```

src/status-report.ts
```diff
@@ -298,19 +294,24 @@ const INCOMPATIBLE_MSG =
  * as failed if the status report failed. This is only expected to be used
  * when sending a 'starting' report.
  *
- * Returns whether sending the status report was successful of not.
+ * This API it calls is private and it is not critical that it succeed:
```
aeisenberg, line 297:
```suggestion
 * The `/code-scanning/analysis/status` endpoint is internal and not critical that it succeeds:
```

src/status-report.ts
```diff
@@ -298,19 +294,24 @@ const INCOMPATIBLE_MSG =
...
+ * https://github.com/github/codeql/issues/15462#issuecomment-1919186317
+ *
+ * Do not worry about whether a report is/isn't submitted.
+ * Just go about your business.
+ * The side effects of this call are not relevant to your program.
```
aeisenberg, line 302:
```suggestion
 * Failures while calling this endpoint are logged as warings.
```

src/upload-lib.ts
```diff
@@ -16,6 +16,11 @@ import { parseRepositoryNwo, RepositoryNwo } from "./repository";
 import * as util from "./util";
 import { SarifFile, ConfigurationError, wrapError } from "./util";
 
+const GENERIC_403_MSG =
+  "The repo on which this action is running is not opted-in to CodeQL code scanning.";
```
aeisenberg, line 20:
```suggestion
  "The repo on which this action is running has not opted-in to CodeQL code scanning.";
```

src/upload-lib.ts
```diff
@@ -16,6 +16,11 @@ import { parseRepositoryNwo, RepositoryNwo } from "./repository";
...
+const GENERIC_404_MSG =
+  "Not authorized to use the CodeQL code scanning feature on this repo.";
```
aeisenberg, line 22: I haven't verified this, but are you sure that you have the correct 404 and 403 error messages? 403 errors usually mean _forbidden_, but you have it here as 404.

Also, _not authorized_ (401) typically means that the user isn't logged in. I think it is better to use the phrase _forbidden_, which correlates better with 403.

```suggestion
const GENERIC_404_MSG =
  "The CodeQL code scanning feature is forbidden on this repository.";
```

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The upload-sarif action should not fail if it can't write to the telemetry api endpoint [#2121](https://github.com/github/codeql-action/pull/2121)
```
aeisenberg, line 9: This is more than just the `upload-sarif` endpoint. I think this reads better.

```suggestion
- The CodeQL action no longer fails if it can't write to the telemetry api endpoint. [#2121](https://github.com/github/codeql-action/pull/2121)
```

aeisenberg (commented):
This is looking great. I have a few copy-editing suggestions and one question about 404/403 errors.

src/upload-lib.ts
jsoref: The strings were merely moved. I make no claim to their accuracy.

aeisenberg (approved):

aeisenberg:
Thanks again for your work on this.

src/status-report.ts
```diff
@@ -298,19 +294,22 @@ const INCOMPATIBLE_MSG =
  * as failed if the status report failed. This is only expected to be used
  * when sending a 'starting' report.
  *
- * Returns whether sending the status report was successful of not.
+ * The `/code-scanning/analysis/status` endpoint is internal and not critical that it succeeds:
```
jsoref, line 297:
```suggestion
 * The `/code-scanning/analysis/status` endpoint is internal and it is not critical that it succeeds:
```

aeisenberg (approved):

aeisenberg (approved):

aeisenberg:
Thanks again! Nice work here.

Issue #2117, opened 2024-02-02
codeql/upload-sarif@v3 action failed: Resource not accessible by integration


SPodjasek:
> I'm opening this issue as requested in #1806

When trying to upload sarif file produced by Docker Scout we get: `Resource not accessible by integration` - despite that `security-events` permission is set to `write`.

Detailed workflow run logs are below. I've stripped output from scout as I believe it's irrelevant.

<details><summary>Logs</summary>
<p>

```
2024-02-02T22:03:48.1355639Z Requested labels: ubuntu-latest
2024-02-02T22:03:48.1355934Z Job defined at: ....
2024-02-02T22:03:48.1356154Z Reusable workflow chain:
2024-02-02T22:03:48.1356234Z .... (6e62641865d79cd11cea291c21405d81fb03275d)
2024-02-02T22:03:48.1356335Z -> .... (3e071b83a90458e94e89a01903cda60650f86a6c)
2024-02-02T22:03:48.1356436Z Waiting for a runner to pick up this job...
2024-02-02T22:03:48.5338786Z Job is waiting for a hosted runner to come online.
2024-02-02T22:03:50.8524838Z Job is about to start running on the hosted runner: GitHub Actions 207 (hosted)
2024-02-02T22:03:54.0039541Z ##[debug]Starting: Build, publish and notify / Docker Scout
2024-02-02T22:03:54.0069923Z ##[debug]Cleaning runner temp folder: /home/runner/work/_temp
2024-02-02T22:03:54.0383199Z ##[debug]Starting: Set up job
2024-02-02T22:03:54.0383989Z Current runner version: '2.312.0'
2024-02-02T22:03:54.0411603Z ##[group]Operating System
2024-02-02T22:03:54.0412393Z Ubuntu
2024-02-02T22:03:54.0413252Z 22.04.3
2024-02-02T22:03:54.0413748Z LTS
2024-02-02T22:03:54.0414183Z ##[endgroup]
2024-02-02T22:03:54.0414786Z ##[group]Runner Image
2024-02-02T22:03:54.0415453Z Image: ubuntu-22.04
2024-02-02T22:03:54.0416046Z Version: 20240126.1.0
2024-02-02T22:03:54.0417552Z Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20240126.1/images/ubuntu/Ubuntu2204-Readme.md
[...]

PR #2126, opened 2024-02-09
Favor `GITHUB_WORKFLOW_REF` over using `actions: read` to calculate the workflow path

jsoref:
Introduced with GHES 3.9:
https://docs.github.com/en/enterprise-server@3.9/actions/learn-github-actions/variables

`GITHUB_WORKFLOW_REF` means that actions don't need to use `actions: read` to determine the path to the running workflow.

This should address the problem in #2117 as long as the user is running on GHES 3.9+ (including GHEC).

It doesn't precisely fix the issue because someone running on GHES 3.8 will still need to add `actions: read` until #2121 makes that code path non fatal.

Note that these aren't mutually exclusive PRs (although as they both impact the changelog, they will conflict there...) -- both changes should be applied -- this simplifies the general logic for the average user (and once GHES 3.8 is sunset, the other code path should be removed entirely).


SPodjasek:
I can confirm that this fixes my issue from #2117 without the need to declare `actions: read` permission:

```diff
diff --git a/.github/workflows/local-cd.yml b/.github/workflows/local-cd.yml
index c6bf452..ee11bac 100644
--- a/.github/workflows/local-cd.yml
+++ b/.github/workflows/local-cd.yml
@@ -236,7 +236,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: build-and-publish
     permissions:
-      actions: read
       contents: read
       packages: read
       pull-requests: write
@@ -282,7 +281,8 @@ jobs:
         id: upload-sarif
         continue-on-error: true
         if: hashFiles('sarif.output.json') != '' && github.event_name != 'pull_request_target'
-        uses: github/codeql-action/upload-sarif@main
+        uses: jsoref/github-codeql-action/upload-sarif@favor-workflow-ref
         with:
           sarif_file: sarif.output.json
```

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Users of GHES3.9+ and GHEC should not need to include `actions: read` permissions to use upload-sarif in private repositories.
```
angelapwen, line 9:
```suggestion
- Users of GHES 3.9+ and GHEC will no longer need to include `actions: read` permissions to use `upload-sarif` in private repositories.
```

src/api-client.ts
```diff
@@ -119,6 +119,16 @@ export async function getGitHubVersion(): Promise<GitHubVersion> {
  * Get the path of the currently executing workflow relative to the repository root.
  */
 export async function getWorkflowRelativePath(): Promise<string> {
+  const workflow_ref = process.env["GITHUB_WORKFLOW_REF"];
+  if (workflow_ref) {
```
angelapwen, line 123: Could you add a code comment here about how this is only supported in GHES 3.9+, so for versions <3.8 we fall back on the previous logic?

angelapwen (commented):
This looks good to me ✨ just a few documentation suggestions, and the changelog merge conflict will need to be addressed of course!

src/api-client.ts
```diff
@@ -119,6 +119,16 @@
  * Get the path of the currently executing workflow relative to the repository root.
  */
 export async function getWorkflowRelativePath(): Promise<string> {
+  const workflow_ref = process.env["GITHUB_WORKFLOW_REF"];
```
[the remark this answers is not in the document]
angelapwen: This can be dismissed as a false positive, but it reminded me that we generally check for if the environment variable value `!== undefined` in case it comes out incorrectly truthy or falsy.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ No user facing changes.
 
 - Update default CodeQL bundle version to 2.16.2. [#2124](https://github.com/github/codeql-action/pull/2124)
 - The CodeQL action no longer fails if it can't write to the telemetry api endpoint. [#2121](https://github.com/github/codeql-action/pull/2121)
+- Users of GHES3.9+ and GHEC will no longer need to include `actions: read` permissions to use `upload-sarif` in private repositories.
```
angelapwen, line 23:
```suggestion
- Users of GHES 3.9+ and GHEC will no longer need to include `actions: read` permissions to use `upload-sarif` in private repositories.
```

angelapwen (commented):
I had been planning to test something off of your forkmyself, but I haven't had time to in the last few days so I thought I'd ask if you were willing to test it and link the runs after testing 😄 

We'd like to make sure that for _reusable workflows_, the workflow path we originally got from the API is definitely the same as the one we're getting from `GITHUB_WORKFLOW_REF`. Basically, if they're not the same, alert categories will change after this change. They _should_ be the same, but we'd like to be certain 😸

jsoref:
@angelapwen: um, if someone gave me a workflow, I'm open to running it.

I haven't managed to get reusable workflows to do anything particularly useful for me, so I don't generally use them myself (I think there are some reusable workflows somewhere in one of the codeql repositories) -- mostly they've made me dizzy (I tried to figure out a way to set up a reusable workflow for check-spelling, and it didn't seem to add any value).

I have very little time tomorrow and won't have time until Tuesday (Monday's a holiday in Canada). I expect to be moderately busy next week, but I can generally run commands if people provide them.

(Sorry, I have no idea how/why I missed your event this morning...)

Nicolestevens657 (approved):

Issue #2139, opened 2024-02-14
Executing the CodeQL GitHub Action within a Docker container results in the detection of only a single CPU, due to the `/sys/fs/cgroup/cpuset.cpus` file being empty.


kouberl-bosch:
Hi, 

when using the github action in a container action with a docker image which has cgroup v2 it cannot be run multithreaded.

Example output:
```
While resolving threads, found a cgroup CPUs file with 32 CPUs in /sys/fs/cgroup/cpuset.cpus.effective.
While resolving threads, found a cgroup CPUs file with 1 CPUs in /sys/fs/cgroup/cpuset.cpus.
```

This happens because at my docker image (ubuntu:22.04) the /sys/fs/cgroup/cpuset.cpus is **empty** which is legit:
see **https://www.kernel.org/doc/Documentation/admin-guide/cgroup-v2.rst**
> If "cpuset.cpus" is empty, the "cpuset.cpus.effective" file shows
	all the CPUs from the parent cgroup that can be available to
	be used by this cgroup.
	
Now this lead to a maximum number of 1 CPU.
Relates to #2083

angelapwen:
Hello! Thank you for reporting. I can look into this for you as I authored https://github.com/github/codeql-action/pull/2083. I'd like to understand why `/sys/fs/cgroup/cpuset.cpus` is reporting 1 CPU here: https://github.com/github/codeql-action/blob/dc983b33c32da66f1acc44a1d5ed69d94e9ffb64/src/util.ts#L469. If I understand correctly, the file exists in your container but is empty. Is that right?

kouberl-bosch:
Yes the file exists, but is empty, I'm using ubuntu 22.04 docker image.

angelapwen:
Ah, I understand why the count is reporting as 1 now. I've put up https://github.com/github/codeql-action/pull/2151 which should address this issue. Thank you!

PR #2151, opened 2024-02-15
Account for existing but empty `cpus` file

angelapwen:
If a CPU file exists but is empty, previously we reported this file with a CPU count of 1, which resulted in a single-threaded run.

Fixes https://github.com/github/codeql-action/issues/2139.


src/util.ts
```diff
@@ -453,9 +453,9 @@ function getCgroupCpuCountFromCpus(
 
   let cpuCount = 0;
   // Comma-separated numbers and ranges, for eg. 0-1,3
-  const cpusString = fs.readFileSync(cpusFile, "utf-8");
+  const cpusString = fs.readFileSync(cpusFile, "utf-8").trim();
```
henrymercer, line 456: My understanding is that if the file is empty, we'll return `0`, but I think we probably want to return `undefined`?

src/util.ts
```diff
@@ -453,9 +453,9 @@ function getCgroupCpuCountFromCpus(
...
   for (const token of cpusString.split(",")) {
-    if (!token.includes("-")) {
+    if (token.length > 0 && !token.includes("-")) {
```
henrymercer, line 458: What does it mean if the token is empty here?  If it's not valid, we might want to log then bail out and return undefined.  If it's valid, is the else block doing the right thing?

henrymercer (commented):
Thanks for addressing this.  Couple of comments, and additionally, could we add a unit test?

src/util.ts
angelapwen: That could work too — we're filtering out all counts that are undefined, or <= 0, so they kind of have the same meaning. But I don't mind making it return undefined if that's more clear.

src/util.ts
angelapwen: If the file is empty here (or just has a newline character as in the linked issue) the `trim` should make it an empty string. Then `.split()` will also just return a single string that's empty. Previously we were incrementing the count because an empty string doesn't include `-`. You're right that we shouldn't be entering the `else` block though... we should just be bailing out. I'll update this and look at adding the unit test 👍

src/util.ts
henrymercer: Ah, that's good that we were already filtering out <=0.  I think I'd find `undefined` slightly more clear, but it's not critical.

src/util.ts
angelapwen: Okay — I added unit tests for the case where the file exists and the number of CPUs should be calculated, the case where the file doesn't exist, and the case where the file exists but only has a newline char in it!

src/util.ts
angelapwen: The updated logic returns `undefined` 😄

henrymercer (approved):
Nice!

PR #2160, opened 2024-02-22
Enable C++ deptrace when using autobuild build mode

henrymercer:
This PR fixes a bug with the experimental internal-only build mode functionality where we wouldn't enable deptrace when extracting C++ using `database trace-command`.  Risk is low since this functionality is guarded behind the internal-only `build-mode` input.


src/analyze.ts
```diff
@@ -191,6 +192,12 @@ export async function runExtraction(
         config.buildMode &&
         (await codeql.supportsFeature(ToolsFeature.TraceCommandUseBuildMode))
       ) {
+        if (
+          language === Language.cpp &&
+          config.buildMode === BuildMode.Autobuild
```
aeisenberg, line 197: So, just to be clear then, by default this value is `undefined`? I think that's the case looking at `init/action.yml`.

aeisenberg (approved):
Yep. This looks fine.

src/analyze.ts
henrymercer: That's right, we won't even reach this branch as `config.buildMode` above will be falsy.

PR #2161, opened 2024-02-22
Only warn about path filters in the Action if the CLI doesn't already warn

henrymercer:
We are moving the notice when a language does not support path filtering into the CLI so that it's available for all customers including those using third-party CI systems.  Therefore, to avoid duplicate warnings, this PR updates the Action such that it only warns for older CLIs that don't support this feature.

This does mean that customers on older versions of the Action using newer CLIs might see the warning twice in the logs.  I don't think this is enough of a problem to be worth addressing.


angelapwen (approved):
Makes sense to me, and agree that the double-warning for folks on older versions isn't too big of a deal 😄

PR #2170, opened 2024-02-26
Add build mode, Actions event name, and languages to the status report

henrymercer:
This PR adds the build mode and Actions event name to the status report.  We also record the languages for all first-party Actions rather than just `init` so that we can more easily use this for incident response.


src/status-report.ts
```diff
@@ -79,6 +80,12 @@ export interface StatusReportBase {
    * telemetry tables.
    */
   job_run_uuid: string;
+  /**
+   * Comma-separated list of languages that analysis was run for.
```
angelapwen, line 89: Looks like we're now sending space-separated list rather than comma-separated: is that intentional? 👀

henrymercer: Thanks for the catch, agreed we should be sending comma-separated for consistency.

angelapwen (approved):
Looks good to me!

PR #2171, opened 2024-02-28
Classify third-party upload `SyntaxError`s as configuration errors

angelapwen:
We frequently see `SyntaxError`s from the upload action that are currently classified as failures. We were previously catching this error in the `countResultsInSarif` method, but this method runs after `validateSarifFileSchema`, which would have thrown a `SyntaxError` before we're able to throw it as an `InvalidRequestError`.

This change moves the error catching to `validateSarifFileSchema` so that we will appropriately classify it as an `InvalidRequestError`.


henrymercer (approved):
LGTM.  As a bonus, we could produce a slightly better error message, e.g. "Could not upload <path> as it is not a valid JSON file."

angelapwen:
Ah.. I think we are already attempting to catch this error, but it's not happening in the right place. Let me try to fix.

henrymercer (approved):

PR #2173, opened 2024-02-28
Mark limits errors from third-party SARIF uploads as configuration errors

henrymercer:
I think this was the intention of the original PR, but the place that we were converting `InvalidSarifUploadError`s to `ConfigurationErrors` wasn't including processing errors.


src/analyze-action.ts
```diff
@@ -286,7 +286,6 @@ async function run() {
         actionsUtil.getRequiredInput("checkout_path"),
         actionsUtil.getOptionalInput("category"),
         logger,
-        { isThirdPartyUpload: false },
```
angelapwen, line 289: I'm not sure I understand why we can remove `isThirdPartyUpload` here: wouldn't this mean that we are also marking first-party upload failures as configuration errors?

angelapwen: That makes sense 👍

angelapwen (approved):
Looks good to me 👍 after this is merged I'll kick off an Action release to get this + the `SyntaxError` changes in.

PR #2189, opened 2024-03-08
Fail `analyze` step in PR check by passing an invalid option to `database finalize`

angelapwen:
The `ram: 1` trick won't work anymore with recent updates to the CLI, so we need to find an alternate way to fail the `analyze` step here. 

### DIFF
--- a/.github/workflows/debug-artifacts-failure.yml
+++ b/.github/workflows/debug-artifacts-failure.yml
@@ -50,9 +50,11 @@ jobs:
         run: ./build.sh
       - uses: ./../action/analyze
         id: analysis
+        env: 
+          # Forces a failure in this step.
+          CODEQL_ACTION_EXTRA_OPTIONS: '{ "database": { "finalize": ["--invalid-option"] } }'
         with:
           expect-error: true
-          ram: 1
   download-and-check-artifacts:
     name: Download and check debug artifacts after failure in analyze
     needs: upload-artifacts


angelapwen:
`setup-swift` failures are upstream failures (https://github.com/swift-actions/setup-swift/issues/666). Hopefully, these are resolved soon — otherwise maybe I'll have to make those checks non-required so I can get this in to unblock https://github.com/github/codeql-action/pull/2185.

.github/workflows/debug-artifacts-failure.yml
```diff
@@ -50,9 +50,11 @@ jobs:
         run: ./build.sh
       - uses: ./../action/analyze
         id: analysis
+        env: 
+          # Forces a failure in this step.
+          CODEQL_ACTION_EXTRA_OPTIONS: "{ \"database\": { \"finalize\": [\"--invalid-option\"] } }"
```
adityasharad, line 55: Observation: This tests the case where `database finalize` fails and we want to check we have debug artifacts. It doesn't test the case where `database run-queries` or `database interpret-results` fails and similarly we want to check debug artifacts. That is okay for now, but we may want to expand the test in future to tamper with each of these commands separately.

adityasharad (approved):

.github/workflows/debug-artifacts-failure.yml
adityasharad: Also I think you could have gotten away without the `\"` escaping by using single quotes to enclose the entire string. See https://github.com/search?type=code&q=CODEQL_ACTION_EXTRA_OPTIONS for some examples.

angelapwen: Makes sense. Ah! I didn't think to search in the entire org (only searched in this repo). I'll update to single quotes to make it more readable.

adityasharad (approved):

PR #2191, opened 2024-03-11
Use the `--sarif-include-query-help` option when supported

henrymercer:
This avoids a deprecation warning appearing in the logs.


src/codeql.ts
```diff
@@ -1389,3 +1394,14 @@ async function isSublanguageFileCoverageEnabled(
     ))
   );
 }
+
+async function getCodeScanningQueryHelpArguments(
+  codeql: CodeQL,
+): Promise<string[]> {
```
aeisenberg, line 1400: Minor: This can return `Promise<string>` and the use above can remove the `...`.

aeisenberg (approved):
LGTM with a minor suggestion.

src/codeql.ts
henrymercer: That's true, but there's also something to be said for having these helper functions consistently return `string[]` 🤔 I think I'll leave as is.

PR #2198, opened 2024-03-14
Improve tracking of autobuild errors

henrymercer:
Improve the way we surface errors while automatically attempting to build code, particularly in circumstances where the autobuilder doesn't produce diagnostic information.

Specifically, extract just the `[ERROR]` part of the stderr stream, and link users to the troubleshooting page in the docs.

Also mark Maven and Gradle build failures as configuration errors.


src/cli-errors.ts
```diff
@@ -145,6 +165,11 @@ export const cliErrorsConfig: Record<
   [CliConfigErrorCategory.InvalidSourceRoot]: {
     cliErrorMessageCandidates: [new RegExp("Invalid source root")],
   },
+  [CliConfigErrorCategory.MavenBuildFailed]: {
+    cliErrorMessageCandidates: [
+      new RegExp("[autobuild] [ERROR] Failed to execute goal"),
```
nickfyson, line 170: I haven't had to wrangle Maven in many a year now, but I take it this is an extremely characteristic string? 😆

src/cli-errors.ts
```diff
@@ -96,6 +107,14 @@ function extractFatalErrors(error: string): string | undefined {
   return undefined;
 }
 
+function extractAutobuildErrors(error: string): string | undefined {
+  const pattern = /.*\[autobuild\] \[ERROR\] (.*)/gi;
+  return (
+    [...error.matchAll(pattern)].map((match) => match[1]).join("\n") ||
```
nickfyson, line 113: Is there any value in bounding the max number of lines we'll return here, or given they're generated by our own script are we confident it'll never be _too_ outlandish...? 🙂

nickfyson (approved):
LGTM

src/cli-errors.ts
henrymercer: I think this is fairly characteristic :)

src/cli-errors.ts
henrymercer: Good point — we later cap what we send to telemetry, but I think it would probably make sense to add a cap here too.

nickfyson (approved):

nickfyson (approved):

PR #2199, opened 2024-03-15
Use `github merge-results` command for merging SARIF files

koesie10:
This will use the `github merge-results` command when the `cli_sarif_merge_enabled` feature flag is enabled and all SARIF files were produced by CodeQL.


src/feature-flags.ts
```diff
@@ -58,6 +59,13 @@ export const featureConfig: Record<
   Feature,
   { envVar: string; minimumVersion: string | undefined; defaultValue: boolean }
 > = {
+  [Feature.CliSarifMerge]: {
+    envVar: "CODEQL_ACTION_CLI_SARIF_MERGE",
+    // This feature is only supported in 2.17.0, but we'll want to test this feature
+    // with nightly builds before that. We'll update this to 2.17.0 once it's released.
+    minimumVersion: undefined,
```
aeisenberg, line 66: I should have mentioned this earlier when you were working on the CLI change, but I wasn't able to give any deep review. We have a better way of handling feature enablement in the CLI.

The `codeql version --format json` command will print out a list of features that the CLI version supports. You can add a field to `com.semmle.cli2.VersionCommand.Features` for merging sarif results. And in the action you no longer need to specify a specific version. Instead, you can call `supportsFeature` on the `codeql` object. This is a cleaner and safer way of checking for feature support. It's a small change to the CLI. 

If implementing this will cause a delay in shipping the merge SARIF feature for the CLI, we can consider backporting to the ongoing CLI release.

koesie10: Thanks, I'll implement this in the CLI. I don't think this will change anything about the timeline for the CLI release since the command would already be part of 2.17.0 and this won't be released for a few weeks.

src/codeql.ts
```diff
@@ -191,6 +195,14 @@ export interface CodeQL {
   ): Promise<void>;
   /** Get the location of an extractor for the specified language. */
   resolveExtractor(language: Language): Promise<string>;
+  /**
+   * Run 'codeql github merge-results'.
+   */
+  mergeResults(
+    sarifFiles: string[],
+    outputFile: string,
+    mergeRunsFromEqualCategory?: boolean,
```
henrymercer, line 204: Nit: Make this `{ mergeRunsFromEqualCategory?: boolean }` so we can see what we're setting to `true` when we call this function.

src/codeql.ts
```diff
@@ -1077,6 +1090,29 @@ export async function getCodeQLForCmd(
...
+
+      for (const sarifFile of sarifFiles) {
+        args.push("--sarif", sarifFile);
+      }
+
+      if (mergeRunsFromEqualCategory) {
+        args.push("--sarif-merge-runs-from-equal-category");
+      }
+
+      await new toolrunner.ToolRunner(cmd, args).exec();
```
henrymercer, line 1114: Use `runTool` instead for some niceties like printing the command to the logs and capturing configuration errors (we should do this for other commands too, but it doesn't have to happen in this PR).

src/feature-flags.ts
```diff
@@ -58,6 +59,12 @@ export const featureConfig: Record<
   Feature,
   { envVar: string; minimumVersion: string | undefined; defaultValue: boolean }
 > = {
+  [Feature.CliSarifMerge]: {
+    envVar: "CODEQL_ACTION_CLI_SARIF_MERGE",
+    // This is guarded by a `supportsFeature` check rather than by a version check.
+    minimumVersion: undefined,
```
henrymercer, line 65: Going forwards, this will probably be the common case as we're preferring checking CLI features to checking version numbers.  We should consider adding a `toolVersion` field to this record so callees can just check the feature.

src/upload-lib.ts
```diff
@@ -48,6 +61,103 @@ function combineSarifFiles(sarifFiles: string[]): SarifFile {
...
+
+    const allRunsCodeQL = sarifObject.runs?.every(
+      (run) => run.tool?.driver?.name === "CodeQL",
+    );
+
+    if (!allRunsCodeQL) {
+      allCodeQL = false;
+      break;
+    }
+  }
```
henrymercer, line 90: Nit: Move this into a helper function for readability.

src/upload-lib.ts
```diff
@@ -48,6 +61,103 @@ function combineSarifFiles(sarifFiles: string[]): SarifFile {
   return combinedSarif;
 }
 
+// Takes a list of paths to sarif files and combines them together using the
+// CLI `github merge-results` command when all SARIF files are produced by
+// CodeQL. Otherwise, it will fall back to combining the files in the action.
+// Returns the contents of the combined sarif file.
+async function combineSarifFilesUsingCLI(
```
henrymercer, line 68: Can we return early if only one SARIF file was specified?

src/upload-lib.ts
```diff
@@ -48,6 +61,103 @@ function combineSarifFiles(sarifFiles: string[]): SarifFile {
...
+    if (!allRunsCodeQL) {
+      allCodeQL = false;
+      break;
+    }
+  }
+
+  if (!allCodeQL) {
+    logger.warning(
+      "Not all SARIF files were produced by CodeQL. Merging files in the action.",
+    );
```
henrymercer, line 95: It might not be clear to users why this is a warning.  Could we make it a debug log, or explain what we want users to do instead?

src/upload-lib.ts
```diff
@@ -48,6 +61,103 @@ function combineSarifFiles(sarifFiles: string[]): SarifFile {
...
+  let tempDir: string = actionsUtil.getTemporaryDirectory();
+
+  const config = await getConfig(tempDir, logger);
+  if (config !== undefined) {
+    codeQL = await getCodeQL(config.codeQLCmd);
+    tempDir = config.tempDir;
+  } else {
+    logger.warning(
+      "Initializing CodeQL since the 'init' Action was not called before this step.",
+    );
```
henrymercer, line 113: Similarly, this could probably be an info statement as it will be the common case for default setup.

src/upload-sarif-action-post-helper.ts
```diff
@@ -0,0 +1,49 @@
...
+  uploadDebugArtifacts: (
+    toUpload: string[],
+    rootDir: string,
+    artifactName: string,
+  ) => Promise<void>,
+) {
+  const tempDir = actionsUtil.getTemporaryDirectory();
+
+  // Upload Actions SARIF artifacts for debugging
+  if (core.isDebug()) {
```
henrymercer, line 18: I think this is a fairly broad behaviour change for the `upload-sarif` Action, since some customers will run with debug logging turned on for all workflows, and we don't want to use up too much of their artifact storage.  Consider putting this behind an environment variable.

henrymercer (commented):
This looks generally great.  Some initial comments:

src/feature-flags.ts
koesie10: Are you suggesting that I add a `toolsFeature` field to this object in this PR?

src/upload-lib.ts
koesie10: Yes, this should be a debug log for now. Once we actually deprecate this behavior, we should make this a warning with more actionable instructions.

src/feature-flags.ts
henrymercer: I'm happy if you'd rather do this in a separate PR, but I think if we rely on checking manually we'll probably make a mistake soon enough.

koesie10: I'm not sure if it makes sense for this specific feature flag though since we usually do not have a `CodeQL` object available in the `upload-sarif` action, so we wouldn't be able to check the `toolsFeature` anyway. We probably also don't want to let the future deprecation behavior depend on which version of CodeQL is being used since the deprecation will happen in the Code Scanning API rather than in CodeQL.

henrymercer: > I'm not sure if it makes sense for this specific feature flag though since we usually do not have a `CodeQL` object available in the `upload-sarif` action, so we wouldn't be able to check the `toolsFeature` anyway.

That makes sense — if the feature flag is disabled via the API, there's no point downloading CodeQL.

> We probably also don't want to let the future deprecation behavior depend on which version of CodeQL is being used since the deprecation will happen in the Code Scanning API rather than in CodeQL.

I think we'd probably want to time the deprecation such that the CLIs that can't merge SARIF runs are no longer supported by the time we deprecate the API.

src/upload-sarif-action-post-helper.ts
```diff
@@ -0,0 +1,52 @@
...
+    toUpload: string[],
+    rootDir: string,
+    artifactName: string,
+  ) => Promise<void>,
+) {
+  const tempDir = actionsUtil.getTemporaryDirectory();
+
+  // Upload Actions SARIF artifacts for debugging
+  if (
+    core.isDebug() &&
```
henrymercer, line 19: Minor: no need to check this if we're checking the environment variable.

src/analyze-action-post.ts
```diff
@@ -7,11 +7,16 @@ import * as core from "@actions/core";
...
+import * as uploadSarifActionPostHelper from "./upload-sarif-action-post-helper";
 import { wrapError } from "./util";
 
 async function runWrapper() {
   try {
     await analyzeActionPostHelper.run(debugArtifacts.uploadSarifDebugArtifact);
+
+    // Also run the upload-sarif post action since we're potentially running
+    // the same steps in the analyze action.
+    await uploadSarifActionPostHelper.run(debugArtifacts.uploadDebugArtifacts);
```
henrymercer, line 19: Could you rename the `run` method to something like `uploadArtifacts`?  In general, we might not want the analyze Action to run the entire post step of `upload-sarif` — for example that step might upload telemetry information.  It would be good to encode this somehow.

src/codeql.ts
```diff
@@ -50,6 +50,10 @@ interface ExtraOptions {
     extractor?: Options;
     queries?: Options;
   };
+  github?: {
+    "*"?: Options;
+    "merge-results"?: Options;
+  };
```
henrymercer, line 56: Not for you to fix, but I don't think this addition does anything.

src/upload-lib.ts
```diff
@@ -48,6 +61,113 @@ function combineSarifFiles(sarifFiles: string[]): SarifFile {
...
+    const allRunsCodeQL = sarifObject.runs?.every(
+      (run) => run.tool?.driver?.name === "CodeQL",
+    );
+
+    if (!allRunsCodeQL) {
+      return false;
+    }
+  }
+
+  return true;
```
henrymercer, line 83: This could probably be written functionally using `sarifFiles.every`, since `Array.every` also short-circuits.

henrymercer (commented):
Looks good to me — just a couple of final comments.

henrymercer (approved):
Great, thanks!

PR #2206, opened 2024-03-20
 Improve error message when using build modes and autobuild fails

henrymercer:
Include a more specific error message tailored to the CodeQL Action.

Also wrap configuration errors in all CLI invocations, since the patterns we've specified in `cli-errors.ts` are specific enough that to be applied to all CLI invocations without creating false negative failures.


angelapwen (approved):
Sounds good re. wrapping all CLI invocations 👍

PR #2213, opened 2024-03-26
Move `@types/node` to bottom of devDependencies

aeisenberg:
This will make it less likely for a dependabot update to cause conflicts when backporting from v3 -> v2. It's a little clunky, but should probably work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,9 +73,6 @@
     "removeNPMAbsolutePaths": "3.0.1",
     "sinon": "^17.0.1",
     "typescript": "^5.4.3",
-
-
-
     "@types/node": "20.9.0"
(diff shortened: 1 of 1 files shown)


angelapwen:
Looks like the dependencies check is failing.. let me add the `update dependencies` label and see if it changes `package-lock.json`.

aeisenberg:
Yeah...this approach won't work. It's failing because we recreate the package.json and check to see if there are any differences with what is checked in. And there always will be since the way we recreate the package.json file, the dependencies are sorted alphabetically.

angelapwen:
Ok, don't think this will work unfortunately 😢

PR #2225, opened 2024-04-04
Gracefully continue if `createStatusReportBase` throws

angelapwen:
Previously, we weren't catching any possible exceptions in `createStatusReportBase` and runs would fail if any of the telemetry sub-items threw exceptions. As telemetry should not block the analysis, we continue here even if the status report throws.


src/analyze-action.ts
```diff
@@ -74,22 +74,24 @@ async function sendStatusReport(
...
+  if (statusReportBase !== undefined) {
+    const report: FinishStatusReport = {
+      ...statusReportBase,
+      ...(stats || {}),
+      ...(dbCreationTimings || {}),
     };
-    await statusReport.sendStatusReport(trapCacheUploadStatusReport);
-  } else {
-    await statusReport.sendStatusReport(report);
+    if (config && didUploadTrapCaches) {
```
aeisenberg, line 83: Do we ever reasonably expect config to be falsy, but didUploadTrapCaches to be `true`? If so, we might want to still upload in that case:

```suggestion
    if (didUploadTrapCaches) {
```

And the change below. But only if this is not already an error state.

src/analyze-action.ts
```diff
@@ -74,22 +74,24 @@ async function sendStatusReport(
...
+      const trapCacheUploadStatusReport: FinishWithTrapUploadStatusReport = {
+        ...report,
+        trap_cache_upload_duration_ms: Math.round(trapCacheUploadTime || 0),
+        trap_cache_upload_size_bytes: Math.round(
+          await getTotalCacheSize(config.trapCaches, logger),
```
aeisenberg, line 88: Only if also adding above.

```suggestion
          await getTotalCacheSize(config?.trapCaches || {}, logger),
```

src/status-report.ts
```diff
@@ -243,103 +244,112 @@ export async function createStatusReportBase(
...
+      statusReport.cause = cause;
+    }
+    if (exception) {
+      statusReport.exception = exception;
+    }
+    if (
+      status === "success" ||
+      status === "failure" ||
+      status === "aborted" ||
+      status === "user-error"
```
aeisenberg, line 322: Should this change to `configuration-error`? (Not in this PR)

aeisenberg (approved):
Nice. A few comments inline. If the answer to my question is "no", then this can go in as-is.

src/analyze-action.ts
angelapwen: I don't think so — if `config` is `undefined`, we should be erroring out much earlier: https://github.com/github/codeql-action/blob/f421cda8e715cc7652d8b748f49574cf286aec12/src/analyze-action.ts#L205.

src/status-report.ts
angelapwen: We could: the `configuration-error` was codified in the overall job status (rather than status for each action, which is what this is). It would be nice to have them use the same syntax. But we would want to make sure that everyone consuming the data are aware about the transition, because we likely have existing monitors/queries/dashboards using `user-error`. I can write up an internal issue for this.

aeisenberg: Sure. I'd say it's low priority since it's just changing a name.

aeisenberg (approved):

PR #2229, opened 2024-04-05
Add the overwrite flag when init-ing databases

aeisenberg:
This ensures that if there is a database already in the filesystem, we can gracefully handle creating a new one. This sometimes happens when a user is on a self-hosted runner and neglects to clean state across runs.


angelapwen (approved):
Great, thank you!

PR #2231, opened 2024-04-09
CI: apply `setup-swift` workaround

redsun82:


henrymercer (approved):
Thanks!  Are you planning to contribute this upstream?

redsun82:
I'm trying to nudge someone to do it more cleanly (my fix is probably POSIX only, and assumes `gunzip` is installed, which is ok for github hosted runners but might be not solid enough for widespread usage). But I might have a shot at it if nobody else steps up.

PR #2234, opened 2024-04-11
Fail more clearly if the `file` program is not installed on the runner

mbg:
This PR checks more clearly whether we can successfully find the location of the `file` program and emits a clearer message in the build log as well as a warning-level diagnostic if it cannot be found. There is also a new integration test which checks that this works as intended.

Since the check for whether the Go workaround is needed takes place before the database gets initialised, this PR also modifies the code for diagnostics to:

- Check if the database exists yet before writing a diagnostic.
- If it does not exist, the diagnostic is added to an array and flushed to disk later once the database is presumed to exist (`init` Action only)


src/diagnostics.ts
```diff
@@ -107,3 +148,18 @@ export function addDiagnostic(
     logger.warning(`Unable to write diagnostic message to database: ${err}`);
   }
 }
+
+/** Writes all unwritten diagnostics to disk. */
+export function flushDiagnostics(config: Config) {
```
henrymercer, line 153: This looks good.  Some high level comments:

- Should we have a method that we call unconditionally at the end of the `init` Action to check that `unwrittenDiagnostics` is empty, and warn if it isn't?
- Once we have that, I think we can make the logging messages debug level.

src/init-action.ts
```diff
@@ -372,6 +374,27 @@ async function run() {
...
           `Failed to determine the location of the Go binary: ${e}`,
         );
+
+        if (e instanceof FileCmdNotFoundError) {
+          addDiagnostic(
+            config,
+            Language.go,
+            makeDiagnostic(
+              "go/workflow/file-program-unavailable",
+              "The `file` program is required, but does not appear to be installed",
```
henrymercer, line 384: Worth pointing out that this requirement is Linux only to avoid potential confusion?

mbg: Not sure that's necessary since the `try`/`catch` block is already guarded by:

```typescript
if (
  config.languages.includes(Language.go) &&
  process.platform === "linux"
) 
```

src/diagnostics.ts
mbg: Good idea, that makes sense to me. I will add something.

src/init-action.ts
henrymercer: Just thinking of how it will display on the tool status page.  Not blocking.

mbg: Ah! I hadn't thought about that. You're right that it may not be clear that the diagnostic originated from a linux run if there are multiple CodeQL workflows on different operating systems. I will add something.

henrymercer (approved):
Very nice, I expect the functionality for writing diagnostics before the database has been initialized will come in handy elsewhere too.

You might want to consider adding a changelog note, otherwise LGTM!

mbg:
Added a changelog entry

henrymercer (approved):

PR #2235, opened 2024-04-12
Improve reliability and performance when using the `autobuild` build mode

henrymercer:
This PR directly traces autobuild scripts when a build mode is specified to the `init` Action.

Before, the `init` Action did not know whether the workflow would call the autobuilder or run manual build scripts, so we always had to start indirect tracing in case we needed to trace manual build steps.

However in new-style workflows where the build mode is specified, this is no longer necessary, and we only need to use indirect tracing when the build mode is `manual`.  For the `autobuild` build mode, we can avoid starting indirect tracing and directly trace the autobuild script using `database trace-command`.  We expect this to provide reliability and performance benefits, which we will validate.

This is behind the `autobuild_direct_tracing_enabled` feature flag.


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
   We recommend removing any references to these from your workflows. For more information, see the release notes for CodeQL Action v3.23.0 and v2.23.0.
 - Automatically overwrite an existing database if found on the filesystem. [#2229](https://github.com/github/codeql-action/pull/2229)
 - Bump the minimum CodeQL bundle version to 2.12.6. [#2232](https://github.com/github/codeql-action/pull/2232)
+- We are rolling out a feature in April 2024 that improves the reliability and performance of analyzing code when using the `autobuild` build mode. This feature uses direct tracing to trace the build when the `autobuild` build mode is specified using the `build-mode` input to the `init` Action. [#2335](https://github.com/github/codeql-action/pull/2335)
```
aeisenberg, line 19: There's a lot going on in this changelog. Can you add some links to external docs around direct tracing. Also, is there a github changelog post for the build-mode changes you can link to?

pr-checks/checks/autobuild-direct-tracing.yml
```diff
@@ -0,0 +1,30 @@
+name: "Autobuild direct tracing"
+description: "An end-to-end integration test of a Java repository built using 'build-mode: autobuild', with direct tracing enabled"
+operatingSystems: ["ubuntu"]
```
aeisenberg, line 3: Should we be testing on windows, too?

pr-checks/checks/autobuild-direct-tracing.yml
```diff
@@ -0,0 +1,30 @@
...
+versions: ["latest", "nightly-latest"]
```
aeisenberg, line 4: Not something to fix in this PR, but how can we ensure that as we release new versions of the CLI, we add them to the matrix? There are other integration tests that only run on `"latest", "nightly-latest"` because when we wrote them, those were the only versions that supported the feature being tested.

Could we add a `minimum-version` property so that tests that specify it run on all CLI versions >= the minimum version?

aeisenberg (commented):
Looks great.

src/codeql.ts
```diff
@@ -681,6 +671,9 @@ export async function getCodeQLForCmd(
       ]);
     },
     async extractUsingBuildMode(config: Config, language: Language) {
+      if (config.buildMode === BuildMode.Autobuild) {
+        applyAutobuildAzurePipelinesTimeoutFix();
+      }
```
angelapwen, line 698: I'm not sure I understand why're now applying this fix only if we're using autobuild/direct tracing 🤔 it looks like we set those options unconditionally in the past.

pr-checks/checks/autobuild-direct-tracing.yml
henrymercer: I like this idea, but I think it's beyond the scope of this PR.  I'll file an issue.

src/codeql.ts
henrymercer: I think this PR should be additive when it comes to applying this fix — previously we applied it to the autobuild Action only, now we also apply it to `database trace-command` calls.

CHANGELOG.md
henrymercer: Fair point — I'll leave the details out of the changelog note, and link some docs about build modes.

pr-checks/checks/autobuild-direct-tracing.yml
henrymercer: Yes, I think this makes sense.  It doesn't really give us any additional confidence that the code changes are correct, but it does give us better confidence that direct tracing is working as expected.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
   We recommend removing any references to these from your workflows. For more information, see the release notes for CodeQL Action v3.23.0 and v2.23.0.
 - Automatically overwrite an existing database if found on the filesystem. [#2229](https://github.com/github/codeql-action/pull/2229)
 - Bump the minimum CodeQL bundle version to 2.12.6. [#2232](https://github.com/github/codeql-action/pull/2232)
-- We are rolling out a feature in April 2024 that improves the reliability and performance of analyzing code when using the `autobuild` build mode. This feature uses direct tracing to trace the build when the `autobuild` build mode is specified using the `build-mode` input to the `init` Action. [#2335](https://github.com/github/codeql-action/pull/2335)
+- We are rolling out a feature in April/May 2024 that improves the reliability and performance of analyzing code when analyzing a compiled language with the `autobuild` [build mode](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages#codeql-build-modes). [#2335](https://github.com/github/codeql-action/pull/2335)
```
angelapwen: Just noticed PR number should be 2235 instead of 2335 😸

angelapwen (approved):

angelapwen (approved):

PR #2245, opened 2024-04-16
Ensure `--overwrite` flag is only passed once

henrymercer:
This fixes a bug where if `CODEQL_ACTION_EXTRA_OPTIONS` specified that `--overwrite` should be passed to `database init`, the `--overwrite` argument would be passed twice, leading to an error.

To accomplish this, I introduce an `ignoringOptions` argument to specify options in `CODEQL_ACTION_EXTRA_OPTIONS` that should be ignored.  It would be nice to have a general solution that didn't require specifying which options should be ignored, but the problem with that is that it does make sense to pass some flags more than once.


angelapwen (approved):
Makes sense — did you include `--expect-discarded-cache` because that's another obvious one that should only be passed once? I wonder if that one should also be part of the changenote (though I guess then the note would need to be more generalized so we don't have to write a note each time we pass `ignoringOptions`).

nickfyson (approved):

henrymercer:
> Makes sense — did you include `--expect-discarded-cache` because that's another obvious one that should only be passed once? I wonder if that one should also be part of the changenote (though I guess then the note would need to be more generalized so we don't have to write a note each time we pass `ignoringOptions`).

I had a quick scan through and that was the most obvious other one to add.  I'll just include `--overwrite` in the changelog note though as I don't think we've advised customers to specify `--expect-discarded-cache` manually.

PR #2251, opened 2024-04-18
Use `merge=union` for `package.json`

angelapwen:
Occasionally the v3 -> v2 mergeback script for releases fails due to a merge conflict in `package.json`.

Specifically, when the line below "@types/node" (currently "@types/semver") is [bumped in the Action](https://github.com/github/codeql-action/blob/294b6df61d4812e7497e15a9d0f240a97f453b45/package.json#L60) by Dependabot, there is a merge conflict between v3 and v2 because node should be version 20 for v3 and version 16 for v2.

Then the call to `npm version` [in the release script](https://github.com/github/codeql-action/blob/1aae1e7090c64083768b852cc5f59346312fb276/.github/update-release-branch.py#L377) fails because the file is no longer valid JSON due to the merge conflict; so the mergeback PR is not created at all and the release conductor needs to resolve the conflict and restart the script locally.

This change adds `merge=union` as a git attribute for `package.json` so that if there is a merge conflict, both changes will be included. This will prevent parsing errors in `package.json` — at least it will be valid JSON — when `npm version` is run (the release conductor will need to manually edit the file to resolve the merge conflict, but that's already part of the autogenerated checklist that's created in the associated PR. 

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,10 @@
 lib/*.js linguist-generated=true
 .github/workflows/__* linguist-generated=true
 
-# Reduce incidence of needless merge conflicts on CHANGELOG.md
+# Reduce incidence of needless merge conflicts on CHANGELOG.md and package.json
 # The man page at
 # https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html
 # suggests that this might interleave lines arbitrarily, but empirically
 # it keeps added chunks contiguous
 CHANGELOG.md                merge=union
+package.json                merge=union


nickfyson (approved):
Makes sense to me!

henrymercer (changes requested):
As you point out, this probably isn't the right way to resolve merge conflicts in general — we'll end up with duplicate keys in the JSON.  I think this has the potential to create some bugs.

For example, if you add a duplicate key to `package.json` and then run `npm version`, it swallows one of the keys.  So we'd never find out about the merge conflict, and might end up with the wrong version of the package.

I think a textual replacement might be the safest way to update the version number in the presence of merge conflicts.

angelapwen:
Ah.. I hadn't realized that `npm version` would then swallow one of the keys. I'll close in favor of a textual replacement fix then. Thanks both for taking a look!

PR #2261, opened 2024-04-25
Log warning if SIP is disabled and CLI version is < 2.15.1

angelapwen:
The `macos-latest` image is now ARM rather than Intel, so we need to change some of our PR Checks. The build tracer on CLI versions before v2.15.1 did not support the ARM machines where System Integrity Protection was disabled, which now includes `macos-latest`.

This change:
- uses `macos-12` Intel runners for any checks where the CLI version is below v2.15.1
- logs a warning if SIP is disabled, and if the CLI version is below v2.15.1

Separately, the `macos-latest` image no longer supports Go on the path by default, so this PR adds `setup-go` to all PR checks analyzing Go. 

I've updated the `Required` PR checks on `main` to include the new ones, but have not updated `v2` or `v3` yet.


src/init-action.ts
```diff
@@ -467,6 +473,18 @@ async function run() {
...
 
+    // For CLI versions <2.15.1, build tracing caused errors in MacOS ARM machines with
+    // System Integrity Protection (SIP) disabled.
+    if (
+      !(await codeQlVersionAbove(codeql, "2.15.1")) &&
+      process.platform === "darwin" &&
+      !(await isSipEnabled(logger))
+    ) {
+      logger.warning(
+        "CodeQL versions 2.15.0 and lower are not supported on MacOS ARM machines with System Integrity Protection (SIP) disabled.",
```
adityasharad, line 484: The warning mentions ARM, but you're not checking `process.arch`.
Either we can change the warning to say macOS in general with SIP disabled is not supported on <=2.15.0 (not strictly true, but I don't know if we fixed other relocation issues that would affect Intel),
or change the code above to check `process.arch` being `arm` or `arm64`.

src/init-action.ts
```diff
@@ -467,6 +473,18 @@ async function run() {
       }
     }
 
+    // For CLI versions <2.15.1, build tracing caused errors in MacOS ARM machines with
+    // System Integrity Protection (SIP) disabled.
+    if (
+      !(await codeQlVersionAbove(codeql, "2.15.1")) &&
```
adityasharad, line 479: Minor, separate: I had to go look at the definition to remind myself whether this was `>` or `>=`. Perhaps we should rename it `codeQlVersionAtLeast` or similar.

pr-checks/sync.py
```diff
@@ -81,7 +89,8 @@ def writeHeader(checkStream):
             # Ensure that this is serialized as a folded (`>`) string to preserve the readability
             # of the generated workflow.
             'if': FoldedScalarString(textwrap.dedent('''
-                    matrix.os == 'macos-latest' && (
+                    (matrix.os == 'macos-latest' || 
+                    matrix.os == 'macos-12') && (
```
adityasharad, line 93: Can we simplify this to just `runner.os == 'macOS'`?

adityasharad (commented):
Thanks - generally looks good but a couple of minor suggestions.

src/init-action.ts
angelapwen: Ah yes, good point 👍 will change it to `process.arch` as I believe that's more accurate to the problem we were looking at in 2.15.1.

src/init-action.ts
angelapwen: Good point, I always double check that it's using `gte` too. I've made the change

pr-checks/sync.py
angelapwen: Yep, done!

src/init-action.ts
```diff
@@ -477,7 +477,7 @@ async function run() {
     // System Integrity Protection (SIP) disabled.
     if (
       !(await codeQlVersionAbove(codeql, "2.15.1")) &&
-      process.platform === "darwin" &&
```
adityasharad: This line needs to remain!

angelapwen: Oh yes, Linux exists 😆

adityasharad (approved):

PR #2265, opened 2024-04-26
Add deprecation warning for merging SARIF files with non-unique categories

koesie10:
This adds a deprecation warning for merging SARIF files with non-unique categories. This is behind a feature flag.

These are the criteria for showing the deprecation warning:
- The feature flag is enabled
- If on GHES: GHES version >= 3.14.0
- The deprecation warning hasn't been shown in the same run
- Non-unique categories are found


src/upload-lib.ts
```diff
@@ -149,6 +238,19 @@ async function combineSarifFilesUsingCLI(
...
 
+    if (
+      await showCombineSarifFilesDeprecationWarning(
+        sarifObjects,
+        features,
+        gitHubVersion,
+      )
+    ) {
+      logger.warning(
+        `Uploading multiple CodeQL runs with the same category is deprecated ${deprecationWarningMessage}. Please update your CodeQL CLI version or update your workflow to set a distinct category for each CodeQL run. For more information, see https://github.blog/changelog/2024-05-06-code-scanning-will-stop-combining-runs-deprecation-notice`,
```
aeisenberg, line 249: Is this the same message as above? If so, probably best to extract to a variable so we don't accidentally change one but not the other.

src/upload-lib.ts
```diff
@@ -65,20 +68,84 @@ function combineSarifFiles(sarifFiles: string[], logger: Logger): SarifFile {
...
+};
+
+function createRunKey(run: SarifRun): SarifRunKey {
+  return {
+    name: run.tool?.driver?.name,
+    fullName: run.tool?.driver?.fullName,
+    version: run.tool?.driver?.version,
+    semanticVersion: run.tool?.driver?.semanticVersion,
+    guid: run.tool?.driver?.guid,
+    automationId: run?.automationDetails?.id,
```
aeisenberg, line 97: Nit: we already know that `run` is defined.

```suggestion
    automationId: run.automationDetails?.id,
```

aeisenberg (commented):
I think this is good and should work. Two minor comments. 

However, I think we should add some tests here. The logic is a bit complex and it would be good to make sure the messages are emitted only when we expect them to be.

src/upload-lib.ts
```diff
@@ -120,6 +120,16 @@ function areAllRunsUnique(sarifObjects: SarifFile[]): boolean {
   return true;
 }
 
+function showCombineSarifFilesDeprecationWarning(
```
angelapwen: nit: maybe `shouldShowCombineSarifFilesDeprecationWarning` so it's more clear that this returns a boolean and isn't meant to actually warn?

angelapwen (commented):
Makes sense to me! Just a small comment on top of Andrew's.

src/upload-lib.ts
koesie10: No, the message is not exactly the same. I'll extract the last part ("For more information, see ..."), but the other parts of the message are different. The first message is for third-party tools, while the second message is more specific to CodeQL-only uploads.

koesie10:
Thanks for the reviews!

> However, I think we should add some tests here. The logic is a bit complex and it would be good to make sure the messages are emitted only when we expect them to be.

I've added tests for the (renamed) `shouldShowCombineSarifFilesDeprecationWarning` function which tests that this function only returns `true` when all conditions are met. Since this is called before showing the message, this should ensure that we never show the message inadvertently.

src/upload-lib.ts
```diff
@@ -65,40 +68,129 @@ function combineSarifFiles(sarifFiles: string[], logger: Logger): SarifFile {
...
+
+  // Do not show this warning on GHES versions before 3.14.0
+  if (
+    githubVersion.type === GitHubVariant.GHES &&
+    semver.lt(githubVersion.version, "3.14.0")
+  ) {
+    return false;
+  }
+
+  // Only give a deprecation warning when not all runs are unique
```
henrymercer, line 143: "and we haven't already shown the warning", to cover the other part of the `&&`.

src/upload-lib.ts
```diff
@@ -149,6 +241,19 @@ async function combineSarifFilesUsingCLI(
...
+    if (
+      await shouldShowCombineSarifFilesDeprecationWarning(
+        sarifObjects,
+        features,
+        gitHubVersion,
+      )
+    ) {
+      logger.warning(
+        `Uploading multiple CodeQL runs with the same category is deprecated ${deprecationWarningMessage}. Please update your CodeQL CLI version or update your workflow to set a distinct category for each CodeQL run. ${deprecationMoreInformationMessage}`,
+      );
```
henrymercer, line 253: Could we clarify this message to say that rather than it being deprecated, it is deprecated for CodeQL CLI version 2.16.6 and earlier?

henrymercer (commented):
Looks good, just a comment about the warning for CodeQL.

koesie10:
> Looks good to me to merge once the changelog post is out.

I think we can actually merge this before the changelog post is out since the deprecation message is behind a feature flag. Does that make sense, or do you think it would be better to wait?

henrymercer:
Ah perfect, just make sure the changelog's out before you start rolling out the feature flag then!  Thanks!

aeisenberg (approved):
LGTM. Thanks for addressing my comments. As Henry mentions, feel free to merge after the changelog post is in.

angelapwen (approved):
Not that you needed this third approval but providing it anyway 😆

PR #2269, opened 2024-04-30
Textually replace CodeQL version in `package.json`

angelapwen:
Textually replace CodeQL version in `package.json` instead of using `npm version`, which occasionally fails due to merge conflicts that arise in the `package.json` process.

I have tested the `replace_version_package_json` function locally against the current copy of `package.json` and it works as expected, only replacing the version number in the file.

### DIFF
--- a/.github/update-release-branch.py
+++ b/.github/update-release-branch.py
@@ -1,5 +1,6 @@
 import argparse
 import datetime
+import fileinput
 import re
 from github import Github
 import json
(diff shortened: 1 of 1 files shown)


henrymercer (dismissed):
Thanks, looks good to me!

.github/update-release-branch.py
```diff
@@ -171,6 +172,19 @@ def get_current_version():
   with open('package.json', 'r') as f:
     return json.load(f)['version']
 
+# `npm version` doesn't always work because of merge conflicts, so we
+# replace the version in package.json textually.
+def replace_version_package_json(prev_version, new_version):
+  prev_line_is_codeql = False
+  for line in fileinput.input('package.json', inplace = True, encoding='utf-8'):
+    if prev_line_is_codeql and f'\"version\": \"{prev_version}\"' in line:
+      print(f'  \"version\": \"{new_version}\",')
```
henrymercer, line 181: Perhaps something like `line.replace(prev_version, new_version)` to account for whitespace potentially changing in the future?

angelapwen: Makes sense 👍 validated this worked so committing!

henrymercer (approved):

PR #2273, opened 2024-05-06
Add a compatibility table to the README

aeisenberg:
Specifies the versions of the action, CLI and GHES that are jointly compatible.

When reviewing, please make sure this is table is clear to read and easy to maintain as we release new versions of the action.


README.md
```diff
@@ -33,6 +36,19 @@ To provide the best experience to customers using older versions of GitHub Enter
...
+## Supported versions of the CodeQL CLI and GitHub Enterprise Server
+
+We typically release new minor versions of the CodeQL Action and CLI when a new minor version of GitHub Enterprise Server (GHES) is released. When a version of GHES is deprecated, the corresponding CodeQL Aciton and CLI releases are deprecated as well.
+
+| CodeQL Action Versions | CodeQL CLI Versions | GHES Versions | Deprecated? |
+|----------------|------------|-----|--|
+| `v3.25`, `v2.25` | ![Supports CodeQL CLI v2.12.6 - v2.17](https://img.shields.io/badge/Supports_CodeQL_CLI-v2.12.6_--_v2.17-blue) | `v3.9` - `v3.12` | No |
+| `v3.23`, `v2.23`, `v3.24`, `v2.24` | ![Supports CodeQL CLI v2.11.6 - v2.16](https://img.shields.io/badge/Supports_CodeQL_CLI-v2.11.6_--_v2.16-blue) | `v3.8` - `v3.11` | No |
+| `v3.22`, `v2.22` | ![Supports CodeQL CLI v2.10.5 - v2.15](https://img.shields.io/badge/Supports_CodeQL_CLI-v2.10.5_--_v2.15-blue) | `v3.7` - `v3.10` | No |
+| `v2.21`          | ![Supports CodeQL CLI v2.9.4 - v2.14](https://img.shields.io/badge/Supports_CodeQL_CLI-v2.9.4_--_v2.14-blue)   | `v3.6` - `v3.9` | No |
```
henrymercer, line 48: Couple of comments:

- I suggest we list recommended CodeQL Action versions in this table, to avoid the difficulties of having to deal with ranges.  We could recommend using `v3` for Dotcom, and then list the specific version that went into GHES for each version of GHES. 
- I think we should be clear that we don't support CodeQL v2.12.5 and earlier in any version of the Action.  I'd suggest dropping this column entirely in favour of a sentence like "All supported versions of the CodeQL Action support CodeQL Bundle v2.12.6 and later".
- We should also note that v2 will no longer receive updates once GHES 3.11 is deprecated.  The first point about only listing a recommended version would make this obvious, since v2 releases would slowly drop off the bottom of the table.

README.md
```diff
@@ -1,3 +1,6 @@
+
+![Supports CodeQL CLI v2.12.6 - v2.17](https://img.shields.io/badge/Supports_CodeQL_CLI-v2.12.6_--_v2.17-blue)
```
henrymercer, line 2: I suggest we use "and later" to avoid having to update this each minor cycle.

README.md
```diff
@@ -33,6 +36,19 @@ To provide the best experience to customers using older versions of GitHub Enter
 
 For more information, see "[Code scanning: deprecation of CodeQL Action v2](https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/)."
 
+## Supported versions of the CodeQL CLI and GitHub Enterprise Server
+
+We typically release new minor versions of the CodeQL Action and CLI when a new minor version of GitHub Enterprise Server (GHES) is released. When a version of GHES is deprecated, the corresponding CodeQL Aciton and CLI releases are deprecated as well.
```
henrymercer, line 41:
```suggestion
We typically release new minor versions of the CodeQL Action and CLI when a new minor version of GitHub Enterprise Server (GHES) is released. When a version of GHES is deprecated, the CodeQL Action and CLI releases that shipped with it are deprecated as well.
```

henrymercer (commented):
Thanks, this will help clear up something that has created confusion.

README.md
aeisenberg: Hmmm...that won't be very nice if you are looking at historical versions. I can add an entry to the contributing guide to make sure this isn't missed.

aeisenberg: Maybe better to just remove the line and have users refer to the table below.

README.md
aeisenberg: I have to be clear to be using "recommended" instead of "supported", and then I think I can make this work.

README.md
```diff
@@ -33,6 +33,22 @@ To provide the best experience to customers using older versions of GitHub Enter
...
+
+We typically release new minor versions of the CodeQL Action and CLI when a new minor version of GitHub Enterprise Server (GHES) is released. When a version of GHES is deprecated, the CodeQL Action and CLI releases that shipped with it are deprecated as well.
+
+| Recommended CodeQL Action | Recommended CodeQL CLI Version | GitHub Environment |
+|---------|----------|--------------|
+| `v3`    | `latest` | `github.com` |
+| `v3.25` | `v2.17`  | `v3.12`      |
+| `v3.24` | `v2.16`  | `v3.11`      |
+| `v3.22` | `v2.15`  | `v3.10`      |
+| `v3.21` | `v2.14`  | `v3.9`       |
```
henrymercer, line 46: - Use the specific patch versions we shipped with each GHES version
- (More controversial) Use "default" instead of "latest" to avoid confusion with `tools: latest`

```suggestion
| `v3` | default (do not pass a `tools` input) | GitHub.com |
| `3.22.12` | `2.15.5` | Enterprise Server 3.12 |
| `2.22.1` | `2.14.6` | Enterprise Server 3.11 |
| `2.20.3` | `2.13.5` | Enterprise Server 3.10 |
| `2.2.9` | `2.12.5` | Enterprise Server 3.9 |
```

aeisenberg: Sure.

aeisenberg:
Oops. I just merged a previous PR that put the changelog note in the wrong place. Let me fix that.

henrymercer (commented):
It looks like there was a merge in the changelog file that didn't go entirely right, otherwise LGTM.

aeisenberg:
I think that was already fixed in [d234275](https://github.com/github/codeql-action/pull/2273/commits/d23427596bdfd376bd20737b9ab57c9617fd061d), unless there's something I missed.

henrymercer:
Oh I see, the merge conflict was actually on `main`.

henrymercer (approved):
Thanks!

Issue #1327, opened 2022-10-26
Clarify documentation for the `tools` field in the `init` action


sidshank:
The CodeQL `init` action YML file has a field called `tools`. The field is optional, but when specified, should reference a URL to the CodeQL tools:

https://github.com/github/codeql-action/blob/40542d38bc4936c2ca7da883aeb050a6081b4b31/init/action.yml#L5

We need clear documentation for our customers on how they might specify the `tools` field to point to  a specific version of CodeQL. For example, if a new version of CodeQL breaks their workflow for some reason, they might want to downgrade to the last working CodeQL version until the issue is resolved.

Specifically, it would be nice to have documentation on how users can derive / construct the URL that needs to be specified for the `tools` field.

brendandburns:
fwiw, making this the complete URL (e.g. `https://github.com/github/codeql-action/releases/download/codeql-bundle-v2.13.4/codeql-bundle-linux64.tar.gz`) works in my setup.

PR #2281, opened 2024-05-09
Add option `tools: linked` for `tools` input of init action.

NlightNFotis:
### Description

The `init` action supports an input `tools`, which can get a range of values, including a special value `latest` that is supposed to force the use of the bundled version of CodeQL for the action.

The use of the word `latest`, however, might hint to users that this is really fetching the newest version of CodeQL, rather than the one associated with the action when it was released.

This PR is adding an option `tools: linked` that is functionally equivalent to `tools: latest`, but is supposed to guide user expectations as to the version that's actually used.     

Fixes https://github.com/github/codeql-core/issues/3256

### Review Guidance

* The basic mechanism that supports the new input option has been added in 80f8aff61df9b64baf81eb92e5feee41836749ca, in the file `src/setup-codeql.ts`.
* Tests for the new option, and the old option working in an equivalent manner have been added in 80f8aff61df9b64baf81eb92e5feee41836749ca, in the file `src/setup-codeql.test.ts`
* Tests for the logging of the version being present in the logs generated by the program have been added in 3a2da5f6fbc083d59fe140fd758936c132c2811d

### Still outstanding

- [ ] Will need to check whether readme or changelog need to be edited.
- [ ] Will need to add documentation for the new option, but I want to do this as a different PR associated with https://github.com/github/codeql-action/issues/1327
- [ ] Need to document some information transfer from Henry that unblocked me while trying to write a stub using Sinon.


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add `tools: linked` option for input of `init` action. [#2281](https://github.com/github/codeql-action/pull/2281)
```
henrymercer, line 9:
```suggestion
- The `tools: latest` input to the `init` Action has been renamed to `tools: linked`. This option specifies that the Action should use the tools shipped at the same time as the Action. The old name will continue to work for backwards compatibility, but we recommend that new workflows use the new name. [#2281](https://github.com/github/codeql-action/pull/2281)
```

henrymercer (commented):
Very nice!  Could you update the docs for `tools` in `init/action.yml` too?

src/setup-codeql.ts
```diff
@@ -292,15 +298,21 @@ export async function getCodeQLSource(
...
    */
-  const forceShippedTools = toolsInput === "latest";
+  const forceShippedTools =
+    toolsInput && CODEQL_BUNDLE_VERSION_ALIAS.includes(toolsInput);
   if (forceShippedTools) {
     logger.info(
       "Overriding the version of the CodeQL tools by the version shipped with the Action since " +
-        `"tools: latest" was requested.`,
+        `"tools: linked" or "tools: latest" was requested. The version shipped with the Action is ` +
+        `${defaultCliVersion.cliVersion}.`,
```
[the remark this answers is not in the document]
aeisenberg: If `latest` is deprecated, we should also create a warning if we see it.

aeisenberg:
Once this is done, we will also need to update the documentation in `docs-internal`. Though, I took a quick look and I can't find any reference to the `tools` input. Is this even documented there? The init action file contains a link to some documentation, but that page doesn't mention `tools` anywhere.

https://github.com/github/codeql-action/blob/main/init/action.yml#L17

For now, maybe we just need to update this input description with the new change and remove the link.

Then later, we should update the `docs-internal` with a full description of `tools`.

src/setup-codeql.ts
```diff
@@ -681,6 +693,10 @@ export async function setupCodeQLBundle(
     logger,
   );
 
+  logger.info(
+    `Using CodeQL CLI version ${source.toolsVersion} from ${source.sourceType}.`,
+  );
```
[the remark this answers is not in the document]
NlightNFotis: Ah, I see - I wasn't aware that we actually can be in a position where we _don't_ have all the information.

I was hoping that by putting the logging into the `getCodeQLBundle`, after `getCodeQLSource` has returned, that I could get away with logging at just one point, and at a time where the dust had settled and a decision on what/where to get codeql from had been made, instead of trying to chase the various return paths in `getCodeQLSource`.

I will revise the approach.

src/setup-codeql.ts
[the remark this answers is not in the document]
NlightNFotis: I wanted to ask actually about the deprecation process for the tool. I'll take we're happy to just `log.warn`?

NlightNFotis:
> Very nice! Could you update the docs for tools in init/action.yml too?

Yes, this is my plan before the end of the ticket as a whole. I was originally thinking of doing this as part of a seperate PR that's designed to specifically address the issue in https://github.com/github/codeql-action/issues/1327.

Will raise this soon - unless there's a preference to address this in this PR? I was thinking that a separate PR will be cleaner, and can afford being more focused on drafting a good general documentation for `tools`, without being subjected to the review cycle/scope of this PR. Any thoughts?

src/setup-codeql.ts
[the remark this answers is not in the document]
henrymercer: My 2¢: it's fine for customers who have already set up `tools: latest` to continue using it.  I don't think it's worth removing support for this for the forseeable future.

henrymercer:
It's not currently documented, so I'm happy with a separate PR if you prefer!

NlightNFotis:
Apologies, I had to force push a `rebase` because of some issues with the `CHANGELOG`. What has changed in 5a08657:

* I moved the logging to the downstream function `getCodeQLSource`, with more specialised log messages at each `return` point.
* Added a deprecation notice for the `tools: latest` input.
* Added tests for the logging remaining valid for more code paths of the `getCodeQLSource` when called with different `toolsInput` arguments from `setupCodeQLBundle` function.

src/setup-codeql.ts
```diff
@@ -281,7 +283,12 @@ export async function getCodeQLSource(
   variant: util.GitHubVariant,
   logger: Logger,
 ): Promise<CodeQLToolsSource> {
-  if (toolsInput && toolsInput !== "latest" && !toolsInput.startsWith("http")) {
+  if (
+    toolsInput &&
+    !CODEQL_BUNDLE_VERSION_ALIAS.includes(toolsInput) &&
+    !toolsInput.startsWith("http")
+  ) {
+    logger.info("Using CodeQL CLI from local path $path");
```
henrymercer, line 291:
```suggestion
    logger.info(`Using CodeQL CLI from local path ${toolsInput}`);
```

src/setup-codeql.ts
```diff
@@ -292,16 +299,27 @@ export async function getCodeQLSource(
...
-        `"tools: latest" was requested.`,
+      `Overriding the version of the CodeQL tools by ${defaultCliVersion.cliVersion}, the version shipped with the Action since ` +
+        `tools: ${toolsInput} was requested.`,
     );
+
+    if (toolsInput === "latest") {
+      logger.warning(
+        "The 'latest' alias for the CodeQL tools has been deprecated. Please use 'linked' instead.",
+      );
+    }
```
henrymercer, line 322: See my previous comment — I'm not sure it's worth creating work for customers to change their workflows to use "linked".

src/setup-codeql.ts
```diff
@@ -424,10 +442,14 @@ export async function getCodeQLSource(
   }
 
   if (codeqlFolder) {
+    const version = cliVersion ?? humanReadableVersion;
+    logger.info(
+      `Using CodeQL CLI version ${version} from toolcache at ${codeqlFolder}`,
+    );
```
henrymercer, line 448: `humanReadableVersion` can still contain something like `0.0.0-<bundle version number>` (it could probably have a better name as it's mainly for telemetry and debug logs).  How about only printing the `version ${cliVersion}` bit if `cliVersion` is defined?

src/setup-codeql.ts
```diff
@@ -452,12 +474,16 @@ export async function getCodeQLSource(
     url = await getCodeQLBundleDownloadURL(tagName!, apiDetails, logger);
   }
 
+  const toolsVersion = cliVersion ?? humanReadableVersion;
+  logger.info(
+    `Using CodeQL CLI version ${toolsVersion} downloaded from ${url}.`,
```
henrymercer, line 479: Likewise here.  Also we might want to rephrase this slightly so it doesn't sound like the CLI has already been downloaded.

henrymercer (commented):
Nice!  Just a couple of final comments.

src/setup-codeql.ts
NlightNFotis: Hmm, thinking out loud here:

Does it make sense to keep the deprecation warning, but soften the wording a bit, to indicate that the preferred option going forward is `linked`, but that `latest` remains compatible?

I'm thinking that if we don't have this bit of documentation (via the warning), from the user's perspective there's a chance of looking at the other bits of documentation we have, which indicate `linked` to be used, compare with their action containing `latest`, and wonder what happened to `latest` and why there's a discrepancy between our documentation and their current version, without any warning.

I believe that raising a warning here at least alerts a user to the fact that this has changed. 

Any thoughts?

henrymercer: That's a good point.  We could say something like "`tools: latest` has been renamed to `tools: linked`, but the previous name is still supported for backwards compatibility. No action is necessary.".

NlightNFotis:
Hi @henrymercer, this is now ready for another round of reviews.

henrymercer (approved):

PR #2282, opened 2024-05-09
End tracing in autobuild Action for workflows that don't specify a build mode

henrymercer:
This lets us achieve some performance and reliability improvements for workflows that run autobuild directly without specifying a build mode.

Also don't call `trace-command` for workflows that don't specify a build mode as the indirect tracer will already have been started in these cases.


angelapwen (approved):
Looks good! The additional log message about unsetting the tracing environment variables will be helpful 👍

PR #2283, opened 2024-05-10
Disable `fail-fast` for PR checks

henrymercer:
We have found `fail-fast` has been getting in the way more than it's been helping recently.  For example:

- Sometimes only a particular set of versions or OSes are broken, and we want to let the other checks complete so we can know the full set.
- We see quite a large degree of parallelism when running the checks, so most of the jobs have nearly finished anyway when they are cancelled by `fail-fast`.
- `fail-fast` deals better with transient failures.


aeisenberg (approved):
⏩

PR #2284, opened 2024-05-10
Add documentation for the `tools` option in the various workflow files

NlightNFotis:
### Description

Documents the `tools:` input and its options for the various workflow files.

Companion PR to https://github.com/github/codeql-action/pull/2281

Fixes https://github.com/github/codeql-action/issues/1327

### DIFF
--- a/.github/actions/check-codescanning-config/action.yml
+++ b/.github/actions/check-codescanning-config/action.yml
@@ -29,7 +29,15 @@ inputs:
   tools:
     required: true
     description: |
-      The url of codeql to use.
+      The version of the CodeQL tools to use. This can be any of the following:
+
+      - A local path to a tarball containing the CodeQL tools, or
+      - A URL to a GitHub release assets containing the CodeQL tools, or
+      - A special value `linked` which is forcing the use of the CodeQL tools
+        that the action has been bundled with.
+
+      If not specified, the Action will check in several places until it finds
+      the CodeQL tools.
 
 runs:
   using: composite
--- a/.github/actions/check-codescanning-config/action.yml
+++ b/.github/actions/check-codescanning-config/action.yml
@@ -29,7 +29,8 @@ inputs:
   tools:
     required: true
     description: |
-      The version of the CodeQL tools to use. This can be any of the following:
+      The version of CodeQL passed to the `tools` input of the init action.
+      This can be any of the following:
 
       - A local path to a tarball containing the CodeQL tools, or
       - A URL to a GitHub release assets containing the CodeQL tools, or
--- a/.github/actions/query-filter-test/action.yml
+++ b/.github/actions/query-filter-test/action.yml
@@ -23,7 +23,15 @@ inputs:
   tools:
     required: true
     description: |
-      The url of codeql to use.
+      The version of the CodeQL tools to use. This can be any of the following:
+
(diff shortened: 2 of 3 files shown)


init/action.yml
```diff
@@ -3,9 +3,16 @@ description: 'Set up CodeQL'
...
-    description: URL of CodeQL tools
+    description: >-
+      The version of the CodeQL tools to use. This can be any of the following:
+      
+      - A local path to a tarball containing the CodeQL tools, or
+      - A URL to a GitHub release assets containing the CodeQL tools, or
+      - A special value `linked` which is forcing the use of the CodeQL tools
+        that the action has been bundled with.
+
+      If not specified, the Action will check in several places until it finds the CodeQL tools.
```
henrymercer, line 14:
```suggestion
    description: >-
      By default, the Action will use the recommended version of the CodeQL
      Bundle to analyze your project. You can override this choice using this
      input. One of:

      - A local path to a CodeQL Bundle tarball, or
      - The URL of a CodeQL Bundle tarball GitHub release asset, or
      - A special value `linked` which uses the version of the CodeQL tools
        that the Action has been bundled with.
```

henrymercer (commented):
Thanks!  I made a suggestion to make it clearer the recommended case of not passing this input to let the CodeQL Action choose which tools to use.

.github/actions/check-codescanning-config/action.yml
```diff
@@ -29,7 +29,15 @@ inputs:
   tools:
     required: true
     description: |
-      The url of codeql to use.
+      The version of the CodeQL tools to use. This can be any of the following:
```
aeisenberg, line 32: This is an action used internally for testing only. Maybe better to change this to:

```suggestion
      The version of CodeQL passed to the `tools` input of the init action.
```

.github/actions/query-filter-test/action.yml
```diff
@@ -23,7 +23,15 @@ inputs:
   tools:
     required: true
     description: |
-      The url of codeql to use.
+      The version of the CodeQL tools to use. This can be any of the following:
```
aeisenberg, line 26: Same comment as above.

init/action.yml
```diff
@@ -3,9 +3,16 @@ description: 'Set up CodeQL'
...
+      By default, the Action will use the recommended version of the CodeQL
+      Bundle to analyze your project. You can override this choice using this
+      input. One of:
+
+      - A local path to a CodeQL Bundle tarball, or
+      - The URL of a CodeQL Bundle tarball GitHub release asset, or
+      - A special value `linked` which uses the version of the CodeQL tools
+        that the Action has been bundled with.
     required: false
-    # If not specified the Action will check in several places until it finds the CodeQL tools.
```
aeisenberg, line 8: I think keeping this line and removing the `#` is useful. At some point, but not here, we should document exactly which places are checked.

henrymercer: The "check in several places" comment feels more about implementation than specification, and in fact this applies even if you do specify an input.  For instance, if you specify a URL we will try to extract the CodeQL version from the URL and use the toolcache if we can find a matching version.  Overall, I'd prefer to remove this in favour of saying we'll use the recommended version if you don't specify an input, but I don't feel that strongly about it.

henrymercer (approved):

PR #2286, opened 2024-05-13
Enable database uploads for GHEC-DR

koesie10:
This will enable CodeQL database uploads on GHEC-DR, in addition to the existing github.com uploads. This does not have any impact on github.com or GHES.


src/database-upload.ts
```diff
@@ -49,8 +55,9 @@ export async function uploadDatabases(
       );
       try {
         await client.request(
-          `POST https://uploads.github.com/repos/:owner/:repo/code-scanning/codeql/databases/:language?name=:name&commit_oid=:commit_oid`,
+          `POST /repos/:owner/:repo/code-scanning/codeql/databases/:language?name=:name&commit_oid=:commit_oid`,
```
robertbrignull, line 58: I tried testing this and hit issues because the `uploadsUrl` ends in a slash and it leads to a double slash at the start of the path.

The API then redirects `from `//repos/...` to `/repos/...` and that redirect itself causes issues because we're using `createReadStream` which can only be read from once, so when after following the redirect it can't redo the re-read the data.

Removing the `/` here works, though I'm not sure if a better solution would be to remove the trailing slash from `uploadsUrl`.

koesie10: Thanks! I've removed the trailing slash from the `baseUrl` passed to Octokit since this matches with what [I've also done in the PR for the extension](https://github.com/github/vscode-codeql/pull/3607/files#diff-d52259eaaa34199e497cd6350d2236fa40610cf4bb909df0604c847a27e9c55d).

robertbrignull (approved):
Good to merge now the flags are enabled. And I did some manual testing yesterday.

PR #2306, opened 2024-05-23
Automatically clean up old TRAP caches

henrymercer:
This PR implements a new feature, behind a flag, that keeps only the newest TRAP cache for each language.

We call the Actions cache APIs directly as these have not yet been added to the `@actions/cache` npm package.

I have added a unit test exercising most of the functionality, and have also tested this manually using a test repo.  I think an integration test is probably not worth it, particularly on this repo where the databases are small enough that TRAP caching skips the upload anyway.


src/trap-caching.test.ts
```diff
@@ -189,3 +194,75 @@ test("download cache looks for the right key and creates dir", async (t) => {
...
+        key: "codeql-trap-1-2.0.0-swift-newest",
+        created_at: "2024-02-22T14:25:00Z",
+        size_in_bytes: 300 * 1024 * 1024,
+      },
+      {
+        id: 7,
+        key: "codeql-trap-1-2.0.0-swift-older",
+        created_at: "2024-02-21T14:25:00Z",
+        size_in_bytes: 300 * 1024 * 1024,
+      },
```
aeisenberg, line 246: This one wasn't deleted. Is it because the timestamp is identical to id 6? Wait...the code should only keep one cache per language. So, I'm not sure what's going on here.

henrymercer: Ah, I see now that the reasoning for this is a bit opaque without looking at the rest of the file.  We're expecting to only clean up TRAP caches for JavaScript and Ruby as these are the only ones that the `getTestConfigWithTempDir` config specifies as languages to analyze in this job.  I'll add a comment to make this clearer.

aeisenberg: Thanks. That makes a lot of sense.

PR #2328, opened 2024-06-11
Fix incompatibility between direct tracing in the `autobuild` Action and specifying a custom working directory

henrymercer:
This PR fixes an issue with the `AutobuildDirectTracing` feature where it would ignore the `working-directory` input to the autobuild Action.  The fix is to pass Node's working directory to the CLI, since we already change directory in Node based on the input.  We also bump the version of the feature flag so that we can avoid re-enabling the flag for old Action versions.

I've added a PR check to validate that the fix works.


pr-checks/checks/autobuild-direct-tracing-with-working-dir.yml
```diff
@@ -0,0 +1,34 @@
...
+    run: |
+      if [[ ! -z "${CODEQL_RUNNER}" ]]; then
+        echo "Expected indirect tracing to be disabled, but the" \
+          "CODEQL_RUNNER environment variable is set."
+        exit 1
+      fi
+  - uses: ./../action/autobuild
+    with:
+      working-directory: autobuild-dir
+  - uses: ./../action/analyze
```
aeisenberg, line 34: I think we want to avoid uploading the database here.

```suggestion
  - uses: ./../action/analyze
     with:
       upload-database: false
```

henrymercer: Hmm, I think this should be the default when we're in test mode.  Let me fix that.

src/database-upload.ts
```diff
@@ -20,6 +20,11 @@ export async function uploadDatabases(
     return;
   }
 
+  if (util.isInTestMode()) {
+    logger.debug("In test mode. Skipping database upload.");
+    return;
+  }
```
aeisenberg, line 26: Nice.

aeisenberg (approved):

aeisenberg (approved):

PR #2329, opened 2024-06-11
Add rollback mechanism for C# buildless

henrymercer:
This PR adds a rollback mechanism for C# buildless similar to the one we added for Java.  When this is enabled, workflows configured to use `build-mode: none` for C# will use `build-mode: autobuild` instead.


src/feature-flags.ts
```diff
@@ -109,6 +110,11 @@ export const featureConfig: Record<
     legacyApi: true,
     minimumVersion: "2.16.1",
   },
+  [Feature.DisableCsharpBuildless]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_DISABLE_CSHARP_BUILDLESS",
+    minimumVersion: undefined,
```
nickfyson, line 116: OOI, why do some need `legacyApi: true` and some not?

nickfyson (approved):
LGTM

src/feature-flags.ts
henrymercer: This is to do with our work to be able to create feature flags without monolith changes.  Features with `legacyApi: true` were added using a monolith PR.  Features with `legacyApi: false` use the new API where the CodeQL Action requests them by name so we don't need to make a monolith PR.

PR #2330, opened 2024-06-11
Use `--force-overwrite` for creating databases

aeisenberg:
Only if the feature exists in the CLI.


src/tools-features.ts
```diff
@@ -8,6 +8,7 @@ export enum ToolsFeature {
   SetsCodeqlRunnerEnvVar = "setsCodeqlRunnerEnvVar",
   TraceCommandUseBuildMode = "traceCommandUseBuildMode",
   SarifMergeRunsFromEqualCategory = "sarifMergeRunsFromEqualCategory",
+  ForceOverwite = "forceOverwite",
```
adityasharad, line 11: typo in both?

aeisenberg: Yikes. Good catch. There was the same typo in our internal PR. I fixed that too.

adityasharad (approved):

PR #2331, opened 2024-06-12
Fix bash syntax

igfoo:
This should stop

    .github/workflows/script/update-node-modules.sh: line 1: [: missing `]'


henrymercer (approved):
Thanks!  My understanding is that we could also use double-brackets, but that's less POSIX-compliant.

PR #2349, opened 2024-06-20
Avoid failing when debug artifacts can't be uploaded

aeisenberg:
Failing to upload debug artifacts should not fail the action since it is possible that the action has already succeeded by the time this failure occurs.


angelapwen (approved):

angelapwen:
Maybe we should add a note in the changelog about this change? 💭

NlightNFotis (dismissed):
Yep, a warning for this is good behaviour IMO.

aeisenberg:
> Maybe we should add a note in the changelog about this change? 💭

Yes. We should. I meant too....but forgot.

angelapwen (approved):

PR #2350, opened 2024-06-24
PR check generator: add `excludeOsAndVersionCombination`

angelapwen:
We often need to exclude specific combinations of CLI versions + operating systems. Previously there was no easy way to do this: we would run all CLI versions specified against all operating systems specified. This change adds the `excludeOsAndVersionCombination` parameter to do just that!

To test, I re-enabled Swift on Linux for CLI v >= 2.17.4, which remained as a to-do item after we disabled those checks in https://github.com/github/codeql-action/pull/2299. Note that `default`, `linked`, and `nightly-latest` are all now current enough to re-enable those. The new generator script allows us to exclude all older CLI versions on Linux. 

I'll update the required PR checks once this PR is approved ✅ 

### DIFF
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -27,6 +27,12 @@ defaultTestVersions = [
     "nightly-latest"
 ]
 
+def is_version_and_os_excluded(version, os, exclude_params):
+    for exclude_param in exclude_params:
+        if exclude_param[0] == os and exclude_param[1] == version:
+            return True
+    return False
+
 # When updating the ruamel.yaml version here, update the PR check in
 # `.github/workflows/pr-checks.yml` too.
 header = """# Warning: This file is generated automatically, and should not be modified.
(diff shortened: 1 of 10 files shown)


pr-checks/sync.py
```diff
@@ -56,27 +62,32 @@ def writeHeader(checkStream):
 for file in (this_dir / 'checks').glob('*.yml'):
     with open(file, 'r') as checkStream:
         checkSpecification = yaml.load(checkStream)
-
     matrix = []
+    excludedVersionsAndOses = checkSpecification.get('excludeOsAndVersionCombination', [])
```
henrymercer, line 66: We could rename to `excludedOsesAndVersions` for clarity since it's OSes first and versions second

pr-checks/sync.py
```diff
@@ -56,27 +62,32 @@ def writeHeader(checkStream):
...
-                })
+        runnerImages = ["ubuntu-latest", "macos-latest", "windows-latest"]            
+        operatingSystems = checkSpecification.get('operatingSystems', ["ubuntu", "macos", "windows"])
+
+        for operatingSystem in operatingSystems:
+            runnerImagesForOs = [image for image in runnerImages if image.startswith(operatingSystem)]
+
+            for runnerImage in runnerImagesForOs:
+                # Skip appending this combination to the matrix if it is explicitly excluded.
+                if is_version_and_os_excluded(version, operatingSystem, excludedVersionsAndOses):
```
henrymercer, line 76: Similarly here consider using the same order of OS and version as in the YAML.

henrymercer (approved):
Just a small suggestion to reduce the likelihood of an argument ordering mismatch

pr-checks/sync.py
angelapwen: Good point, I actually misordered it myself when testing initially 😆

pr-checks/sync.py
```diff
@@ -27,6 +27,12 @@
     "nightly-latest"
 ]
 
+def is_os_and_version_excluded(version, os, exclude_params):
```
henrymercer, line 30: Final suggestion: match the order of the arguments too, e.g. `is_os_and_version_excluded(os, version, exclude_params):`

angelapwen: Good call ✨

henrymercer (approved):

angelapwen:
Updated the branch protection rules accordingly for `main`, `releases/v2`, and `releases/v3`.

Issue #2354, opened 2024-06-27
[bug] CodeQL error says goto `https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/automatic-build-failed.`


SwuduSusuwu:
[WeblQL gives a clickable link which ends with `.`]
If you remove the trailing `.` the URL ( https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/automatic-build-failed ) launches.
`https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/automatic-build-failed.` is from https://github.com/SwuduSusuwu/SubStack/actions/runs/9693135643/job/26747940322
```
  [2024-06-27 08:20:32] [build-stderr] cpp/autobuilder: using built-in deptrace DB
  [2024-06-27 08:20:33] [build-stderr] cpp/autobuilder: launched deptrace server
  [2024-06-27 08:20:34] [build-stderr] cpp/autobuilder: Build will be attempted from '.' even though no build system was found there.
  [2024-06-27 08:20:34] [build-stdout] ~/work/SubStack/SubStack ~/work/SubStack/SubStack
  [2024-06-27 08:20:34] [build-stdout] ~/work/SubStack/SubStack
  [2024-06-27 08:20:34] [build-stderr] cpp/autobuilder: No supported build system detected.
  [2024-06-27 08:20:34] [build-stderr] cpp/autobuilder: deptrace was enabled.
  [2024-06-27 08:20:34] [build-stderr] cpp/autobuilder: autobuild summary.
  Error: 6-27 08:20:34] [ERROR] Spawned process exited abnormally (code 1; tried to run: [/opt/hostedtoolcache/CodeQL/2.17.5/x64/codeql/tools/linux64/preload_tracer, /opt/hostedtoolcache/CodeQL/2.17.5/x64/codeql/cpp/tools/autobuild.sh])
  A fatal error occurred: Exit status 1 from command: [/opt/hostedtoolcache/CodeQL/2.17.5/x64/codeql/cpp/tools/autobuild.sh]
[...]

PR #2355, opened 2024-06-27
Make all instances of doc links in logs clickable

angelapwen:
Fixes https://github.com/github/codeql-action/issues/2354. 

This PR updates all log messages with links to GitHub docs so that they are not succeeded by punctuation marks, in order to make sure the links are clickable from our logs. 


src/codeql.test.ts
```diff
@@ -912,8 +912,8 @@ test("runTool summarizes autobuilder errors", async (t) => {
       instanceOf: CommandInvocationError,
       message:
         "We were unable to automatically build your code. Please provide manual build steps. " +
-        "For more information, see " +
-        "https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/automatic-build-failed. " +
+        "See https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/automatic-build-failed " +
```
NlightNFotis, line 915: I'm wondering whether it makes sense to export these into their own variables, and even better perhaps in their own file, something like `src/docs-urls.ts` or something similar.

Doing so would have the benefit of them being more centrally organised (so we can track changed URLs to one place and do a one-move global replace if anything changes).

src/autobuild.ts
```diff
@@ -139,14 +139,14 @@ export async function setupCppAutobuild(codeql: CodeQL, logger: Logger) {
...
-            ? ` To override this, set the ${envVar} environment variable to 'true' in your workflow (see ${envDoc}).`
+            ? ` To override this, set the ${envVar} environment variable to 'true' in your workflow (see ${envDoc} for more information).`
             : ""
         }`,
       );
       core.exportVariable(envVar, "false");
     } else {
       logger.info(
-        `Enabling ${featureName}. This can be disabled by setting the ${envVar} environment variable to 'false' (see ${envDoc}).`,
+        `Enabling ${featureName}. This can be disabled by setting the ${envVar} environment variable to 'false' (see ${envDoc} for more information).`,
```
NlightNFotis, line 149: In some of these strings we have the URL reference inside of a parenthesis, whereas in others we end the sentence with a period and start a new one.

I'm curious if there are any benefits to being more consistent on that front.

NlightNFotis (approved):
LGTM broadly, just a couple of nits - don't need to be implemented within the PR.

henrymercer (dismissed):
Nice!  Non-blocking, optional: This could be something we create a CodeQL query for to help make sure we follow this pattern going forward.

src/codeql.test.ts
angelapwen: I did this in an enum, similar to the way we have `EnvVar` for shared environment variables 😄 it was a really good idea — from there I found other instances where we had docs links that ended with periods!

src/autobuild.ts
angelapwen: I definitely don't see any downsides to standardizing our docs links — I removed all the parenthetical URL references!

angelapwen:
> Nice! Non-blocking, optional: This could be something we create a CodeQL query for to help make sure we follow this pattern going forward.

Thanks!! I've waited too long to get back to this PR 😆 so I won't write the CodeQL query this time. I think the new `DocUrl` enum out should make it a lot clearer moving forward as well.

NlightNFotis (approved):

PR #2356, opened 2024-06-27
Better handling for required workflows

marcogario:
Return the workflow path from the run rather than trying to fetch the workflow information when the workflow information is missing or refers to a required workflow.

This fixes the error:
> Error: As of January 2024, this feature is fully deprecated and creating required workflows is only available with repository rulesets. All existing workflows have been automatically migrated to rulesets. Learn more about rulesets: https://github.blog/2023-10-11-enforcing-code-reliability-by-requiring-workflows-with-github-repository-rules/

I've tested this by setting up a required workflow using the action version from this version and confirmed it works.

Note: We are reporting the path as if it exists in the current repo, but that is not the case. Logic that relies on making a distinction here might be affected.


src/api-client.ts
```diff
@@ -139,6 +139,14 @@ export async function getWorkflowRelativePath(): Promise<string> {
   );
   const workflowUrl = runsResponse.data.workflow_url;
 
+  const requiredWorkflowRegex = /\/repos\/[^/]+\/[^/]+\/actions\/required_workflows\/[^/]+/;
+  if (!workflowUrl || workflowUrl === "" || requiredWorkflowRegex.test(workflowUrl)) {
```
aeisenberg, line 143: The `workflowUrl === ""` is redundant.
```suggestion
  if (!workflowUrl || requiredWorkflowRegex.test(workflowUrl)) {
```

src/api-client.ts
```diff
@@ -139,6 +139,14 @@ export async function getWorkflowRelativePath(): Promise<string> {
...
+    // For required workflows, the workflowUrl is invalid so we cannot fetch more informations
+    // about the workflow.
+    // However, the path is available in the original response.
+    return runsResponse.data.path as string;
```
aeisenberg, line 147: `runsResponse.data.path` Why isn't this value always correct? Is it because there are edge cases in normal workflow runs where the workflow may have been moved?

marcogario: I don't know. I think it should be correct, but I can try to dig into the server-side logic to try to be sure. Right now this was meant to be backwards compatible.

henrymercer (dismissed):
LGTM, this is clearly correct.  As Andrew mentions, it would be nice to see if we can use `runsResponse.data.path` directly, but we should test this thoroughly to avoid stale tips.

henrymercer:
@marcogario Unfortunately the rebuild workflow doesn't know how to fix linter errors — would you mind addressing these manually?

henrymercer (approved):

PR #2361, opened 2024-07-01
Refactoring: Simplify upload SARIF flow and reuse loaded flags

henrymercer:
Minor refactoring to simplify the flow for uploading SARIF files, and to reuse the loaded feature flags if we've already loaded them from disk or from the API.


src/upload-lib.ts
```diff
@@ -618,16 +585,16 @@ async function uploadFiles(
...
+  const checkoutURI = fileUrl(checkoutPath);
 
   const payload = buildPayload(
-    commitOid,
-    ref,
+    await actionsUtil.getCommitOid(checkoutPath),
+    await actionsUtil.getRef(),
     analysisKey,
-    analysisName,
+    category,
```
aeisenberg, line 594: Previously, this was `analysisName`, but that doesn't look right since the `util.getRequiredEnvParam("GITHUB_WORKFLOW")` was being passed in. Now you're using `category`. This seems to be correct now, but will that mess up anything with our teleemtry since we're changing what we're stuffing into the `analysis_name` field of the payload?

aeisenberg (commented):
Looks like a general improvement, but just want to make sure about the payload change.

src/upload-lib.ts
henrymercer: You're right, let's keep it the same.

aeisenberg (approved):

angelapwen:
💭 The `submit SARIF after failure` workflow on `main` seems to be [failing now](https://github.com/github/codeql-action/actions/runs/9842773192) after this was merged 😢 (though I'm not sure why it didn't fail in this PR's checks).

aeisenberg:
The latest run on main is passing, but I see the [run](https://github.com/github/codeql-action/actions/runs/9842726577) associated with https://github.com/github/codeql-action/pull/2364 is failing. I'm rerunning since it might be transient.

aeisenberg:
...and it passed! So, guessing the error came from the server side.

PR #2368, opened 2024-07-10
Add codeql-version to outputs

mschwager:
Fixes #2367.

### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [X] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

Regarding the CHANGELOG check above, my understanding was that this would be [automatically populated on release](https://github.com/github/codeql-action/blob/main/CONTRIBUTING.md#releasing-write-access-required)?

PR #2372, opened 2024-07-12
Fix changelog entry section

henrymercer:
I just noticed that the changelog entry for https://github.com/github/codeql-action/pull/2368 is in the wrong section — this PR fixes that.  It's a common mistake with automatic merge resolution in that file — we should consider dropping unreleased changelog entries in a folder instead like we do for the CLI.


aeisenberg (approved):
Oops. Thanks. I should have caught that.

PR #2394, opened 2024-07-26
Extend init complete status report

rvermeulen:
To gain insights into how users are using CodeQL in combination with packs, query filters, and registries we extend the init complete status report. This report now includes the packs, query-filters, and registries that are used in the CodeQL setup.


aeisenberg (commented):
Looks good. This PR is missing the updated map files and minified JS files.

aeisenberg:
The rebuild action is failing because this is in a fork. @rvermeulen, if you remove and then add the `Rebuild` label, a job will run to automatically transpile the javascript (it fails for me). Or you can do it locally.

src/init-action.ts
```diff
@@ -174,18 +181,46 @@ async function sendCompletedStatusReport(
...
+        packs[config.languages[0]] = config.originalUserInput.packs;
+      } else {
+        packs = config.originalUserInput.packs;
+      }
+    }
+
+    if (config.augmentationProperties.packsInput) {
+      packs[config.languages[0]] ??= [];
+      packs[config.languages[0]].push(
+        ...config.augmentationProperties.packsInput,
```
aeisenberg, line 201: This is modifying the underlying config object since `packs` is a reference, not a copy. On lines 192 and 194, you can wrap the RHS in  a call to `cloneObject`. https://github.com/github/codeql-action/blob/main/src/codeql.ts#L1347

aeisenberg (commented):
Apologies...something I didn't catch the first time I reviewed.

src/init-action.ts
rvermeulen: Good catch. I will reuse `cloneObject`.

aeisenberg (approved):

src/config-utils.ts
```diff
@@ -881,6 +881,15 @@ function parseRegistries(
   }
 }
 
+export function parseRegistriesWithoutCredentials(
+  registriesInput?: string,
+): RegistryConfigNoCredentials[] | undefined {
+  return parseRegistries(registriesInput)?.map((r) => {
+    const { url, packages } = r;
+    return { url, packages };
+  });
```
angelapwen, line 890: I'm not sure I understand the logic here 🤔 do you mind explaining it here or in a code comment?

src/init-action.ts
```diff
@@ -85,12 +86,19 @@ interface InitWithConfigStatusReport extends InitStatusReport {
...
   trap_cache_languages: string;
   /** Size of TRAP caches that we downloaded, in bytes. */
   trap_cache_download_size_bytes: number;
   /** Time taken to download TRAP caches, in milliseconds. */
   trap_cache_download_duration_ms: number;
+  /** Stringified JSON array of registry configuration objects, from the 'registries' config field
+  or workflow input. **/
+  registries: string;
+  /** Stringified JSON object representing a query-filters, from the 'query-filters' config field. **/
+  query_filters: string;
```
angelapwen, line 101: 💭 Is it possible that any of these field values are undefined or empty before we pass them to `JSON.stringify`? If so, what's the output & is that we want?

angelapwen (approved):
Just a few questions from my end 😄

src/config-utils.ts
rvermeulen: Yes, we use destructuring assignment to unpack the properties of a registry we need to create a new `RegistryConfigNoCredentials` object.

Initially I had:

```
const { token: _, ...registryWithoutCredentials } = r;
```

but this triggered an ESLint unused variable error for `_`.
In hindsight, selecting the properties we actually care about instead of removing current sensitive properties is a better solution in case we add more sensitive properties in the future or rename the current sensitive property.

src/init-action.ts
rvermeulen: Yes, they might be undefined, but the value of those properties cannot be.
That being said, let me explicitly use their initialization values according to their type, that is an empty array for `registries` and `query_filters`.

aeisenberg (approved):

src/config-utils.ts
angelapwen: Ah, I see! That makes sense 👍 I was missing the context that we were selecting for `url` and `packages` (it looked to me like we were just returning everything in `registriesInput`)

angelapwen (approved):
Thanks for answering my questions ❣️

angelapwen (approved):

PR #2399, opened 2024-07-31
Add warning and update PR checks for Swift on Linux

angelapwen:
We are deprecating Swift support on Linux for code scanning (changenote in #2403).

This PR logs a warning for users in the `init` Action if it's determined that Swift on Linux is being run.

This change also updates the manual build script `build.sh` which we also use in internal repositories for testing: the `swift` command will only be run on MacOS rather than on both MacOS and Linux.

As a result, we can also update our PR checks that only test on Linux, to remove the `setup-swift` Action step as it is no longer needed; and other checks that look for the existence of Swift databases on Linux.

I've removed `Swift analysis using a custom build command` check on Linux images from required checks for `main`, `releases/v3`, and `releases/v2` already as well. 


pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -1,28 +1,30 @@
 name: "Multi-language repository"
-description: "An end-to-end integration test of a multi-language repository using automatic language detection"
+description: "An end-to-end integration test of a multi-language repository using automatic language detection for MacOS"
```
henrymercer, line 2:
```suggestion
description: "An end-to-end integration test of a multi-language repository using automatic language detection"
```

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -1,28 +1,30 @@
...
+    if: runner.os == 'Linux'
+    with:
+      db-location: "${{ runner.temp }}/customDbLocation"
+      # We manually exclude Swift from the languages list here, as it is not supported on Ubuntu
+      languages: cpp,csharp,go,java,javascript,python,ruby 
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
+
+  - uses: ./../action/init
+    id: init-macos
+    if: runner.os == 'macOS'
```
henrymercer, line 19: Alternatively we could make the languages input conditional, e.g. something like `languages: ${{ runner.os == 'Linus' && 'cpp,csharp,go,java,javascript,python,ruby' || '' }}`

henrymercer (approved):
Looks good, a couple of suggestions

pr-checks/checks/multi-language-autodetect.yml
angelapwen: 🤔 I added the bit about MacOS because the test actually no longer tests autodetect for MacOS (since I had to manually exclude Swift from the `languages` input).

pr-checks/checks/multi-language-autodetect.yml
angelapwen: Ah that's a good point! I had thought that an empty `languages` input might cause us to run with no languages but it looks like we guard against that and turn autodetect on: https://github.com/github/codeql-action/blob/5cdd182212d3c2e1b78ee867e1dc499ddc60c0eb/src/config-utils.ts#L384-L392 🧠  I'll make the change 👍

src/init-action.ts
```diff
@@ -333,6 +333,15 @@ async function run() {
       );
     }
 
+    if (
+      config.languages.includes(Language.swift) &&
+      process.platform === "linux"
+    ) {
+      logger.warning(
+        `Swift analysis on Ubuntu runner images is no longer supported. Please migrate to a  acOS runner if this affects you.`,
```
aeisenberg, line 341:
```suggestion
        `Swift analysis on Ubuntu runner images is no longer supported. Please migrate to a macOS runner if this affects you.`,
```

aeisenberg (dismissed):
Thanks. Looks good after the typo is fixed.

src/init-action.ts
angelapwen: Whoops. Thanks!

angelapwen:
Huh.. not sure why the `Unit Test` workflow isn't being picked up. I'll push an empty commit to try to force it to run.

aeisenberg (approved):

PR #2404, opened 2024-08-02
Registries Proxy: Support feeding a base64 encoded configuration

marcogario:
Extend the start-proxy action to support passing the configuration as a base64 encoded variable. The reason to do so, is that this information is being passed by Actions as a secret. Actions recommends against passing structured data as a secret because it becomes harder to scrub it. Therefore, we send the information as a simple string.

We introduce the new input `registries_credentials` in parallel with the previous input to make it easier to test this while we are still building the remaining parts of the system. We will deprecate and remove the old property (`registry_secrets`) in the future. Note that this entire action is meant to be for internal use only.

Finally, in this PR I also refactored a bit the logic to more clearly distinguish the key steps, perform some validation of the input, and thread the logger for increase observability.


src/start-proxy-action.ts
```diff
@@ -23,12 +24,17 @@ export type CertificateAuthority = {
 
 export type Credential = {
   type: string;
-  host: string;
+  host?: string;
+  url?: string;
```
aeisenberg, line 28: Is it an error to have both `host` and `url` defined?

src/start-proxy-action.ts
```diff
@@ -89,46 +95,41 @@ function generateCertificateAuthority(): CertificateAuthority {
...
 
 async function runWrapper() {
+  const logger = getActionsLogger();
+
+  // Setup logging for the proxy
   const tempDir = actionsUtil.getTemporaryDirectory();
-  const logFilePath = path.resolve(tempDir, "proxy.log");
-  const input = actionsUtil.getOptionalInput("registry_secrets") || "[]";
-  const credentials = JSON.parse(input) as Credential[];
+  const proxyLogFilePath = path.resolve(tempDir, "proxy.log");
```
aeisenberg, line 102: In debug mode, this file should be uploaded as an artifact.

src/start-proxy-action.ts
```diff
@@ -158,15 +159,57 @@ async function runWrapper() {
...
   }
 }
 
+// getCredentials returns registry credentials from action inputs.
+// It prefers `registries_credentials` over `registry_secrets`.
+// If neither is set, it returns an empty array.
+function getCredentials(): Credential[] {
+  const encodedCredentials = actionsUtil.getOptionalInput("registries_credentials");
+  if (encodedCredentials !== undefined) {
+    const credentialsStr = Buffer.from(encodedCredentials, "base64").toString();
```
aeisenberg, line 179: This seems like some extra hoops to go through for the workflow file to send base64 encoded credentials. Does it make more sense to accept non-encoded credentials and only base64 encode them in the action?

start-proxy/action.yml
```diff
@@ -6,6 +6,9 @@ inputs:
     description: The URLs and credentials of package registries
     required: false
     default: "[]"
+  registries_credentials:
+    description: Base64 encoded JSON configuration for the URLs and credentials of the package registries
```
aeisenberg, line 10: Is there a link or some more information you can put here? Otherwise it's opaque to users how this input is structured. Since this is still an experimental action, not high priority right now.

src/start-proxy-action.ts
marcogario: No. If both are defined, the [current logic](https://github.com/github/dependabot-update-job-proxy/blob/360169018ca47384b0fe80420c7271a3af927d0b/handlers/maven_repository.go#L53) tries to match on the url first, and then the host.
You do need at least one, so I should probably add a check for that.

src/start-proxy-action.ts
marcogario: Indeed! Do you have a pointer on how to do this? I was refactoring this code, so I assumed this was working but testing it out it seems it does not.

src/start-proxy-action.ts
marcogario: Indeed, there are some extra steps, but I think they are justified.

The proxy takes a JSON object as previously defined in `registry_secrets`. The problem is that [Actions recommends ](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#naming-your-secrets)not using structured data within secrets:

> To help ensure that GitHub redacts your secret in logs, avoid using structured data as the values of secrets. 

Therefore, we encode it as a single string to make scrubbing possible.

The underlying reason we need a struct is that (due to the integration with default setup) we need to provide a potentially arbitrary number of credentials, and we cannot easily know which ones/how many beforehand.

start-proxy/action.yml
marcogario: I don't think users will need to use this input. Is there a way to denote it as "internal"?

aeisenberg: Maybe just doing this?

```suggestion
    description: [Internal] Base64 encoded JSON configuration for the URLs and credentials of the package registries
```

Am I correct in that it's unlikely users will be directly accessing this action?

marcogario: Yes, the entire action is meant to be internal. I'll add that to the description.

marcogario:
@aeisenberg this is ready for another review. I am also assigning @aibaars so he can shepherd this to completion in case there are issues to address.

aeisenberg (approved):
Awesome. Thanks. Looks great.

aibaars (approved):

PR #2408, opened 2024-08-05
Remove support for CodeQL 2.13.4 and earlier

henrymercer:
Follow up to https://github.com/github/codeql-action/pull/2375 where we announced that support for CodeQL v2.13.4 and earlier would be removed after these versions were deprecated alongside GHES 3.9.

We'll need to update the set of required PR checks before merging this.


angelapwen (approved):
Very nice 😌 thanks for all the 🧹 as always!!

PR #2414, opened 2024-08-07
Set build tracing environment variables after init step

angelapwen:


angelapwen:
Closing — we'll need to do this a different way as we do want the environment variables to be set right after the `init` step for a potential manual build step between `init` and `analyze`.

PR #2428, opened 2024-08-14
Unset `DYLD_INSERT_BINARIES` when unneeded

angelapwen:
Previously, the tracer environment variables were set for the
current process, and for future steps, in the init action. In
certain scenarios (such as on MacOS ARM runners with System
Integrity Protection disabled), these environment variables are
not unset by the system. In particular, the `DYLD_INSERT_BINARIES`
variable interferes with later system calls.

This change unsets the `DYLD_INSERT_BINARIES` variable for the
current process in init. It also unsets the variables either at the
beginning of autobuild, or analyze, if autobuild has not run.


src/init-action.ts
```diff
@@ -611,9 +612,18 @@ async function run() {
       logger,
     );
     if (tracerConfig !== undefined) {
+      const tracerEnvVarNames: string[] = [];
```
adityasharad, line 615: Do we need this variable? Is it not just `Object.keys(tracerConfig.env)`?

src/tracer-config.ts
```diff
@@ -128,3 +128,14 @@ export async function getCombinedTracerConfig(
 
   return mainTracerConfig;
 }
+
+export function unsetTracerEnvVarsForCurrentProcess(): void {
+  const tracerEnvVarNames: object = JSON.parse(
+    process.env["CODEQL_ACTION_TRACER_ENV_VAR_NAMES"] ?? "null",
+  );
+  if (tracerEnvVarNames !== null) {
+    for (const envVar of Object.keys(tracerEnvVarNames)) {
```
adityasharad, line 137: I do not think the types work out here.
- `tracerEnvVarNames` was originally `string[]`
- `process.env.TRACER_ENV_VAR_NAMES` is `JSON.stringify(tracerEnvVarNames)`, so it will be a string containing a JSON-encoded `string[]` array
- `tracerEnvVarNames` in this step is `JSON.parse(process.env.TRACER_ENV_VAR_NAMES)`, so it will be `string[]`
- `Object.keys(tracerEnvVarNames))` will be the keys of an array, so `['0', '1', ... ]`

I think it is enough to verify it is an array, and iterate over the array.

src/tracer-config.ts
```diff
@@ -128,3 +128,14 @@ export async function getCombinedTracerConfig(
...
+      process.env[envVar] = "";
```
adityasharad, line 138: Two questions:
1. Do we want to set to the empty string, rather than unsetting the variable from the environment entirely?
2. Can we document in comments the reasoning you had from the internal issue? I think the idea is that we want to unset the variables for our own steps, but ensure the variables are correctly propagated to subsequent steps (which might be builds). So we use `core.exportVariable` (which sets both for current step and future steps) and then remove from the current step's environment.

src/init-action.ts
angelapwen: Ah yes, that'd do it!

src/tracer-config.ts
angelapwen: 1. Good point, I can use `delete` instead to unset the variable entirely. 
2. Yep, can do! I have the reasoning in the commit message/PR description for now but will add it to the code comments.

angelapwen:
Getting some unbound variable errors from the autobuilders. I'll test out only unsetting the variable that we need to fix this issue.

angelapwen:
Hrm... hang on, still getting [warnings related to the insertion when running on `macos-latest`](https://github.com/github/codeql-action/actions/runs/10396716311/job/28791237857#step:5:611) so moving back to draft.

angelapwen:
Closing as unsetting this variable doesn't seem to have done what we'd hoped. We will stop running the system binary instead for now: https://github.com/github/codeql-action/pull/2434

PR #2430, opened 2024-08-19
Fix writing diagnostics on Windows

henrymercer:
Fix an issue where the CodeQL Action could not write diagnostics on Windows by removing colons from the generated filename so it is valid on Windows.


nickfyson (dismissed):
LGTM

nickfyson (approved):

henrymercer:
Checked that the diagnostics are being written correctly in the PR checks.  We may want to add an automated test too, but I'm not sure it's worth it.

PR #2434, opened 2024-08-20
Stop checking disk usage for MacOS ARM with SIP disabled

angelapwen:
On MacOS ARM machines where SIP is disabled, after the build tracer is initialized in the `init` Action, we receive warnings when we run send status reports due to the `df` binary. This change will make it so that we no longer run `df` for those machines.


aeisenberg (approved):
Nice

src/util.ts
```diff
@@ -1013,14 +1014,23 @@ export interface DiskUsage {
...
 export async function checkDiskUsage(
-  logger?: Logger,
+  logger: Logger,
 ): Promise<DiskUsage | undefined> {
   try {
+    // We avoid running the `df` binary under the hood for macOS ARM runners with SIP disabled.
+    if (
+      process.platform === "darwin" &&
+      (process.arch === "arm" || process.arch === "arm64") &&
+      !(await isSipEnabled(logger))
```
aibaars, line 1024: @angelapwen 
I think `isSipEnabled` should not be called after the tracer has been initialised. Perhaps run it once in `init` and store the result with `core.saveState()` . The `csrutil` command is arm64e just like `df`.

angelapwen: @aibaars thanks!! `core.saveState()` as far as I'm aware only works within an Action and its wrappers (eg. the post-hook) — usually we use an environment variable to pass state between steps. I've done that in https://github.com/github/codeql-action/pull/2434.

PR #2441, opened 2024-08-22
Only run check SIP enablement once in `init` step

angelapwen:
In https://github.com/github/codeql-action/pull/2434 we stopped calling the `df` system binary on MacOS ARM runners where SIP was disabled, but the actual call to `csrutil` itself was run again and is also a binary 😸 

This change makes it so that `csrutil` is only called once, in the `init` step before the build tracer environment variables are set, and then its result is saved in an environment variable. 


src/util.ts
```diff
@@ -1113,11 +1113,17 @@ export function cloneObject<T>(obj: T): T {
...
-// Integrity Protection is enabled.
-export async function isSipEnabled(
+// The first time this function is called, it runs `csrutil status` to determine
+// whether System Integrity Protection is enabled; and saves the result in an
+// environment variable. Afterwards, simply return the value of the environment
+// variable.
+export async function checkSipEnablement(
   logger: Logger,
 ): Promise<boolean | undefined> {
+  if (process.env[EnvVar.IS_SIP_ENABLED] !== undefined) {
```
henrymercer, line 1123: We may also want to check that it's not the empty string (or defensively that it's either "true" or "false")

henrymercer (dismissed):

src/util.ts
angelapwen: I can check if it's "true" or "false" 👍 good suggestion!

src/util.ts
```diff
@@ -1113,11 +1113,20 @@ export function cloneObject<T>(obj: T): T {
...
+// The first time this function is called, it runs `csrutil status` to determine
+// whether System Integrity Protection is enabled; and saves the result in an
+// environment variable. Afterwards, simply return the value of the environment
+// variable.
+export async function checkSipEnablement(
   logger: Logger,
 ): Promise<boolean | undefined> {
+  if (
+    process.env[EnvVar.IS_SIP_ENABLED] !== undefined &&
+    process.env[EnvVar.IS_SIP_ENABLED] in ["true", "false"]
```
henrymercer, line 1125: This is a TypeScript gotcha: `in` just looks at keys, e.g. `"true" in {"true": 1, "false": 1}` is true, but `"true" in ["true", "false"]` is false.

```suggestion
    ["true", "false"].includes(process.env[EnvVar.IS_SIP_ENABLED])
```

henrymercer (approved):

src/util.ts
angelapwen: Thanks for the catch!!

henrymercer (approved):

PR #2442, opened 2024-08-23
Add support for using zstd-compressed nightly bundles

henrymercer:
We are currently experimenting with using bundles compressed using Zstandard (zstd) rather than gzip to potentially save a significant chunk of the time we spend extracting the bundle.  This could lead to a noticeable analysis speedup, particularly for small codebases where fixed costs represent a higher proportion of the total runtime.

This PR is one of the first steps.  Specifically:

- We allow users to pass zstd bundles using the `tools` input to `init` 
- We add support for decompressing these zstd bundles.  This works by letting tar auto-detect the compression method, i.e. instead of running `tar xz`, `x` for extract, `z` for gzip, we just run `tar x`.
- We add telemetry for the compression method used and the version of tar (this informs us whether zstd should be available).
- We download zstd bundles in PR checks to test out this functionality (and hopefully speed up the PR checks!)


angelapwen (approved):
Looks good, couple of questions:
- is there a reason we're only supporting for nightly bundles so far? (maybe I've missed the context behind that) 
- should we expose this in an `experimental` changenote or will we do that once we've deemed the experiment a success?

henrymercer:
Thanks for the review!

>     * is there a reason we're only supporting for nightly bundles so far? (maybe I've missed the context behind that)

Only that we haven't had a stable release with a zstd bundle yet.  My thinking was that it would be easier to write and test the functionality for stable versions once we have a stable zstd bundle.

>     * should we expose this in an `experimental` changenote or will we do that once we've deemed the experiment a success?

I was thinking no since this only applies to nightly bundles which we haven't advertised before, but I don't feel strongly about it.

aeisenberg:
> > * should we expose this in an `experimental` changenote or will we do that once we've deemed the experiment a success?
> 
> I was thinking no since this only applies to nightly bundles which we haven't advertised before, but I don't feel strongly about it.

When we're confident about this being a positive improvement, we will go through a proper release channel just to make sure we put out a changelog note and update documentation.

PR #2443, opened 2024-08-23
Add `config-file` to init status report

dbartol:
Testing using a private branch confirm that we're already receiving the telemetry correctly.


aeisenberg:
So, this will add the path to the config file to our telemetry? If so, how useful is this for us? More precisely, what questions will this help us answer? Would we get the same value by adding a boolean field something like `uses_config_file: true/false`?

angelapwen:
> So, this will add the path to the config file to our telemetry? If so, how useful is this for us? More precisely, what questions will this help us answer? Would we get the same value by adding a boolean field something like `uses_config_file: true/false`?

From internal issue: it looks like we want to know:
- if a config file was used
- whether the config file was remote or local

I guess having the path will answer both of those questions. Or we could alternatively have 2 fields in the status report, a boolean for if a config file was used and an enum/string. 

The latter would probably make it easier to query (so we have the logic determining whether the file was remote or local in the Action, rather than in the query). On the other hand, we've already merged the changes to the schema/API so changing the shape of the data we send through would require updating those.

dbartol:
We already have to do a similar local vs. remote parsing in the Kusto query for the `queries` property, so we might as well do the same for the config file path.

angelapwen (approved):

PR #2451, opened 2024-08-28
Change "recommended" to "minimum"

aeisenberg:
The wording was causing confusion.


rvermeulen:
For posterity, following the recommendation and pinning the recommended version (e.g., `v3.25.11`) of the Action on a GHES instance that uses the [CodeQL Action sync tool](https://github.com/github/codeql-action-sync-tool/) could lead to an error where the CodeQL bundle wasn't found because only the latest bundle (i.e., the default bundle for `v3` at the time of syncing ) for the major version was synchronized.

aeisenberg:
Maybe we should explicitly state something like: we recommend pinning to the major version tag `v3` even on GHES in order to get the most recent updates to the action.

henrymercer (commented):
I like the idea here.  Couple of points:

- How do we clarify that Enterprise Server 3.10 does not support CodeQL Action v3, because it does not support running Actions using Node 20?  And Enterprise Server 3.11 only supports CodeQL Action v3 if you pull in v3 using GitHub Connect?
- Does the Dotcom entry still make sense?  It might be worth removing it.

aeisenberg:
> * How do we clarify that Enterprise Server 3.10 does not support CodeQL Action v3, because it does not support running Actions using Node 20?  And Enterprise Server 3.11 only supports CodeQL Action v3 if you pull in v3 using GitHub Connect?

I'll add a new column "Notes" to add this info.


> Does the Dotcom entry still make sense? It might be worth removing it.

No. Now that it is explicitly "minimum" version, I agree that this should be removed.

README.md
```diff
@@ -33,20 +33,19 @@ To provide the best experience to customers using older versions of GitHub Enter
...
 |---------|----------|--------------|
-| `v3` | default (do not pass a `tools` input) | GitHub.com |
-| `v3.25.11` | `2.17.6` | Enterprise Server 3.14 |
-| `v3.24.11` | `2.16.6` | Enterprise Server 3.13 |
-| `v3.22.12` | `2.15.5` | Enterprise Server 3.12 |
-| `v2.22.1` | `2.14.6` | Enterprise Server 3.11 |
-| `v2.20.3` | `2.13.5` | Enterprise Server 3.10 |
-
-CodeQL Action `v2` will stop receiving updates when GHES 3.11 is deprecated.
+| `v3.25.11` | `2.17.6` | Enterprise Server 3.14 | Does not support CodeQL Action v2 |
```
henrymercer, line 42: Do you have a link for this?  I didn't realise that GHES 3.14 already dropped support for Node 16 Actions.

henrymercer (dismissed):

README.md
aeisenberg: Sorry. I was mistaken about this. I'll remove the note section.

```suggestion
| `v3.25.11` | `2.17.6` | Enterprise Server 3.14 |  |
```

angelapwen (approved):

PR #2461, opened 2024-09-04
Bump artifacts dependencies

angelapwen:


angelapwen:
Okay... it looks like uploading to a different artifact directory (name) each time will work, but this will break how some of our PR checks look for debug artifacts so I'll have to update those too.

Also, I'm trying to think of any other consumers we have that might expect the debug directories to look a certain way 🤔

src/debug-artifacts.ts
```diff
@@ -47,16 +58,29 @@ export async function uploadDebugArtifacts(
...
-      sanitizeArifactName(`${artifactName}${suffix}`),
-      toUpload.map((file) => path.normalize(file)),
-      path.normalize(rootDir),
-      {
-        continueOnError: true,
-        // ensure we don't keep the debug artifacts around for too long since they can be large.
-        retentionDays: 7,
-      },
-    );
+    if (ghVariant === GitHubVariant.GHES) {
```
aeisenberg: Do you need to check for the GHES version here?

angelapwen:
I split out a bunch of the refactoring into https://github.com/github/codeql-action/pull/2475 so that this PR can stick to bumping the dependencies.

src/debug-artifacts.ts
angelapwen: I don't think so — no versions of GHES support the new artifacts dependencies yet 🤦

angelapwen:
Closing in favor of https://github.com/github/codeql-action/pull/2482.

PR #2474, opened 2024-09-10
Always upload eslint.sarif

aeisenberg:


aeisenberg:
@rvermeulen, can I get another review?

angelapwen (approved):
This means that it will always upload even if previous steps fail, right? 👍

aeisenberg:
> This means that it will always upload even if previous steps fail, right? 👍

Yes.

Here's an example of this working: https://github.com/github/codeql-action/pull/2473

PR #2475, opened 2024-09-11
Refactor: prepare debug artifacts for `artifact` upgrades

angelapwen:
As we prepare to upgrade to `upload-artifacts@v4`, `download-artifacts@v4`, or `@actions/artifacts@v2`, we noticed that there's a breaking change in that we can no longer upload to the same artifact name anymore: https://github.com/actions/upload-artifact/issues/478 — once an artifact upload is completed, that artifact directory is immutable. 

This PR prepares all our uses of the artifact dependencies for an easier switchover, but does not yet bump the dependencies for the switchover. The change should be primarily a refactor: the existing debug artifacts PR checks continue to pass.

- Previously, we uploaded SARIF artifacts in the `analyze-post` step and database and log artifacts in the `init-post` step. As we migrate to the updated `artifact` dependencies, we want switch to uploading _all_ artifacts in the `init-post` step. This should be safe as it always runs last.
  - In order to upload all artifacts in one go and maintain the artifacts at the root of the debug directory, we first move SARIF artifacts to the database directory. This should not affect any other consumers of the SARIF file as this occurs in the `init-post` step.
- Previously, we also uploaded combined SARIF artifacts, under the same artifact name, in both the `analyze-post` step and the `upload-sarif-post` step. This change ensures that these artifacts are uploaded at most once — in `analyze-post` if it is a first-party run and `upload-sarif-post` if it is a third-party run.
- The combined SARIF artifacts are now uploaded under the artifact name `combined-sarif-artifacts` rather than `upload-debug-artifacts` for clarity.

- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if nexessary.


src/upload-sarif-action-post.ts
```diff
@@ -6,14 +6,19 @@
...
 import { wrapError } from "./util";
 
 async function runWrapper() {
   try {
-    await uploadSarifActionPostHelper.uploadArtifacts(
-      debugArtifacts.uploadDebugArtifacts,
-    );
+    // Upload SARIF artifacts if we determine that this is a third-party analysis run.
+    // For first-party runs, this artifact will be uploaded in the `analyze-post` step.
+    if (process.env[EnvVar.INIT_ACTION_HAS_RUN] !== "true") {
```
aeisenberg, line 17: Is this file ever imported by another typescript file? It looks like this file is only referenced from `upload-sarif/action.yml`. If so, I don't think we need the if statement around the call to upload the files.

src/analyze-action-post-helper.ts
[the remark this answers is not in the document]
angelapwen: Good point 👍 I can delete it.

src/debug-artifacts.ts
```diff
@@ -23,6 +24,67 @@ export function sanitizeArifactName(name: string): string {
...
+    ) {
+      const sarifFile = path.resolve(analyzeActionOutputDir, `${lang}.sarif`);
+      // Move SARIF to DB location so that they can be uploaded with the same root directory as the other artifacts.
+      if (fs.existsSync(sarifFile)) {
+        const sarifInDbLocation = path.resolve(
+          config.dbLocation,
+          `${lang}.sarif`,
+        );
+        fs.renameSync(sarifFile, sarifInDbLocation);
+        filesToUpload = filesToUpload.concat(sarifInDbLocation);
```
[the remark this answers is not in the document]
angelapwen: Yep, makes sense — I had been meaning to make this change!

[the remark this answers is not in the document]
angelapwen: Is there any reason to use the `...` operator when I know it's a single file like the `sarifInDbLocation` variable?

src/upload-sarif-action-post.ts
angelapwen: It's not imported elsewhere, but`uploadSarifActionPostHelper.uploadArtifacts()` is and I'm trying to make sure it's called at most once — it's called once here and once in `analyze-post`. So in `upload-sarif-post` we only upload these artifacts if we're in third-party analysis; and in `analyze-post` we only upload if we're in first-party analysis.

src/analyze-action-post.ts
```diff
@@ -5,20 +5,20 @@
...
-
-    // Also run the upload-sarif post action since we're potentially running
-    // the same steps in the analyze action.
-    await uploadSarifActionPostHelper.uploadArtifacts(
-      debugArtifacts.uploadDebugArtifacts,
-    );
+    // Upload SARIF artifacts if we determine that this is a first-party analysis run.
+    // For third-party runs, this artifact will be uploaded in the `upload-sarif-post` step.
+    if (process.env[EnvVar.INIT_ACTION_HAS_RUN] === "true") {
+      await uploadSarifActionPostHelper.uploadArtifacts(
```
henrymercer, line 18: Appreciate that this is existing code, but I think it's a bit confusing that the `analyze-post` Action calls into the `upload-sarif-post` helper library.  Perhaps we could move this function to `debug-artifacts` instead?

Though I actually wonder whether we can remove this functionality altogether.  To my knowledge we used it while migrating the SARIF merging to the CLI, but we haven't needed it since.

src/debug-artifacts.ts
```diff
@@ -23,6 +24,65 @@ export function sanitizeArifactName(name: string): string {
...
+      fs.lstatSync(analyzeActionOutputDir).isDirectory()
+    ) {
+      const sarifFile = path.resolve(analyzeActionOutputDir, `${lang}.sarif`);
+      // Move SARIF to DB location so that they can be uploaded with the same root directory as the other artifacts.
+      if (fs.existsSync(sarifFile)) {
+        const sarifInDbLocation = path.resolve(
+          config.dbLocation,
+          `${lang}.sarif`,
+        );
+        fs.renameSync(sarifFile, sarifInDbLocation);
```
henrymercer, line 48: Minor: Consider copying rather than renaming, just to minimise side effects.

henrymercer (commented):
Thanks for separating out the PRs — this is easy to review!

src/analyze-action-post.ts
angelapwen: Yeah, I found that a bit confusing as well. I can make the move 👍 

I looked back at the code added to [migrate the SARIF merging to the CLI](https://github.com/github/codeql-action/pull/2199/commits/e20c2732953c51f43160fe8cdc654b6d6d1c241e) and it looks like the CLI version that can merge SARIF is 2.17.0. So it seems like we probably need to keep supporting the SARIF merge in the Action (+ uploading its related debug artifacts) until we no longer support CLI versions < 2.17.0. (cc @koesie10 to correct me if I'm wrong!)

angelapwen: Moved the code to `debug-artifacts` and re-named the artifact name to `combined-sarif-artifacts` for clarity!

src/debug-artifacts.ts
angelapwen: Done!

henrymercer (approved):

PR #2477, opened 2024-09-13
Fix incorrect documentation about the `token` input to the Actions.

chrisgavin:
This documentation is currently misleading as it implies you can use any token here. In reality, this Action calls API endpoints that only accept tokens from the GitHub Actions app, so you should basically never override it.

I believe all the other Actions in this repository will work with an arbitrary token if one needs to be provided for cross-repository access, but these two call the SARIF upload endpoint which requires an Actions token.

### DIFF
--- a/analyze/action.yml
+++ b/analyze/action.yml
@@ -74,7 +74,7 @@ inputs:
     required: true
     default: "true"
   token:
-    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow must have the `security-events: write` permission."
+    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow must have the `security-events: write` permission. Most of the time it is advisable to avoid specifying this input so that the workflow falls back to using the default value."
     required: false
     default: ${{ github.token }}
   matrix:
--- a/analyze/action.yml
+++ b/analyze/action.yml
@@ -74,7 +74,7 @@ inputs:
     required: true
     default: "true"
   token:
-    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow should have the `security-events: write` permission."
+    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow must have the `security-events: write` permission."
     required: false
     default: ${{ github.token }}
   matrix:
--- a/analyze/action.yml
+++ b/analyze/action.yml
@@ -74,7 +74,7 @@ inputs:
     required: true
     default: "true"
   token:
-    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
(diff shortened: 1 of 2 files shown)


analyze/action.yml
```diff
@@ -74,7 +74,7 @@ inputs:
     required: true
     default: "true"
   token:
-    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
+    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow should have the `security-events: write` permission."
```
aeisenberg, line 77:
```suggestion
    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow must have the `security-events: write` permission."
```

upload-sarif/action.yml
```diff
@@ -20,7 +20,7 @@ inputs:
     description: "The sha of the HEAD of the ref where results will be uploaded. If not provided, the Action will use the GITHUB_SHA environment variable. If provided, the ref input must be provided as well. This input is ignored for pull requests from forks."
     required: false
   token:
-    description: "GitHub token to use for authenticating with this instance of GitHub. The token needs the `security-events: write` permission."
+    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow should have the `security-events: write` permission."
```
aeisenberg, line 23:
```suggestion
    description: "GitHub token to use for authenticating with this instance of GitHub. The token must be the built-in GitHub Actions token, and the workflow must have the `security-events: write` permission."
```

aeisenberg (commented):
I think `security-events` is required.

aeisenberg:
I wonder if we should also include the guidance that it's best to avoid using this input and so the workflow falls back to using the actions-provided token.

chrisgavin:
That might make sense, though I'm not sure how exactly to phrase it. There are some pretty specific situations where it is useful to be able to set these inputs, but in general they probably shouldn't be used.

aeisenberg:
What do you think of this?

> Most of the time it is advisable to avoid specifying this input so that the workflow falls back to using the default value.

If you need to specify the input, you will be an advanced user anyway, so I don't think this would be confusing.

aeisenberg (approved):
Nice.

jsoref:
This is very confusing since the underlying api clearly says it's possible:
https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#upload-an-analysis-as-sarif-data--fine-grained-access-tokens

Oh well, at least the documentation now warns that this doesn't work.

I'd almost suggest that the description should say:
> If you think you want to use a custom token you should probably talk to the endpoint directly instead of using this action.

* For `analyze/action.yml` that would mean `with: { "upload": "never" }` followed by a manual call to the endpoint.
* For `upload-sarif/action.yml` which is the one I actually use, I have no idea what to do since I'm _pretty_ sure that the action is actually rewriting my sarif file with some extra hints, and I guess I'd need to find a way to ask code like that to do that rewrite and then manually call the endpoint directly.

PR #2481, opened 2024-09-13
Use generated token on checkout

rvermeulen:
The script `.github/update-release-branch.py` uses the `git` command to push changes. Therefore we need to ensure that `git` authenticates with a token that has the `workflows` write permision.

This change restore the GitHub token used by the script to access the API and applies the `workflows` write permission to the token used by `git`.


dbartol:
Where does the `workflows: write` permission get applied?

rvermeulen:
This happens on the GitHub App that generates the token.

PR #2482, opened 2024-09-13
Bump artifact dependencies if `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` enabled

angelapwen:
We need to bump `actions/upload-artifact` and `actions/download-artifact` to v4 in our PR checks, as v3 will be deprecated in November: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ 

To do so, we also need to bump `@actions/artifact` to v2, but it is not yet supported on GHES: https://www.npmjs.com/package/@actions/artifact#v2---whats-new

This change would immediately cause customers using v3 of `actions/download-artifact` in their workflows to see failures, so we are gating it behind an opt-in feature flag for another month. Customers will be able to set the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable to true in their workflows as they upgrade `actions/download-artifact` to `v4`.

At the beginning of November, we will set the default value of the feature flag to `true` and the upgraded version of the artifact dependencies will be the default. 

This PR:
- imports `@actions/artifact@v1` as `artifact-legacy` and imports `@actions/artifact@v2` as `artifact`
- adds a feature flag, defaulting to false, that can be enabled by setting the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable to true 
- uses the legacy version of the `artifact` client on GHES or when the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` flag is false; and otherwise the upgraded version
- adds a PR check that sets `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` to true, bumping `upload-artifact` and `download-artifact` to v4.


angelapwen:
💭 People who have workflows using `download-artifacts@v3` will no longer be able to access the debug artifacts that way (they will need to use `@v4`) — I'm not sure how many users do that as usually the debug artifacts are more of a manual inspection sort of thing. If we think enough people do that we could add a changenote.

src/analyze-action-post.ts
```diff
@@ -5,16 +5,23 @@
...
+  const logger = getActionsLogger();
   try {
     // Upload SARIF artifacts if we determine that this is a first-party analysis run.
     // For third-party runs, this artifact will be uploaded in the `upload-sarif-post` step.
     if (process.env[EnvVar.INIT_ACTION_HAS_RUN] === "true") {
-      await debugArtifacts.uploadCombinedSarifArtifacts();
+      const config = await getConfig(getTemporaryDirectory(), logger);
+      if (config !== undefined) {
+        await debugArtifacts.uploadCombinedSarifArtifacts(config);
+      }
```
aeisenberg: Should we warn when `config === undefined`?

src/debug-artifacts.ts
```diff
@@ -148,16 +159,29 @@ export async function uploadDebugArtifacts(
...
-    );
+    if (ghVariant === GitHubVariant.GHES) {
+      await artifactLegacy.create().uploadArtifact(
+        sanitizeArifactName(`${artifactName}${suffix}`),
+        toUpload.map((file) => path.normalize(file)),
+        path.normalize(rootDir),
+        {
+          continueOnError: true,
+          // ensure we don't keep the debug artifacts around for too long since they can be large.
+          retentionDays: 7,
```
aeisenberg: I think the arguments here and below are the same. If so, I'd like to make that explicit by only specifying them once. And if they aren't the same, I'd like it to be clear where the differences are. You could probably do something like:

```js

const artifactUploader = (ghVariant === GitHubVariant.GHES 
  ? artifactLegacy.create()
  : new artifact.DefaultArtifactClient())

artifactUploader.uploadArtifact(...args...);
```

src/start-proxy-action-post.ts
```diff
@@ -31,18 +32,36 @@ async function runWrapper() {
...
+        `Did not upload debug artifacts because cannot determine the GitHub variant running.`,
+      );
+      return;
+    }
     try {
-      await artifactLegacy
-        .create()
-        .uploadArtifact(
+      if (config.gitHubVersion.type === GitHubVariant.GHES) {
+        await artifactLegacy
```
aeisenberg: Same here as above.

aeisenberg (commented):
Nice. Some suggestions inline.

>  +2,174,754 −72,459

Productive day! 🎉 😄

src/debug-artifacts.ts
angelapwen: They're almost the same, except the new client no longer accepts `continueOnError` as an option 😬 I might be able to use the ternary operator to append the `continueOnError` only for the legacy client.

angelapwen: Actually, it looks like the default behavior of the legacy client is to set `continueOnError` to true. So we don't have to specify it and the args can be the same!

src/analyze-action-post.ts
angelapwen: I'm not sure actually: elsewhere when we see that `config` is `undefined`, it indicates that the `init` step didn't succeed successfully. Since this is the `analyze-post` step, the `init` step should have succeeded. So it seems unlikely that anyone will ever end up in this state.

That made me think of something else though.. will post a separate comment.

src/upload-sarif-action-post.ts
```diff
@@ -5,16 +5,23 @@
...
+  const logger = getActionsLogger();
   try {
     // Upload SARIF artifacts if we determine that this is a third-party analysis run.
     // For first-party runs, this artifact will be uploaded in the `analyze-post` step.
     if (process.env[EnvVar.INIT_ACTION_HAS_RUN] !== "true") {
-      await debugArtifacts.uploadCombinedSarifArtifacts();
+      const config = await getConfig(getTemporaryDirectory(), logger);
+      if (config !== undefined) {
+        await debugArtifacts.uploadCombinedSarifArtifacts(config);
+      }
```
angelapwen: Thinking out loud: Since `config` gets set in `init`, and this is only supposed to run in third-party analysis runs, `config` will _always_ be undefined here. So I think we actually need to determine the GH variant separately here.

angelapwen:
Holding on this as I investigate internal reports that #2475 has caused some regressions

src/debug-artifacts.ts
angelapwen: Done in the new commits I pushed.

src/upload-sarif-action-post.ts
angelapwen: Done in the new commits I pushed.

src/debug-artifacts.ts
```diff
@@ -246,16 +260,35 @@ export async function uploadDebugArtifacts(
...
-  await artifactLegacy.create().uploadArtifact(
+  // `@actions/artifact@v2` is not yet supported on GHES so the legacy version of the client will be used on GHES
+  // until it is supported. We also use the legacy version of the client if the feature flag is disabled.
+  const artifactUploader =
+    ghVariant !== GitHubVariant.GHES &&
+    (await features.getValue(Feature.ArtifactUpgrade))
+      ? new artifact.DefaultArtifactClient()
+      : artifactLegacy.create();
+
+  const artifactUploaderArgs: [
```
henrymercer: Here we're constructing the arguments in a separate array — since we're only using these arguments once, could we pass them directly in the call to `artifactUploader.uploadArtifact`?

src/feature-flags.ts
```diff
@@ -40,6 +40,7 @@ export interface FeatureEnablement {
  * Legacy features should end with `_enabled`.
  */
 export enum Feature {
+  ArtifactUpgrade = "artifact_upgrade",
```
henrymercer: Nit: consider `artifact_v3_upgrade` in case there's another one in the future.

src/debug-artifacts.ts
```diff
@@ -246,16 +260,35 @@ export async function uploadDebugArtifacts(
     }
   }
 
-  await artifactLegacy.create().uploadArtifact(
+  // `@actions/artifact@v2` is not yet supported on GHES so the legacy version of the client will be used on GHES
+  // until it is supported. We also use the legacy version of the client if the feature flag is disabled.
+  const artifactUploader =
```
henrymercer: Could we add some logging explaining which version we chose and why?

src/start-proxy-action-post.ts
```diff
@@ -31,18 +41,51 @@ async function runWrapper() {
...
-            retentionDays: 7,
-          },
-        );
+      // `@actions/artifact@v2` is not yet supported on GHES so the legacy version of the client will be used on GHES
+      // until it is supported. We also use the legacy version of the client if the feature flag is disabled.
+      const artifactUploader =
+        config?.gitHubVersion.type !== GitHubVariant.GHES &&
+        (await features.getValue(Feature.ArtifactUpgrade))
+          ? new artifact.DefaultArtifactClient()
+          : artifactLegacy.create();
```
henrymercer: This looks identical to the code in `debug-artifacts` — can we reuse it?

src/start-proxy-action-post.ts
```diff
@@ -31,18 +41,51 @@ async function runWrapper() {
...
+
+      const artifactUploaderArgs: [
```
henrymercer: Same question here around the indirection

.github/workflows/debug-artifacts-upgrade.yml
```diff
@@ -0,0 +1,99 @@
+# Checks logs, SARIF, and database bundle debug artifacts exist and are accessible
+# with download-artifact@v4 when CODEQL_ACTION_ARTIFACT_UPGRADE is set to true.
+name: PR Check - Debug artifact upload using artifact@v2
```
henrymercer, line 3: Nice :)

CHANGELOG.md
```diff
@@ -6,7 +6,10 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add support for `@actions/artifact@v2` for uploading debug artifacts, for customers not running on GitHub Enterprise Server. [#2482](https://github.com/github/codeql-action/pull/2482)
```
henrymercer, line 9: I would lead with compatibility with the `actions/download-artifact@v4` Action.  The npm package is comparatively less well known and it's confusing that the version number is different.

CHANGELOG.md
```diff
@@ -6,7 +6,10 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
...
+  - Only workflows that use `actions/download-artifact` to programmatically download the CodeQL Action debug artifacts will be affected: with the legacy artifact client, only `actions/download-artifact@v3` will be compatible. With the upgraded artifact client, only `actions/download-artifact@v4` will be compatible.
+  - To opt in to the upgraded artifact client, set the `CODEQL_ACTION_ARTIFACT_UPGRADE` environment variable to true and bump to `actions/download-artifact@v4`.
```
henrymercer, line 11: I would be stronger here — if users are programmatically consuming artifacts, we recommend they set this environment variable to true and bump to `actions/download-artifact@v4` to avoid breakage when v4 becomes the default.

henrymercer (commented):
Thanks for the extra work here to get this behind a feature flag and provide a smooth upgrade path! ✨

src/debug-artifacts.ts
angelapwen: Oh yes you're right! I think I had initially constructed them separately to push an extra property on under certain conditions. Changed!

src/feature-flags.ts
angelapwen: Done! It's `v2` though 😸

src/debug-artifacts.ts
angelapwen: Done!

src/start-proxy-action-post.ts
angelapwen: Makes sense 👍 done

CHANGELOG.md
angelapwen: Sounds good — reworked the changenote with your comment here and below!

.github/workflows/debug-artifacts-upgrade.yml
```diff
@@ -0,0 +1,99 @@
...
+      matrix:
+        version:
+        - stable-v2.13.5
+        - stable-v2.14.6
+        - stable-v2.15.5
+        - stable-v2.16.6
+        - stable-v2.17.6
+        - default
+        - linked
+        - nightly-latest
```
aeisenberg, line 34: Are we not testing 2.18 or 2.19? How will we maintain this list going forward? Same question for the set of versions below.

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add support for using `actions/download-artifact@v4` to programmatically consume CodeQL Action debug artifacts. Customers who use `actions/download-artifact` should set the `CODEQL_ACTION_ARTIFACT_UPGRADE` environment variable to true and bump `actions/download-artifact@v3` to `actions/download-artifact@v4` to avoid breakage. All customers, except those running on GitHub Enterprise Server, will be opted-in to this change in early November. [#2482](https://github.com/github/codeql-action/pull/2482)
+  - This is because `actions/upload-artifact@v3` and `actions/download-artifact@v3` will be deprecated at the end of November. See [GitHub Changelog](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).
+  - This change is currently unavailable for GitHub Enterprise Server customers, as the compatible npm `@actions/artifact@v2` used to upload the artifacts is not yet supported on GHES.
```
aeisenberg, line 11: Nit: markdown requires 4 spaces for indenting.

```suggestion
    - This is because `actions/upload-artifact@v3` and `actions/download-artifact@v3` will be deprecated at the end of November. See [GitHub Changelog](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).
    - This change is currently unavailable for GitHub Enterprise Server customers, as the compatible npm `@actions/artifact@v2` used to upload the artifacts is not yet supported on GHES. 
```

src/debug-artifacts.ts
```diff
@@ -246,16 +263,51 @@ export async function uploadDebugArtifacts(
...
+  features: FeatureEnablement,
+): Promise<artifact.ArtifactClient | artifactLegacy.ArtifactClient> {
+  if (ghVariant === GitHubVariant.GHES) {
+    logger.info(
+      "Uploading debug artifacts using the `@actions/artifact@v1` client because the `v2` version is not yet supported on GHES.",
+    );
+    return artifactLegacy.create();
+  } else if (!(await features.getValue(Feature.ArtifactV2Upgrade))) {
+    logger.info(
+      "Uploading debug artifacts using the `@actions/artifact@v1` client because the value of the relevant feature flag is false. To use the `v2` version of the client, set the `CODEQL_ACTION_ARTIFACT_V2_UPGRADE` environment variable to true.",
```
aeisenberg, line 302: Nice explanation. 👍🏼

aeisenberg (commented):
Looks good. The test makes sense, but I just want to make sure it is easy to maintain.

.github/workflows/debug-artifacts-upgrade.yml
angelapwen: We try to keep the list of versions in the checks that don't use the PR check generator consistent with the `sync.py` file: https://github.com/github/codeql-action/blob/46e0c78da9edf293aeab3d4d62cf1a7b7534c6a0/pr-checks/sync.py#L9-L28. When we update the list in that file, we also update the manually created PR checks. 

`default` is currently 2.19.0, so looks like we'll need to add 2.18.6 to the list. I'll open up a separate internal issue to track that.

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Add support for using `actions/download-artifact@v4` to programmatically consume CodeQL Action debug artifacts. Customers who use `actions/download-artifact` should set the `CODEQL_ACTION_ARTIFACT_UPGRADE` environment variable to true and bump `actions/download-artifact@v3` to `actions/download-artifact@v4` to avoid breakage. All customers, except those running on GitHub Enterprise Server, will be opted-in to this change in early November. [#2482](https://github.com/github/codeql-action/pull/2482)
```
henrymercer, line 9: I think this environment variable changed.  Also should the environment variable also mention "v4" so we're mentioning v4 everywhere?

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
...
+    - This is because `actions/upload-artifact@v3` and `actions/download-artifact@v3` will be deprecated at the end of November. See [GitHub Changelog](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).
+    - This change is currently unavailable for GitHub Enterprise Server customers, as the compatible npm `@actions/artifact@v2` used to upload the artifacts is not yet supported on GHES.
```
henrymercer, line 11: Perhaps instead say `actions/download-artifact@v4` is not yet compatible with GHES, again so we're mentioning v4 everywhere.

src/debug-artifacts.ts
```diff
@@ -299,6 +285,31 @@ export async function uploadDebugArtifacts(
...
+// `@actions/artifact@v2` is not yet supported on GHES so the legacy version of the client will be used on GHES
+// until it is supported. We also use the legacy version of the client if the feature flag is disabled.
+export async function getArtifactUploaderClient(
+  logger: Logger,
+  ghVariant: GitHubVariant,
+  features: FeatureEnablement,
+): Promise<artifact.ArtifactClient | artifactLegacy.ArtifactClient> {
+  if (ghVariant === GitHubVariant.GHES) {
+    logger.info(
+      "Uploading debug artifacts using the `@actions/artifact@v1` client because the `v2` version is not yet supported on GHES.",
```
henrymercer: Consider referring instead to the versions of download-artifact here, so we're using the same version numbers throughout.

henrymercer (commented):
This looks great.  Just a couple of comments about whether we can avoid the confusion between the npm and Action versions by always talking about the Action versions, in line with the public changelog posts.

CHANGELOG.md
angelapwen: Thanks for catching! I updated all the references to `v4` as you suggested, and just commented why we're using `v4` when it looks like a `v2` import in the code comments.

CHANGELOG.md
angelapwen: Good idea 👍

angelapwen:
There is some kind of unrelated CI incident related to Docker so will have to re-run these checks later~

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 
 ## [UNRELEASED]
 
+- Add support for using `actions/download-artifact@v4` to programmatically consume CodeQL Action debug artifacts. Customers who use `actions/download-artifact` should set the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable to true and bump `actions/download-artifact@v3` to `actions/download-artifact@v4` to avoid breakage. All customers, except those running on GitHub Enterprise Server, will be opted-in to this change in early November. [#2482](https://github.com/github/codeql-action/pull/2482)
+    - This is because `actions/upload-artifact@v3` and `actions/download-artifact@v3` will be deprecated at the end of November. See [GitHub Changelog](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).
+    - This change is currently unavailable for GitHub Enterprise Server customers, as `actions/upload-artifact@v4` and `actions/download-artifact@v4` are not yet compatible on GHES.
```
henrymercer, line 11:
```suggestion
- Add support for using `actions/download-artifact@v4` to programmatically consume CodeQL Action debug artifacts. 

  Starting November 30, 2024, GitHub.com customers will [no longer be able to use `actions/download-artifact@v3`](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). Therefore, to avoid breakage, customers who programmatically download the CodeQL Action debug artifacts should set the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable to `true` and bump `actions/download-artifact@v3` to `actions/download-artifact@v4` in their workflows.
  
  This change is currently unavailable for GitHub Enterprise Server customers, as `actions/upload-artifact@v4` and `actions/download-artifact@v4` are not yet compatible with GHES. 
```

src/feature-flags.ts
```diff
@@ -40,7 +40,7 @@ export interface FeatureEnablement {
  * Legacy features should end with `_enabled`.
  */
 export enum Feature {
-  ArtifactV2Upgrade = "artifact_v2_upgrade",
+  ArtifactV4Upgrade = "artifact_v2_upgrade",
```
henrymercer: Should we also update the name of the feature flag?

src/debug-artifacts.ts
```diff
@@ -287,24 +287,26 @@ export async function uploadDebugArtifacts(
...
     logger.info(
-      "Uploading debug artifacts using the `@actions/artifact@v1` client because the `v2` version is not yet supported on GHES.",
+      "Debug artifacts can be consumed with `actions/download-artifact@v3` because the `v4` version is not yet compatible on GHES.",
     );
     return artifactLegacy.create();
-  } else if (!(await features.getValue(Feature.ArtifactV2Upgrade))) {
+  } else if (!(await features.getValue(Feature.ArtifactV4Upgrade))) {
     logger.info(
-      "Uploading debug artifacts using the `@actions/artifact@v1` client because the value of the relevant feature flag is false. To use the `v2` version of the client, set the `CODEQL_ACTION_ARTIFACT_V2_UPGRADE` environment variable to true.",
+      "Debug artifacts can be consumed with `actions/download-artifact@v3` because the value of the relevant feature flag is false. To use the `actions/download-artifact@v4`, set the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable to true.",
```
henrymercer: I'd suggest we drop "because the value of the relevant feature flag is false" here, since customers shouldn't need to know about feature flags.

henrymercer (dismissed):
Nice!  A couple of small suggestions, but this looks good!

src/feature-flags.ts
angelapwen: Good point — forgot to do that because we don't really use that value. Fixed!

src/debug-artifacts.ts
angelapwen: Makes sense 👍 to them it's just behind an environment variable. Done!

henrymercer (dismissed):
Looks good, just one merge conflict to resolve in the changelog.

henrymercer (approved):

PR #2486, opened 2024-09-16
Improve the robustness of creating and uploading debug artifacts

henrymercer:
Tolerate failures when creating and uploading debug artifacts, and add a fallback such that if bundling the database with the CLI fails, we will fall back to creating a partial database bundle.

Also use a helper function to simplify the common pattern `wrapError(e).message`.


src/debug-artifacts.ts
```diff
@@ -65,63 +66,127 @@ export async function uploadCombinedSarifArtifacts() {
   }
 }
 
-export async function uploadAllAvailableDebugArtifacts(
+/**
+ * Try to get the SARIF result path for the given language.
+ *
+ * If an error occurs, log it and return an empty list.
+ */
```
aeisenberg, line 73: This comment and the function name are a little misleading because this function has a side effect that is not mentioned or implied.

aeisenberg (commented):
Just a jsdoc comment that can be improved. Otherwise, 👍🏼 .

angelapwen (approved):
Also looks good to me! Makes sense, the original refactor dropped some of the error catching where the original debug artifact upload methods were called 👍

src/debug-artifacts.ts
henrymercer: Good point, I'll get this in and address this as follow up.

PR #2487, opened 2024-09-16
[SARIF validation] turn URI errors into warnings

rvermeulen:
This PR extends the list of string `format`'s for which we turn validation errors into warnings to ensure we don't fail on invalid URIs.


src/upload-lib.ts
```diff
@@ -449,11 +449,20 @@ export function validateSarifFileSchema(sarifFilePath: string, logger: Logger) {
...
   // Filter errors related to invalid URIs in the artifactLocation field as this
   // is a breaking change. See https://github.com/github/codeql-action/issues/1703
-  const errors = (result.errors || []).filter(
-    (err) => err.argument !== "uri-reference",
+  const warningAttributes = ["uri-reference", "uri"];
+  const errors = (result.errors ?? []).filter(
+    (err) =>
+      !(
+        err.name === "format" &&
+        typeof err.argument === "string" &&
```
aeisenberg, line 457: This line is redundant since `warningAttributes` is an array of strings.

```suggestion
```

src/upload-lib.ts
```diff
@@ -449,11 +449,20 @@ export function validateSarifFileSchema(sarifFilePath: string, logger: Logger) {
...
+        warningAttributes.includes(err.argument)
+      ),
   );
-  const warnings = (result.errors || []).filter(
-    (err) => err.argument === "uri-reference",
+  const warnings = (result.errors ?? []).filter(
+    (err) =>
+      err.name === "format" &&
+      typeof err.argument === "string" &&
```
aeisenberg, line 464: Same

```suggestion
```

aeisenberg (approved):
Two nits. Non-blocking.

src/upload-lib.ts
rvermeulen: The type of `argument` is `any` so we get an error from ESLint if we don't enforce that.

```
Unsafe argument of type any assigned to a parameter of type string.
```

I added the type check to exclude the cases where `argument` is not a `string`.

aeisenberg: Makes sense.

aeisenberg (dismissed):

aeisenberg:
Rebuilding the JS.

aeisenberg:
The rebuild action is failing since it's in a fork. @rvermeulen, if you add the `Rebuild` label, the js will be rebuilt, so you don't need to do it locally.

rvermeulen:
Mmm, I rebuild it locally. Let me look at the failures.

aeisenberg (approved):

PR #2488, opened 2024-09-17
Improve logging when preparing and uploading debug artifacts

henrymercer:
Improve the logging during these steps so it's easier to tell what's going on and easier to debug in the case of a failure.


src/debug-artifacts.ts
```diff
@@ -172,9 +178,10 @@ export async function tryUploadAllAvailableDebugArtifacts(
       }
 
       // Add database bundle
-      filesToUpload.push(
-        ...(await tryBundleDatabase(config, language, logger)),
-      );
+      const databaseBundle = await tryBundleDatabase(config, language, logger);
+      if (databaseBundle) {
+        filesToUpload.push(databaseBundle);
+      }
```
NlightNFotis: I'm curious whether it would make sense, if nothing came out of this operation (`tryBundleDatabase` returned `undefined`), to emit a warning?

The aim would be to make it easier to debug in the future and allow the user some visibility in its internals in case it goes wrong.

src/debug-artifacts.ts
```diff
@@ -233,23 +246,16 @@ export async function uploadDebugArtifacts(
...
-      sanitizeArtifactName(`${artifactName}${suffix}`),
-      toUpload.map((file) => path.normalize(file)),
-      path.normalize(rootDir),
-      {
-        continueOnError: true,
-        // ensure we don't keep the debug artifacts around for too long since they can be large.
-        retentionDays: 7,
-      },
-    );
-  } catch (e) {
```
NlightNFotis, line 226: Since we removed this `catch`, is it correct that now we're not suppressing the failure and instead surface it, failing the run?

Why did we change this behaviour? The change makes sense to me, but it looks potentially breaking for some flows.

NlightNFotis (approved):
LGTM, just a couple of questions.

src/debug-artifacts.ts
henrymercer: We currently log the warning in the `tryBundleDatabase` method.

Another approach would be to have `tryBundleDatabase` return a `Result` like type to capture the success value (a string with the bundled database path), or a failure value (the error).  Then the callee could do the error handling.

I don't have particularly strong feelings either way — let me know what you think :)

src/debug-artifacts.ts
henrymercer: We're now catching the exception in the callers.  The main reason for changing it is to provide a more descriptive error message, since the caller knows what _type_ of debugging artifact we were trying to upload e.g. combined SARIF file, or the regular debug artifacts.

src/debug-artifacts.ts
NlightNFotis: I'm usually quite keen on as pure functions as I can get, so my default preference would be a `Result` return and the handling of the error to go to the caller. That would also simplify the control flow in `tryBundleDatabase`, and give more flexibility to the caller with regard to what to do with an error.

However, this is just more of a philosophical position - don't let me impose extra work on this PR if it's finished. This is good to go in as is IMO.

henrymercer: I think that's fair.  The advantage is that the caller has more choice over what to do with the error.  The disadvantage is that there's more code as we have to deal with errors in both the callee and the caller.

I'm going to leave this for now, but we can bear this in mind the next time we look at this code.

PR #2502, opened 2024-09-23
Add a feature flag to use a bundle compressed using Zstandard when setting up the default tools

henrymercer:
This PR adds a feature flag to download a CodeQL Bundle compressed using Zstandard when setting up the default tools.

If we determine that `tar` is new enough, we will download the zstd-compressed bundle.

We add a fallback mechanism: if setting up the zstd-compressed bundle fails, we will fall back to the gzip-compressed bundle.

We add PR checks to validate both setting up a zstd-compressed bundle when the feature flag is enabled, and the fallback mechanism.


pr-checks/checks/zstd-bundle-fallback.yml
```diff
@@ -0,0 +1,74 @@
+name: "Zstandard bundle fallback"
+description: "Tests the fallback for when downloading a Zstandard-compressed CodeQL Bundle fails"
```
aeisenberg, line 2: Copy-edit

```suggestion
description: "Tests the fallback when downloading a Zstandard-compressed CodeQL Bundle fails"
```

pr-checks/checks/zstd-bundle.yml
```diff
@@ -0,0 +1,63 @@
+name: "Zstandard bundle"
+description: "Tests the feature flag that downloads a Zstandard-compressed CodeQL Bundle by default"
+versions:
+  - linked
+operatingSystems:
+  - ubuntu
```
aeisenberg, line 6: Is there a windows and mac test where the zstd bundle is downloaded?

src/setup-codeql.ts
```diff
@@ -722,3 +815,15 @@ function sanitizeUrlForStatusReport(url: string): string {
...
 }
+
+async function useZstdBundle(
+  cliVersion: string,
+  features: FeatureEnablement,
+  tarSupportsZstd: boolean,
+): Promise<boolean> {
+  return (
+    tarSupportsZstd &&
+    semver.gte(cliVersion, "2.19.0") &&
```
aeisenberg, line 826: Can you extract this as a constant and place it somewhere near the declaration of the feature flag in `feature-flags.ts`? This way at least the version and the flag would be within reading distance of each other.

pr-checks/checks/zstd-bundle.yml
henrymercer: These are tested in the nightly checks, but we can test here too defensively.

aeisenberg (approved):
LGTM!

NlightNFotis (approved):
LGTM

PR #2514, opened 2024-10-02
Capture stderr from extracting Zstandard bundles

henrymercer:
We use the standard toolcache API to decompress bundles, but when decompression fails, we only get the exit code from the toolcache API, which isn't very helpful for debugging.

Therefore this PR calls `tar` directly when decompressing Zstandard bundles to provide better quality failure information.  A large part of the PR is a refactor to expose the functionality we use for invoking and capturing errors from the CLI in a generic way.


src/tar.ts
```diff
@@ -84,24 +89,77 @@ export async function isZstdAvailable(
...
+      return await toolcache.extractTar(tarPath);
     case "zstd":
-      // By specifying only the "x" flag, we ask tar to autodetect the
-      // compression method.
-      return await toolcache.extractTar(path, undefined, "x");
+      if (!tarVersion) {
+        throw new Error(
+          "Could not determine tar version, which is required to extract a Zstandard archive.",
+        );
+      }
```
angelapwen, line 107: Does throwing here prevent us from falling back on the gzip version?

angelapwen (approved):
Refactoring makes sense 👍 just a question!

src/tar.ts
henrymercer: Good idea to double check this — this should indeed be covered by the fallback in `setupCodeQLBundle`.

PR #2526, opened 2024-10-04
Check zstd is on the path in addition to tar version

henrymercer:
`tar` calls out to `zstd` to decompress a `.tar.zst`, so we should check the `zstd` binary is accessible as part of determining whether to download a `.tar.zst` bundle.


src/tar.ts
```diff
@@ -53,36 +67,40 @@ async function getTarVersion(): Promise<TarVersion> {
...
+          foundZstdBinary,
           version: tarVersion,
         };
       default:
         assertNever(type);
     }
   } catch (e) {
-    logger.error(
+    logger.warning(
       "Failed to determine tar version, therefore will assume zstd may not be available. " +
```
aeisenberg, line 100: Minor: I think this is clearer.

```suggestion
      "Failed to determine tar version, therefore will assume zstd is not available. " +
```

src/tar.ts
```diff
@@ -53,36 +67,40 @@ async function getTarVersion(): Promise<TarVersion> {
...
         `The underlying error was: ${e}`,
     );
-    return { available: false };
+    return { available: false, foundZstdBinary };
```
aeisenberg, line 103: What's the use case for knowing if zstd is available if we can't figure out what tar version is available? 

Actually, I'm not sure what the use case for having `foundZstdBinary` as part of the interface is. Is this for telemetry?

henrymercer: > What's the use case for knowing if zstd is available if we can't figure out what tar version is available?

If we can't figure out what tar version is available, we want to fall back to gzip.

> Actually, I'm not sure what the use case for having `foundZstdBinary` as part of the interface is. Is this for telemetry?

Yes, this is just to help us understand how much of the time zstd wasn't available due to the tar version and how much of the time it wasn't available due to not being able to find a zstd binary.

aeisenberg: Got it.

aeisenberg (approved):

PR #2536, opened 2024-10-09
Python: Handle new default for standard library extraction

yoff:
When you review this, could you also confirm my thinking on the following?
- With the change to `feature-flags.ts`:
  - `Feature.CodeqlActionPythonDefaultIsToNotExtractStdlib` will be(come) true iff
      (((the feature flag is on AND NOT the disabling feature flag is on) OR the environment variable is set to true) AND NOT the environment variable is set to false) AND `ToolsFeature.PythonDefaultIsToNotExtractStdlib` is true.
  - That is, this one entry is enough to make the action react to both the feature flag, the disabling feature flag (`x_disabled`), and the environment variable. And the tools feature takes precedence, then the environment variable, then the disabling feature flag, then the feature flag.
  - The environment variable `"CODEQL_ACTION_DISABLE_PYTHON_STANDARD_LIBRARY_EXTRACTION"` is now available to the action.
  - People can write something in their workflow file to turn off the feature.
- With the changes to `src/tools-features.ts`:
  - `ToolsFeature.PythonDefaultIsToNotExtractStdlib` is now received from the CLI as long as the entry in there is called
     `pythonDefaultIsToNotExtractStdlib`
  - if the CLI is not new enough to define this feature, `ToolsFeature.PythonDefaultIsToNotExtractStdlib` will simply be off.


tausbn (approved):
Looks good to me! 👍 
(However, I'm not familiar with most aspects of how `codeql-action` works, so we should probably get someone from the core team to give the go-ahead before merging.)

src/init-action.ts
```diff
@@ -611,6 +611,24 @@ async function run() {
...
+      await codeql.supportsFeature(
+        ToolsFeature.PythonDefaultIsToNotExtractStdlib,
+      )
+    ) {
+      // We are in the case where the default has switched to not extracting the stdlib.
+      if (
+        !(await features.getValue(
+          Feature.CodeqlActionPythonDefaultIsToNotExtractStdlib,
+          codeql,
+        ))
```
henrymercer, line 624: Minor: you could use `await codeql.supportsFeature` here too for consistency

henrymercer (approved):
LGTM, both on the thinking and the code :)

PR #2537, opened 2024-10-10
Don't use Zstandard bundles on Windows

henrymercer:
In testing, gzip performs better than Zstandard on Windows.


pr-checks/checks/zstd-bundle.yml
```diff
@@ -4,7 +4,6 @@ versions:
   - linked
 operatingSystems:
   - macos
-  - windows
```
aeisenberg, line 7: Do we have a test that ensures zstd is not used on windows?

aeisenberg (commented):
Wild...I didn't expect that.

aeisenberg (approved):
Nice.

pr-checks/checks/zstd-bundle.yml
henrymercer: Let's add one :)

PR #2550, opened 2024-10-17
Fix name of Python stdlib extraction feature flag

henrymercer:
I missed this in my initial review of #2536.  Also add a unit test to prevent us from making this mistake again.


yoff (dismissed):
Thanks for catching this. I think it is great that you added a test for this also 💪 (If I misunderstood the naming convention, others might too.)

NlightNFotis (dismissed):

yoff (approved):
This currently has little effect due to the way the Python extractor interprets environment variables, but it will allow us to make a more standard interpretation in the future. (This is not unlikely, if we start moving to extractor options.)

Issue #2553, opened 2024-10-20
`upload-sarif post-action step failed: Input required and not supplied: token`


jsoref:
I'm about to release a new version of check-spelling, and I'm tripping on https://github.com/actions/runner/issues/3514 in this action ([`github/codeql-action/upload-sarif@v3`](https://github.com/github/codeql-action/blob/v3/upload-sarif/action.yml)).

A run of check-spelling using the current github/codeql-action/upload-sarif@v3 triggers:
https://github.com/check-spelling-sandbox/check-spelling/actions/runs/11422357281
<img width="1324" alt="image" src="https://github.com/user-attachments/assets/67e8f35d-ff13-40ce-9efa-3cc971054a81">

https://github.com/check-spelling-sandbox/check-spelling/actions/runs/11422357281/job/31780250586#step:9:1
```
Post check-spelling
Post job cleanup.
Post job cleanup.
Error: upload-sarif post-action step failed: Input required and not supplied: token
```

The error message is wrong -- the action was called with a token. But actions/runner does not reliably send inputs to actions when they run in `post`. As of today (until someone fixes https://github.com/actions/runner/issues/3514), each action must save its inputs from the `pre` or `main` step using the save-state magic and then retrieve the value from the state environment variable when running in the pre stage instead of using inputs.

hvitved:
@chrisgavin (on-call for code scanning): I believe this is one for you?

chrisgavin:
FWIW, here's where I got to before I realized there was already a PR. :sweat_smile:

https://github.com/github/codeql-action/compare/main...persist-inputs

The main thing I did differently was saving all the inputs in one go, rather than individually as they are accessed. This would prevent a case where an input is only used in the post-action step and therefore not accessed and persisted in the main action.

NlightNFotis:
Happy to take this over as I'm shield this week for our team.

@jsoref I've raised a PR with @chrisgavin 's fix and your `CHANGELOG` entry ported to it (I've maintained your authorship, let me know if you don't want this and I can drop the entry/write a new one).

Would it be possible to test our solution against your original issue and let us know if this is working as intended for you?

I slightly prefer @chrisgavin 's solution for now, but we can fall back to your original PR if there's any issue with this one instead.

jsoref:
[...]

PR #2554, opened 2024-10-20
Work around broken github/runner `input` handling for post steps

jsoref:
Address #2553 until the underlying github/runner is fixed everywhere (including GHES which is over a year from now).

- [x] Confirm this change is backwards compatible with existing workflows. -- State is supposed to be stored on a [per action basis](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#sending-values-to-the-pre-and-post-actions), so barring _other_ bugs in github/runner, this change should do the right thing.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary. -- This is really a bug fix that users shouldn't care about
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary. -- I've added an entry since it's where I'd look (other than just searching /issues/) for such a thing when I trip over it and am wondering if it's been "fixed".


jsoref:
Closing in favor of #2557

PR #2559, opened 2024-10-21
Add support for diff-informed queries

cklin:


src/actions-util.ts
```diff
@@ -161,6 +162,147 @@ export const determineBaseBranchHeadCommitOid = async function (
   }
 };
 
+/**
+ * Deepen the git history of the given ref by one level. Errors are ignored.
+ */
+export const deepenGitHistory = async function () {
```
henrymercer, line 168: Two comments about the additions to this file in general:

- Why completely ignore errors?  Does it make sense to at least debug log them?
- Nit, optional: This PR adds a bunch of functionality that interacts with git — now might be a good time to pull this functionality out into its own file.

src/actions-util.ts
```diff
@@ -161,6 +162,147 @@ export const determineBaseBranchHeadCommitOid = async function (
...
+  try {
+    await runGitCommand(
+      getOptionalInput("checkout_path"),
```
henrymercer, line 171: It might be worth noting that this function (and others) only work when called from `analyze` or `upload-sarif`.

src/actions-util.ts
```diff
@@ -161,6 +162,147 @@ export const determineBaseBranchHeadCommitOid = async function (
...
+  fromRef: string,
+  toRef: string,
+): Promise<string[] | undefined> {
+  let stdout = "";
+  try {
+    stdout = await runGitCommand(
+      getOptionalInput("checkout_path"),
+      [
+        "-c",
+        "core.quotePath=false",
```
henrymercer, line 231: Could you explain the difference between this approach and using the [`-z` flag](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--z)?

src/analyze-action.ts
```diff
@@ -260,6 +262,18 @@ async function run() {
...
+    const diffRangePackDir =
+      pull_request === undefined
+        ? undefined
+        : await setupDiffInformedQueryRun(
+            pull_request.base.ref as string,
+            pull_request.head.ref as string,
+            codeql,
+            logger,
+            features,
+          );
```
henrymercer, line 275: Nit: we could abbreviate this to `pull_request && (await ...)`.  Though generally I'm of the view that the linter should have these opinions, so feel free to leave as is.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+ *
+ * @param baseRef The base branch name, used for calculating the diff range.
+ * @param headRef The head branch name, used for calculating the diff range.
+ * @param logger
+ * @returns An array of tuples, where each tuple contains the absolute path of a
+ * file, the start line and the end line (both 1-based and inclusive) of an
+ * added or modified range in that file. Returns `undefined` if the action was
+ * not triggered by a pull request or if there was an error.
+ */
+async function getPullRequestEditedDiffRanges(
```
henrymercer, line 281: Do any of these steps take a long time?  If so, we could consider adding some more logging to explain what's happening.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+  baseRef: string,
+  headRef: string,
+  codeql: CodeQL,
+  logger: Logger,
+  features: FeatureEnablement,
+): Promise<string | undefined> {
+  if (!(await features.getValue(Feature.DiffInformedQueries, codeql))) {
+    return undefined;
+  }
+  logger.startGroup("Generating diff range extension pack");
```
henrymercer, line 259: You can use `withGroup` if you like.  That ensures the group is ended properly if there's an exception, and would simplify the code slightly.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+ * @returns An array of tuples, where each tuple contains the absolute path of a
+ * file, the start line and the end line (both 1-based and inclusive) of an
+ * added or modified range in that file. Returns `undefined` if the action was
+ * not triggered by a pull request or if there was an error.
+ */
+async function getPullRequestEditedDiffRanges(
+  baseRef: string,
+  headRef: string,
+  logger: Logger,
+): Promise<Array<[string, number, number]> | undefined> {
```
henrymercer, line 285: I would personally prefer an object here like `{ path: string, startLine: number; endLine: number; }`, just so it's always clear what these values mean.  I don't think there should be a significant performance penalty to using an object rather than an array here.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+  await actionsUtil.gitFetch(baseRef, ["--depth=1"]);
+  // Step 3: Fetch the PR topic branch history, stopping when we reach commits
+  // that are reachable from the base branch head.
+  await actionsUtil.gitFetch(headRef, [`--shallow-exclude=${baseRef}`]);
+  // Step 4: Fetch the base branch history, stopping when we reach commits that
+  // are reachable from the PR topic branch head.
+  await actionsUtil.gitFetch(baseRef, [`--shallow-exclude=${headRef}`]);
+  // Step 5: Deepen the history so that we have the merge bases between the base
+  // branch and the PR topic branch.
+  await actionsUtil.deepenGitHistory();
```
henrymercer, line 314: This might not be important yet, so feel free to leave this, but just putting it out there: can we parallelize any of these steps, e.g. fetching some refs at the same time?  I don't suppose we want parallel invocations of `git`, but perhaps we can fetch more things in the same `git` invocation.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+
+  // To compute the exact same diff as GitHub would compute for the PR, we need
+  // to use the same merge base as GitHub. That is easy to do if there is only
+  // one merge base, which is by far the most common case. If there are multiple
+  // merge bases, we stop without producing a diff range.
+  const mergeBases = await actionsUtil.getAllGitMergeBases([baseRef, headRef]);
+  logger.info(`Merge bases: ${mergeBases.join(", ")}`);
+  if (mergeBases.length !== 1) {
+    return undefined;
+  }
```
henrymercer, line 324: This might be a good place for a debug log.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+    headRef,
+  );
+  if (diffHunkHeaders === undefined) {
+    return undefined;
+  }
+
+  const results = new Array<[string, number, number]>();
+
+  let changedFile = "";
+  for (const line of diffHunkHeaders) {
```
henrymercer, line 337: In general this seems to keep going when it sees certain kinds of unexpected items in the diff.  It seems that there's a risk of missing changed files — should we be more robust here, or fail more eagerly, or is that not important for now?

src/actions-util.ts
```diff
@@ -161,6 +162,147 @@ export const determineBaseBranchHeadCommitOid = async function (
   }
 };
 
+/**
```
henrymercer, line 165: Nit: same comment about considering moving this bunch of self-contained functionality to a new file.

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+  - addsTo:
+      pack: codeql/util
+      extensible: restrictAlertsTo
+    data:
+`;
+
+  let data = ranges
+    .map((range) => {
+      return `      - ["${range[0]}", ${range[1]}, ${range[2]}]\n`;
+    })
```
henrymercer, line 413: Nit: again, linter should be catching this, but I think we could simplify this:

```suggestion
    .map((range) => `      - ["${range[0]}", ${range[1]}, ${range[2]}]\n`)
```

src/analyze.ts
```diff
@@ -234,21 +235,225 @@ async function finalizeDatabaseCreation(
...
+    // no edited lines would exclude (instead of accepting) all alerts.
+    data = '      - ["", 0, 0]\n';
+  }
+
+  const extensionContents = header + data;
+  const extensionFilePath = path.join(diffRangeDir, "pr-diff-range.yml");
+  fs.writeFileSync(extensionFilePath, extensionContents);
+  logger.info(
+    `Wrote pr-diff-range extension pack to ${extensionFilePath}:\n${extensionContents}`,
+  );
```
henrymercer, line 426: I don't see a limit on the size of the extension contents — this could be a very long log line.  I think it's fine to leave this in while we're testing this internally, but we should revisit before turning it on more broadly.

henrymercer (commented):
Thanks!  I have a bunch of comments and questions — feel free to only address the ones that make sense.

src/actions-util.ts
cklin: The error is indeed logged centrally within `runGitCommand()`.  By "Errors are ignored", I was trying to convey that errors are not reported back to the caller.  I have update the function documentation to clarify the behavior.

Moving the git interactions to its own file sounds like a good idea, though trying to incorporate that into this PR is a bit messy.  Can I defer that to a dedicated PR after this one merges?

src/actions-util.ts
cklin: I added a note to the function documentation, here and for other git functions.

src/actions-util.ts
cklin: The `-z` flag affects only specific modes: `--raw`, `--numstat`, `--name-only` and `--name-status`.  Since we need the actual diff line ranges (as opposed to only the file-level information), we cannot use any of these modes, and as a result specifying `-z` here has no effect.

src/analyze-action.ts
cklin: Thanks for the suggestion.  Done.

src/analyze.ts
cklin: In my experience they tend to be very fast, though I have tried it on only small repositories.  Once we are in staff ship we will have a better idea about performance in more realistic scenarios.

src/analyze.ts
cklin: Thanks for the suggestion.  Done.

src/analyze.ts
cklin: Thanks for the suggestion.  Done.

src/analyze.ts
cklin: I thought about it but cannot see an opportunity.  Steps 3 and 4 have no dependencies and can in principle be done in parallel.  But they require different `--shadow-exclude` flags, which I think prevents them from being combined in the same command.

src/analyze.ts
cklin: Thanks for the suggestion.  Done.

src/analyze.ts
cklin: Can you tell me more about the risk you see, and how the code can be made more robust?

src/actions-util.ts
cklin: Please see response to your earlier mention of the same suggestion.

src/analyze.ts
cklin: Thanks for the suggestion.  Done.

src/analyze.ts
cklin: I changed the logging level to `debug`, so that even though it can still be a long log line, at least it is confined to debug logging runs.

cklin:
> Thanks! I have a bunch of comments and questions — feel free to only address the ones that make sense.

Thanks for the detailed comments and helpful suggestions!  PTAL.

src/actions-util.ts
henrymercer: Yes, that's fine!

src/analyze.ts
henrymercer: Here are some examples where we seem to find unexpected items in the diff:

- `// Not an actual file path; probably /dev/null`
- `Failed to parse diff hunk header line: ${line}. Skipping.`

In these cases we seem to be encountering something unexpected, but we log or ignore the error and carry on.  My question is could these cases where the diff isn't quite as we expect cause us to miss some files that have been changed by the PR, and therefore cause us to miss alerts?  If so, we'd probably want to return `undefined` and fall back to a full analysis.

henrymercer (dismissed):
There's one remaining thread [here](https://github.com/github/codeql-action/pull/2559#discussion_r1812447496), but otherwise LGTM!

src/analyze.ts
cklin: That is a great observation.  Thank you!  I have updated the PR accordingly.

cklin:
> There's one remaining thread [here](https://github.com/github/codeql-action/pull/2559#discussion_r1812447496), but otherwise LGTM!

I updated the PR, which implements your suggestion and also fixes a minor bug in the handling of deletion hunks (where git reports the insertion line range as 0).  See [diff for the latest force-push](https://github.com/github/codeql-action/compare/38487cda349ce47af6dfe796af0740a1545662e4..94b5d396bc9fdecd0470274a36585a3224b4267f) for details.

henrymercer (approved):
Thanks @cklin!

nickfyson (approved):

PR #2564, opened 2024-10-22
Bundle installation performance improvements

henrymercer:
Make some performance improvements to bundle installation:

- Only stream the download and extraction of bundles on Linux.
- Only pass the archive to `tar` via stdin when we're streaming the download and extraction.
- Bump the high water mark when streaming the download and extraction of the bundle to 16 MB to hopefully reduce context switching.


aeisenberg (approved):
Seems reasonable. Do you have an idea how much time is being saved?

henrymercer:
Good question:

- Passing the archive via stdin seems to slow things down on macOS and Windows — on macOS by about 5s (~25%), on Windows it is difficult to get a precise measurement since we only use zstd when explicitly specified.
- Streaming on macOS (~20s) is faster than passing the archive via stdin (~28s), but both methods have bad p95 performance (several minutes), so it is more reliable to download in full and pass the file to tar (~23s).
- The high water mark I didn't observe anything over a small number of testing runs, but will see if there's a difference at larger scale.

PR #2565, opened 2024-10-23
Add Feature.CppBuildModeNone

calumgrant:
Part of https://github.com/github/codeql-core/issues/4624

### Merge / deployment checklist

- [ ] Confirm this change is backwards compatible with existing workflows.
- [ ] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [ ] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #2568, opened 2024-10-24
C++: Handle codeql_action_cpp_build_mode_none feature flag

calumgrant:

Continues https://github.com/github/codeql-action/pull/2565, tidying up the feature flag options and setting the environment variables.


src/init-action.ts
```diff
@@ -546,6 +546,20 @@ async function run() {
...
+    if (config.languages.includes(Language.cpp)) {
+      const bmn_var = "CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE";
+      if (process.env[bmn_var]) {
+        logger.info("CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE is already set");
+      } else if (await features.getValue(Feature.CppBuildModeNone, codeql)) {
+        logger.info("Enabling C++ build-mode: none");
+        core.exportVariable(bmn_var, "true");
+      } else {
+        logger.info("Disabling C++ build-mode: none");
+        core.exportVariable(bmn_var, "false");
```
aeisenberg, line 559: I assume the reason why you are doing this is because the extractor requires this same env var to be set?

src/init-action.ts
```diff
@@ -546,6 +546,20 @@ async function run() {
       }
     }
 
+    // Set CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE
+    if (config.languages.includes(Language.cpp)) {
+      const bmn_var = "CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE";
```
aeisenberg, line 551: This is probably a linter error. Here's a more canonical name.

```suggestion
      const bmnVar = "CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE";
```

aeisenberg: Is this simpler? `features.getValue` is guaranteed to return a boolean and if the env var is already set, it will use that as an override.

```js
const bmnVar = await features.getValue(Feature.CppBuildModeNone, codeql);
core.exportVariable(bmnVar);
logger.info(`Setting C++ build-mode: none to ${bmnVar}`);
```

src/init-action.ts
calumgrant: It's not required, just setting it for completeness.

src/init-action.ts
```diff
@@ -546,6 +546,16 @@ async function run() {
       }
     }
 
+    // Set CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE
+    if (config.languages.includes(Language.cpp)) {
+      const bmnVar = "CODEQL_EXTRACTOR_CPP_BUILD_MODE_NONE";
```
henrymercer, line 551: If you think we might want users to interact with this option, we should make this an extractor option by changing the name to `CODEQL_EXTRACTOR_CPP_OPTION_BUILD_MODE_NONE` and including an entry in `codeql-extractor.yml`.  That gives customers more options for configuring it.  If this is only going to be used by the Action, then we can leave it as is.

henrymercer (approved):

src/init-action.ts
calumgrant: This is internal and short-lived, so I don't think we should give customers the option of configuring it.

PR #2570, opened 2024-10-28
 Improve robustness of extracting the bundle tag name

henrymercer:
Previously, we'd extract the wrong tag for a tools URL like `https://github.com/org/codeql-bundle-testing/releases/download/codeql-bundle-v2.19.0/codeql-bundle-linux64.tar.zst`.  This approach fixes that case while avoiding assuming too much about the structure of the tools URL.


dbartol (approved):
LGTM

PR #2571, opened 2024-10-29
Add the action stub

aeisenberg:
This is required for the migration of the codeql-action to an immutable action.

We can delete this after the migration is done, or just keep it so people have an extra way to understand how to use the action.


action.yml
```diff
@@ -0,0 +1,11 @@
+name: 'CodeQL: Stub'
+description: "Stub: Don\'t use this action directly. Read [the documentation](https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning-with-codeql) instead."
```
henrymercer, line 2:
```suggestion
description: "Stub: Don't use this action directly. Read [the documentation](https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning-with-codeql) instead."
```

henrymercer (dismissed):
Have we run the tool before on repos containing multiple Actions?  If not, it might be worth doing a dry run on a fork.

aeisenberg:
> Have we run the tool before on repos containing multiple Actions? If not, it might be worth doing a dry run on a fork.

@Jcambass has already mentioned this should not be a problem.

henrymercer (approved):

PR #2572, opened 2024-11-01
Add support for `actions` language

dbartol:
This PR adds support for the new `actions` CodeQL language, which will be part of the upcoming 2.19.3 release of the CodeQL bundle. In addition to adding `actions` to the list of supported languages, this PR also fixes a couple of issues I encountered while testing the new language:

The `CODEQL_ACTION_EXTRA_OPTIONS` environment variable, used for setting additional CLI options, now supports YAML, rather than just JSON. This makes it easier to specify complex options in a workflow file without complicated escaping.

I added support for the `kind` property for pack registries, to allow downloading packs from GitHub repo-based registries, in addition to the existing Docker-based registries.


src/config-utils.ts
```diff
@@ -64,6 +64,9 @@ export interface RegistryConfigNoCredentials {
 
   // List of globs that determine which packs are associated with this registry.
   packages: string[] | string;
+
+  // Kind of registry, either "github" or "docker". Default is "docker".
+  kind?: "github" | "docker";
```
henrymercer, line 69: I wonder if folks might confuse "docker" with docker hub when I think we're talking about GHCR.  We could say something like "repository" / "container"?  I'm not sure what's best.

src/languages.ts
```diff
@@ -1,5 +1,6 @@
 // All the languages supported by CodeQL
 export enum Language {
+  actions = "actions",
```
henrymercer, line 3: Thought: is it worth adding a warning in the `init` Action when using this language that `actions` isn't officially supported yet?

henrymercer (approved):

src/config-utils.ts
dbartol: The `docker` value is already baked into the CLI. We could consider changing it there (adding `ghcr` or whatever as an alias), but I don't think it's worth the effort. `docker` is the default, so nobody ever has to actually specify it.

src/languages.ts
dbartol: They'll already have to enable the experimental language support in the CLI by setting `CODEQL_ENABLE_EXPERIMENTAL_FEATURES=true`, so I think they'll already know it's not officially supported yet.

src/config-utils.ts
aeisenberg: Maybe just adding a comment to say that the only docker registry we support is GHCR. This is public code and I wouldn't want to give any false impressions.

src/config-utils.ts
```diff
@@ -64,6 +64,9 @@ export interface RegistryConfigNoCredentials {
 
   // List of globs that determine which packs are associated with this registry.
   packages: string[] | string;
+
+  // Kind of registry, either "github" or "docker". Default is "docker".
```
aeisenberg, line 68:
```suggestion
  // Kind of registry, either "github" or "docker". Default is "docker".
  // The only docker registry that codeql supports is the GitHub Container Registry.
```

src/config-utils.ts
dbartol: Added a comment

src/config-utils.ts
dbartol: Sorry, didn't see this suggestion until I'd already added a different comment.

aeisenberg: I think what you have now is fine. Thanks for updating.

aeisenberg (approved):

PR #2573, opened 2024-11-01
Download Zstandard bundles on GHES too by default

henrymercer:
We're seeing an average 2x speed up by using bundles compressed using Zstandard.  Now that this feature has stabilised, this PR rolls it out to GHES, cleans up the feature flags, and removes a mechanism to fall back to gzip that has not been needed.


aeisenberg (approved):
With this PR, users will no longer be able to control if they use zstd or tar. I think that's fine since it is not relevant for them if everything works anyway.

henrymercer:
Yeah agreed.  Customers can still specify a tools URL if they want to use gzip or zstd.  In general I think we want to have flexibility to change the default way the tools are downloaded and installed, e.g. so we can move to per-language bundles.

PR #2582, opened 2024-11-07
Temporarily include the Actions extractor with the Action

dbartol:
This PR embeds the new experimental `actions` extractor to the CodeQL Action repo itself, and updates the `init` action to add that directory to `--search-path` if the `actions` language is enabled.

This is a hack to make the `actions` extractor available without it actually being in the bundle. This is needed for some internal testing over the next couple of weeks, after which the extractor will be in the bundle and we can remove this hack.

The extractor sources embedded here are just the current version copied directly from `github/codeql`.

I've already verified that this works correctly on a test repo with Advanced Setup.


aeisenberg:
This looks reasonable. Can you share a link to the run in the test repo?

henrymercer (approved):
This seems like a good approach to me given the size of the Actions extractor.

The internal test run looks good, so approving.

dbartol:
Successful run in `semmle-code`: https://github.com/github/semmle-code/actions/runs/11731165575/job/32680646659

PR #2583, opened 2024-11-08
Upgrade workflows to `actions/upload-artifact@v4`

jsoref:
- Split from #2578 per https://github.com/github/codeql-action/pull/2578#issuecomment-2463427214

- [x] Confirm this change is backwards compatible with existing workflows. -- This change is *not* backwards compatible with existing workflows (if they use the debug artifact feature which I've never seen outside of this repository)

### DIFF
--- a/.github/workflows/__config-export.yml
+++ b/.github/workflows/__config-export.yml
@@ -70,7 +70,7 @@ jobs:
           output: ${{ runner.temp }}/results
           upload-database: false
       - name: Upload SARIF
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: config-export-${{ matrix.os }}-${{ matrix.version }}.sarif.json
           path: ${{ runner.temp }}/results/javascript.sarif
--- a/.github/workflows/__diagnostics-export.yml
+++ b/.github/workflows/__diagnostics-export.yml
@@ -82,7 +82,7 @@ jobs:
           output: ${{ runner.temp }}/results
           upload-database: false
       - name: Upload SARIF
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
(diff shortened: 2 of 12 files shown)


jsoref:
Note: I've left a file that tests the old v3 behavior. It's clearly called legacy, and I saw no particular reason to remove it yet as I'm not removing the code that it validates and someone who has access to GHES could conceivably want to test it.

This PR gave me enough insight that I could probably write some code to let that workflow skip doing real work unless it was running under GHES, although the lazy test would be to check the server URL as I did in my initial work.

angelapwen:
👋 thanks again for splitting this out @jsoref!

The default value for the feature flag is only used when there is no corresponding feature flag value from the GitHub feature flag API: https://github.com/github/codeql-action/blob/main/src/feature-flags.ts#L276-L293 so the default value change won't actually enable the behavior here. 

I've begun rolling out the feature flag to `true` (currently at 10% of repos) and once that rollout is complete, we can set the default value to `true`. We prefer to roll out with the GitHub internal feature flag API as we have a mechanism that allows us to roll back for specific repositories or organizations. So let's hold the default value change until the rollout via the internal flag API is complete.

Note that the PR checks you are updating can be changed to `@v4` at any time: they are uploading SARIF rather than our debug artifacts so are unaffected by the feature flag. 

(The PR checks related to debug artifacts, https://github.com/github/codeql-action/blob/main/.github/workflows/debug-artifacts-failure.yml using v4 and https://github.com/github/codeql-action/blob/main/.github/workflows/debug-artifacts-legacy.yml using v3 are unaffected by the default value/feature flag API value because I have set the `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable in those workflows.)

jsoref:
Sigh. The naming is confusing. I'll split out just the workflow upload artifact version changes to another PR.

angelapwen (approved):
Thank you for the contributions, as always 😄

PR #2597, opened 2024-11-12
Dependency caching: small output improvements

mbg:
This PR makes two small changes related to dependency caching to improve the Action's output during normal operation:

- We intercept `ReserveCacheError` errors when saving caches, since these are expected when a cache already exists for a given cache key. We log a message indicating this instead of propagating the error itself.
- We allow warnings from `tryGetFolderBytes` to be suppressed, which would otherwise result in check annotations for a workflow run. This occurs during normal operation, since we try to calculate the cache size for all folders that may need to be cached for a given language, whether they exist or not. If they do not exist, that's not a problem for us.


aeisenberg (approved):
Nice.

PR #2599, opened 2024-11-12
workaround: disable streaming when downloading codeql bundle

### LINKED ISSUE #2593: Enabling zstd bundles for GHES seems to have broken proxy support on Enterprise Cloud with self-hosted runners
It looks like a recent change to enable zstd bundles on GHES has broken CodeQL scanning for anyone on Enterprise Cloud with self-hosted runners that are behind a proxy that is configured by the normal `https_proxy` type system variables.

I did some investigation to understand why, and it appears that in the commit https://github.com/github/codeql-action/commit/33f2dc57a419743d4e5f95dab4045a64d0d76809 the code path used now always uses the streaming version as the feature flag that previously prevented it has been removed.

The main problem seems to be that the change from v3.27.0 to v3.27.1 has switched http libraries and broken the previous proxy support.

Specifically, in the working v3.27.0 version, the feature flag checked here:
https://github.com/github/codeql-action/blob/v3.27.0/lib/tools-download.js#L59
wasn't triggered and therefore it used this code path:
https://github.com/github/codeql-action/blob/v3.27.0/lib/tools-download.js#L73
which downloads via the `tool-cache`:
https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts
and this respects the system proxy variables as you can see in:
https://github.com/actions/toolkit/blob/main/packages/http-client/src/proxy.ts

Unfortunately, now the feature flag is removed in v3.27.1:
https://github.com/github/codeql-action/blob/v3.27.1/lib/tools-download.js#L62
it always uses the streaming method, but this does not use the [toolkit http-client](https://github.com/actions/toolkit/tree/main/packages/http-client) that includes system proxy support but instead appears to use the `follow_directs` library here:
https://github.com/github/codeql-action/blob/v3.27.1/lib/tools-download.js#L62
which appears to be a small wrapper over the basic node fetch functionality and does not support proxies as far as I can make out.

This ultimately has meant that we've had to revert to v3.27.0 functionality to get CodeQL scanning in our Enterprise Cloud self-hosted runners working again.

As a short term fix, is it possible this feature flag can be brought back until proxy support has been added to the new streaming method?

NlightNFotis:
## Description

This PR disables streaming the extraction of the CodeQL bundle as a temporary workaround to the issue described in https://github.com/github/codeql-action/issues/2593.

A proper fix will be implemented later so that we can stream while respecting proxy settings.

Fixes #2593


henrymercer (commented):
Can you reintroduce the `CODEQL_ACTION_ZSTD_BUNDLE_STREAMING_EXTRACTION` environment variable so the PR check `pr-checks/checks/zstd-bundle-streaming.yml` will continue to test that streaming works?  Otherwise LGTM.

src/tools-download.ts
```diff
@@ -87,7 +87,9 @@ export async function downloadAndExtract(
 
   const compressionMethod = tar.inferCompressionMethod(codeqlURL);
 
-  if (compressionMethod === "zstd" && process.platform === "linux") {
+  // TODO: Re-enable streaming when we have a more reliable way to respect proxy settings.
+  // eslint-disable-next-line no-constant-condition, no-constant-binary-expression
+  if (false && compressionMethod === "zstd" && process.platform === "linux") {
```
aeisenberg, line 92: Can you turn `false` into an environment variable? This way at least we can selectively enable it.

Maybe something like:

```suggestion
  if (process.env[EnvVar.STREAM_BUNDLE_DOWNLOAD] !== "true" && compressionMethod === "zstd" && process.platform === "linux") {
```

And then add the variable to the `EnvVar` class?

NlightNFotis:
@aeisenberg @henrymercer , thank you both for your reviews. I've implemented Henry's request which I found to be in the intersection of what both of you suggested. Would it be possible to have another look at this?

src/codeql.test.ts
```diff
@@ -39,6 +41,15 @@ setupTests(test);
 
 let stubConfig: Config;
 
+// TODO: Remove when when we no longer need to pass in features
```
aeisenberg, line 44: Can you link to the issue for this? I think it's this one: https://github.com/github/codeql-action/issues/2600

src/setup-codeql.test.ts
```diff
@@ -25,6 +27,14 @@ import {
 
 setupTests(test);
 
+// TODO: Remove when when we no longer need to pass in features
```
aeisenberg, line 30: Same...can you link to https://github.com/github/codeql-action/issues/2600?

aeisenberg (approved):
Looks good after you can update the comments to link to the appropriate issue that tracks the TODO.

src/setup-codeql.test.ts
```diff
@@ -25,6 +27,14 @@ import {
...
+const expectedFeatureEnablement: FeatureEnablement = initializeFeatures(
```
henrymercer, line 31: Nit: update `initializeFeatures` to return `FeatureEnablement` so we can remove this cast.

henrymercer (dismissed):

NlightNFotis:
@henrymercer @aeisenberg sorry for this, can you please re-approve, because I've added the link that Andrew asked for and it automatically dismissed the reviews.

henrymercer (approved):

PR #2611, opened 2024-11-18
Throw configuration error when `tar` is not available

angelapwen:
We recently had an internal monitor trigger when `tar` was not found on the PATH of a self-hosted runner. This change checks whether `tar` is available on the PATH and throws a configuration error when it is not. Small refactoring: I moved the `isBinaryAccessible` method to `util` now that it is used in multiple places.


aeisenberg (approved):
Nice.

Issue #2600, opened 2024-11-12
Enable streaming while respecting proxy settings


NlightNFotis:
This is for the followup work to what's done in https://github.com/github/codeql-action/pull/2599 so that we provide a longer term fix for the issue described in https://github.com/github/codeql-action/issues/2593.

In https://github.com/github/codeql-action/pull/2599 we opted to deactivate the path that was using streaming, but the longer term, more robust fix, is to use streaming while detecting and respecting proxy settings that the client has setup.

PR #2624, opened 2024-11-28
Respect proxy settings when streaming the download and extraction of the CodeQL bundle

NlightNFotis:
## Description

This PR is adding support for respecting proxy settings while downloading the CodeQL bundle in streaming mode.

Fixes https://github.com/github/codeql-action/issues/2600

## Guidance

* The code changes required to support the feature have been added in 0dc76a996d2e7b4c024bc76368ecb70c76bb62ed.
* It's introducing a `justfile`, which is a script interpreted by https://github.com/casey/just, a command action runner. It allows us to create an python virtual environment and run the `sync.py` to update dependencies in one step, which is very useful for someone trying to interact with it the first time.
  * If `just` is not present, the file is irrelevant (it's just a text file) but can act as documentation for the steps to invoke manually.
* I have initially tested this works locally, by setting `tinyproxy` and writing a function that invoked `downloadAndExtractZstdWithStreaming` and observing that we go through the proxy.
  * This function was subsequently thrown away after the first tests locally validated the design.
* For CI testing, I have made changes to the `test_proxy.yml` workflow, so that we download the latest nightly bundle, which by default has the extension `.tar.zst` which goes through the code path introduced.
  * This can be seen working [here](https://github.com/github/codeql-action/actions/runs/12184584143/job/33988811971#step:7:35) (latest CI run at the time of writing).
* This required some changes to the `sync.py` file in 9c48c8bf180ad9a20f3a2322dfdf1abe591e177b in order to support a step that is being used to initialise the container image (because the Ubuntu container is leaned out and doesn't come with the list of pre-installed software the runners usually come prepared with).
* The diff (~31k lines) is mostly comprised of a couple of packages added in 8f2cb3a9314157cca05deb1ddf96f0aaff71ab94. The actual code changes introduced by this PR are in the other commits, and are much smaller in terms of loc changed. 


src/tools-download.ts
```diff
@@ -89,41 +89,42 @@ export async function downloadAndExtract(
 
   const compressionMethod = tar.inferCompressionMethod(codeqlURL);
 
-  // TODO: Re-enable streaming when we have a more reliable way to respect proxy settings.
-
-  if (
-    (await features.getValue(Feature.ZstdBundleStreamingExtraction)) &&
```
aeisenberg, line 95: I wonder if we should keep this flag for now just in case there's another problem with streaming.

NlightNFotis:
The streaming of the bundle using the Proxy worked fine in the last CI run: https://github.com/github/codeql-action/actions/runs/12146133664/job/33869336618?pr=2624#step:7:33

I'm now in the process of tidying this up (lots of experimental commits), and incorporating conflicting changes from `main`.

pr-checks/test-proxy.patch
aeisenberg, line 1: Is there a way to change sync.py so that you don't need to patch after running it?

NlightNFotis: Yes, this is the last thing I need to do here - I will first have a look if I can have a lighter touch action to move this forward (perhaps move initialisation into a `Dockerfile` and use that to initialise the proxy actions), otherwise even if I need to change `sync.py` I think the changes needed are not large - I just need to find a way to customise insertion order of some steps, so I can create some `before_steps` or `after_steps` sections (might not reflect the current symmetry with the action workflow, but I think the added flexibility might be worth it).

I will add a comment outlining the approach I took as soon as I finalise it.

src/tools-download.ts
NlightNFotis: I'm confident we are good without a flag to manage the risk.

To explain why: because the conflicts introduced by a couple of other PRs were extensive, and I couldn't sort out a clean commit history, I opted to instead just manually recreate the PR on top of `main` (my code changes are small, so easy to manually replay).

In doing so, I made a mistake: I initialised the HTTP client (the one that respects the proxy settings) and passed it to the _headers_ by accident (instead of the `http.get` options). What happened is [here](https://github.com/github/codeql-action/actions/runs/12166081189/job/33931523101#step:7:35), but short story, because of the streaming download now being inside of a fallback `try`/`catch`, the failure to download the CodeQL pack with streaming resulted in falling back to downloading it *and then* extracting it.

I've since fixed this, but this little mistake validated the fallback mechanism working as intended, so I _believe_ that this won't cause an incident like last time - if it fails, worst case scenario is that the mechanism won't work as expected and fall back to the previous mechanism (download and then extract), with the impact being that we lose a few seconds per run.

Any thoughts on this?

aeisenberg: That makes sense to me. I'll leave it to your judgment.

pr-checks/test-proxy.patch
NlightNFotis: Okay, I've tried a few approaches and the simples one was to just add a `container-init-steps` property in the yaml file, which is then interpreted by `sync.py`.

Using a dockerfile for a custom image proved to be more of a hassle than I thought. We don't control it directly, but instead need to have a JSON specially crafted, that contains a reference to the `Dockerfile`, which is then interpreted by `actions` when starting to build the image (or run steps) (relevant docs: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/customizing-the-containers-used-by-jobs#about-container-customization)

NlightNFotis:
Something that I forgot to mention: this PR is **not** removing the feature flag: there are 2 references in the code (after this PR gets merged) which are very easy to remove and do not impact functionality.

There are also two more references in a couple of test actions, which would require a bit more thought as to how to handle (we could just remove the environment variable reference in them, or just delete them given that this is more or less tested in the new `test_proxy` workflow).

I wanted to do any changes to the feature flag/env var as a different PR, and keep the changes here scoped down to the proxy settings and their handling by the streaming download handler.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 ## [UNRELEASED]
 
 - We are rolling out a change in December 2024 that will extract the CodeQL bundle directly to the toolcache to improve performance. [#2631](https://github.com/github/codeql-action/pull/2631)
+- We have added support for respecting the proxy settings present in environment variables in a runner when streaming the download and extraction of the CodeQL bundle. [#2624](https://github.com/github/codeql-action/pull/2624)
```
henrymercer, line 10: I'd suggest we phrase this as a bug, since streaming should have respected proxy settings from the get go:

```suggestion
- Fixed an issue where streaming the download and extraction of the CodeQL bundle did not respect proxy settings. [#2624](https://github.com/github/codeql-action/pull/2624)
```

pr-checks/checks/test-proxy.yml
```diff
@@ -1,10 +1,20 @@
...
 description: "Tests using a proxy specified by the https_proxy environment variable"
-versions: ["linked"]
+versions: ["linked", "nightly-latest"]
 operatingSystems: ["ubuntu"]
 container:
   image: ubuntu:22.04
-  options: --dns 127.0.0.1
+  options: --dns 8.8.8.8
+container-init-steps:
+  name: Set up GitHub CLI
```
henrymercer, line 9: How come we didn't need to do this before, but we need to now?

pr-checks/checks/test-proxy.yml
```diff
@@ -1,10 +1,20 @@
 name: "Proxy test"
 description: "Tests using a proxy specified by the https_proxy environment variable"
-versions: ["linked"]
+versions: ["linked", "nightly-latest"]
 operatingSystems: ["ubuntu"]
 container:
   image: ubuntu:22.04
-  options: --dns 127.0.0.1
+  options: --dns 8.8.8.8
```
henrymercer, line 7: Why are we making this change?

pr-checks/checks/test-proxy.yml
```diff
@@ -1,10 +1,20 @@
 name: "Proxy test"
 description: "Tests using a proxy specified by the https_proxy environment variable"
-versions: ["linked"]
+versions: ["linked", "nightly-latest"]
```
henrymercer, line 3: This test was passing before when we were streaming zstd bundles with a CodeQL Action version that wasn't respecting proxy settings, presumably because the CodeQL Action could still access the bundle without the proxy.

Can we update the test to ensure that we're downloading the bundle via the proxy?

src/tools-download.ts
```diff
@@ -94,39 +94,40 @@ export async function downloadAndExtract(
...
+
+      return {
+        compressionMethod,
+        toolsUrl: sanitizeUrlForStatusReport(codeqlURL),
+        ...makeStreamedToolsDownloadDurations(combinedDurationMs),
+      };
+    }
+  } catch (e) {
+    core.warning(
+      `Failed to download and extract CodeQL bundle using streaming. Falling back to downloading the bundle before extracting.`,
```
henrymercer, line 128: Have you tested this fallback?  For instance, what happens if we partially extract the CodeQL bundle while streaming, but something goes wrong?  Is it worth cleaning up the destination directory in this case or is that handled elsewhere?

henrymercer (commented):
Thanks for this fix!  I've added a couple of questions / suggestions:

pr-checks/checks/test-proxy.yml
```diff
@@ -1,10 +1,20 @@
...
+container-init-steps:
+  name: Set up GitHub CLI
+  run: |
+    apt update
+    apt install -y curl libreadline8 gnupg2 software-properties-common zstd
+    curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+    apt-key add /usr/share/keyrings/githubcli-archive-keyring.gpg
+    apt-add-repository https://cli.github.com/packages
+    apt install -y gh
+  env: {}
```
aeisenberg, line 17: Is this line useful?

pr-checks/justfile
```diff
@@ -0,0 +1,6 @@
+# Perform all necessary steps to update the PR checks
+update-pr-checks:
```
aeisenberg, line 2: Presumably, this makes it easier to run `sync.py`. Can you add some instructions on how to do this? You could remove this file from the PR and move it to a new one.

CHANGELOG.md
NlightNFotis: Done.

pr-checks/checks/test-proxy.yml
NlightNFotis: The reason for that was intriguing and surprising for me to discover:

In the `prepare-test` action, we would pass in `inputs.version` as `linked`, which would result in `tools-url=linked` being written to the `$GITHUB_OUTPUT` variable _and then exit_.

When we passed in `nightly-latest` in this PR, it had the surprising effect of following a different path inside the same file - one that would result in [executing `gh release list`](https://github.com/github/codeql-action/blob/main/.github/actions/prepare-test/action.yml#L63) later.

This all would be fine, but these steps are running in an `Ubuntu 22.04` container, which has been crafted to be as leaned out as possible - hence, it doesn't come prepackaged with a lot of the usual goodies the runner images come with.

This all means that if we *don't* install the GitHub CLI before that step containing the action `prepare-test` runs, the action will fail with the error message being `failed to locate gh` (I paraphrase, but the error is failing to find the binary).

The container also lacks a whole swath of other stuff we also need to install to be able to add the repository from which to fetch the `gh` CLI from: `sudo`, `curl`, etc.

The step added takes care of that, before the other steps are run.

src/tools-download.ts
NlightNFotis: Yes, I'm quoting from a previous message of mine that probably got lost because of the volume of messages here:

> I'm confident we are good without a flag to manage the risk.

> To explain why: because the conflicts introduced by a couple of other PRs were extensive, and I couldn't sort out a clean commit history, I opted to instead just manually recreate the PR on top of main (my code changes are small, so easy to manually replay).

> In doing so, I made a mistake: I initialised the HTTP client (the one that respects the proxy settings) and passed it to the headers by accident (instead of the http.get options). What happened is [here](https://github.com/github/codeql-action/actions/runs/12166081189/job/33931523101#step:7:35), but short story, because of the streaming download now being inside of a fallback try/catch, the failure to download the CodeQL pack with streaming resulted in falling back to downloading it and then extracting it.

> I've since fixed this, but this little mistake validated the fallback mechanism working as intended, so I believe that this won't cause an incident like last time - if it fails, worst case scenario is that the mechanism won't work as expected and fall back to the previous mechanism (download and then extract), with the impact being that we lose a few seconds per run (compared to a failed run).

Does this answer your question satisfactorily?

pr-checks/checks/test-proxy.yml
NlightNFotis: I'm not sure what was causing it, but when the `localhost` was being used as the DNS, it failed to resolve the address for the Ubuntu sources database, which failed the step that needed to do `apt update` before downloading the missing packages for the container initialisation.

pr-checks/checks/test-proxy.yml
NlightNFotis: Yes, I'm using it to ensure that the step runs with a clean environment, shadowing the environment available to the other steps, which contain the proxy settings that cause this particular step (the container initialisation) to fail.

pr-checks/justfile
aeisenberg: This is fine as long as it is documented.

NlightNFotis: Done in [a89fbc8](https://github.com/github/codeql-action/pull/2624/commits/a89fbc80a21d76b6e2304d8ccf53ae01036b94ab)

pr-checks/checks/test-proxy.yml
henrymercer: Excellent, thank you for the detailed explanation!  Could you comment that this is necessary to use `version: nightly-latest`?

pr-checks/checks/test-proxy.yml
henrymercer: Out of interest, can we fall back to the network's default DNS settings by removing the `--dns` arguments altogether rather than hardcoding Google DNS?

src/tools-download.ts
henrymercer: Thank you, that answers the first part of my question.  I still wonder if it is worth cleaning up the destination directory as part of the `catch` in case we managed to extract some of the CodeQL bundle but not a complete CodeQL bundle.

pr-checks/checks/test-proxy.yml
NlightNFotis: We can definitely test this out quickly. I will do so now, and update with the result.

pr-checks/checks/test-proxy.yml
NlightNFotis: Yes, added documentation for that in 51e71f8.

src/tools-download.ts
NlightNFotis: I agree this is a good idea, and something prudent for us to do before we retry.

Added in [4c20d4f](https://github.com/github/codeql-action/pull/2624/commits/4c20d4f58ac2e73745e8a46eb4050bdee9e92115).

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ Note that the only difference between `v2` and `v3` of the CodeQL Action is the
 ## [UNRELEASED]
 
 - We are rolling out a change in December 2024 that will extract the CodeQL bundle directly to the toolcache to improve performance. [#2631](https://github.com/github/codeql-action/pull/2631)
-- We have added support for respecting the proxy settings present in environment variables in a runner when streaming the download and extraction of the CodeQL bundle. [#2624](https://github.com/github/codeql-action/pull/2624)
+- Fixed an issue where streaming the download and extraction of the CodeQL bundle did not respect proxy settings. [#2624](https://github.com/github/codeql-action/pull/2624)
+- Update default CodeQL bundle version to 2.20.0. [#2636](https://github.com/github/codeql-action/pull/2636)
```
aeisenberg, line 11: Was this line added due to a bad merge?

src/tools-download.ts
```diff
@@ -127,7 +127,11 @@ export async function downloadAndExtract(
     core.warning(
       `Failed to download and extract CodeQL bundle using streaming. Falling back to downloading the bundle before extracting.`,
     );
-    core.warning(getErrorMessage(e));
+    core.warning(`Error: ${getErrorMessage(e)}`);
```
aeisenberg, line 130: Don't prefix a warning with `Error`.  

```suggestion
    core.warning(`Warning: ${getErrorMessage(e)}`);
```

But, do we even need the prefix at all?

Maybe reverting this is better:

```suggestion
    core.warning(getErrorMessage(e));
```

pr-checks/readme.md
```diff
@@ -0,0 +1,23 @@
+# PR Checks
+
+This folder contains the code supporting the workflows run when a PR is created.
+
+## Update
```
aeisenberg, line 5: I think we can simplify this. I can give some suggestions, but not quite right now.

aeisenberg: See here for an alternative readme: https://github.com/github/codeql-action/compare/NlightNFotis/detect_use_proxy_when_streaming...aeisenberg/justfile-readme?quick_pull=1#diff-b90d48e45a6813fad7195d74075f05bdf05919291a02116fe8a6b67691b1266a

CHANGELOG.md
NlightNFotis: Hm, I will have a look.

This was reported as a conflict in the GitHub UI, but when I did a `git merge` locally, it had no problem automatically resolving the conflict. Perhaps it did a weird thing because of my `.gitconfig`.

src/tools-download.ts
NlightNFotis: Thank you for catching this.

Seeing it reported now, I agree it looks a bit silly.

My rationale for changing this was that this was rendering an error that was generated in the process of unpacking as a simple warning, with no other context whatsoever, which could be confusing to a user when interpreting the runner log ([example](https://github.com/github/codeql-action/actions/runs/12166081189/job/33931523101#step:7:35)).

I'm still motivated to decorate it somehow, as reverting may not satisfy this concern, but I will think about a cleaner decoration that doesn't look weird/silly.

pr-checks/readme.md
NlightNFotis: Thank you for this! I agree it looks cleaner.

Will `cherry-pick` this in.

src/tools-download.ts
NlightNFotis: Okay, this is what I came up with: [88bcf64](https://github.com/github/codeql-action/pull/2624/commits/88bcf64e02c4dffb7bded80efda37efcdb69b07a)

I've reworded the warnings to flow a bit more naturally for me, and to allow me to embed the error message in an indicative but non-invasive way.

Is this workable for you @aeisenberg ? Or do you have a strong preference for the previous state of affairs?

CHANGELOG.md
NlightNFotis: I cannot find precisely what went wrong here, as the merge log is past my terminal character limit. I remember it saying that it resolved it automatically using the `ort` method, but it apparently went wrong there.

This is now fixed in [9e8cd42](https://github.com/github/codeql-action/pull/2624/commits/9e8cd42adc33391cddeaec148d34c6cd8b1d1d65).

Going forward I will be more careful validating these merges went right.

aeisenberg (approved):
Nice.

PR #2633, opened 2024-12-03
Set `ArtifactV4Upgrade`'s default value to true

angelapwen:
`artifacts@v3` has an official deprecation date of December 5. Technically, merging this PR does not actually change much — the feature flags will first check the GitHub API for the value of the feature flag (which has been rolled out at 100% enabled for several weeks). 

In the future, we will remove the actual feature flag in GitHub's internal feature flag portal. Once this is done, the default `true` value set in this PR will be actually used in production. We don't want to yet remove the feature flag from the internal feature flag portal, as old versions of the CodeQL Action will use a default value of `false` if the flag is not found from the GItHub API, and use the deprecated version of `actions@v3`. Therefore we will only be able to remove the feature flag from the internal feature flag portal once we no longer support these older versions of the Action.

Note that GHES customers will continue to use `artifacts@v3` until it is supported on GHES. 


aeisenberg (approved):
LGTM!

PR #2634, opened 2024-12-04
Remove `artifact_v4_upgrade` feature flag logic and related PR checks

angelapwen:
I was reminded that we can set the value in the API response to be `true` for old versions of the CodeQL Action; once that change is in, this PR can be merged.

The PR removes all mention of the feature flag, so the new artifact dependency will always be used except on GHES.

The `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` environment variable is now deprecated as well so it is removed from the PR checks that included it. The `debug-artifacts-legacy` PR check, that makes sure that the deprecated legacy version of the client works, is now also removed as we no longer can set `CODEQL_ACTION_ARTIFACT_V4_UPGRADE` to false. 


aeisenberg (approved):

angelapwen:
I will merge this once the monolith PR change is deployed.

PR #2644, opened 2024-12-11
Use an app token for triggering a release

aeisenberg:
We need to do this because using a default token will not recursively trigger a new workflow run.

The only way to test this is to merge to main and do a release.

### DIFF
--- a/.github/workflows/post-release-mergeback.yml
+++ b/.github/workflows/post-release-mergeback.yml
@@ -162,11 +162,18 @@ jobs:
             --assignee "${GITHUB_ACTOR}" \
             --draft
 
+      - name: Generate token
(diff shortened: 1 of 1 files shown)


henrymercer (commented):
The indentation doesn't look quite right, but otherwise this looks good!

aeisenberg:
Fixed the indentation. After merging, I will do a release to see if this fixes the problem.

NlightNFotis (approved):
Code-wise, this looks good.

The only question is whether the permissions are adequate for performing the release, but I guess we won't know until we try it.

aeisenberg:
> the permissions are adequate for performing the release

I believe so. We need `contents: write` for this to work. And the app provides it.

PR #2655, opened 2024-12-16
Bump the minimum supported version of CodeQL to 2.15.5

aeisenberg:

### DIFF
--- a/.github/workflows/__go-indirect-tracing-workaround-diagnostic.yml
+++ b/.github/workflows/__go-indirect-tracing-workaround-diagnostic.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            version: stable-v2.14.6
+            version: default
     name: 'Go: diagnostic when Go is changed after init step'
     permissions:
       contents: read
--- a/.github/workflows/__go-indirect-tracing-workaround-no-file-program.yml
+++ b/.github/workflows/__go-indirect-tracing-workaround-no-file-program.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            version: stable-v2.14.6
+            version: default
     name: 'Go: diagnostic when `file` is not installed'
     permissions:
       contents: read
--- a/.github/workflows/__go-indirect-tracing-workaround.yml
+++ b/.github/workflows/__go-indirect-tracing-workaround.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            version: stable-v2.14.6
+            version: default
     name: 'Go: workaround for indirect tracing'
     permissions:
       contents: read
--- a/README.md
+++ b/README.md
@@ -81,9 +81,8 @@ We typically release new minor versions of the CodeQL Action and Bundle when a n
 | `v3.25.11` | `2.17.6` | Enterprise Server 3.14 | |
 | `v3.24.11` | `2.16.6` | Enterprise Server 3.13 | |
 | `v3.22.12` | `2.15.5` | Enterprise Server 3.12 | |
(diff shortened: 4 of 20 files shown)


aeisenberg:
@mbg, there are three workflows that are failing now that I'm deprecating 2.14.6. It looks like you created these workflows and hard-coded 2.14.6. Is there another, non-deprecated version we can use? Is it the case that now that 2.14.6 is deprecated, we can remove the workaround?

mbg:
I think intention with pinning that version was that we would fix/improve the tracer not long after, and so would no longer need the workaround then. However, that never happened, so the workaround is still needed to this day (AFAIK). That means we can either:

- Bump the pinned version to the current version (or any other non-deprecated one)
- Not pin it to a specific version, and then pin it to one once the tracer can deal with statically linked binaries

aeisenberg:
OK. Thanks. I'll pin it to 2.20.0.

aeisenberg:
Actually, if I pin to `default` and the workaround is implemented, then we will see a test failure more quickly and will be reminded to change or remove the tests. I changed to that.

.github/workflows/__go-indirect-tracing-workaround-no-file-program.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            version: stable-v2.14.6
+            version: default
```
zutshisunakshi, line 31: same comment as above

.github/workflows/__go-indirect-tracing-workaround.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            version: stable-v2.14.6
+            version: default
```
zutshisunakshi, line 31: same comment as above

README.md
```diff
@@ -81,9 +81,8 @@ We typically release new minor versions of the CodeQL Action and Bundle when a n
 | `v3.25.11` | `2.17.6` | Enterprise Server 3.14 | |
 | `v3.24.11` | `2.16.6` | Enterprise Server 3.13 | |
 | `v3.22.12` | `2.15.5` | Enterprise Server 3.12 | |
-| `v2.22.1` | `2.14.6` | Enterprise Server 3.11 | Supports CodeQL Action v3, but did not ship with CodeQL Action v3. For more information, see "[Code scanning: deprecation of CodeQL Action v2](https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/#users-of-github-enterprise-server-311)." |
 
-CodeQL Action v2 will stop receiving updates when GHES 3.11 is deprecated.
+CodeQL Action v2 has stopped receiving updates now that GHES 3.11 is deprecated.
```
zutshisunakshi, line 85: Does this need to be part of changelog update?

aeisenberg: I jumped the gun on this change by just a little since v2 is not yet officially deprecated. This is something `@angelapwen` will be working on shortly. GHES 3.11 is getting deprecated on December 19.

I made the change since I thought it would be confusing to remove the line about GFHES 3.11 and keep the line about v2 unchanged.

The last update to the v2 branch should only go in the v2 branch, so I won't put it the changelog on the main branch.

.github/workflows/__go-indirect-tracing-workaround-diagnostic.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            version: stable-v2.14.6
+            version: default
```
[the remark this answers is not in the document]
aeisenberg: Versions for the other workflows are handled by the `sync.py` script.

angelapwen (commented):
Looks fine to me, but we want to make sure that we don't merge before Dec 19, right?

angelapwen (approved):
:merge-it:

PR #2660, opened 2024-12-19
Don't error in the unusual case that a file is on the search path

henrymercer:
In the unusual case that a file is on the search path, log a warning but don't throw.

I've accomplished this by switching our `which` implementation to use the standard Actions one within `@actions/io` but have also submitted a PR to `safe-which`: https://github.com/chrisgavin/safe-which/pull/14.


angelapwen (approved):
Looks good — also looks like your PR to `safe-which` is merged, so not sure if we want to just use that?

henrymercer:
My thinking was that it makes slightly more sense to use `@actions/io` for consistency with other first-party Actions, but I don't think there's a huge difference.

PR #2672, opened 2025-01-06
Include registry type in `proxy_urls` output

mbg:
This is a follow-up to #2652 to include the registry type in the `proxy_urls` output. Since private registry configurations are org-wide, we may end up with registry configurations in a given workflow run that are not relevant to the particular language being analysed. Including the feed type in the list of configured registries allows extractors to identify the ones that relevant to them.


mbg:
CI failures appear to be unrelated.

aeisenberg:
> CI failures appear to be unrelated.

CI will be passing with the release of v2.20.1 of the CLI.

src/start-proxy-action.ts
```diff
@@ -174,7 +174,10 @@ async function startProxy(
 
     const registry_urls = config.all_credentials
       .filter((credential) => credential.url !== undefined)
-      .map((credential) => credential.url);
+      .map((credential) => ({
+        type: credential.type,
+        url: credential.url,
+      }));
```
aeisenberg, line 180: This changes the shape of the `proxy_urls` output. Previously, it was just an array of strings. Is there anything that consumes `proxy_urls` that will break because of this change?

aeisenberg:
Also, maybe the description of the output should be updated as well? https://github.com/github/codeql-action/blob/main/start-proxy/action.yml#L27.

src/start-proxy-action.ts
mbg: I don't think so. I only added this before Christmas for a feature I am working on, and I don't expect anyone else to be using it yet. This action is also marked as experimental, so I wouldn't worry about this potentially being a breaking change externally.

aeisenberg (approved):

PR #2680, opened 2025-01-10
Filter registry credentials by language

marcogario:
Introduce the `language` parameter for the start-proxy action so we know which credentials we should use. 

By specifying the language with disambiguating cases where the same registry has distinct credentials. For example, in Artifactory it is possible to use a token that is scoped to the nuget feed, while the maven repository uses the username/password combination. 


src/start-proxy-action.ts
```diff
@@ -17,6 +17,12 @@ const PROXY_USER = "proxy_user";
 const KEY_SIZE = 2048;
 const KEY_EXPIRY_YEARS = 2;
 
+const LANGUAGE_TO_REGISTRY_TYPE = {
+  "java-kotlin": "maven_repository",
```
aeisenberg, line 21: Should you add `kotlin` as a standalone language?

Also there is `languages.ts`, which is where we've traditionally added language-specific logic. I think I'd prefer to move this declaration to that file.

aeisenberg: I'm fine if the logic doesn't go into language.ts.

src/start-proxy-action.ts
```diff
@@ -17,10 +18,18 @@ const PROXY_USER = "proxy_user";
...
   csharp: "nuget_feed",
+  javascript: "npm_registry",
+  python: "python_index",
+  ruby: "rubygems_server",
+  rust: "cargo_registry",
+  // We do not have an established proxy type for these languages, thus leaving empty.
+  actions: "",
+  cpp: "",
+  go: "",
+  swift: "",
```
aeisenberg, line 32: nit:

```suggestion
  actions: undefined,
  cpp: undefined,
  go: undefined,
  swift: undefined,
```

src/start-proxy-action.ts
```diff
@@ -198,7 +207,11 @@ function getCredentials(logger: Logger): Credential[] {
     "registries_credentials",
   );
   const registrySecrets = actionsUtil.getOptionalInput("registry_secrets");
-  const language = actionsUtil.getOptionalInput("language");
+  const languageString = actionsUtil.getOptionalInput("language");
+  const language = languageString ? parseLanguage(languageString) : undefined;
+  const registryTypeForLanguage = language
+    ? LANGUAGE_TO_REGISTRY_TYPE[language]
+    : undefined;
```
aeisenberg, line 214: nit: If you apply the suggestion above, then this becomes simpler:

```suggestion
  const registryTypeForLanguage = LANGUAGE_TO_REGISTRY_TYPE[language];
```

aeisenberg (approved):
Looks good. I have a suggestion, but it is purely stylistic and non-blocking.

src/start-proxy-action.ts
marcogario: I still need the check on `language` itself, though.

PR #2699, opened 2025-01-16
getDiffRanges: better fallback for absent patch

cklin:
It was pointed out to me in a meeting today that we do not necessarily need to fall back to full analysis even if we cannot obtain the patch for a specific file (for example, if the diff is too large).  Instead, we can continue with diff-informed analysis, using an (already supported) special diff range to cover that entire file.

This PR implements that improvement.


henrymercer (approved):
Makes sense 👍

angelapwen (approved):

PR #2712, opened 2025-01-22
Temporarily disable uploading debug artifacts

angelapwen:
Existing PR checks for debug artifacts will be failing and this is expected. 

I have temporarily removed them from branch protection rules. We should re-add them when we restore functionality.


aeisenberg (approved):
LGTM, assuming that eslint doesn't complain about unreachable code.

PR #2725, opened 2025-01-24
Add actions analysis to code scanning

aeisenberg:
This enables the new actions analysis queries for this repository.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -119,7 +119,7 @@ jobs:
       uses: ./init
       with:
         languages: actions
-        config-file: ./.github/codeql/codeql-config.yml
+        config-file: ./.github/codeql/codeql-actions-config.yml
     - name: Perform CodeQL Analysis
       uses: ./analyze
       with:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -70,7 +70,7 @@ jobs:
         echo "Suggested matrix config for analysis job: $VERSIONS_JSON"
         echo "versions=${VERSIONS_JSON}" >> $GITHUB_OUTPUT
 
-  build:
+  analyze-javascript:
     needs: [check-codeql-versions]
     strategy:
       fail-fast: false
@@ -81,7 +81,7 @@ jobs:
 
     permissions:
       contents: read
-      security-events: write # needed to upload results
+      security-events: write
 
     steps:
     - name: Checkout
@@ -100,3 +100,27 @@ jobs:
       uses: ./analyze
       with:
         category: "/language:javascript"
+
+
+  analyze-actions:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+
+    permissions:
+      contents: read
+      security-events: write
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Initialize CodeQL
+      uses: ./init
+      with:
+        languages: actions
+        config-file: ./.github/codeql/codeql-config.yml
+    - name: Perform CodeQL Analysis
+      uses: ./analyze
+      with:
+        category: "/language:actions"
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -19,6 +19,11 @@ env:
 jobs:
   # Identify the CodeQL tool versions to use in the analysis job.
   check-codeql-versions:
+    strategy:
+      fail-fast: false
+      matrix:
+        language: javascript, actions
```
angelapwen, line 25:
```suggestion
        language: ['javascript', 'actions']
```

I think it's not liking the syntax: https://github.com/github/codeql-action/actions/runs/12957765224/workflow

aeisenberg: Yep!

angelapwen:
I'm actually a little confused by this workflow file but I think we actually need to add the matrixed language variable in the 2nd `build` job rather than the first `check-codeql-versions` job:

https://github.com/github/codeql-action/blob/e7c0c9d71b7bd108fd12e06b56fc58d3d154164d/.github/workflows/codeql.yml#L93 and maybe https://github.com/github/codeql-action/blob/e7c0c9d71b7bd108fd12e06b56fc58d3d154164d/.github/workflows/codeql.yml#L102 to update the category?

I'm not sure we care about the languages specified in `check-codeql-versions` job but maybe it doesn't hurt to include both `actions` and `javascript` there too.

aeisenberg:
Hmmm...I think you're right. I was just being lazy, making changes, and hoping it would work out. I should be thinking deeper about this.

It might just be easier to create a separate job for the actions run. I don't think we need to matrix it over multiple OSes.

aeisenberg:
Seems like it worked!

.github/workflows/codeql.yml
```diff
@@ -100,3 +100,27 @@ jobs:
       uses: ./analyze
       with:
         category: "/language:javascript"
+
+
+  analyze-actions:
+    runs-on: ubuntu-latest
```
angelapwen, line 106: Do we also want this job to depend on `check-codeql-versions`? I guess it's not necessary, as long as one of them depends on it?

aeisenberg: I was thinking that we don't need it.

This workflow is matrixed on multiple versions of CodeQL mainly as an integration test. Since we're using the different versions for analyzing JavaScript, I was thinking that we don't need to do the same for actios. Though, if you can think of a reason, we can change this.

aeisenberg:
I changed the name of the job from `build` to `analyze`. So, there are still 12 expected checks. If we want to get this through, then I'll update the required jobs.

.github/workflows/codeql.yml
```diff
@@ -100,3 +100,27 @@ jobs:
...
+      security-events: write
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Initialize CodeQL
+      uses: ./init
+      with:
+        languages: actions
+        config-file: ./.github/codeql/codeql-config.yml
```
dbartol, line 122: Using the config file means that we're adding a couple directories to the list of ignored paths. The current behavior of the extractor is that if any of the path include/exclude properties are set, we use _only_ those include/exclude settings, and ignore the default Actions path settings. Is that what we want here? (Or, from another angle, is that the right behavior of the Actions extractor?)

dbartol (dismissed):
One comment/question inline, not blocking

.github/workflows/codeql.yml
aeisenberg: Good point. In this case, we don't want to use the custom paths-ignore and just rely on the default. I can't think of a reason why a user would want the non-default version. I'll create a separate actions config and use that one instead.

aeisenberg:
Once this workflow is approved, I'll update the required checks to get this mergeable.

aeisenberg:
@angelapwen or @dbartol can I get another review?

angelapwen (approved):
Looks good to me!

PR #2726, opened 2025-01-25
Ensure artifacts are only uploaded in safe situations

aeisenberg:
Turns on uploading of artifacts again but only if CLI version is >= 2.20.3. ~I implemented the check using our feature flag functionality. I was on the fence about this since it makes the PR more complex. However, it does give us more flexibility when controlling artifact uploads.~

I've changed the PR so that it checks versions directly.

Also, I renamed the two workflows that were previously disabled. This way we will not accidentally enable the old workflows for previous versions of the action.


.github/workflows/debug-artifacts-failure-safe.yml
```diff
@@ -19,6 +19,14 @@ on:
...
 jobs:
   upload-artifacts:
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+        - stable-v2.20.3
+        - default
+        - linked
+        - nightly-latest
```
aeisenberg, line 29: We are now matrixing the build because we want to be sure it is working in the versions we know about.

src/debug-artifacts.test.ts
```diff
@@ -20,16 +20,82 @@ test("sanitizeArtifactName", (t) => {
...
+      ["hucairz"],
+      "i-dont-exist",
+      "artifactName",
+      GitHubVariant.DOTCOM,
+      undefined,
+    );
+    t.is(
+      uploaded,
+      "upload-failed",
+      "Expect failure to upload artifacts since root dir does not exist",
```
angelapwen, line 58: I'm a bit confused by these tests: it looks like this one simultaneously tests when no CodeQL version is used & when the root directory doesn't exist? 

Maybe the test name could be `uploadDebugArtifacts when no codeql version is used and root directory does not exist`?

src/debug-artifacts.test.ts
```diff
@@ -20,16 +20,82 @@ test("sanitizeArtifactName", (t) => {
...
+      ["hucairz"],
+      "i-dont-exist",
+      "artifactName",
+      GitHubVariant.DOTCOM,
+      "2.20.2",
+    );
+    t.is(
+      uploaded,
+      "upload-not-supported",
+      "Expect failure to upload artifacts since root dir does not exist",
```
angelapwen, line 98: I think this test failure should be because the CodeQL version is old and not supported here?

src/debug-artifacts.test.ts
```diff
@@ -20,16 +20,82 @@ test("sanitizeArtifactName", (t) => {
...
+    );
+    t.is(
+      uploaded,
+      "upload-failed",
+      "Expect failure to upload artifacts since root dir does not exist",
+    );
+  });
+});
+
+test("uploadDebugArtifacts when new codeql version is used", async (t) => {
```
angelapwen, line 63: Maybe the test name could be `uploadDebugArtifacts when new codeql version is used and root directory does not exist`?

angelapwen (approved):
Looks good — just some non-blocking comments about giving the tests a bit more clarity!

src/debug-artifacts.test.ts
aeisenberg: I will add some more explanation for these tests. Ideally, I would have been able to mock the actual call to upload artifacts, but this was not easy to do without introducing a new dependency. So, all this test does is ensure that the call to upload artifacts is reached (even though it fails).

We test that the artifacts are actually uploaded in the integration test workflows.

src/debug-artifacts.test.ts
angelapwen: I think this comment is still valid on this test — the failure should be because the upload is not supported on this version of the CLI, right?

aeisenberg: Oops. Yes. You're right about this one. I'll change it.

angelapwen (approved):
Looks good!!

PR #2765, opened 2025-02-14
Perform consistent diff-informed alert filtering in the action

cklin:
This PR updates the SARIF upload code path to perform consistent diff-informed alert filtering.  With this change, a PR analysis with diff-informed analysis enabled will return only alerts that are within the diff range, regardless of whether the QL queries have been adapted to be diff-informed.


src/upload-lib.ts
```diff
@@ -848,3 +850,45 @@ export class InvalidSarifUploadError extends Error {
...
+          // CodeQL always uses forward slashes as the path separator, so on Windows we
+          // need to replace any backslashes with forward slashes.
+          const locationPath = path
+            .join(checkoutPath, locationUri)
+            .replaceAll(path.sep, "/");
+          return diffRanges.some(
+            (range) =>
+              range.path === locationPath &&
+              ((range.startLine <= locationStartLine &&
+                range.endLine >= locationStartLine) ||
```
henrymercer, line 885: What about if the location's start line isn't within the range but its end line is?  If this is intentional, it might be worth adding a comment to make explicit that this case is filtered.

src/diff-filtering-utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  logger: Logger,
+): DiffThunkRange[] | undefined {
+  const jsonFilePath = getDiffRangesJsonFilePath();
+  if (!fs.existsSync(jsonFilePath)) {
+    logger.debug(`Diff ranges JSON file does not exist at ${jsonFilePath}`);
+    return undefined;
+  }
+  const jsonContents = fs.readFileSync(jsonFilePath, "utf8");
+  logger.debug(
+    `Read pr-diff-range JSON file from ${jsonFilePath}:\n${jsonContents}`,
```
henrymercer, line 39: It would be more robust to log only the first n entries, in case we have a very large PR.

src/diff-filtering-utils.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function writeDiffRangesJsonFile(
+  logger: Logger,
+  ranges: DiffThunkRange[],
+): void {
+  const jsonContents = JSON.stringify(ranges, null, 2);
+  const jsonFilePath = getDiffRangesJsonFilePath();
+  fs.writeFileSync(jsonFilePath, jsonContents);
+  logger.debug(
+    `Wrote pr-diff-range JSON file to ${jsonFilePath}:\n${jsonContents}`,
```
henrymercer, line 25: Same here

henrymercer (approved):
Looks good, a couple of suggestions:

src/upload-lib.ts
```diff
@@ -848,3 +850,45 @@ export class InvalidSarifUploadError extends Error {
     super(message);
   }
 }
+
+function filterAlertsByDiffRange(logger: Logger, sarif: SarifFile): SarifFile {
+  const diffRanges = readDiffRangesJsonFile(logger);
+  if (!diffRanges) {
```
henrymercer, line 856: An empty array is truthy in ECMAScript.  In the case that the PR has no changes, it might be better to test `diffRanges?.length` instead.

```suggestion
  if (!diffRanges?.length) {
```

src/upload-lib.ts
cklin: Thanks for the suggestion!  I added comments to document the intended behavior of the filtering, by way of referencing the `restrictAlertsTo` extensible predicate in QL.

src/upload-lib.ts
cklin: Thanks for the suggestion!  Applied.

src/diff-filtering-utils.ts
cklin: I will keep the logging as-is for now.  Since the JSON file contains only the line ranges, hopefully it will still be manageable even for a very large PR.

henrymercer (approved):
Thanks for the additional clarifications, LGTM

PR #2768, opened 2025-02-17
Pass `--ignore-zeros` to `tar` when decompressing `zstd`-compressed tarballs

smowton:
See comment in the diff for full explanation.


henrymercer (approved):
Thanks for the thorough investigation and explanation!  This looks good to me assuming checks pass.

Do you have an opinion about whether this means we should revert https://github.com/github/codeql-action/pull/2767 and return to preferring bsdtar on macOS systems?  If we revert that PR, we should add a changenote to this one.

smowton:
No strong opinion -- using `gtar` is also fine. This should probably be done anyhow since it means we'll work better with BSD tar whenever it's the only one around.

PR #2774, opened 2025-02-19
Fix sync recipes and add base `justfile`

redsun82:
Both the justfile and the pre-commit configuration for the `pr-check` sync were broken:
* justfiles run recipes one line at a time in a fresh shell, so the venv activation was not working
* the pre-commit config was relying on an installed `ruamel.yaml` pakcage, but the default one installable via `apt` on Ubuntu 24.04 is old and generates different output (with formatting differences).

Now:
* the venv dance is put in a separate bash script
* both just and pre-commit will use that same script, so both problems will be fixed

As a bonus, a root `justfile` is added exposing the `update-pr-checks` recipes plus a `build` one. Running `just` without arguments will also now call the default `sync` recipes that will call both of the above.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,7 +1,10 @@
+# Sync generated files (javascript and PR checks)
 sync: build update-pr-checks
 
+# Perform all necessary steps to update the PR checks
 update-pr-checks:
     pr-checks/sync.sh
 
+# Transpile typescript code into javascript
 build:
     npm run build


justfile
```diff
@@ -0,0 +1,7 @@
+sync: build update-pr-checks
```
NlightNFotis, line 1: Would it be possible to add a comment on top of the command?

Because we're probably going to forget which one is which after a couple months, so `just --list` will probably be our fallback then, and it renders the comment as inline help when invoked.

NlightNFotis (dismissed):
Thank you for fixing the `justfile`.

I seem to recall it working locally when I was last working on this a few months ago, but it might have been incidental?

justfile
redsun82: Good point!

NlightNFotis (approved):
(Reapproving after review automatically went stale).

Thanks for your work on this @redsun82 , I appreciate it 👍🏻

redsun82:
> Thank you for fixing the `justfile`.
> 
> I seem to recall it working locally when I was last working on this a few months ago, but it might have been incidental?

@NlightNFotis Yeah, thinking back that was working because:
* on macOS, `sh` is actually an alias of `bash` (I'm on Linux where that's not the case)
* you had the proper package available to `python3` outside of the venv, so the last command was also working despite being run outside of the venv

PR #2785, opened 2025-02-26
Warn about small amounts of MB, not GB

igfoo:
The number of GB is at most 2, and can be tiny. MB gives a more comprehensible range of values.


henrymercer (approved):
Makes sense, thank you!

PR #2787, opened 2025-02-27
Remove bundled copy of `actions` extractor

dbartol:
In #2582, I temporarily added an embedded version of the `actions` extractor into the CodeQL Action itself, to work around the extractor being omitted from the CLI bundle. That's been fully corrected as of CLI 2.20.4, so we should no longer need the embedded copy.

I'm ignoring the possibility of Advanced Setup users having pinned `tools:` to a specific version, since the Actions analysis is only in preview anyway.


dbartol:
To be safe, I checked our telemetry and there have been zero runs in the last 14 days that analyzed `actions` _and_ used a CLI older than 2.20.4.

dbartol:
🤦‍♂️ The `autobuild.sh` file in the extractor in the CLI doesn't have the execute bit set. I'll have to fix it in the CLI and wait for the newer version to be propagated.

aeisenberg:
I will pick this PR up again.

aeisenberg (dismissed):
@henrymercer, I forgot about this PR. Since this is not really my responsibility any more, I'm hoping you can merge this when you're able.

CHANGELOG.md
```diff
@@ -4,7 +4,10 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The CodeQL Action no longer includes its own copy of the extractor for the `actions` language, which is currently in public preview.
+  The `actions` extractor has been included in the CodeQL CLI since v2.20.4. If your workflow has enabled the `actions` language _and_ you have pinned
+  your `tools:` property to a specific version of the CodeQL CLI earlier than v2.20.4, you will need to update to at least CodeQL v2.20.4 or disable
+  `actions` analysis.
```
henrymercer, line 10:
```suggestion
  The `actions` extractor has been included in the CodeQL CLI since v2.20.6. If your workflow has enabled the `actions` language _and_ you have pinned
  your `tools:` property to a specific version of the CodeQL CLI earlier than v2.20.6, you will need to update to at least CodeQL v2.20.6 or disable
  `actions` analysis.
```

henrymercer (approved):

PR #2788, opened 2025-02-27
Use embedded `actions` extractor only for old CLI versions

dbartol:
When trying to remove the embedded `actions` extractor, we discovered that the extractor in the actual CLI download is missing the execute bit on its scripts. That's fixed in 2.20.6, so I've changed the Action to only use the embedded extractor for pre-2.20.6 versions.


angelapwen (approved):
Makes sense to me 👍 do we want to add a changenote for this or are we skipping it because it's experimental?

dbartol:
>do we want to add a changenote for this 

For this change, there should be no user-visible impact, so I'm skipping the change note. When we do finally remove the embedded extractor entirely, I'll add a change note, although we don't expect any users to actually notice the breaking change.

angelapwen:
Sounds good to me!! 🚢

angelapwen (approved):

PR #2801, opened 2025-03-07
Basic support for overlay databases

cklin:
This PR implements basic support for overlay databases.  It does not handle the uploading and downloading of overlay-base databases; for now we are assuming that the workflow will provide the necessary functionality.


cklin:
I thought about it over the weekend and decided to make some changes.  So taking the PR back to draft mode for now.

cklin:
I split the large "Support overlay database creation" commit into smaller commits.  In addition to making the PR easier to review, the split also allows me to remove the code changes for file change detection, which is a functionality not used in this PR.  As a result the PR as a whole is now smaller too.

src/init.ts
```diff
@@ -79,13 +85,55 @@ export async function initConfig(
...
 
+export async function getOverlayDatabaseMode(
+  codeqlVersion: string,
+  config: configUtils.Config,
+  sourceRoot: string,
+  logger: Logger,
+): Promise<OverlayDatabaseMode> {
+  const overlayDatabaseMode = process.env.CODEQL_OVERLAY_DATABASE_MODE;
+
+  if (
```
henrymercer, line 96: Optional nit: consider using a `switch` statement with `assertNever` so if we add new cases to the `OverlayDatabaseMode` enum, the type checker will remind us to cover them here.

It's optional since we probably won't be adding new cases to this enum.

src/codeql.ts
```diff
@@ -606,12 +609,20 @@ export async function getCodeQLForCmd(
         ? "--force-overwrite"
         : "--overwrite";
 
+      if (overlayDatabaseMode === OverlayDatabaseMode.Overlay) {
```
henrymercer, line 612: Similar nit here

src/analyze-action.ts
```diff
@@ -295,12 +295,16 @@ async function run() {
       logger,
     );
 
+    const cleanupLevel =
+      actionsUtil.getOptionalInput("cleanup-level") || "brutal";
```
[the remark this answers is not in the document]
cklin: When creating an overlay-base database, the only cleanup levels that make sense are `overlay` and `none`.  So there is room for selecting a more sensible default cleanup level based on the overlay mode.

At this point we have not decided how we want the overlay database mode selection to work for the long run, which would affect how we communicate the selected overlay database mode to the analyze action.  For the purpose of this PR, we are assuming that the invoking workflow handles all the complexity so that we can minimize the action changes, both in terms of new code and new design assumptions / commitments.

So I will keep your suggestion in mind and leave this PR as-is.

src/init.ts
cklin: Thanks for the suggestion!  I looked at both places and decided that the benefit does not quite justify the increased code verbosity.  Hopefully, if we should add a new value to the `OverlayDatabaseMode` enum, it would be apparent that we need to change where we make the `OverlayDatabaseMode` selection (here) and where we use the `OverlayDatabaseMode` value (in `databaseInitCluster`).

src/codeql.ts
cklin: Thanks—responded above.

PR #2816, opened 2025-03-19
Overlay databases: use --overlay-changes

cklin:
This PR changes overlay database creation to use the `--overlay-changes` flag. It also implements Git-based file change detection to generate the list of files to extract for the overlay database.


src/git-utils.ts
```diff
@@ -253,6 +253,44 @@ export const getGitRoot = async function (
...
+  // Without the --full-name flag, the path is relative to the current working
+  // directory of the git command, which is basePath.
+  const stdout = await runGitCommand(
+    basePath,
+    ["ls-files", "--recurse-submodules", "--format=%(objectname)_%(path)"],
+    "Cannot list Git OIDs of tracked files.",
+  );
+
+  const fileOidMap: { [key: string]: string } = {};
+  const regex = /^([0-9a-f]{40})_(.+)$/;
```
NlightNFotis, line 278: Is it possible to provide a sample of what this matches as a comment above the variable definition?

NlightNFotis (commented):
From a cursory look, this looks alright, but I would feel much more confident approving this if it came accompanied with some tests.

src/git-utils.ts
cklin: I added a comment as suggested.

cklin:
@NlightNFotis I added comments and tests as you suggested.  PTAL.  Thanks!

NlightNFotis (approved):
LGTM, what I've seen makes sense to me. Thank you for the work here.

PR #2817, opened 2025-03-21
Support diff-informed queries under Default Setup

cklin:
This PR extends the diff-informed queries functionality to also work under the Default Setup.


src/analyze.ts
```diff
@@ -254,6 +254,25 @@ async function finalizeDatabaseCreation(
   };
 }
 
+interface PullRequestBranches {
```
NlightNFotis, line 257: Thank you for this refactoring. I always appreciate bundling information into its own semantic bundle.

Would it be possible to get some documentation above the interface as to what it's supposed to signify? I think it's clear from the context, but it might make future mantainer's life better.

I think some of that documentation could be sourced from the previous `@param` documentation in the `getPullRequestEditedDiffRanges` function.

NlightNFotis (approved):
Some minor comments, thank you for your work here.

src/analyze.ts
```diff
@@ -270,6 +270,19 @@ function getPullRequestBranches(): PullRequestBranches | undefined {
...
+  // but it should set CODE_SCANNING_REF and CODE_SCANNING_BASE_BRANCH.
+  const codeScanningRef = process.env.CODE_SCANNING_REF;
+  const codeScanningBaseBranch = process.env.CODE_SCANNING_BASE_BRANCH;
+  if (codeScanningRef && codeScanningBaseBranch) {
+    return {
+      base: codeScanningBaseBranch,
+      // PR analysis under Default Setup analyzes the PR head commit instead of
+      // the merge commit, so we can use the provided ref directly.
+      head: codeScanningRef,
+    };
```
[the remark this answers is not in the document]
cklin: I added two info log statements:
- One to indicate that we are not performing diff-informed analysis because we are not analyzing a PR
- One to indicate that we are analyzing a PR, and what refs are being used for diff range calculation

NlightNFotis (approved):

src/analyze.ts
cklin: I thought about it and decided that additional documentation would not  have provided much value.

The interface is basically what the definition says: a record for storing the base and head branches of a pull request.  Repeating the same thing in English does not seem to be all that useful.

Previous `@param` documentation for `getPullRequestEditedDiffRanges()` contains some useful information, and I have already incorporated them into code comments in `getPullRequestBranches()`.  That documentation does not quite fit at the `PullRequestBranches` level, because the `pull_request` context is now only one of two possible sources for obtaining branch information.  (When we set the head branch using `CODE_SCANNING_REF`, we are relying on the Default Setup workflow to set the environment variable correctly, without requiring the variable to use a head ref.)

All that is explained in `getPullRequestBranches()`, next to the code that creates `PullRequestBranches` values.  I think that is a better place for that explanation because it is easier to keep the code comments up-to-date if the code needs to be changed in the future.

PR #2818, opened 2025-03-21
Diff-informed analysis: fix empty PR handling

cklin:
This PR fixes how diff-informed analysis handles empty pull requests.


angelapwen (approved):
Thanks for the descriptive comment 👍

src/analyze.ts
```diff
@@ -527,6 +527,15 @@ function writeDiffRangeDataExtensionPack(
     return undefined;
   }
 
+  if (ranges.length === 0) {
+    // An empty diff range means that there are no added or modified lines in
+    // the pull request. But the `restrictAlertsTo` extensible predicate
+    // interprets an empty data extension differently, as an indication that
+    // all alerts should be included. So we need to specifically set the diff
+    // range to a non-empty list that cannot match any alert location.
+    ranges = [{ path: "", startLine: 0, endLine: 0 }];
```
adityasharad, line 536: I was wondering whether we have QL logic that looks for locations overlapping with the diff range. But I think that is not possible with this choice of range.

This looks right as long as we don't try to check the `path` against actual paths in the source archive.

adityasharad (approved):

src/analyze.ts
cklin: When start line and end line are both 0, the QL logic that matches location overlapping with the diff range is a simple join on the absolute path.  Since an absolute path in a PR diff cannot be empty, such a join should very efficiently discard all valid potential alert locations.

PR #2828, opened 2025-03-26
Add default config file location for init action

GeekMasher:
This pull request includes an update to the `src/init-action.ts` file to improve the handling of the configuration file in the `run` function. The most important change is the addition of logic to set a default configuration file location if none is provided.

Improvements to configuration file handling:

* [`src/init-action.ts`](diffhunk://#diff-87bf9e0a2a9244ba1f26eb01d404deaccbc9580479d2e2530a4ff2fd5bd91565L301-R318): Changed the `configFile` variable to be mutable and added logic to check for a default configuration file location at `.github/codeql/codeql.yml` within the `GITHUB_WORKSPACE` if no configuration file is specified. If the default configuration file exists, it sets `configFile` to this location and logs the path. Otherwise, it logs that no configuration file was found.


src/init-action.ts
```diff
@@ -298,7 +298,22 @@ async function run() {
...
+      "codeql",
+      "codeql.yml",
+    );
+
+    if (fs.existsSync(defaultLocation)) {
+      configFile = defaultLocation;
+      logger.info(
+        `Using default config file location: ${path.resolve(configFile)}`,
+      );
+    }
```
NlightNFotis, line 315: I'm curious whether it would make sense to add a log message if the default file fails to be found, to warn both the users and us when debugging that something has gone very wrong.

NlightNFotis (commented):
Thank you for the contribution, this looks reasonable to me.

I _think_ we also need a changelog entry for this change, and it also _might_ be something that might need a docs change, but I will let someone else confirm that.

src/init-action.ts
GeekMasher: I actually added it in the first commit but thought it would almost always trigger a log that no config was used. I'm okay with either personally

NlightNFotis:
Hi @GeekMasher , this PR will need the latest `main` merged in (or rebased) and the `just sync` instruction to be re-run.

If you're busy with other stuff, please do let us know and we can shepherd this in for you.

GeekMasher:
@NlightNFotis Great, I have pulled in and updated the `lib` now. Let me know if you need anything else

GeekMasher:
Removing for now until we have a better pattern to support this

PR #2830, opened 2025-03-26
getFileDiffsWithBasehead(): use CODE_SCANNING_REPOSITORY if present

cklin:
This PR updates `getFileDiffsWithBasehead()` so that it uses `CODE_SCANNING_REPOSITORY` if present, which makes it possible to override the repository that contains the commits being analized.


src/analyze.ts
```diff
@@ -390,7 +390,10 @@ async function getFileDiffsWithBasehead(
   branches: PullRequestBranches,
   logger: Logger,
 ): Promise<FileDiff[] | undefined> {
-  const repositoryNwo = getRepositoryNwo();
+  const repositoryNwo = getRepositoryNwoFromEnv(
+    "CODE_SCANNING_REPOSITORY",
```
angelapwen, line 394: Perhaps a small code comment here indicating that `CODE_SCANNING_REPOSITORY` is intended to override `GITHUB_REPOSITORY` would be useful for other maintainers!

angelapwen (approved):
Makes sense!! Just a small comment about documentation 😸

src/analyze.ts
cklin: Thanks for the suggestion!  I added the comment.

angelapwen (approved):

PR #2831, opened 2025-03-27
Respect `exclude-from-incremental` query tag for diff-informed analysis

cklin:
This PR changes the action to add the following settings to the Code Scanning configuration file when performing diff-informed analysis:

```yaml
query-filters:
    - exclude:
        tags: exclude-from-incremental
```

Code Scanning configuration file for non-diff-informed analysis remains unchanged.

Commit-by-commit review recommended.  The first few commits are preparatory; they change code structure while preserving existing code behavior.  The second-last commit adds the new functionality, and the last commit rebuilds the js files.


angelapwen (dismissed):
Logic looks 👍! 

Non-blocking nit: could `shouldPerformDiffInformedQueries` return a boolean and then the `branches` value be returned elsewhere? Or alternatively maybe we could rename the function something even longer and gnarlier, like `getBranchesIfDiffInformedQueriesEnabled` 😬

henrymercer:
Yeah, or just `checkDiffInformedQueries` — we don't have to explain the return type in the function name, but `should` is usually associated with a boolean return type.

cklin:
Thank you for your suggestions @angelapwen and @henrymercer!

I updated the PR, and there are now two helper functions.
- `shouldPerformDiffInformedAnalysis()` returns `Promise<boolean>`
- `getDiffInformedAnalysisBranches()` returns `Promise<PullRequestBranches | undefined>`

angelapwen (approved):
Thank you 🙇‍♀️

PR #2832, opened 2025-03-28
fix: change regex matching for API error to not contain regex boundaries

NlightNFotis:
## Description

This fixes an issue with a regex pattern match on API errors failing due to regex boundaries on the pattern, which would affect telemetry.

**Note**: This change doesn't affect users and doesn't require any changes to the `readme` or the `changelog`.


henrymercer (approved):
Thanks for adding tests!

PR #2833, opened 2025-03-31
feat: further error re-classification 

NlightNFotis:
## Description

This PR adds some further error classification for issues we observe in telemetry, along with a re-classification of `InvalidSarifUploadError`s as `user-error`s.

**NOTE**: This is an internal/telemetry change, and doesn't affect users in any way that would induce the need to update either the readme or the changelog file.


NlightNFotis:
Hi @henrymercer, would it be possible to get a review on this one?

I'd like some further eyes on f21cf0bbd70a0b319d33ec5c58541cb691ade920, as I'm fairly confident the functionality works, but I'm not sure if we are to consider `InvalidSarifUploadError`s as `user-errors` (I've seen discussions leaning both ways).

src/api-client.ts
```diff
@@ -245,6 +245,9 @@ export function wrapApiConfigurationError(e: unknown) {
     if (
       e.message.includes("API rate limit exceeded for installation") ||
       e.message.includes("commit not found") ||
+      e.message.includes("Bad credentials") ||
+      e.message.includes("Not Found") ||
+      e.message.includes("Resource not accessible by integration") ||
```
henrymercer, line 250: There's an opportunity to provide a better error message here too, for example for bad credentials / not found we could ask whether the job has contents: read and security-events: write permissions.

src/upload-lib.ts
```diff
@@ -734,23 +734,32 @@ export async function waitForProcessing(
...
-function shouldConsiderInvalidRequest(processingErrors: string[]): boolean {
+export function shouldConsiderInvalidRequest(
+  processingErrors: string[],
+): boolean {
   return processingErrors.every(
     (error) =>
       error.startsWith("rejecting SARIF") ||
       error.startsWith("an invalid URI was provided as a SARIF location") ||
       error.startsWith("locationFromSarifResult: expected artifact location") ||
+      error.startsWith("SyntaxError: Unexpected end of JSON input") ||
```
henrymercer, line 762: How's this possible if we're validating the SARIF against the JSON schema before we try to upload it?  I'd be surprised if this was going wrong, but is there a bug in the way we combine the SARIF files?

aeisenberg: Are we getting this error when we're doing the validation locally? And if so, does this function catch the error?

If we're getting this error from the server, I would say this is a bug.

NlightNFotis: Henry is correct, I misunderstood the error location and thought this was coming from the backend. Judging from the stack trace I've seen, this appears to be coming from the `uploadFiles` function calling `validateSarifFileSchema` on each input file, which then fails the validation.

In this case, the `upload-sarif-action.ts` would be the best location for that, and we also probably want to categorise as a `user-error` the case where _we ourselves_ are not the producers of a SARIF file with syntax errors.

I have now corrected this at [b53826d](https://github.com/github/codeql-action/pull/2833/commits/b53826d56df00da26d3b54ec351a2be08c3cb953)

src/api-client.ts
NlightNFotis: Thanks for the suggestion, I have implemented this at [55ee663](https://github.com/github/codeql-action/pull/2833/commits/55ee663d5f438278289baa303efa7a22e417c9ea).

src/status-report.ts
```diff
@@ -173,7 +174,14 @@ export function getActionsStatus(
   otherFailureCause?: string,
 ): ActionStatus {
   if (error || otherFailureCause) {
-    return error instanceof ConfigurationError ? "user-error" : "failure";
+    if (
+      error instanceof ConfigurationError ||
+      error instanceof InvalidSarifUploadError
```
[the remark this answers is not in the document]
NlightNFotis: I put some thought into trying to come up with a schema for identifying how to handle this case, and I came to realisation that perhaps the best way to implement this is by passing `getActionsStatus` an extra parameter indicating whether the analysis is a third party one or not. This would also make it easier to extend it in the future based on this information.

I have implemented this approach in [55ee663](https://github.com/github/codeql-action/pull/2833/commits/55ee663d5f438278289baa303efa7a22e417c9ea).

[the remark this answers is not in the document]
henrymercer: I'd prefer to keep `getActionsStatus` simple and delegate the task of deciding whether errors are user errors or not to the part of the code where the relevant logic is implemented.

For example, the `upload-sarif` Action we should catch any `InvalidSarifUploadError`s and rethrow them as a `ConfigurationError` if it's a third-party run.  So by the time we reach the top-level status reporting logic, we should only be dealing with `InvalidSarifUploadError`s that aren't user errors.

If we're seeing `InvalidSarifUploadError`s from third-party runs, there's a bug in the `upload-sarif` Action — we should fix that to make sure it's wrapping those errors in `ConfigurationError`s.

src/resolve-environment-action.ts
```diff
@@ -82,9 +83,12 @@ async function run() {
         `Failed to resolve a build environment suitable for automatically building your code. ${error.message}`,
       );
 
+      const isThirdPartyAnalysis = !isFirstPartyAnalysis(
+        ActionName.ResolveEnvironment,
+      );
```
aeisenberg, line 88: Can you convert this to a function in `status-report.ts`? You are duplicating and negating the call to `isFirstPartyAnalysis` in enough places that it might be cleaner to just extract it to a separate function.

henrymercer (commented):
(posting this review so Slack doesn't keep reminding me)

See https://github.com/github/codeql-action/pull/2833#discussion_r2023237528

src/status-report.ts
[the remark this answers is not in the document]
NlightNFotis: Thank you, I have re-implemented it this way, managing to repurpose some of the code already written to simplify the code and its handling: [498c7f3](https://github.com/github/codeql-action/pull/2833/commits/498c7f37e85d2d0fe6bac32a3c71708870a27bee)

src/upload-sarif-action.ts
```diff
@@ -104,9 +104,16 @@ async function run() {
     }
     await sendSuccessStatusReport(startedAt, uploadResult.statusReport, logger);
   } catch (unwrappedError) {
+    // This is testing the error to check if it belongs to one of two categories we reliably
+    // know to be configuration errors in certain cases.
+    const configurationErrorCandidate =
+      // There was a problem uploading the SARIF file (perhaps rejected by the backend)
+      unwrappedError instanceof upload_lib.InvalidSarifUploadError ||
+      // There was a problem validating the JSON (SARIF) file.
+      unwrappedError instanceof SyntaxError;
```
henrymercer, line 113: There's a small chance that by catching any SyntaxError here we miss something genuine.  If we put a try/catch around the JSON parse block specifically, and throw an `InvalidSarifUploadError` if there was invalid JSON, we can avoid that situation.

henrymercer (commented):
Thanks for addressing the feedback!  Looks good, just one suggestion.

src/upload-sarif-action.ts
NlightNFotis: Thanks, this is a great suggestion! Added in [676a422](https://github.com/github/codeql-action/pull/2833/commits/676a422916fb4b124d13126326cd03f440e112d6)

henrymercer (approved):

PR #2839, opened 2025-04-01
Clean-up logic for overriding proxy

marcogario:
In https://github.com/github/codeql-action/pull/2741, we added this logic due to an issue with the CLI. The CLI has since been fixed and I think it is cleaner to remove this workaround. I am open to hear otherwise and close the PR instead.


henrymercer (commented):
Makes sense.  Can we gate this behind a version check so that we keep the same behaviour for older CLI versions?

src/analyze-action.ts
```diff
@@ -244,6 +235,16 @@ async function run() {
       );
     }
 
+    // Unset the CODEQL_PROXY_* environment variables, as they are not needed
+    // and can cause issues with older CodeQL CLIs.
+    // Check for CODEQL_PROXY_HOST: and if it is empty but set, unset it.
+    if (process.env.CODEQL_PROXY_HOST === "" && !(await util.codeQlVersionAtLeast(codeql, "2.20.7"))) {
```
aeisenberg, line 241: Can you extract this to a constant and place it in `feature-flags.ts`? 

Ideally, we would have pushed this into the CLI itself and added added it to `isSupportedToolsFeature`, but it's a bit late for that now. Using a hard-coded feature version is best then.

Hmmm...looking through the code, I see that we haven't been to good about moving all hard-coded versions to the feature-flags.ts file. Still, that's something that we've tried to do in the past.

src/analyze-action.ts
```diff
@@ -244,6 +235,16 @@ async function run() {
       );
     }
 
+    // Unset the CODEQL_PROXY_* environment variables, as they are not needed
+    // and can cause issues with older CodeQL CLIs.
+    // Check for CODEQL_PROXY_HOST: and if it is empty but set, unset it.
```
henrymercer, line 240:
```suggestion
    // Older CLIs can run into issues when the CODEQL_PROXY_* environment variables are empty but set.
```

henrymercer (approved):
Agree with Andrew's suggestion here.  LGTM otherwise.

henrymercer (approved):

PR #2842, opened 2025-04-04
Raise the file limit for debug artifacts by producing zip64 files where necessary

henrymercer:
The previous library only supported the original zip format, which is limited to 65535 files.  This caused issues producing debug artifacts for codebases with large numbers of files.


src/debug-artifacts.ts
```diff
@@ -344,9 +344,21 @@ async function createPartialDatabaseBundle(
   if (fs.existsSync(databaseBundlePath)) {
     await del(databaseBundlePath, { force: true });
   }
-  const zip = new AdmZip();
-  zip.addLocalFolder(databasePath);
-  zip.writeZip(databaseBundlePath);
+  const output = fs.createWriteStream(databaseBundlePath);
+  const zip = archiver("zip");
```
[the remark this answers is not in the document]
henrymercer: We only want to use zip64 when necessary as some older platforms don't support it.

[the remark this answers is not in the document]
henrymercer: We have a couple of PR checks to validate that the debug artifacts work.

src/debug-artifacts.ts
```diff
@@ -344,9 +344,21 @@ async function createPartialDatabaseBundle(
...
+
+  zip.on("error", (err) => {
+    throw err;
+  });
+
+  zip.on("warning", (err) => {
+    throw err;
+  });
```
aeisenberg, line 356: Two questions:

1. In what cases will there be a warning and is this really something we want to raise an error on?
2. What will happen if an error is thrown? Where will it be caught and will it be properly handled?

aeisenberg: Looking at the sample code in the packge, the recommendation is something like this:
```
  if (err.code === "ENOENT") {
    // log warning
  } else {
    // throw error
    throw err;
  }
```

I'll update.

src/debug-artifacts.ts
```diff
@@ -344,9 +344,21 @@ async function createPartialDatabaseBundle(
...
-  zip.writeZip(databaseBundlePath);
+  const output = fs.createWriteStream(databaseBundlePath);
+  const zip = archiver("zip");
+
+  zip.on("error", (err) => {
+    throw err;
+  });
+
+  zip.on("warning", (err) => {
+    throw err;
```
aeisenberg, line 355:
```suggestion
    // Ignore ENOENT warnings. There's nothing anyone can do about it.
    if (err.code !== "ENOENT") {
      throw err;
    }
```

aeisenberg:
I pushed the change since Henry is on vacation now.

angelapwen (approved):
:shipit:

PR #2871, opened 2025-04-25
Add `proxy_binary` input to `start-proxy` action

mbg:
This is useful for testing new versions of the proxy before release.


src/start-proxy-action.ts
```diff
@@ -119,7 +119,9 @@ async function runWrapper() {
   };
 
   // Start the Proxy
-  const proxyBin = await getProxyBinaryPath();
+  const proxyBin =
+    actionsUtil.getOptionalInput("proxy_binary") ??
+    (await getProxyBinaryPath());
```
aeisenberg, line 124: What sort of error is emitted if `proxy_binary` doesn't exist or isn't executable? Is there a try-catch somewhere that emits a meaningful error message in this case?

start-proxy/action.yml
```diff
@@ -2,6 +2,10 @@ name: "CodeQL: Start proxy"
 description: "[Experimental] Start HTTP proxy server. This action is for internal GitHub used only and will change without notice."
 author: "GitHub"
 inputs:
+  proxy_binary:
+    description: >-
+      A path for the proxy binary to use.
+    required: false
```
aeisenberg, line 8: This is a question, not anything definitive. We're now giving the user an option of specifying an executable. Are there any security vulnerabilities that this opens up? Maybe the answer is "no" since this is already running in an action, which already allows execution of arbitrary code.

mbg: I figured we allow the same sort of thing with the `tools` input to the `init` Action, so that option is already there. Although I didn't check whether we perform any kind of validation there to verify that we were are running is safe.

PR #2876, opened 2025-05-01
 Do not fail diff informed analyses when analyze is run twice in the same job

henrymercer:
Tolerate the diff range extension pack directory already existing and overwrite it, to avoid failing in jobs that call `analyze` more than once.

Also defer the computation of the dependency caching default cache config to avoid getting the value of the `RUNNER_TEMP` environment variable before it can be overridden by test logic.

And finally disable diff-informed analyses within the config parsing tests, since these tests do not account for the additional query filter added by diff-informed analyses.


nickrolfe (approved):
Looks sensible.

Issue #2890, opened 2025-05-06
No way to configure CODEQL_THREADS with an environment variable


austinpray-mixpanel:
The source says this action respects both CODEQL_RAM and CODEQL_THREADS

https://github.com/github/codeql-action/blob/5eb3ed6614230b1931d5c08df9e096e4ba524f21/lib/init-action.js#L315C12-L319

this is true for CODEQL_RAM

https://github.com/github/codeql-action/blob/5eb3ed6614230b1931d5c08df9e096e4ba524f21/lib/init-action.js#L320

but not true for CODEQL_THREADS

https://github.com/github/codeql-action/blob/5eb3ed6614230b1931d5c08df9e096e4ba524f21/lib/init-action.js#L322

Is this an oversight or is there a good reason for this?

My use case is I'm running this on a big 48 core kubernetes but the codeql runner pod is only allowed to use 16 cores. The autodetection is not factoring in the pod limits, it's looking at the node's available resources. I want to hint to codeql that it only has 16 threads available via the CODEQL_THREADS env var.

austinpray-mixpanel:
Spiked up a PR for this https://github.com/github/codeql-action/pull/2891

mbg:
> The source says this action respects both CODEQL_RAM and CODEQL_THREADS

I think there's a key misunderstanding here: the comment in https://github.com/github/codeql-action/blob/5eb3ed6614230b1931d5c08df9e096e4ba524f21/src/init-action.ts#L538-L542 talks about the CodeQL _CLI_, not the CodeQL _Action_. In other words, it is the _CLI_ that respects these environment variables and they are set by the _Action_ based on the corresponding Actions inputs (if set).

That said, I am unsure why we respect the existing value of `CODEQL_RAM` (if set) over the explicit input to the Action and not `CODEQL_THREADS` (if set).

austinpray-mixpanel:
Thank you for the quick response!

More concrete details:

```yaml
name: "GHAS JS CodeQL"

on:
  push:
    branches: [ "master" ]
    paths:
      - '**.js'
      - '**.jsx'
      - '**.ts'
      - '**.tsx'
      - '**.html'
      - '.github/workflows/ghas-js-codeql.yaml'
  pull_request:
    branches: [ "master" ]
    paths:
      - '**.js'
      - '**.jsx'
      - '**.ts'
      - '**.tsx'
      - '**.html'
      - '.github/workflows/ghas-js-codeql.yaml'
[...]

PR #2891, opened 2025-05-06
Allow configuring CODEQL_THREADS with an env var

austinpray-mixpanel:
closes https://github.com/github/codeql-action/issues/2890


aeisenberg (commented):
Thanks for your PR. I don't think this is harmful or could break anything, but it seems a bit redundant. 

Can you try this alternate approach?

In the `init/actions.yml` file add `default: "${{ env.CODEQL_THREADS }}"` to the `threads` input. This should be sufficient.

austinpray-mixpanel:
> Thanks for your PR. I don't think this is harmful or could break anything, but it seems a bit redundant.
> 
> Can you try this alternate approach?
> 
> In the `init/actions.yml` file add `default: "${{ env.CODEQL_THREADS }}"` to the `threads` input. This should be sufficient.

Yep I can give this a shot. My initial change was just trying to match up the way the ram variable was treated

aeisenberg:
> Yep I can give this a shot. My initial change was just trying to match up the way the ram variable was treated

Great. Let me know if this works for you. Now that I'm engaged, I'll be able to respond more quickly to this PR.

austinpray-mixpanel:
@aeisenberg lmk what you think 👍

aeisenberg (approved):
This looks right to me.

austinpray-mixpanel:
@aeisenberg ah I don't think you can use env variable expressions in inputs

```
Error: /Users/runner/work/codeql-action/codeql-action/./../action/init/action.yml (Line: 118, Col: 14): Unrecognized named-value: 'env'. Located at position 1 within expression: env.CODEQL_THREADS

```

aeisenberg:
Oh...that's really unfortunate. Apologies for the confusion. I'm asking internally what's going on since this is a pattern that's [being used elsewhere](https://github.com/search?q=path%3Aaction.yml%20%22default%3A%20%24%7B%7B%20env.%22&type=code).

aibaars:
> Oh...that's really unfortunate. Apologies for the confusion. I'm asking internally what's going on since this is a pattern that's [being used elsewhere](https://github.com/search?q=path%3Aaction.yml%20%22default%3A%20%24%7B%7B%20env.%22&type=code).

Most of those examples commented out the `default` field with a `#` though. I'm not sure why the `env` context is not allowed, perhaps it is not available until a workflow starts to "run".

austinpray-mixpanel:
In any case it probably makes sense to have the threads and ram behavior configuration be exactly the same so there aren't any surprises
image

henrymercer:
Thanks for the contribution @austinpray-mixpanel!  Could you follow the instructions in https://github.com/github/codeql-action/blob/main/CONTRIBUTING.md#common-tasks to rebuild the project?

austinpray-mixpanel:
@henrymercer done 👍

henrymercer (approved):

aeisenberg (approved):

austinpray-mixpanel:
Thank you all!

PR #2914, opened 2025-05-30
Allow using new CodeQL languages without updating the CodeQL Action

henrymercer:
This PR revives #779, adding a few new features:

- Correctly handle language aliases (these didn't exist at the time)
- Determine whether or not a language is traced by looking for its `tracing-config.lua` file — this is what we do in the CLI too
- Simplify the parsing logic in `config-utils.ts`

Commit-by-commit review recommended as one commit has a large diff from renaming the `Language` enum to `KnownLanguage`.

Caveats:

- We don't attempt to make language handling in the `start-proxy` Action generic, because this Action can't assume access to the CodeQL CLI, which is the source of truth.  However this is probably OK because the default behaviour of this Action, i.e. not to proxy anything, is OK for new languages.

### DIFF
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -1,4 +1,4 @@
-/** A language supported by CodeQL. */
+/** A language to analyze with CodeQL. */
 export type Language = string;
 
 /**


src/analyze.ts
```diff
@@ -177,14 +177,14 @@ export async function runExtraction(
...
+    if (await shouldExtractLanguage(codeql, config, language)) {
       logger.startGroup(`Extracting ${language}`);
-      if (language === Language.python) {
+      if (language === String(KnownLanguage.python)) {
         await setupPythonExtractor(logger);
       }
       if (config.buildMode) {
         if (
-          language === Language.cpp &&
+          language === String(KnownLanguage.cpp) &&
```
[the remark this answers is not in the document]
henrymercer:
```suggestion
          language === KnownLanguage.cpp &&
```

mbg (dismissed):
Looks good! My only (non-blocking) suggestion would be to add some documentation to the new functions (and possibly some of the updated ones that didn't already have any), but generally it's relatively obvious what they do so that's not urgent.

src/languages.ts
```diff
@@ -1,6 +1,10 @@
-// All the languages supported by CodeQL
+/**
+ * A CodeQL language.
+ *
+ * To facilitate adding new languages, this is _not_ an exhaustive list of
+ * languages supported by CodeQL.
```
mbg: Minor: I don't think it's clear who "To facilitate adding a new language" is aimed at. How about "This is _not_ an exhaustive list of languages supported by CodeQL and it is not required that a new language has to be added here"?

src/start-proxy.ts
```diff
@@ -11,6 +11,49 @@ export type Credential = {
...
 };
 
+/*
+ * Language aliases supported by the start-proxy Action.
+ *
+ * In general, the CodeQL CLI is the source of truth for language aliases, and to
+ * allow us to more easily support new languages, we want to avoid hardcoding these
+ * aliases in the Action itself.  However this is difficult to do in the start-proxy
+ * Action since this Action does not use CodeQL, so we're accepting some hardcoding
+ * for this Action.
```
mbg, line 23: This doesn't have to be addressed as part of this PR, but we only use the `start-proxy` in CodeQL (default setup) workflows. The `start-proxy` steps follows the `init` step, so we _could_ rely on that here if we wanted to.

src/start-proxy.ts
```diff
@@ -11,6 +11,49 @@ export type Credential = {
...
+  // See if it's an exact match
+  if (language in Language) {
+    return language as Language;
+  }
+
+  // Check language aliases, but return the original language name,
+  // the alias will be resolved later.
+  if (language in LANGUAGE_ALIASES) {
+    return LANGUAGE_ALIASES[language];
+  }
```
mbg: The comment seems to contradict what is happening here, since the resolved alias is returned here?

src/config-utils.ts
```diff
@@ -383,22 +362,24 @@ export async function getRawLanguages(
...
-  let autodetected: boolean;
-  if (rawLanguages.length) {
-    autodetected = false;
-  } else {
-    autodetected = true;
-
-    // Obtain all languages in the repo that can be analysed
-    rawLanguages = (await getLanguagesInRepo(repository, logger)) as string[];
+  // If the user has specified languages, use those.
+  if (languagesFromInput.length) {
```
mbg: I don't love this check relying on the truthiness of what `length` is. Adding `> 0` would probably be nicer. Could we skip the `|| ""` in `const languagesFromInput = (languagesInput || "")` entirely if we check here more explicitly whether `languagesInput` is something meaningful?

src/analyze-action.ts
```diff
@@ -151,7 +157,7 @@ function doesGoExtractionOutputExist(config: Config): boolean {
  * whether any extraction output already exists for Go.
  */
 async function runAutobuildIfLegacyGoWorkflow(config: Config, logger: Logger) {
-  if (!config.languages.includes(Language.go)) {
+  if (!config.languages.includes(KnownLanguage.go)) {
```
mbg, line 161: All of these `includes` checks probably need the same treatment as the one for `rust`

src/analyze-action.ts
```diff
@@ -95,7 +95,10 @@ async function sendStatusReport(
         ...report,
         trap_cache_upload_duration_ms: Math.round(trapCacheUploadTime || 0),
         trap_cache_upload_size_bytes: Math.round(
-          await getTotalCacheSize(Object.values(config.trapCaches), logger),
+          await getTotalCacheSize(
+            Object.values(config.trapCaches).filter((x) => x !== undefined),
+            logger,
+          ),
```
mbg: OOC what motivated this change as part of this commit?

src/autobuild.ts
```diff
@@ -37,7 +37,7 @@ export async function determineAutobuildLanguages(
     async (language) => await codeql.isTracedLanguage(language),
   );
 
-  if (!autobuildLanguages) {
+  if (!autobuildLanguages.length) {
```
mbg: This might need a comment (and also as with the other check like this, I'd prefer something more explicit than the truthiness of `length`).

src/init-action.ts
```diff
@@ -342,6 +342,31 @@ async function run() {
     }
     core.endGroup();
 
+    // Set CODEQL_ENABLE_EXPERIMENTAL_FEATURES for rust
```
mbg: How do the changes here interact with https://github.com/github/codeql-action/pull/2960?

mbg (commented):
Thanks for working on this and getting it all up-to-date after the other recent changes! Sorry for the premature review earlier as well. This broadly look good. I have a few suggestions and questions.

Should all the `.includes` checks for languages get the same treatment as that one check for `rust`?

src/languages.ts
henrymercer: Good call, I've updated to make this clearer.

src/start-proxy.ts
henrymercer: I agree with your italicisation :) We could rely on that ordering, but I'd prefer not to.

src/start-proxy.ts
henrymercer: Indeed, fixed!

src/config-utils.ts
henrymercer: I don't feel strongly about this, but I can see how a numerical check could be more readable, so have changed.

Regarding the `|| ""`, that's there since `getOptionalInput` returns undefined if the value is the empty string.

src/analyze-action.ts
henrymercer: This one isn't necessary since by the time we have a parsed configuration object, `config.languages` is an array.

src/init-action.ts
henrymercer: I don't see any problematic interactions, though there might be a couple of merge conflicts to address.

src/analyze-action.ts
henrymercer: This is because the type of `trapCaches` is a `Record<Language, string>` and as of this PR `Language` is now non-exhaustive.  I've updated the type of `trapCaches` to `{ [language: Language]: string }` so that we no longer need this check.

src/languages.ts
```diff
@@ -1,11 +1,12 @@
+/** A language supported by CodeQL. */
+export type Language = string;
```
mbg: Minor: Technically this now represents any language identifier, whether supported or not

mbg (approved):
Thanks for addressing the feedback! I think the change of type for the trap caches is good. :shipit:

PR #2917, opened 2025-06-03
Add new `quality-queries` input

mbg:


init/action.yml
```diff
@@ -83,6 +83,9 @@ inputs:
   queries:
     description: Comma-separated list of additional queries to run. By default, this overrides the same setting in a configuration file; prefix with "+" to use both sets of queries.
     required: false
+  quality-queries:
+    description: Comma-separated list of code quality queries to run.
```
esbena: - Does this also support the `+` semantics of `queries` above?
- Does the user need to know what the difference is between these two inputs? What happens if a query is present in both? What if a non-quality query is present in the quality suite?...

src/codeql.ts
```diff
@@ -1219,13 +1219,19 @@ async function generateCodeScanningConfig(
...
       augmentedConfig.queries = (augmentedConfig.queries || []).concat(
-        config.augmentationProperties.queriesInput,
+        config.augmentationProperties.queriesInput || [],
+        config.augmentationProperties.qualityQueriesInput || [],
       );
     } else {
-      augmentedConfig.queries = config.augmentationProperties.queriesInput;
+      augmentedConfig.queries = (
+        config.augmentationProperties.queriesInput || []
+      ).concat(config.augmentationProperties.qualityQueriesInput || []);
```
esbena: Nit: deduplication. Perhaps introduce a shared variable.

init/action.yml
mbg: Sorry, I thought I had replied to this.

> Does this also support the + semantics of queries above?

No, I omitted that for now to keep it simple, but I can add it if we think it makes sense. AFAIK it's not something that we would be using internally right now.

> Does the user need to know what the difference is between these two inputs? 

In default setup, we provide the inputs. For advanced setup, the user can specify this. In the future, we might want a simpler "on" switch for quality queries, but I don't think that would change us wanting to have this.

> What happens if a query is present in both? 

Then the resulting alerts should go in both SARIF uploads.

> What if a non-quality query is present in the quality suite?...

My understanding is that this isn't necessarily something we should be concerned with in the action right now. If we get an input to `quality-queries`, we should assume that those are quality queries. The backend filters out things that it doesn't think are quality queries (that's already happens, since the backend only gets one big SARIF file with all alerts right now).

esbena: OK. Would a user want to know any of these subtleties? Alternative: Should the input perhaps be marked as internal for the time being?

starcke: > My understanding is that this isn't necessarily something we should be concerned with in the action right now. If we get an input to quality-queries, we should assume that those are quality queries. The backend filters out things that it doesn't think are quality queries (that's already happens, since the backend only gets one big SARIF file with all alerts right now).

This is correct, the backend filters the SARIF based on a list of quality queries it knows about.

mbg: > Alternative: Should the input perhaps be marked as internal for the time being?

Happy to mark as internal or experimental for now.

mbg: I have pushed a commit which marks it as internal.

mbg:
> LGTM with the caveat that the splitting strategy for normal and quality queries still hasn't been done.
Depending on how that is done, the quality-queries and queries inputs may have to be propagated further into the action later, so the design now may as well take that into account.

Yes, that will be in a separate PR. The inputs are propagated already through the configuration file that is restored in other steps (`config.augmentationProperties.qualityQueriesInput` is available later).

sins921 (approved):

PR #2930, opened 2025-06-12
Register credentials extracted from proxy configuration as secrets

mbg:
This will cause the credentials extracted from the proxy configurations to subsequently be masked in the Actions log, reducing the probability of accidental leakage.


src/start-proxy.ts
```diff
@@ -63,6 +65,14 @@ export function getCredentials(
 
   const out: Credential[] = [];
   for (const e of parsed) {
+    // Mask credentials to reduce chance of accidental leakage in logs.
+    if (e.password !== undefined) {
```
navntoft, line 69: Consider testing for `null` as well to guard against future changes. Better safe than sorry in these cases...

navntoft (approved):

src/start-proxy.ts
mbg: I don't disagree, but we don't seem to check for `null` anywhere else (including elsewhere here where we inspect other members of the same JSON object). 

I think we can do a much better job at validating the JSON value we get in general (e.g. check that it actually is an array, rather than just assuming it), but that may be a larger chunk of work. I am not sure if it makes sense to change this one now and not the others - what do you think?

navntoft: Hmm, good question. I was about to write that it is probably more important to be consistent and make a sweep later with proper validation.

But: What is the impact if the structure of the JSON changes and makes our assumptions false in other places vs. here? In other places something might just crash, which we can hopefully detect and fix, but here we risk leaking secrets.

Your call though, as I haven't looked at the other places, and don't know the severity of the potential leak...

src/start-proxy.ts
```diff
@@ -63,6 +65,14 @@
 
   const out: Credential[] = [];
   for (const e of parsed) {
+    // Mask credentials to reduce change of accidental leakage in logs.
+    if (e.password !== undefined) {
+      core.setSecret(e.password);
```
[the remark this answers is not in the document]
mbg: This looks like a FP to me.

PR #2931, opened 2025-06-12
Fail `build.sh` if any command in it fails

mbg:
This PR updates `build.sh` so that it fails if any step in the script fails. Previously, the script would continue running and thus mask unexpected failures. For example, Go is no longer available by default on runners and so an explicit `setup-go` step was required. However, because the script didn't fail, we didn't notice that failure until now.

To address the Go issue, I updated `sync.sh` to understand a new `installGo` input that injects the `setup-go` step into the workflow.

I tried adding it to the local `prepare-test` action first, but this resulted in a number of further issues:

- Adding `setup-go` there, which has a post-action, forced the entire `prepare-test` action to have a post-action, which caused problems because the workflow also moves the location of the `prepare-test` action to outside of the workspace.
- Moving the local actions back to the workspace then caused a bunch of further issues with workflows not expecting there to be a meaningful `.github` folder in the workspace.

### DIFF
--- a/pr-checks/checks/all-platform-bundle.yml
+++ b/pr-checks/checks/all-platform-bundle.yml
@@ -3,12 +3,13 @@ description: "Tests using an all-platform CodeQL Bundle"
 versions: ["nightly-latest"]
 operatingSystems: ["ubuntu"]
 useAllPlatformBundle: "true"
+installGo: "true"
 steps:
   - id: init
     uses: ./../action/init
     with:
       # Swift is not supported on Ubuntu so we manually exclude it from the list here
-      languages: cpp,csharp,go,java,javascript,python,ruby 
+      languages: cpp,csharp,go,java,javascript,python,ruby
       tools: ${{ steps.prepare-test.outputs.tools-url }}
   - name: Build code
     shell: bash
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -108,6 +108,22 @@ for file in (this_dir / 'checks').glob('*.yml'):
         },
     ]
 
+    installGo = False
+    if checkSpecification.get('installGo'):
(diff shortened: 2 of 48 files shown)


pr-checks/sync.py
```diff
@@ -108,6 +108,22 @@ def writeHeader(checkStream):
         },
     ]
 
+    installGo = False
+    if checkSpecification.get('installGo'):
+        installGo = True if checkSpecification['installGo'].lower() == "true" else False
```
redsun82, line 113: if we just use
```suggestion
    installGo = checkSpecification.get('installGo')
```
then we can use a more natural
```yaml
installGo: true
```
in the test specification, without using stringly typed values. Although I see we use `"true"` for other cases as well.

pr-checks/checks/all-platform-bundle.yml
```diff
@@ -3,12 +3,13 @@ description: "Tests using an all-platform CodeQL Bundle"
 versions: ["nightly-latest"]
 operatingSystems: ["ubuntu"]
 useAllPlatformBundle: "true"
+installGo: "true"
```
redsun82, line 6: bike shedding: personally I would favour kebab-case (`install-go`) for yaml configuration keys, but seeing that the existing convention is camelcase I guess we should go for it

mbg: Yeah, I went with this for consistency. I don't see any strong reason to change this convention across the board.

pr-checks/sync.py
mbg: Yeah, like with the other point, I went with this to mirror the existing inputs. I would be happy to change this though if we can't think of a downside / motivation for why the others are strings.

PR #2935, opened 2025-06-16
Produce separate SARIF file for `quality-queries` alerts

mbg:
Follows on from #2917.

This PR modifies the action to produce separate SARIF files for code quality queries (as specified as arguments to the `quality-queries` input) and uploads them to the code quality API. The approach here is that:

- We remove the logic from #2917 that adds the `quality-queries` input to the code scanning configuration file that's used when initialising the database.
- That means that the quality queries are no longer included in the `config-queries.qls` that's generated by the CLI.
- When we call `database run-queries`, and if quality queries are enabled, we explicitly specify `config-queries.qls` (which the CLI normally does by default if nothing is specified) + the quality queries.
- Since the quality queries are only specified as argument to `database run-queries`, the existing call to `database interpret-results` generates a SARIF that only includes results relating to security queries (i.e. `config-queries.qls`).
- We make an additional call to `database interpret-results` for queries that were specified as arguments to the `quality-queries` input (if any). This results in SARIF files for those queries.
- We then upload them to the CQ API endpoint.

### Notes

- The changes aim to be as backwards-compatible as possible. That includes:
  - SARIF files are output to the same place as before (e.g. as opposed to having separate subdirectories for security and quality SARIFs). SARIF files containing quality results are identified by the `.quality.sarif` extension.
- We don't support `query-filter` options with respect to code quality results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Experimental: When the `quality-queries` input for the `init` action is provided with an argument, separate `.quality.sarif` files are produced and uploaded for each language with the results of the specified queries. Do not use this in production as it is part of an internal experiment and subject to change at any time. [#2376](https://github.com/github/codeql-action/pull/2935)
 
 ## 3.29.1 - 27 Jun 2025
 


src/analyze.ts
```diff
@@ -561,6 +561,34 @@ extensions:
...
+ * If `query` is the name of a default query suite, it is resolved into the corresponding
+ * query suite name for the given `language`. Otherwise, `query` is returned as is.
+ *
+ * @param language The language for which to resolve the default query suite name.
+ * @param query The string that potentially contains the name of a default query suite.
+ * @returns Returns the resolved query suite name, or the unmodified input.
+ */
+export function resolveQuerySuiteAlias(
+  language: Language,
+  query: string,
```
esbena: Nit: I think should be named `suite` instead of `query`, but if there's precedence for using the `query` name elsewhere, then just keep it.

src/analyze.ts
```diff
@@ -561,6 +561,34 @@ extensions:
...
 }
 
+// A set of default query suite names that are understood by the CLI.
+export const defaultSuites: Set<string> = new Set([
+  "security-experimental",
+  "security-extended",
+  "security-and-quality",
+  "code-quality",
+  "code-scanning",
+]);
```
esbena, line 571: This is my least favorite part of the PR. As it leads to double maintenance.

It would be great if the same resolution that the ordinary `.queries` uses could be reused here. In the interest of time, we don't need to do it right now though. Could you perhaps create an issue with a list of things to follow up on later?

src/codeql.ts
```diff
@@ -159,7 +159,11 @@ export interface CodeQL {
   /**
    * Run 'codeql database run-queries'.
    */
-  databaseRunQueries(databasePath: string, flags: string[]): Promise<void>;
+  databaseRunQueries(
+    databasePath: string,
+    flags: string[],
+    queries?: string[],
```
esbena, line 167: this deserves a jsdoc comment.
I'm guessing that the semantics is that she CLI will magically look up the query suite to use if nothing explicitly is provided by the user, and not that it will run zero queries.

src/upload-lib.ts
```diff
@@ -306,12 +306,19 @@ function getAutomationID(
   return api.computeAutomationID(analysis_key, environment);
 }
 
+// Enumerates API endpoints that accept SARIF files.
+export enum SARIF_UPLOAD_TARGET {
```
esbena: Nit: `SARIF_UPLOAD_ENDPOINT` is a more standard naming scheme for `<method> <path>` strings like this.

src/upload-lib.ts
```diff
@@ -376,14 +380,21 @@ export interface UploadResult {
   sarifID: string;
 }
 
+export const qualityIsSarif = (name: string) => name.endsWith(".quality.sarif");
+export const defaultIsSarif = (name: string) =>
+  name.endsWith(".sarif") && !qualityIsSarif(name);
```
esbena: could we move these definitions into `UploadTarget` instances below, so you *always* access them through their associated target. I'd like a very strong coupling between the components of such targets to avoid a fragment being used wrongly.

That is: I love seeing `sentinelPrefix: string = CodeScanningTarget.sentinelPrefix,` below rather than a direct reference to a `codeScanningSentinelPrefix`

src/upload-lib.ts
```diff
@@ -394,15 +405,18 @@ export function findSarifFilesInDir(sarifPath: string): string[] {
   return sarifFiles;
 }
 
-function getSarifFilePaths(sarifPath: string) {
+export function getSarifFilePaths(
+  sarifPath: string,
+  isSarif: (name: string) => boolean = defaultIsSarif,
```
esbena: Rambling:

Generally, I'd like us to not introduce these opinionated default arguments, but I suppose it simplifies the diff, so lets just keep it.

The reason I find defaults opinionated is that they aren't "safe and intuitive" values like `0, "", []`, but rather a specific historic behaviour.

src/upload-lib.ts
```diff
@@ -567,6 +581,30 @@ export function buildPayload(
   return payloadObj;
 }
 
+// Represents configurations for different services that we can upload SARIF to.
+export interface UploadTarget {
```
esbena, line 581: regarding the target/endpoint comment above: this is a much better place to call something target. I'd probably prefer `SarifUploadConfig` but that is splitting hairs.

esbena, line 581: I really like to see this config pattern used here. Good choice.

src/upload-lib.ts
```diff
@@ -567,6 +581,30 @@ export function buildPayload(
...
+  name: string;
+  target: SARIF_UPLOAD_TARGET;
+  sarifFilter: (name: string) => boolean;
+  sentinelPrefix: string;
+}
+
+// Represents the Code Scanning upload target.
```
esbena, line 588: Stray thought. I have seen both line comments and jsdoc comments on your new functions. Is it deliberate that you do not use jsdoc here? If yes, OK.

src/upload-sarif-action.ts
```diff
@@ -83,15 +83,33 @@ async function run() {
...
+      sarifPath,
+      checkoutPath,
+      category,
       features,
       logger,
     );
     core.setOutput("sarif-id", uploadResult.sarifID);
 
+    // If there are `.quality.sarif` files in `sarifPath`, then upload those to the code quality service.
+    const qualitySarifFiles = upload_lib.getSarifFilePaths(sarifPath);
```
esbena: Bug: `getSarifFilePaths(sarifPath);` is this missing a quality specialization through some argument, or is there some hidden state elsewhere?

I think this should be `getSarifFilePaths(sarifPath,         upload_lib.CodeQualityTarget.sarifFilter)`,

src/upload-lib.ts
esbena: Also. Please access the default through the `Target` object, as mentioned above.

src/upload-lib.ts
```diff
@@ -567,6 +581,30 @@ export function buildPayload(
   return payloadObj;
 }
 
+// Represents configurations for different services that we can upload SARIF to.
+export interface UploadTarget {
+  name: string;
+  target: SARIF_UPLOAD_TARGET;
+  sarifFilter: (name: string) => boolean;
```
esbena: Nit: a `filter` is typically a function of `Array -> SubsetArray`, whereas this is a `predicate`. The relation being that a filter is instantiated with a predicate.

esbena (commented):
Nice, and pleasantly surgical.

I think this does the right thing, except for a missing quality.sarif predicate in the very end.

Other than that it's mostly nits, and some concerns about maintenance in the long term. We have already discussed them elsewhere. Perhaps we should update the discussion document with our current choice for query resolution.

src/upload-sarif-action.ts
mbg: Probably more than the bug itself, this shows that there's lacking test coverage for this part. I will see what I can add without too much refactoring, or track the need for this in an issue.

src/upload-lib.ts
mbg: I think the existing codebase is fairly inconsistent in this regard. I probably just subconsciously mirrored what adjacent definitions do. I'll make this more consistent for my changes.

src/upload-lib.ts
mbg: > Generally, I'd like us to not introduce these opinionated default arguments, but I suppose it simplifies the diff, so lets just keep it.

Primarily, it just made working on this simpler since it required fewer initial changes until the semantics were right. However, I think it's a good idea to remove them since it could highlight places where the argument might need to be different.

More secondary is that I still have the issue from a few weeks ago in the back of my mind, where changing a function signature here broke our workflows elsewhere.

src/upload-lib.ts
mbg: Absolutely, I think that's a good idea. The only reason it's not already in `UploadTarget` is because I added this part before `UploadTarget` was a thing, and then didn't think to just move it into that. Thanks for noticing!

src/analyze.ts
mbg: I agree that having this in the action isn't great - this is one of the bits I was thinking of yesterday when we were talking. I'll create an issue for it.

src/analyze.ts
mbg: Yeah, I wasn't happy with the naming for `query` (here and possibly elsewhere) while working on it. I think I'll go with `maybeSuite` since it _could_ be any number of things and, in this function, we care only about what it if it is a (default) query suite.

src/upload-sarif-action.ts
mbg: I have added a test workflow for this, separate from the existing `sarif-upload` one. That's so that the existing test checks the case where there's only security SARIFs and the new one tests the case where there are both security and quality SARIFs.

src/upload-lib.ts
```diff
@@ -306,12 +306,19 @@ function getAutomationID(
   return api.computeAutomationID(analysis_key, environment);
 }
 
+// Enumerates API endpoints that accept SARIF files.
+export enum SARIF_UPLOAD_ENDPOINT {
+  CODE_SCANNING_UPLOAD_TARGET = "PUT /repos/:owner/:repo/code-scanning/analysis",
+  CODE_QUALITY_UPLOAD_TARGET = "PUT /repos/:owner/:repo/code-quality/analysis",
+}
+
```
esbena:
```suggestion
export enum SARIF_UPLOAD_ENDPOINT {
  CODE_SCANNING = "PUT /repos/:owner/:repo/code-scanning/analysis",
  CODE_QUALITY = "PUT /repos/:owner/:repo/code-quality/analysis",
}

```

esbena (dismissed):
thanks! There's one naming nit left, but it is quite minor.

src/upload-lib.ts
mbg: Makes sense, done!

esbena (approved):

jsoref:
This appears to talk about another github api endpoint. Is it documented / will it be documented?

mbg:
@jsoref It is not documented. This is part of an internal experiment and so is no use to anyone but us right now.

PR #2941, opened 2025-06-23
Add workflow for updating release used by `start-proxy`

mbg:
This PR adds a new workflow which automates the creation of PRs that update which release is used by the `start-proxy` action to pull binaries of the `update-job-proxy` from. 

Previously, I manually created PRs [such as this one](https://github.com/github/codeql-action/pull/2869) for that. The new workflow here creates similar ones, [such as this example](https://github.com/github/codeql-action/pull/2944).

This PR currently contains a commit that adds a `push` trigger (limited to this branch) for testing, which should be removed before merging this PR.

### DIFF
--- a/.github/workflows/update-proxy-release.yml
+++ b/.github/workflows/update-proxy-release.yml
@@ -29,6 +29,13 @@ jobs:
             exit 1
           fi
 
+      - name: Check that the release exists
+        shell: bash
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        run: |
+          (gh release view --repo ${{ github.event.repository.full_name }} --json "assets" "$RELEASE_TAG" && echo "Release found.") || exit 1
+
       - name: Install Node
         uses: actions/setup-node@v4
 
--- a/.github/workflows/update-proxy-release.yml
+++ b/.github/workflows/update-proxy-release.yml
@@ -49,7 +49,7 @@ jobs:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         run: |
           set -exu
-          pr_title="Update release used by `start-proxy` to $RELEASE_TAG"
+          pr_title="Update release used by \`start-proxy\` to $RELEASE_TAG"
           pr_body=$(cat << EOF
             This PR updates the `start-proxy` action to use the private registry proxy binaries that
             are attached as release assets to the `$RELEASE_TAG` release.
--- a/.github/workflows/update-proxy-release.yml
+++ b/.github/workflows/update-proxy-release.yml
@@ -1,5 +1,8 @@
 name: Update dependency proxy release assets
 on:
+  push:
+    branches:
+      - mbg/update-proxy-binaries # for testing
   workflow_dispatch:
     inputs:
       tag:
@@ -16,7 +19,7 @@ jobs:
       contents: write # needed to push the updated files
       pull-requests: write # needed to create the PR
     env:
-      RELEASE_TAG: ${{ inputs.tag }}
+      RELEASE_TAG: ${{ inputs.tag || 'codeql-bundle-v2.22.0' }}
     steps:
       - name: Install Node
         uses: actions/setup-node@v4
--- a/.github/workflows/update-proxy-release.yml
+++ b/.github/workflows/update-proxy-release.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Push changes and open PR
         shell: bash
         env:
-          BRANCH: "dependency-proxy/$RELEASE_TAG"
+          BRANCH: "dependency-proxy/${{ env.RELEASE_TAG }}"
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         run: |
           set -exu
--- a/.github/workflows/update-proxy-release.yml
+++ b/.github/workflows/update-proxy-release.yml
@@ -40,7 +40,7 @@ jobs:
         shell: bash
         run: |
           NOW=$(date +"%Y%m%d%H%M%S") # only used to make sure we don't fetch stale binaries from the toolcache
-          sed -i '' "s|https://github.com/github/codeql-action/releases/download/codeql-bundle-[0-9.]*/|https://github.com/github/codeql-action/releases/download/$RELEASE_TAG/|g" ./src/start-proxy-action.ts
+          sed -i '' "s|https://github.com/github/codeql-action/releases/download/codeql-bundle-v[0-9.]*/|https://github.com/github/codeql-action/releases/download/$RELEASE_TAG/|g" ./src/start-proxy-action.ts
           sed -i '' "s/\"v2.0.[0-9]*\"/\"v2.0.$NOW\"/g" ./src/start-proxy-action.ts
 
       - name: Push changes and open PR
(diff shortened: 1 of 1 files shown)


.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,78 @@
...
+jobs:
+  update:
+    name: Update code and create PR
+    timeout-minutes: 15
+    runs-on: macos-latest
+    permissions:
+      contents: write # needed to push the updated files
+      pull-requests: write # needed to create the PR
+    env:
+      RELEASE_TAG: ${{ inputs.tag || 'codeql-bundle-v2.22.0' }}
```
esbena: What is the plan for this default value moving forward?

.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,78 @@
...
+      tag:
+        description: "The tag of CodeQL Bundle release that contains the proxy binaries as release assets"
+        type: string
+        required: true
+
+jobs:
+  update:
+    name: Update code and create PR
+    timeout-minutes: 15
+    runs-on: macos-latest
```
esbena: is macos required?

esbena (commented):
Is there prior art for doing releases like this? 

I'm not a huge fan of doing text replacements in typescript files with workflow inputs...

.github/workflows/update-proxy-release.yml
mbg: That's part of the test commit, which I noted in the PR description will be removed. So there will be no default value, and the `tag` input is required for the `workflow_dispatch` event.

.github/workflows/update-proxy-release.yml
mbg: I noticed that the some other workflows which run `npm` commands (i.e. `pr-checks.yml` and `update-dependencies.yml`) run on macos rather than ubuntu, and so mirrored that here in case there was a good reason. I suppose it wouldn't hurt to try this on an ubuntu runner to see if it still works OK.

mbg:
> I'm not a huge fan of doing text replacements in typescript files with workflow inputs...

Agreed, perhaps a check that the tag is in the expected format before proceeding would be good?

> Is there prior art for doing releases like this?

Depends on what you are asking about specifically. This PR just automates a few steps we'd have to perform manually otherwise. The whole arrangement with how the `update-job-proxy` binaries are made available to the action has always been a bit temporary, but we haven't prioritised improvements here (see the linked epic).

esbena:
> Agreed, perhaps a check that the tag is in the expected format before proceeding would be good?

Yes. Let's add that in. It's also nice to have to prevent silly bugs.

> Depends on what you are asking about specifically.

NP. Thanks for the extra context.

.github/workflows/update-proxy-release.yml
mbg: Seems to work fine on `ubuntu-latest` 👍🏻

mbg:
> Yes. Let's add that in. It's also nice to have to prevent silly bugs.

Done, along with an additional check to make sure that the release actually exists. We could take this further to check that the release assets include the three expected files, but this is probably enough sanity checking in the workflow for now -- I am not expecting this to be used much. I'd only expect an increase in usage if we automate the process end-to-end, but we may well choose to store the proxy binaries elsewhere then.

.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,100 @@
...
+          fi
+
+          echo "target_branch=dependency-proxy/$RELEASE_TAG" >> $GITHUB_OUTPUT
+
+      - name: Check that the release exists
+        shell: bash
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        run: |
+          (gh release view --repo ${{ github.event.repository.full_name }} --json "assets" "$RELEASE_TAG" && echo "Release found.") || exit 1
```
esbena:
```suggestion
          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
        run: |
          (gh release view --repo "$GITHUB_REPOSITORY" --json "assets" "$RELEASE_TAG" && echo "Release found.") || exit 1
```
This does two things: 
- quotes the repository name, which is just a good habit
- eliminates the dynamic construction of bash syntax, which is a must to me

Given the constraints on repository names, this is probably not needed though

.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Install Node
+        uses: actions/setup-node@v4
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # ensure we have all tags and can push commits
+          ref: main
```
esbena, line 46: If we move this checkout earlier than the `Check that the release exists` step, then `gh` will pick up the repository to query automatically, allowing us to drop the `--repo`. Checkout is fast enough that it can go before the error checking IMO.

.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          NOW=$(date +"%Y%m%d%H%M%S") # only used to make sure we don't fetch stale binaries from the toolcache
+          sed -i "s|https://github.com/github/codeql-action/releases/download/codeql-bundle-v[0-9.]\+/|https://github.com/github/codeql-action/releases/download/$RELEASE_TAG/|g" ./src/start-proxy-action.ts
+          sed -i "s/\"v2.0.[0-9]\+\"/\"v2.0.$NOW\"/g" ./src/start-proxy-action.ts
+
+      - name: Compile TypeScript and commit changes
+        shell: bash
+        run: |
+          set -exu
+          git checkout -b "${{ steps.checks.outputs.target_branch }}"
```
esbena:
```suggestion
        shell: bash
        env:
          TARGET_BRANCH: ${{ steps.checks.outputs.target_branch }}
        run: |
          set -exu
          git checkout -b "$TARGET_BRANCH"
``````

File: .github/workflows/update-proxy-release.yml [hunk 6]
```diff
@@ -0,0 +1,100 @@
...
+
+
+            Please do the following before merging:
+
+            - [ ] Verify that the changes to the code are correct.
+            - [ ] Mark the PR as ready for review to trigger the CI.
+          EOF
+          )
+
+          git push origin "${{ steps.checks.outputs.target_branch }}"
```
esbena: As above: use an environment variable instead of creating bash syntax dynamically.

.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,100 @@
...
+          gh pr create \
+            --head "${{ steps.checks.outputs.target_branch }}" \
```
esbena: ditto

.github/workflows/update-proxy-release.yml
```diff
@@ -0,0 +1,100 @@
...
+            --base "main" \
+            --title "${pr_title}" \
+            --body "${pr_body}" \
+            ${{ (github.event_name == 'workflow_dispatch' && '--draft') || '--dry-run' }}
```
esbena: ditto

.github/workflows/update-proxy-release.yml
mbg: I had considered that, but explicitly adding the `--repo` argument isn't much of an inconvenience so it makes more sense to me to perform the check first.

mbg:
First workflow run on `main`: https://github.com/github/codeql-action/actions/runs/15848800951/job/44677031711

Resulting PR: https://github.com/github/codeql-action/pull/2947

Looks like it is working fine outside of test conditions, too.

PR #2945, opened 2025-06-23
Basic support for overlay PR analysis

cklin:


src/config-utils.ts
```diff
@@ -711,6 +767,115 @@ function parseQueriesFromInput(
...
+    return nonOverlayAnalysis;
+  }
+
+  if (buildMode !== BuildMode.None) {
+    logger.warning(
+      `Cannot build an ${overlayDatabaseMode} database because ` +
+        `build-mode is set to "${buildMode}" instead of "none". ` +
+        "Falling back to creating a normal full database instead.",
+    );
+    return nonOverlayAnalysis;
```
nickrolfe: I think we need to check whether PR analyses for dynamic languages like Ruby actually have the build mode defined as `None`, or if it's left undefined.

src/config-utils.ts
```diff
@@ -711,6 +767,115 @@ function parseQueriesFromInput(
...
+    modeEnv === OverlayDatabaseMode.OverlayBase ||
+    modeEnv === OverlayDatabaseMode.None
+  ) {
+    overlayDatabaseMode = modeEnv;
+    logger.info(
+      `Setting overlay database mode to ${overlayDatabaseMode} ` +
+        "from the CODEQL_OVERLAY_DATABASE_MODE environment variable.",
+    );
+  } else if (
+    ["github", "dsp-testing"].includes(repository.owner) &&
```
nickrolfe, line 823: Can you explain why we can't just use the feature flag to restrict overlay analysis to these orgs?

src/overlay-database-utils.ts
```diff
@@ -122,3 +126,210 @@ function computeChangedFiles(
...
+    logger.warning("Downloaded overlay-base database failed validation");
+    return false;
+  }
+
+  logger.info(`Successfully downloaded overlay-base database to ${dbLocation}`);
+  return true;
+}
+
+function generateCacheKey(config: Config, codeQlVersion: string): string {
+  const sha = process.env.GITHUB_SHA || "unknown";
```
nickrolfe: If the `GITHUB_SHA` env var is somehow not set, should we instead skip uploading to the actions cache?

src/overlay-database-utils.ts
```diff
@@ -122,3 +126,210 @@ function computeChangedFiles(
...
+  if (isInTestMode()) {
+    logger.debug(
+      "In test mode. Skip downloading overlay-base database from cache.",
+    );
+    return false;
+  }
+
+  const dbLocation = config.dbLocation;
+  const codeQlVersion = (await codeql.getVersion()).version;
+  const restoreKey = getCacheRestoreKey(config, codeQlVersion);
```
nickrolfe, line 277: I don't follow how this can work, since the `saveKey` computation appends either a SHA or `unknown` to the result of `getCacheRestoreKey()`, and we're not appending that suffix here.

nickrolfe (commented):
I have a few questions/comments, but overall it looks very sensible.

src/config-utils.ts
cklin: This PR does not fully implement all the functionality that we planned, so it is considered an early prototype and suitable for internal testing only.  So we want to be sure that we never attempt to perform overlay analysis on user repositories _with this implementation_, even if a user were to pin their workflow to this PR after merge, and then we enable all feature flags.

That is why I am using an explicit allowlist here instead of relying on feature flags.

src/overlay-database-utils.ts
cklin: Good point.  In addition to what you said, there is another reason why we should not be relying on `GITHUB_SHA`: the environment variable, even if it exists, may not necessarily correspond to the commit that is analyzed.  It is better to call `getCommitOid()` instead.

src/overlay-database-utils.ts
cklin: This code relies on the actions cache feature of performing restores using prefix match on the cache key:
https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

cklin:
Thank you for your feedback @nickrolfe!

I will take the PR back to draft to address comments and resolve merge conflicts.  Once that is done, I will mark the PR as ready for review again.

src/config-utils.ts
cklin: I updated the code so that it checks for build mode only for traced languages.

src/overlay-database-utils.ts
cklin: I updated the code to use `getCommitOid()`.

cklin:
@nickrolfe The PR is ready for another look.

src/analyze-action.ts
```diff
@@ -291,8 +295,15 @@ async function run() {
...
+    // An overlay-base database should always use the 'overlay' cleanup level
+    // to preserve the cached intermediate results.
+    //
+    // Note that we may be overriding the 'cleanup-level' input parameter.
     const cleanupLevel =
-      actionsUtil.getOptionalInput("cleanup-level") || "brutal";
+      config.augmentationProperties.overlayDatabaseMode ===
+      OverlayDatabaseMode.OverlayBase
+        ? "overlay"
+        : actionsUtil.getOptionalInput("cleanup-level") || "brutal";
```
redsun82, line 306: Wouldn't it be better to raise a configuration error if we explicitly set anything else than `overlay` or nothing if we are using overlay database? I think we shouldn't silently ignore an explicit parameter by the user

src/config-utils.ts
redsun82: this seems to be worth a `TODO` comment :slightly_smiling_face:

src/feature-flags.ts
```diff
@@ -142,6 +144,11 @@ export const featureConfig: Record<
     envVar: "CODEQL_ACTION_EXTRACT_TOOLCACHE",
     minimumVersion: undefined,
   },
+  [Feature.OverlayAnalysis]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_OVERLAY_ANALYSIS",
+    minimumVersion: CODEQL_OVERLAY_MINIMUM_VERSION,
```
redsun82, line 156: Theoretically as stated on line 86, we should
> Prefer using `ToolsFeature`s for future flags

Maybe we're not very thorough with that, and as far as I can see we did not add `overlayDatabases` to the list of CLI features that are listed in `codeql version --format=json`. We should have, right?

src/config-utils.ts
mbg: There is also a potential concern here for outside of dotcom if this ends up in a release, since it is possible for customers on other instances to have organisations with specific names. Two options:

- If you want to get this merged, then I think it would be good to put this behind a FF now, which then also helps us going forward.
- If it's really for internal testing only, then you may not need to get this PR merged at all, if you are able to reference this branch directly in testing workflows

nickrolfe (approved):
I re-reviewed the parts I previously commented on, and I'm happy with the changes.

src/actions-util.ts
```diff
@@ -363,3 +364,48 @@ export const restoreInputs = function () {
...
+  head: string;
+}
+
+/**
+ * Returns the base and head branches of the pull request being analyzed.
+ *
+ * @returns the base and head branches of the pull request, or undefined if
+ * we are not analyzing a pull request.
+ */
+export function getPullRequestBranches(): PullRequestBranches | undefined {
```
mbg, line 379: It would be good to have unit tests for this which exercise the different conditions.

src/analyze-action.ts
mbg: I agree with @redsun82. It might be surprising if the `cleanup-level` option is silently overriden. An error if `cleanup-level` is set and overlay databases are enabled would make sense.

src/analyze.ts
```diff
@@ -6,7 +6,11 @@ import * as io from "@actions/io";
 import del from "del";
 import * as yaml from "js-yaml";
 
-import * as actionsUtil from "./actions-util";
+import {
+  getRequiredInput,
+  getTemporaryDirectory,
+  PullRequestBranches,
+} from "./actions-util";
```
mbg, line 13: Minor: I note that this change results in two unrelated changes in this file. We have imports like `import * as actionsUtil` in a lot of places already, so I am not sure there's a good reason for changing this here as part of this PR?

src/feature-flags.ts
mbg: Agreed that it would be better to use the CLI's features list than set a `minimumVersion` here.

src/init-action.ts
```diff
@@ -297,6 +299,11 @@ async function run() {
 
   const configFile = getOptionalInput("config-file");
 
+  const sourceRoot = path.resolve(
+    getRequiredEnvParam("GITHUB_WORKSPACE"),
+    getOptionalInput("source-root") || "",
+  );
```
mbg, line 308: It may be worth adding a short comment here to explain that `resolve` respects the intended semantics of `source-root` (relative to the workspace root by default, but can be absolute).

src/config-utils.ts
mbg: Sorry, I only reviewing this fully now, so I think I missed that you have already added the FF and reasoned why you should have both checks. I am not sure we should really be concerned with users pinning feature branches. That's not a concern for default setup, and if they did it with an advanced setup, then they are probably aware that it is at their own risk (they could also just enable the environment variable).

src/overlay-database-utils.ts
```diff
@@ -122,3 +126,218 @@ function computeChangedFiles(
...
+
+  logger.info(
+    `Looking in Actions cache for overlay-base database with restore key ${restoreKey}`,
+  );
+
+  try {
+    const foundKey = await withTimeout(
+      MAX_CACHE_OPERATION_MS,
+      actionsCache.restoreCache([dbLocation], restoreKey),
```
mbg, line 286: Quick note here: `restoreCache` treats the `dbLocation` as essentially part of the cache key. In other words, it will _not_ restore the cache if `dbLocation` is different from the path that was used to store the cache. I am not sure whether `dbLocation` is currently stable like that or not. If it is already stable between runs, then that may not be a guarantee that we have previously made and relied on.

src/overlay-database-utils.ts
```diff
@@ -122,3 +126,218 @@ function computeChangedFiles(
...
+  // The restore key (prefix) specifies which cached overlay-base databases are
+  // compatible with the current analysis: the cached database must have the
+  // same cache version and the same CodeQL bundle version.
+  //
+  // Actions cache supports using multiple restore keys to indicate preference.
+  // Technically we prefer a cached overlay-base database with the same SHA as
+  // we are analyzing. However, since overlay-base databases are built from the
+  // default branch and used in PR analysis, it is exceedingly unlikely that
+  // the commit SHA will ever be the same, so we can just leave it out.
+  const languages = [...config.languages].sort().join("_");
```
mbg, line 341: Henry has an [open PR](https://github.com/github/codeql-action/pull/2914) to open up languages more, which might affect this in the future. Not necessarily something we need to consider here, but may need to consider there.

src/overlay-database-utils.ts
```diff
@@ -122,3 +126,218 @@ function computeChangedFiles(
...
+  return `${CACHE_PREFIX}-${CACHE_VERSION}-${languages}-${codeQlVersion}-`;
```
mbg, line 342: I am wondering if we should include something to identify the workflow here (e.g. the analysis category). For advanced setup, it is possible to have multiple workflows for the same language.

mbg (commented):
Thank you for preparing this PR -- I am very excited for this functionality! 💖 

As I suggested in my earlier comment, if the main goal is to test this internally for now, then it may be best to leave this PR unmerged and reference the PR branch explicitly in test workflows. Have you done any testing like that yet? If not, it would be good to do some of that in any case.

If we want to merge this, then the changes will end up in the next release. Before we do that, I think we should address a few points:

- Before we merge this, we should add some test coverage. At this stage, it would be good to gain confidence that none of the new logic interferes with existing functionality when the feature is disabled. The existing PR checks do that to an extent, but we should verify that we cover all cases. In the long run, we will want test coverage for when the feature is enabled as well.
- There are general concerns around us using the Actions cache, similar to what we have for TRAP caching and dependency caching. I.e. we might be eating away at the cache quota that may already be in use by other workflows in a given repository. Can we mitigate our impact here? For example, by ensuring that we only have one cache at a time or limiting the size of the caches we store. This may not need to be addressed at this point if we only plan to use this internally for now, but it should be addressed before we ship it more widely, especially if the goal is to have it on by default at some point.

src/config-utils.ts
cklin: > this seems to be worth a TODO comment 🙂

Yes, that is a good suggestion.  I will do that.

> There is also a potential concern here for outside of dotcom if this ends up in a release, since it is possible for customers on other instances to have organisations with specific names.

Thanks for bringing that up—I haven't considered that possibility.  That said, we do have a feature flag on top of the allow list (your option 1 above), so GHES users should never reach this check.

> If it's really for internal testing only, then you may not need to get this PR merged at all, if you are able to reference this branch directly in testing workflows

The internal testing can include staff-ship to internal repositories, and it would be ideal if we can do so through feature flags without requiring workflow file changes.

> I am not sure we should really be concerned with users pinning feature branches.

Once this PR merges, a user could get this code by pinning a commit on `main`, which I think is a use case we should support.

src/analyze-action.ts
cklin: I am a bit conflicted about what to do here.

On the one hand, silently ignoring explicit user input (which I am doing here) is not considered proper behavior.

On the other hand, it is important for the action to maintain backward compatibility.  If we were to throw an error for users setting `cleanup-level` to a value other than `overlay`, we would break those (currently working) workflows when we roll out overlay analysis.  I can't think of any reason for users to set `cleanup-level` to `overlay`, especially since it is currently not even a documented allowed value.

Another option is to not build an overlay-base database if the user specifies a `cleanup-level` that is not `overlay`.  There is no good way to do that because the decision to build the overlay-base database was made in the `init` step, and the `cleanup-level` input is in the `analyze` step.  We could obey the `cleanup-level` input and not upload the overlay-base database if the input is not `overlay`, which means that the workflow will not benefit from overlay analysis.  It becomes a matter of which we think the user would care more about: getting the benefit of overlay analysis when we roll it out, or having the `cleanup-level` input strictly obeyed?

Of these three options, I think the currently implemented behavior is the best choice.  We can make the behavior change more explicit, by updating the `cleanup-level` description to state that it is advisory, and (if you consider it appropriate) to add a changelog entry stating the same.  At the end of the day, the original assumption when we added `cleanup-level` (that database cleanup is a choice that has no impact on how the action itself functions) will no longer hold, and I think it is not unreasonable for the meaning of the input to change accordingly.

src/feature-flags.ts
cklin: The CLI interface for overlay analysis has not yet been completely finalized, which is one of the reasons why we have both the feature flag check and the explicit allowlist for internal orgs.  We can use `ToolsFeature`s, but that would require us to come up with a new `ToolsFeature` if there should be an upcoming change that is not backwards compatible.  Setting a minimum version number makes it easier to iterate.

There is also the question of how much we want to lean on the version or feature check.  Support for overlay analysis is not just a feature of the CLI—it is a feature of the CLI and the compiled queries.  `ToolsFeature` tells us whether we can pass the overlay analysis flags to the CLI and not get an error.  But it does not tell us whether the query packs are ready for overlay analysis.  By setting a minimum version number, we can version not only the CLI but also the precompiled query packs included in the bundle, which helps us avoid unintended enablement.

(In the long run, we do plan to make the CLI robust enough so that it can correctly handle existing pre-compiled query packs for overlay analysis.  For now internal testing focuses on query packs included in the bundle, and the minimum version serves as a useful check.)

src/init-action.ts
cklin: This is not new code—it is existing code that was moved to make `sourceRoot` available for `initConfig()`.  But I can add the comment you requested.

src/actions-util.ts
cklin: This is not new code—it is existing code that was moved to allow greater reuse.

src/analyze.ts
cklin: I don't think there is a compelling reason for the change either.  I am not even sure the change was intentional—it might have been made by copilot or IDE auto-completion, which I then let stand because it did not add that much noise to justify the additional work to preserve the existing import style, which seems just as arbitrary.

If this becomes a concern, we might be able to configure the linter to insist on one import style or the other.

src/overlay-database-utils.ts
cklin: Thank you—that is a good observation.  I added a note to the internal tracking issue for consideration later.

src/overlay-database-utils.ts
cklin: Thank you—that is a good observation. I added a note to the internal tracking issue for consideration later.

cklin:
@mbg Thank you for your review!

> * Before we merge this, we should add some test coverage. At this stage, it would be good to gain confidence that none of the new logic interferes with existing functionality when the feature is disabled. The existing PR checks do that to an extent, but we should verify that we cover all cases. In the long run, we will want test coverage for when the feature is enabled as well.

I agree that more tests are a good idea.  Can you say more about what you mean by "gain confidence that none of the new logic interferes with existing functionality" and "cover all cases", so that I have a better estimate of the effort required?

> * There are general concerns around us using the Actions cache, similar to what we have for TRAP caching and dependency caching. I.e. we might be eating away at the cache quota that may already be in use by other workflows in a given repository. Can we mitigate our impact here? For example, by ensuring that we only have one cache at a time or limiting the size of the caches we store. This may not need to be addressed at this point if we only plan to use this internally for now, but it should be addressed before we ship it more widely, especially if the goal is to have it on by default at some point.

Thank you for bringing that up.  I added a note to the internal tracking issue for consideration later.

src/analyze-action.ts
mbg: Thanks for the thoughtful discussion of this! 

Firstly, I am thinking that, regardless of which behaviour we go with, we should implement a warning that warns of incompatible `cleanup-level` and overlay database settings. 

Beyond that, my feeling is that we should preserve existing behaviour, at least until we have given enough notice that a change is coming. That is to say, I'd probably have a preference for "We could obey the cleanup-level input and not upload the overlay-base database if the input is not overlay, which means that the workflow will not benefit from overlay analysis.". 

I would say that is acceptable because:

- We will have a warning which notifies the user (but not an error that breaks existing workflows).
- We can assume that `cleanup-level` isn't super commonly used, and most users will still benefit from overlay databases anyway, particularly in default setup where no `cleanup-level` is specified.
- We don't override user-configured behaviour.
- Since `cleanup-level` is used in advanced setup, users can change this setting if they have it configured and they want to take advantage of overlay databases.

We can then, after some time, revisit this to decide whether we want to override the `cleanup-level` setting if overlay databases are enabled.

src/init-action.ts
mbg: Thanks! Yes, I saw that it had been moved, but ultimately that doesn't change my thinking that this would be a useful thing to have!

src/actions-util.ts
mbg: I noticed -- we probably should have added some tests back then, so now may be a good opportunity to do that.

src/feature-flags.ts
mbg: I think I am happy with that justification for leaving this as is for now.

src/config-utils.ts
mbg: > The internal testing can include staff-ship to internal repositories, and it would be ideal if we can do so through feature flags without requiring workflow file changes.

Agreed.

> Once this PR merges, a user could get this code by pinning a commit on main, which I think is a use case we should support.

I am a bit conflicted about this. On one hand, I agree that a user could do this and the "future FF enablement on an old pinned version" scenario is an interesting one to consider. On the other hand, I am not sure how much we have considered this scenario in the past -- possibly a question for @aeisenberg or @henrymercer. I'm slightly leaning towards thinking that the FF alone is probably enough, particularly since a user who pins this version could also just set the environment variable to get this prototype implementation, but no strong feelings. 

FYI we also have a FF in default setup for internal things. That isn't propagated to the action right now in any way, but it could be. Perhaps if we feel that this kind of feature gating is necessary for experimental features, then we could instead do some work to start using that.

mbg:
> Can you say more about what you mean by "gain confidence that none of the new logic interferes with existing functionality" and "cover all cases".

Broadly, we should check that `getOverlayDatabaseMode()` doesn't turn on this feature unexpectedly. I.e. we should have tests that cover all the cases it uses for deciding whether to turn the feature on or not. That might be doable with some unit tests, or might need an extra `pr-check`.

cklin:
@mbg

I made some of the requested changes ([diff from force pushes sans rebase](https://github.com/github/codeql-action/compare/a8a0cc829c32c0d4a9922ab977a94023e611e448..55ea4c300dac27722dcb689bfcb0a342f993611e)).

- Added comment on `sourceRoot` definition
- Added comment on repo owner allowlist
- Added unit tests for `getPullRequestBranches()`
- Added unit tests for `getOverlayDatabaseMode()`

These are the changes I consider (a) in-scope for this PR and (b) there is clear consensus on how to proceed.

The main unresolved issue is how we want to handle the `cleanup-level` input.  My reading of the internal discussion is that that input is considered advisory, and it is reasonable to override the user input when appropriate.  That said, I am happy to make additional changes to try to mitigate the situation:

- Update the description of `cleanup-level` to specify that it is advisory
- Remove the default `brutal` value of `cleanup-level`, and document the default as automatic selection
- Emit a warning when we override the user-specified `cleanup-level` value

What do you think?

src/actions-util.test.ts
```diff
@@ -1,12 +1,47 @@
...
+import { getPullRequestBranches, isAnalyzingPullRequest } from "./actions-util";
 import { computeAutomationID } from "./api-client";
 import { EnvVar } from "./environment";
 import { setupTests } from "./testing-utils";
 import { initializeEnvironment } from "./util";
 
 setupTests(test);
 
+function withMockedContext<T>(mockPayload: any, testFn: () => T): T {
+  const originalContext = github.context;
```
mbg: Minor: Any reason for saving the entire `context` even though only the `payload` needs to be restored later?

src/actions-util.test.ts
```diff
@@ -58,6 +93,88 @@ test("computeAutomationID()", async (t) => {
   );
 });
 
+test("getPullRequestBranches() with pull request context", (t) => {
```
mbg, line 96: Minor: perhaps it would be good to have a test case for a non-`pull_request` event (e.g. `push`) to check that `isAnalyzingPullRequest` returns false there.

mbg (approved):
Thanks for making these changes, the new tests are nicely implemented! ✨ 

I think I am happy to approve this in its current state. I only had a couple of minor thoughts about the tests, which are not urgent. I have also responded to the point about `cleanup-level` below, where one of the points you listed could be done as part of this PR.

> The main unresolved issue is how we want to handle the cleanup-level input. My reading of the internal discussion is that that input is considered advisory, and it is reasonable to override the user input when appropriate. 

I think we should consider this a bit more before reaching a decision. We can proceed with this PR though, since it is an experimental implementation for internal testing that's gated behind an allow list and FF. However, I would like to refrain from making changes that affect all users until we have reached a decision. 

The current implementation only overrides the `cleanup-level` if overlay databases are enabled, so it might be OK to mostly leave that as-is for now. Implementing the warning if the `cleanup-level` gets overridden would probably still be useful though. 

If we decide to go ahead with public-facing changes to the `cleanup-level` input, then we can make those in a separate PR.

src/actions-util.test.ts
cklin: Thanks—I implemented the suggested change.

src/actions-util.test.ts
cklin: Thanks—I implemented the suggested change.

PR #2959, opened 2025-07-01
Remove support for combining SARIF runs with non-unique categories

koesie10:
This removes support for combining SARIF runs with non-unique categories, and will instead throw a configuration error. This behavior is only enabled behind a feature flag.

This is intentionally not backwards compatible when the feature flag is enabled, as this [was announced over a year ago](https://github.blog/changelog/2024-05-06-code-scanning-will-stop-combining-runs-from-a-single-upload) and a deprecation warning is shown for every upload that would run into this error.

Screenshot 2025-07-02 at 11 33 41


src/upload-lib.ts
```diff
@@ -145,6 +145,52 @@ export async function shouldShowCombineSarifFilesDeprecationWarning(
...
+    !(await shouldDisableCombineSarifFiles(
+      sarifObjects,
+      features,
+      githubVersion,
+    ))
+  ) {
+    return;
+  }
+
+  // TODO: Update this changelog URL to the correct one when it's published.
```
mbg, line 163: Minor: if not already done, we should track this somewhere.

src/upload-lib.ts
```diff
@@ -167,11 +213,17 @@ async function combineSarifFilesUsingCLI(
...
+      : "and will be removed in July 2025";
   const deprecationMoreInformationMessage =
     "For more information, see https://github.blog/changelog/2024-05-06-code-scanning-will-stop-combining-runs-from-a-single-upload";
 
   if (!areAllRunsProducedByCodeQL(sarifObjects)) {
+    await throwIfCombineSarifFilesDisabled(
+      sarifObjects,
+      features,
+      gitHubVersion,
+    );
```
mbg, line 225: I assume the deprecation warning a few lines further down is not necessary if the conditions for this are true. Should it be removed now or later?

mbg (approved):
Thanks for taking care of this and including a bunch of tests! This generally looks good to me.

src/upload-lib.ts
koesie10: 👍, I've linked an internal issue that tracks this task.

src/upload-lib.ts
koesie10: That's true, but we can only remove the deprecation warning when the action drops support for GHES 3.17. In all GHES versions between 3.14 (including) and 3.18 (excluding), the action should still show the deprecation warning and should never give this error. Until support for GHES 3.17 is dropped, customers can still sync their action from dotcom to their GHES instance and expect it to work without breaking changes.

mbg: Makes sense -- probably also worth tracking then

PR #2960, opened 2025-07-03
Rust: remove shipped feature flag

redsun82:


src/init-action.ts
```diff
@@ -588,32 +587,25 @@ async function run() {
       core.exportVariable(bmnVar, value);
     }
 
-    // For rust: set CODEQL_ENABLE_EXPERIMENTAL_FEATURES, unless codeql already supports rust without it
+    // For rust, if running a version prior to public preview (2.22.1) set CODEQL_ENABLE_EXPERIMENTAL_FEATURES
     if (
       config.languages.includes(Language.rust) &&
-      !(await codeql.resolveLanguages()).rust
+      !(await codeql.resolveLanguages()).rust // means codeql already supports rust without any intervention
```
mbg: Minor: I'd prefer this comment to be part of the comment above the `if` rather than at the end of this line, if you think that it is needed at all.

src/init-action.ts
```diff
@@ -588,32 +587,25 @@ async function run() {
       core.exportVariable(bmnVar, value);
     }
 
-    // For rust: set CODEQL_ENABLE_EXPERIMENTAL_FEATURES, unless codeql already supports rust without it
+    // For rust, if running a version prior to public preview (2.22.1) set CODEQL_ENABLE_EXPERIMENTAL_FEATURES
```
mbg: I am not sure that this comment accurately describes what is currently implemented. As far as I can tell you check whether the actual CLI version is less than `minVer` and if so throw a `ConfigurationError`. That means we never get to the `CODEQL_ENABLE_EXPERIMENTAL_FEATURES` check. 

If that is the case, and that behaviour is unintentional, then you should obviously fix it, but I'd also like to perhaps see this refactored with a test to make sure it does the right thing.

mbg (changes requested):

src/init-action.ts
redsun82: there are 2 versions involved:
* prior to 2.19.3, there is no rust support whatsoever, which means that if we are requesting rust we indeed need to throw a configuration error
* between 2.19.3 and 2.22.1, rust support is experimental, which means the variable needs to be set explicitly. What I now see, is that if `!(await codeql.resolveLanguages()).rust`, that actually entails already that `CODEQL_ENABLE_EXPERIMENTAL_FEATURES` is not set, otherwise `codeql resolve languages` would provide an entry for `rust`, which makes the `process.env[envVar] !== "true"` check redundant.
* after 2.22.1, `(await codeql.resolveLanguages()).rust` will always be set

If we forego the `"Experimental Rust analysis enabled"` log line entirely (there's probably no burning need for it), we can simplify by getting rid of the second if. And the comment needs a rewrite to be clearer

redsun82: I'm considering to just not mention the 2.19.3 range, and go for a simple `Rust analysis is supported from version 2.22.1`. No need for the general public to know about the experimental phase I guess

src/init-action.ts
```diff
@@ -630,25 +629,20 @@ async function run() {
       core.exportVariable(bmnVar, value);
     }
 
-    // For rust, if running a version prior to public preview (2.22.1) set CODEQL_ENABLE_EXPERIMENTAL_FEATURES
+    // For rust, if `codeql` does not support it, we are in two possible situations:
+    // * either we run `codeql` prior to 2.19.3, which means no rust support is possible
+    // * or we run a version between 2.19.3 and 2.22.1 (excluded), when rust support was
+    //   experimental and required explicitly setting CODEQL_ENABLE_EXPERIMENTAL_FEATURES
+    // However, there is not need to let users know about the experimental feature flag at
```
mbg: Typo: this should read "no need", rather than "not need"

src/init-action.ts
```diff
@@ -630,25 +629,20 @@ async function run() {
...
+    // this point, if they want rust support they should chose a version which supports it
+    // publicly.
```
mbg: This might break Rust analysis for users who have pinned `v3` (or similar) of the Action as well as a CodeQL CLI `>= 2.19.3 && < 2.22.1` and who used `CODEQL_ENABLE_EXPERIMENTAL_FEATURES`. I agree that it's not unreasonable to ask users to upgrade to a newer CLI version in that case, but perhaps we should give them some warning first before breaking their workflows over night when the next Action release happens? That said, it's not clear to me that a warning in the output would necessarily be noticed. I don't know if we have any data to know how many people would be affected.

mbg (approved):
I think this is OK to merge if you are happy to break Rust analyses for those older versions of the CLI. See my thoughts in the comment I added. There's also one typo in the comment.

src/init-action.ts
redsun82: in that situation (version `2.19.3..2.22.1` _and_ `CODEQL_ENABLE_EXPERIMENTAL_FEATURES=true`) `codeql resolve languages` does give rust, which means that `(await codeql.resolveLanguages()).rust` is not undefined. The only change then is that users lose the `Experimental Rust analysis enabled` log line, but no functional change happens.

redsun82: uh, actually, now I see that that log line was dead code anyway!

src/init-action.ts
redsun82:
```suggestion
    // However, there is no need to let users know about the experimental feature flag at
```

mbg (approved):

henrymercer:
https://github.com/github/codeql-action/pull/3014 should ease the merge conflict in compiled code pain here.

henrymercer:
(Though we'll have to do the merge first, to get the updated workflow in the branch)

redsun82:
I've noticed that https://github.com/github/codeql-action/pull/2914 duplicated the rust specific bock before `initConfig`. That got me thinking, that actually since the feature flag has been rolled out, the effective version from which we must support Rust to be backward compatible is `2.19.3`, which means that we must set the env variable in versions between 2.19.3 and 2.22.1. I've changed the code accordingly in the if block before `initConfig`, and removed the later if block which becomes unneeded.

src/init-action.ts
```diff
@@ -359,37 +358,29 @@ async function run() {
     }
     core.endGroup();
 
-    // Set CODEQL_ENABLE_EXPERIMENTAL_FEATURES for Rust. We need to set this environment
-    // variable before initializing the config, otherwise Rust analysis will not be
-    // enabled.
+    // Set CODEQL_ENABLE_EXPERIMENTAL_FEATURES for Rust if between 2.19.3 and 2.22.1 (excluded)
```
mbg: This is slightly ambiguous: does the "(excluded)" refer to both `2.19.3` and `2.22.1` or only the latter?

src/init-action.ts
```diff
@@ -359,37 +358,29 @@ async function run() {
...
+    // We need to set this environment variable before initializing the config, otherwise Rust
+    // analysis will not be enabled.
```
mbg: Could you add a small bit to this comment to remind us why it will not be enabled otherwise? I.e. is the logic for that elsewhere in the Action, the CLI, the extractor, ...?

src/init-action.ts
```diff
@@ -359,37 +358,29 @@ async function run() {
...
+    // Initially this was driven by a feature flag which has been rolled out already.
```
mbg: Minor: This can probably be removed. I don't see much value in documenting what used to be the case here.

mbg (commented):
As I wrote elsewhere, I am a bit concerned that we have gone through a few iterations of the logic here, approved it, and then changed our minds about how this should work. 

That signals to me that we:

- Don't necessarily know what this should do and, while these changes make sense in isolation, we could still be missing something. It might be good to outline internally in an issue what different scenarios we need to consider for this (e.g. permutations of: Advanced/Default Setup, GH-hosted runner vs self-hosted, different CLI versions, Rust with and without other languages, ...). I think having a matrix of scenarios to consider like that might be more broadly useful a well.
- Don't have any test coverage that would tell us if the logic here doesn't do the right thing. I believe that we can pin specific CLI versions in the PR checks. So perhaps it would make sense to have a PR check that tests this with different permutations of relevant CLI versions (`< 2.19.3`, `>= 2.19.3 && < 2.22.1`, and `>= 2.22.1`)? Or perhaps a combination of unit tests (to check the general logic of this) and PR check (to check that the analysis for Rust works when it should).

src/environment.ts
```diff
@@ -114,4 +114,9 @@ export enum EnvVar {
    * Useful for testing purposes where multiple caches may be stored in the same repository.
    */
   DEPENDENCY_CACHING_PREFIX = "CODEQL_ACTION_DEPENDENCY_CACHE_PREFIX",
+
+  /**
+   * Whether to enable experimental extractors for CodeQL.
+   */
+  EXPERIMENTAL_FEATURES = "CODEQL_ACTION_EXPERIMENTAL_FEATURES",
```
redsun82: ha, well, I should have double checked what copilot was writing here. Good thing someone suggested to put in more testing :wink:

redsun82: (this was causing a failure in the 2.19.3 case, which is now fixed)

src/init-action.ts
```diff
@@ -360,37 +359,29 @@ async function run() {
...
+      if (semver.lt(actualVer, experimental)) {
         throw new ConfigurationError(
-          `Experimental rust analysis is supported by CodeQL CLI version ${minVer} or higher, but found version ${actualVer}`,
+          `Rust analysis is supported by CodeQL CLI version ${experimental} or higher, but found version ${actualVer}`,
         );
       }
-      logger.info("Experimental rust analysis enabled");
+      if (semver.lt(actualVer, publicPreview)) {
+        core.exportVariable(EnvVar.EXPERIMENTAL_FEATURES, "true");
+        logger.info("Experimental rust analysis enabled");
```
mbg: Minor: Should "rust" be capitalised here?

redsun82: Makes sense

mbg (approved):

PR #2967, opened 2025-07-10
Overlay: additional feature flags

cklin:
This PR updates the action to check additional per-language features for enabling overlay analysis.

See back-linked issue for additional context.


src/feature-flags.ts
```diff
@@ -502,18 +622,29 @@ class GitHubFeatureFlags {
...
-          features: featuresToRequest,
-        },
-      );
-      const remoteFlags = response.data as GitHubFeatureFlagsApiResponse;
+        .map(([f]) => f);
+
+      const FEATURES_PER_REQUEST = 25;
+      const featureChunks: string[][] = [];
+      while (featuresToRequest.length > 0) {
+        featureChunks.push(featuresToRequest.splice(0, FEATURES_PER_REQUEST));
```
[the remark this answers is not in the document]
cklin: `featuresToRequest` is used only for populating `featureChunks`, so it is safe to modify.

src/config-utils.ts
```diff
@@ -195,7 +196,7 @@ export interface AugmentationProperties {
   /**
    * Extra query exclusions to append to the config.
    */
-  extraQueryExclusions?: ExcludeQueryFilter[];
+  extraQueryExclusions: ExcludeQueryFilter[];
```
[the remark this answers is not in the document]
cklin: All uses of `extraQueryExclusions` are in this repository, and they have all been checked and updated.  Furthermore, `AugmentationProperties` is never deserialized from files, so there is no concern for compatibility with earlier versions.

src/config-utils.ts
```diff
@@ -768,6 +647,78 @@ function parseQueriesFromInput(
...
+    const codeScanningFeature =
+      OVERLAY_ANALYSIS_CODE_SCANNING_FEATURES[language];
+    if (
+      codeScanningFeature &&
+      (await features.getValue(codeScanningFeature, codeql))
+    ) {
+      enableForCodeScanningOnly = true;
+      continue;
+    }
+    return false;
```
nickrolfe, line 706: If I'm reading this correctly, then we won't check the code-scanning variant for a language if the main feature flag for that language is enabled. If, for example, the `overlay_analysis_java` flag is already enabled, then toggling `overlay_analysis_code_scanning_java` will have no effect. That seems reasonable, but we should take care that anyone enabling feature flags is aware of this.

src/feature-flags.ts
```diff
@@ -502,18 +622,29 @@ class GitHubFeatureFlags {
...
+          "GET /repos/:owner/:repo/code-scanning/codeql-action/features",
+          {
+            owner: this.repositoryNwo.owner,
+            repo: this.repositoryNwo.repo,
+            features: chunk.join(","),
+          },
+        );
+        const chunkFlags = response.data as GitHubFeatureFlagsApiResponse;
+        remoteFlags = { ...remoteFlags, ...chunkFlags };
+      }
```
nickrolfe, line 646: This chunking seems like the riskiest part of the PR, since it could affect all uses of the Action. I don't see anything wrong with it, and I see you've added a test (with some API-mocking) for it, but do we have a way to see it working in practice before merging/deploying?

nickrolfe (approved):

src/feature-flags.ts
cklin: All the PR checks run with the changes in this PR, so any outright failure would show up on CI checks.

For reference, [the commit without request chunking that I first pushed to this PR](https://github.com/github/codeql-action/commit/1ab4d5876d29c763b660b9922675aa4b2d1c0c0b) had 244 check failures.  It was hard to miss.

In addition to passing CI checks, [I reran one of the workflows with debug logs](https://github.com/github/codeql-action/actions/runs/16227798371/job/45827765182?pr=2967).  In the "Initialize CodeQL" step, we can see in the debug logs that the action was able to obtain status for all declared features (under "Loaded the following default values for the feature flags from the Code Scanning API"), which lends further support that the request chunking is working as intended.

cklin:
@mbg I'd also appreciate your review and approval on this PR before I merge.

src/feature-flags.ts
mbg: I am also wondering what the performance impact of this is (both for users and the backend). If I follow our implementation correctly, we always fetch the values for all FFs that the Action knows of. So that means, once merged, the Action will fetch the status of all of these FFs for every run for everyone. I imagine it's fine if it's one request with a few FFs, but if this now spans multiple requests for every CodeQL analysis, that could quickly add up?

cklin: > So that means, once merged, the Action will fetch the status of all of these FFs for every run for everyone.

The Action fetches the status of all declared `Feature`s for every run for everyone.  This PR does not change that behavior—that will continue to be the case even if we do not merge this PR.

What this PR changes is to support fetching feature status in batches of 25 in accordance with the API limit.  In practical terms, where "fetching the status of all declared features" used to require 1 API call, now it requires 2 API calls.  Each call takes less than 1 second on my laptop, so adding one more call should not have any perceivable impact on user experience.  With the volume of Code Scanning runs, I don't think it will have perceivable impact on backend load either.

src/feature-flags.test.ts
```diff
@@ -131,6 +133,29 @@ test("Features use default value if they're not returned in API response", async
...
+        status: requestedFeatures.length <= 25 ? 200 : 400,
+        messageIfError: "Can request a maximum of 25 features.",
+        data: {},
+      };
+    });
+
+    // We only need to call getValue once, and it does not matter which feature
+    // we ask for. Under the hood, the features library will request all features
+    // from the API.
+    const feature = Feature.OverlayAnalysis;
```
mbg: Given that `Feature` elements are intended to be removed when the feature has fully shipped, we would need to update this test when `OverlayAnalysis` is removed. How about something like this instead:
```suggestion
    const feature = Object.values(Feature)[0];
```
However, I see that we similarly pick specific elements elsewhere in the tests (and that they have to be updated when they are removed). It's your choice if you want to fix this in this case or let us deal with it more generally.

src/feature-flags.ts
mbg: > The Action fetches the status of all declared Features for every run for everyone. This PR does not change that behavior—that will continue to be the case even if we do not merge this PR.

Sorry for the ambiguity in what I wrote. What I meant with "all these FFs" was "all the new FFs introduced by this PR". I am happy for you to keep this as is as long as you're happy that doubling the number of requests to this endpoint isn't a problem.

src/feature-flags.test.ts
cklin: Thank you for the suggestion.  I applied it in the [Replicate "too many feature flags" error in test](https://github.com/github/codeql-action/pull/2967/commits/3eaefb4deb0ab7420fb76c2a72212d5b86309a46) commit.

PR #2969, opened 2025-07-14
Fix parsing of GHES pre-release versions

koesie10:
This fixes parsing of GHES pre-release versions for the combining SARIF files check. The version `3.18.0.pre1` is not valid semver, so this just transforms it to `3.16.0-pre1`.


src/util.ts
```diff
@@ -1132,6 +1132,14 @@ export function checkActionVersion(
   }
 }
 
+export function parseGhesVersion(version: string): semver.SemVer {
+  // GHES pre-release versions are in the format "3.18.0.pre1", which is not a valid semver version.
```
mbg: Minor: It might be useful to move a version of this comment into a JSDoc for this function, so that it's easier to see what the purpose of this function is from use sites.

mbg (approved):
Looks like a reasonable fix for this scenario! Thanks for implementing and thanks to @cklin for spotting it.

Just one minor suggestion, otherwise LGTM.

PR #2970, opened 2025-07-14
Enable Feature.DiffInformedQueries

cklin:
This PR updates `Feature.DiffInformedQueries` to be enabled on GHES versions 3.19 or higher.


src/diff-informed-analysis-utils.ts
```diff
@@ -37,6 +41,14 @@ export async function getDiffInformedAnalysisBranches(
     return undefined;
   }
 
+  const gitHubVersion = await getGitHubVersion();
+  if (
+    gitHubVersion.type === GitHubVariant.GHES &&
+    semver.lt(parseGhesVersion(gitHubVersion.version), "3.19.0")
```
mbg: Would this benefit from the new function added in https://github.com/github/codeql-action/pull/2972?

mbg (approved):
LGTM, thanks for changing this and adding tests! Just one minor question / suggestion.

src/diff-informed-analysis-utils.ts
cklin: Certainly, especially since that PR removed `parseGhesVersion()`, so this code no longer builds.

cklin: PR updated to use `satisfiesGHESVersion()`.

PR #2972, opened 2025-07-15
Ignore pre-release parts when comparing GHES versions

koesie10:
This will change our GHES version comparison to always compare without taking into consideration the pre-release part of the version, and dropping any unrecognized semver parts.

- See https://github.com/github/codeql-action/pull/2969


src/util.ts
```diff
@@ -1133,20 +1133,26 @@ export function checkActionVersion(
...
+  githubVersion: GitHubVersion,
+  range: string,
+): boolean {
+  if (githubVersion.type !== GitHubVariant.GHES) {
+    return false;
   }
-  return new semver.SemVer(version);
+
+  const semverVersion =
+    semver.coerce(githubVersion.version) ?? new semver.SemVer("0.0.0");
```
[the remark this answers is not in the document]
mbg: `satisfiesGHESVersion` implicitly has the behaviour that:

- If `range` is a `<` check, we default to `true`
- If `range` is a `>` check, we default to `false`

I agree with Copilot that this might shoot us in the foot in the future if we try to use `satisfiesGHESVersion` elsewhere and this isn't clear.

I see that we do a similar thing in `checkActionVersion`, but the check condition is right next to it so it's more obvious that `0.0.0` is chosen as a default to cause the check to fail by default. 

Perhaps the default could just be a parameter of `satisfiesGHESVersion` that can be set to `0.0.0` at the current call sites or set to `undefined` if `satisfiesGHESVersion` should fail if `semver.coerce` fails.

mbg (commented):
Thanks for putting this together! This probably makes sense, since I can't imagine that we'd want to check for a particular pre-release version. Dropping the pre-release for the comparison makes sense to me. 

One suggestion about the potential footgun that Copilot highlighted as well.

src/util.ts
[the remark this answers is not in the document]
koesie10: That's a good point. It should never happen that `semver.coerce` fails, but if it does `0.0.0` is not always the desired behavior. I've added a new `defaultIfInvalid` argument to the function that will be returned when the version is invalid/not a GHES version. This should ensure that the call site can determine what should happen (in this case not show the warning/throw an error).

src/upload-lib.ts
```diff
@@ -132,7 +132,7 @@ export async function shouldShowCombineSarifFilesDeprecationWarning(
   // Do not show this warning on GHES versions before 3.14.0
   if (
     githubVersion.type === GitHubVariant.GHES &&
-    satisfiesGHESVersion(githubVersion, "<3.14")
+    satisfiesGHESVersion(githubVersion, "<3.14", true)
```
mbg: This now means that `satisfiesGHESVersion` returns `true` if it's not a GHES version, rather than `false` as before. I am guessing that wasn't intentional?

src/upload-lib.ts
```diff
@@ -177,7 +177,7 @@ async function shouldDisableCombineSarifFiles(
 ) {
   if (githubVersion.type === GitHubVariant.GHES) {
     // Never block on GHES versions before 3.18.
-    if (satisfiesGHESVersion(githubVersion, "<3.18")) {
+    if (satisfiesGHESVersion(githubVersion, "<3.18", true)) {
```
mbg: Same here.

src/util.ts
```diff
@@ -1136,17 +1136,23 @@ export function checkActionVersion(
...
  * taking into account that a range like >=3.18 will also match the GHES 3.18
  * pre-release/RC versions.
+ *
+ * When the given `githubVersion` is not a GHES version, or if the version
+ * is invalid, this will return `defaultIfInvalid`.
  */
 export function satisfiesGHESVersion(
   githubVersion: GitHubVersion,
   range: string,
+  defaultIfInvalid = false,
```
mbg: Minor: Probably best to leave out the default value, since it's not needed for any kind of backwards compatibility and an explicit value is provided everywhere the function is used.

src/upload-lib.ts
koesie10: Good spot, I've changed the function to accept the version string instead of the `GithubVersion` object to make this more explicit (the function should never be called when the version is not GHES anyway).

src/util.ts
```diff
@@ -1141,15 +1141,11 @@ export function checkActionVersion(
  * is invalid, this will return `defaultIfInvalid`.
  */
 export function satisfiesGHESVersion(
-  githubVersion: GitHubVersion,
+  ghesVersion: string,
```
mbg, line 1144: Good idea!

mbg (dismissed):
Thanks! This looks good now.

mbg:
There's a merge conflict with `lib/upload-lib.js.map` which I think you can resolve by merging / rebasing and then running `npm run build` again

koesie10:
Thanks for the reviews! I've rebased and squashed all three commits into one to make this easier to cherry-pick for the backport.

PR #2975, opened 2025-07-17
Overlay: report telemetry

cklin:
This PR adds telemetry fields for overlay analysis.
- `analysis_is_overlay`
- `analysis_builds_overlay_base_database`
- `overlay_base_database_download_size_bytes`
- `overlay_base_database_download_duration_ms`


src/analyze.ts
```diff
@@ -617,12 +629,20 @@ export async function runQueries(
...
     config.augmentationProperties.overlayDatabaseMode ===
     OverlayDatabaseMode.Overlay
   ) {
+    statusReport.analysis_is_overlay = true;
     incrementalMode.push("overlay");
+  } else if (
+    config.augmentationProperties.overlayDatabaseMode ===
+    OverlayDatabaseMode.OverlayBase
+  ) {
+    statusReport.analysis_builds_overlay_base_database = true;
```
mbg: Minor: perhaps this would be slightly nicer as, similar to what we already have for `analysis_is_diff_informed`: 
```suggestion
  } 
  
  statusReport.analysis_builds_overlay_base_database =
    config.augmentationProperties.overlayDatabaseMode ===
    OverlayDatabaseMode.OverlayBase;
```

src/overlay-database-utils.ts
```diff
@@ -235,41 +235,47 @@ export async function uploadOverlayBaseDatabaseToCache(
   return true;
 }
 
+export interface OverlayBaseDatabaseDownloadStats {
+  databaseSizeBytes: number;
+  databaseDownloadDurationMillis: number;
```
mbg: Minor: elsewhere we use `Ms` as suffix rather than `Millis`. Might be nicer to use that for consistency.

src/overlay-database-utils.ts
```diff
@@ -235,41 +235,47 @@ export async function uploadOverlayBaseDatabaseToCache(
...
+ * @returns A promise that resolves to download statistics if an overlay-base
+ * database was successfully downloaded, or undefined if the download was
+ * either not performed or failed.
  */
 export async function downloadOverlayBaseDatabaseFromCache(
   codeql: CodeQL,
   config: Config,
   logger: Logger,
-): Promise<boolean> {
+): Promise<undefined | OverlayBaseDatabaseDownloadStats> {
```
mbg: Minor: Conventionally the `undefined` seems to come last. There's nothing wrong with this I think, but it's just a bit unusual. I see we have this order in a couple of other places in the codebase though.

mbg: A bigger question here is probably why the return type is used to indicate failure at all. Wouldn't it be better to have this as `Promise<OverlayBaseDatabaseDownloadStats>` and reject if the download was either not performed or failed?

src/overlay-database-utils.ts
```diff
@@ -312,11 +323,27 @@ export async function downloadOverlayBaseDatabaseFromCache(
...
     logger.warning("Downloaded overlay-base database failed validation");
-    return false;
+    return undefined;
+  }
+
+  const databaseSizeBytes = await tryGetFolderBytes(dbLocation, logger);
+  if (databaseSizeBytes === undefined) {
+    logger.warning(
+      "Filesystem error while accessing downloaded overlay-base database",
+    );
```
mbg: `tryGetFolderBytes` will log a warning (since you don't set `quiet=true`) before returning `undefined`, so this will always lead to two warnings for the same problem, which I believe will show up as separate workflow annotations.

I'd suggest either removing this warning or setting `quiet=true` in the `tryGetFolderBytes` call.

src/overlay-database-utils.ts
```diff
@@ -312,11 +323,27 @@ export async function downloadOverlayBaseDatabaseFromCache(
...
+    // The problem that warrants reporting download failure is not that we are
+    // unable to determine the size of the database. Rather, it is that we
+    // encountered a filesystem error while accessing the database, which
+    // indicates that an overlay analysis will likely fail.
+    logger.warning("Downloaded overlay-base database is invalid");
```
mbg: Similarly here: it would be better to merge this into the previous `.warning` call if you want both messages. Otherwise, maybe remove the previous one or make it a `.debug` level message?

mbg (commented):
This generally looks good, assuming all the necessary backend work is also done to support this. Thanks also for including some tests.

There are no blocking issues. I flagged up a couple points that might be good to deal with before merging this though, as well as some other trivial points.

src/analyze.ts
cklin: Thanks—I implemented the suggestion.

src/overlay-database-utils.ts
cklin: Thanks—I implemented the suggestion.

src/overlay-database-utils.ts
cklin: I moved `undefined` to the last.

The question of using a rejected promise instead of promise of `undefined` is interesting.  There are three cases:
1. We are not supposed to download, so we didn't
2. We are supposed to download, but something went wrong
3. We are supposed to download, and the download completed successfully

The caller does not care about the distinction between 1 and 2, and the function returns `undefined` for both of these cases.  We could use a rejected promise for both 1 and 2, and it would work, but the control flow would be a bit strange—it is like throwing an exception (and then catch it gracefully) to skip a feature that is disabled by flag.

An alternative is return `undefined` for 1 and a rejected promise for 2.  That avoids the aforementioned awkwardness, at the cost of making the caller more complicated, having to check for both `undefined` and add a `catch` block and do the same thing in both cases.

I think the current implementation is better than these two alternatives, but I can be persuaded otherwise.

mbg: You could use a rejected promise with different values for 1 and 2 respectively that the caller can then check for (these can be arbitrary values). It is indeed a bit odd to handle such cases with `catch`, although you could have something as simple as `.catch(console.debug)` (or similar) after the call to `downloadOverlayBaseDatabaseFromCache` if there are appropriate messages given to `throw`. 

Alternatively, the feature check could be moved out from `downloadOverlayBaseDatabaseFromCache` to the caller.

I don't feel strongly about this though and am be happy for it to stay as-is.

src/overlay-database-utils.ts
cklin: I think both messages are worth preserving because they tell the user different things.
- The warning from `tryGetFolderBytes` explains what went wrong.  This is important if the user were to fix the underlying technical problem.
- The warning here in `downloadOverlayBaseDatabaseFromCache` explains the consequence of the failure (that the downloaded overlay-base database cannot be accessed) so that the user can assess the severity and impact of the problem.

That said, I agree that they do not both have to be warnings.  I downgraded this message to `info`.

src/overlay-database-utils.ts
cklin: I think this extra `warning` call is an oversight on my part, and it should not be here.  I have removed it.

Thanks for catching this!

src/overlay-database-utils.ts
mbg: Perhaps it would be reasonable for `tryGetFolderBytes` to return the error rather than `undefined` and then leave it up to the caller to decide what to do with that (e.g. include it in a combined log message with the consequence + underlying problem). 

Not something you need to act on if you're happy to downgrade this to an `info`-level message.

src/overlay-database-utils.ts
cklin: In that case, let's keep the current code.

src/overlay-database-utils.ts
cklin: Thanks.  I already downgraded the log from `warning` and `info`, and I am happy with that change.

PR #2987, opened 2025-07-23
Treat processing error for multiple runs with the same category as configuration error

mbg:
This will result in it being reported as a user error rather than a failure


koesie10 (approved):
Thanks!

PR #2988, opened 2025-07-23
Disable combining runs within a single file

koesie10:
This will throw a `ConfigurationError` when a single SARIF file contains multiple runs with the same category.


src/upload-lib.ts
```diff
@@ -206,9 +206,6 @@ async function combineSarifFilesUsingCLI(
   logger: Logger,
 ): Promise<SarifFile> {
   logger.info("Combining SARIF files using the CodeQL CLI");
-  if (sarifFiles.length === 1) {
```
koesie10, line 209: This case was never actually hit, so I've removed this.

mbg (approved):

PR #2993, opened 2025-07-25
Overlay: check query packs for compatibility

cklin:
This PR updates the `init` action to prevent overlay analysis (or overlay-base database construction) when one of the query packs involved does not support overlay analysis with the installed CodeQL CLI.

- @alexet Please review the aspects concerning interaction with the CodeQL bundle
- @mbg Please review all other aspects of the change


cklin:
> It doesn't feel to me that resolve packs is quite the right command (although it looks like it will work).It feels like we should run `resolve queries` with `--format startingpacks` to find the query packs we care about, and then just use those.
> 
> But I am also not fully sure of all the subtleties of either choice.

Thanks for the suggestion!

When I first looked into the problem, I considered this approach but was unable to get it to work.  While `codeql resolve queries` with `--format startingpacks` can map queries to their containing query packs, there does not seem to be a way to get the list of queries that would be used in the analysis and feed that into `codeql resolve queries`.  I could see that `codeql database init` already computes that information (though a deep-plumbing subcommand) for internal use, but no easy way to obtain that information from outside the CodeQL CLI.  Re-interpreting the Code Scanning config file in the action seems like a lot of work for relatively little gain.

Prompted by your suggestion, I looked again.  This time I noticed that `codeql database init` actually generates `config-queries.qls` suite file with the list of queries, so all I need to do is to feed that suite file into `codeql resolve queries`, and I can get the list of query packs to check for overlay compatibility.

(There is the slight complication that `codeql resolve queries --format startingpacks` would also return query packs that have not been compiled.  I will have to make the action ignore those query packs when checking for overlay compatibility.)

cklin:
> It doesn't feel to me that resolve packs is quite the right command (although it looks like it will work).It feels like we should run `resolve queries` with `--format startingpacks` to find the query packs we care about, and then just use those.

I updated the PR to use this approach to identify the query packs to check for overlay compatibility.  Related changes:
- I removed the commit that adds `resolvePacks()` to the `CodeQL` interface
- Instead, there is now a commit to add `resolveQueriesStartingPacks()`
- To avoid confusion, I removed the unused `resolveQueries()` from `CodeQL`
- While I am at it, I also removed the unused `packDownload()` from `CodeQL`

The other commits remain the same as before.  PTAL.

src/codeql.ts
```diff
@@ -218,6 +218,7 @@ export interface CodeQL {
 export interface VersionInfo {
   version: string;
   features?: { [name: string]: boolean };
+  overlayVersion?: number;
```
mbg, line 226: Minor: I know the other fields here don't have documentation, but perhaps it would be useful to add a comment for `overlayVersion` to document what purpose it serves so that it's easy to look up when working on the action code.

src/codeql.ts
```diff
@@ -944,6 +952,24 @@ export async function getCodeQLForCmd(
...
+        "resolve",
+        "queries",
+        "--format=startingpacks",
+        ...getExtraOptionsFromEnv(["resolve", "queries"]),
+        ...queries,
+      ];
+      const output = await runCli(cmd, codeqlArgs, { noStreamStdout: true });
+
+      try {
+        return JSON.parse(output) as string[];
```
mbg, line 994: Not specific to this PR: I see that we follow the same `JSON.parse(output) as type` pattern throughout the implementation here, but that doesn't actually give us any guarantees that the result of `JSON.parse` is compatible with the `type` here. This may effectively delay an error until a later point. Where possible, we should probably check that the result actually what we expect (and not just valid JSON) as done in e.g. https://github.com/github/codeql-action/pull/2956.

I don't think this has to be addressed here, but we might want to look into improving this throughout in the future.

src/init.ts
```diff
@@ -104,6 +105,114 @@ export async function runInit(
...
+    return false;
+  }
+
+  for (const language of config.languages) {
+    const suitePath = path.join(
+      config.dbLocation,
+      language,
+      "temp",
+      "config-queries.qls",
+    );
```
mbg: I think this is now at least the second place where we construct this path (after my changes in https://github.com/github/codeql-action/pull/2935) so we may want to add a utility function that constructs this path for a given config+language somewhere.

cklin:
The latest force push updated `CODEQL_OVERLAY_MINIMUM_VERSION` to 2.22.3 at Alex's request.

src/init.ts
```diff
@@ -104,6 +105,114 @@ export async function runInit(
...
+    const packDirs = await codeql.resolveQueriesStartingPacks([suitePath]);
+    for (const packDir of packDirs) {
+      if (
+        !checkPackForOverlayCompatibility(packDir, codeQlOverlayVersion, logger)
+      ) {
+        return false;
+      }
+    }
```
mbg: Minor: perhaps this could be nicer as something like:

```typescript
if (packDirs.some((packDir) => !checkPackForOverlayCompatibility(packDir, codeQlOverlayVersion, logger))) {
  return false;
}
```
You could potentially also chain something together with the outer loop.

src/init.ts
```diff
@@ -101,7 +89,114 @@ export async function runInit(
...
+function checkPackForOverlayCompatibility(
+  packDir: string,
+  codeQlOverlayVersion: number,
+  logger: Logger,
+): boolean {
+  try {
+    const qlpackPath = path.join(packDir, "qlpack.yml");
+    const qlpackContents = yaml.load(
+      fs.readFileSync(qlpackPath, "utf8"),
+    ) as any;
```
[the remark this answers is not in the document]
mbg: I agree with Copilot here - it would be good to have typings for (at least) the parts of the format that are used here. 

This might also make the tests a little nicer since you can then have test objects of this type that you can serialise, rather than hard-coded strings.

src/init.ts
```diff
@@ -104,6 +105,114 @@ export async function runInit(
...
+      return true;
+    }
+
+    const packInfoPath = path.join(packDir, ".packinfo");
+    if (!fs.existsSync(packInfoPath)) {
+      logger.warning(
+        `The query pack at ${packDir} ` +
+          "does not have a .packinfo file. This pack is too old to support " +
+          "overlay analysis.",
+      );
```
mbg: Minor: This log message makes an assumption about why there is no `.packinfo` file, but there could be other reasons. This might be misleading when troubleshooting problems in the future, so it might be best to avoid drawing conclusions. How about just "[..] a .packinfo file and therefore does not support overlay analysis."?

src/init.ts
```diff
@@ -101,7 +88,149 @@ export async function runInit(
...
+        `The query pack ${packReference} at ${packInfo.path} ` +
+          "does not have a .packinfo file. This pack is too old to support " +
+          "overlay analysis.",
+      );
+      return false;
+    }
+
+    const packInfoFileContents = JSON.parse(
+      fs.readFileSync(packInfoPath, "utf8"),
+    );
```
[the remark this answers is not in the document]
mbg: Agreed with Copilot. It would be good to distinguish errors while reading/parsing JSON from other errors.

src/init.ts
```diff
@@ -104,6 +105,114 @@ export async function runInit(
...
+    const packOverlayVersion = packInfoFileContents.overlayVersion;
+    if (typeof packOverlayVersion !== "number") {
+      logger.warning(
+        `The query pack at ${packDir} is not compatible with overlay analysis.`,
+      );
```
mbg: Minor: This is very non-specific. How about "The query pack at ${packDir} does not contain the required `overlayVersion` property."?

src/init.ts
```diff
@@ -104,6 +105,114 @@ export async function runInit(
...
+      return false;
+    }
+
+    if (packOverlayVersion !== codeQlOverlayVersion) {
+      logger.warning(
+        `The query pack at ${packDir} was compiled with ` +
+          `overlay version ${packOverlayVersion}, but the CodeQL CLI ` +
+          `supports only overlay version ${codeQlOverlayVersion}. The ` +
```
mbg: The wording here suggests that `codeQlOverlayVersion` is smaller than `packOverlayVersion`, but the check just tests that they are not equal. Should the condition be different or the wording of the warning?

src/init.ts
```diff
@@ -78,21 +78,40 @@ export async function runInit(
   apiDetails: GitHubApiCombinedDetails,
   logger: Logger,
 ): Promise<TracerConfig | undefined> {
-  fs.mkdirSync(config.dbLocation, { recursive: true });
```
mbg, line 78: Just noting this for future reference: it doesn't look like `dbLocation` is used by `generateRegistries`, so moving this seems fine.

src/init-action.ts
```diff
@@ -733,6 +734,36 @@ async function run() {
...
     );
 
+    // To check custom query packs for compatibility with overlay analysis, we
+    // need to first initialize the database cluster, which downloads the
+    // user-specified custom query packs. But we also want to check custom query
+    // pack compatibility first, because database cluster initialization depends
+    // on the overlay database mode. The solution is to initialize the database
+    // cluster first, check custom query pack compatibility, and if we need to
+    // revert to `OverlayDatabaseMode.None`, re-initialize the database cluster
+    // with the new overlay database mode.
```
mbg, line 776: It seems like a good bit of effort in the last couple of commits here is spent on enabling this. I am sure you've already thought about this, but is there any way that we could change the overlay database mode for the database cluster after it has been initialised? Or alternatively, shuffle things around so that we can download the user-specified custom query packs?

I am trying to understand the design space here a bit better. I am also thinking about concurrent efforts we'll have to look into related to code quality, which might mean that we have to break up what happens when a database is initialised anyway. It might be that we can find a solution for both in one go depending on what timelines permit.

src/codeql.ts
cklin: Done.

src/codeql.ts
cklin: Acknowledged.  I will leave it to you to decide if and when you want to perform that overall cleanup.

src/init.ts
cklin: Furthermore, `codeql database init` generally expects the database path to be non-existent and reports an error when the database path already exists.  (There are some exceptions, such as when initializing an overlay database.)  So I would be very surprised if `generateRegistries()` writes anything into `dbLocation`.

cklin:
I responded to some of the comments, and will follow up on the remaining ones tomorrow.

src/init.ts
cklin: Done.

src/init.ts
cklin: Done.

src/init.ts
cklin: How about I change the message as follows?

> The query pack at ${packDir} does not have a .packinfo file, so it cannot support overlay analysis.  Recompiling the query pack with the latest CodeQL CLI should solve this problem.

I don't think even the original message was that misleading.  It explains the detected problem (that there is no .packinfo file), and then it offers the most likely cause.  I grant that it is theoretically possible for the problem to be something else, but to me all other possibilities seem extremely unlikely.  (Do you have specific other reasons in mind?)

Without the suggested likely cause, the error message would be meaningless to the vast majority of Code Scanning users.  Nobody knows what a `.packinfo` file is and what can be done to address the situation that it is missing.  If they file a support ticket, the first thing we would tell them is "your pack is probably too old and you should try to recompile it".  Including that suggestion in the error message would save this entire round-trip of communication.

src/init.ts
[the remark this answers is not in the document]
cklin: The overall catch block already includes the error message, which would make it clear what the nature of the problem is.  And a malformed `.packinfo` file should be an exceedingly rare occurrence: it would have caused the CodeQL CLI to fail with an error during pack resolution, and I don't remember ever getting a report of that happening.

Can you say more about what problems would be solved by adding a catch block specifically for JSON parsing?

src/init.ts
cklin: Mentioning the `overlayVersion` property would make the message more specific, but it would not add any useful information.

Code Scanning users do not know what the `overlayVersion` property is or what it does.  It is an implementation detail used to indicate presence and compatibility of overlay support.  If a customer files a support ticket asking about the `overlayVersion` property being absent, our answer would be "that means your pack is not compatible with overlay analysis".

So I don't see how adding that detail would be useful.

src/init.ts
cklin: The `!==` comparison is intended.  If the pack overlay version is lower than the CLI overlay version, overlay analysis is not guaranteed to work.  If the pack overlay version is higher than the CLI overlay version, overlay analysis is also not guaranteed to work.

We don't plan to ever reduce overlay versions with a new CodeQL CLI release, so the most common case would be the CLI having a higher overlay version that the pack.  But I think the error message works both ways:

> The query pack at ${packDir} was compiled with overlay version 6, but the CodeQL CLI supports only overlay version 4.  The query pack needs to be recompiled to support overlay analysis.

and

> The query pack at ${packDir} was compiled with overlay version 4, but the CodeQL CLI supports only overlay version 6.  The query pack needs to be recompiled to support overlay analysis.

src/init-action.ts
cklin: I responded in more detail in the internal tracking issue, but the short answer is that neither of the suggested approach would be practical.  Nonetheless thank you for the suggestions!

src/init.ts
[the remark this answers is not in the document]
cklin: Done.

cklin:
I responded to all comments (responses to copilot-initiated comments might be visible only under the original copilot reviews).

Also rebased the PR against current main to resolve merge conflicts.

src/codeql.ts
mbg: Thanks for adding that very detailed comment, that's very helpful in understanding the intended semantics!

src/init.ts
mbg: Ah, I see. I think the real problem here is that "only" can be ambiguous and could either be interpreted as "the versions supported by this CLI go up to only version X right now and version Y is greater" or "the version X must be an exact match for Y". I had interpreted it as the former, while you meant the latter. 

If you don't plan to support version mixing, then perhaps the wording could be improved to be less ambiguous. How about simply dropping the "only":

> The query pack at ${packDir} was compiled with overlay version 6, but the CodeQL CLI supports overlay version 4. The query pack needs to be recompiled to support overlay analysis.

cklin:
Another rebase to resolve merge conflicts.

src/init.ts
mbg: I see Copilot agreed with my comment on this 😅 

It might not be useful for users to know this implementation detail, but it would be helpful for us when troubleshooting issues related to this. It might be clear to you (and perhaps even me) what this error means since we know where in the code it is raised, but someone else might need to review the code to figure that out.

Just to be clear, I don't care whether you explicitly mention `overlayVersion` in the error or not. But it would be good to have a message that points more clearly at there being something wrong with the file (for whatever reason) than a compatibility issue for an entirely unspecified reason.

src/init.ts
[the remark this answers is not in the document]
mbg: On second thought, I think it's probably fine. You're right that whatever errors come out of here are probably distinguishable based on the result of `util.getErrorMessage(e)` in the existing `catch` block.

cklin:
@mbg Friendly ping

src/init.ts
mbg: I think your proposed change is good. Proposing a possible resolution, rather than suggesting a cause might allow most people to resolve this issue themselves if they run into it as well.

The main scenario I was thinking of is users who create their own query packs and (perhaps inadvertently) use an older CLI for that. That said, I am not sure whether users who create their own packs often `bundle` them, so as you say this is probably not a likely issue.

src/config-utils.test.ts
mbg, line 1: The PR removes `resolveQueries` and adds `resolveQueriesStartingPacks`. For the corresponding tests, the ones for `resolveQueries` are removed, but none are added for `resolveQueriesStartingPacks`. Do you think it would make sense to add some?

src/init-action.ts
mbg: Thanks for the responses. I am happy to keep this as-is since it shouldn't be a common occurrence, and responded to the internal issue with more detailed thoughts.

src/init-action.ts
```diff
@@ -744,15 +746,60 @@ async function run() {
...
+      });
+      await runDatabaseInitCluster(
+        databaseInitEnvironment,
+        codeql,
+        config,
+        sourceRoot,
+        "Runner.Worker.exe",
+        qlconfigFile,
+        logger,
+      );
```
mbg, line 798: Do you think it's worth adding a test that exercises this logic? I don't see anything wrong with things as they are, but it might be worthwhile to have to check that we don't break this sequence, especially since an error in `cleanupDatabaseClusterDirectory` would be fatal. I don't feel strongly about it.

src/init.ts
cklin: Done.

src/init.ts
cklin: Done.

src/init.ts
cklin: Done.

src/config-utils.test.ts
cklin: I did not remove any `resolveQueries` tests.  What I removed was unnecessary `resolveQueries` and `packDownload` stubs in config loading tests.  Those tests are still working the same way they used to: the removed `resolveQueries` and `packDownload` stubs has no effect on the tests because the tested code does not call either of those functions.

`resolveQueriesStartingPacks` is just a wrapper that calls the corresponding CodeQL CLI subcommand, and I don't think tests for that function would add much value.

src/init-action.ts
cklin: I added a new PR check to exercise this re-initialization logic.

cklin:
@mbg Thanks for your comments.  I think I responded to them all.

There has been a few force pushes to the PR branch, both to resolve merge conflicts and to iterate on the new database re-init PR check.  The following link contains all the relevant changes since your last review:

https://github.com/github/codeql-action/compare/37f26d0f026df3e20f15eb7bf68a3bb8d096418f..e47147711bab2cae6089f637baa13b51de265126

Edit: I updated the comparison link because I added a commit to support the `codeql-pack.yml` pack file name.

src/init.ts
```diff
@@ -101,7 +89,119 @@ export async function runInit(
...
+  packDir: string,
+  codeQlOverlayVersion: number,
+  logger: Logger,
+): boolean {
+  try {
+    const qlpackPath = path.join(packDir, "qlpack.yml");
+    const qlpackContents = yaml.load(
+      fs.readFileSync(qlpackPath, "utf8"),
+    ) as QlPack;
+    if (!qlpackContents.buildMetadata) {
```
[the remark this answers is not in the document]
mbg: Doesn't matter since the outer try/catch block will catch this and log it.

mbg (dismissed):
Thanks for making those final changes!

mbg (approved):

PR #3003, opened 2025-08-05
Rewrite legacy SARIF categories for CQ 

mbg:
In some Default Setup workflows, an old language category identifier may be given to the Action to ensure backwards compatibility for Code Scanning analyses. This doesn't apply to CQ analyses. However, since there's only one workflow for both, the best place where this can be addressed is in the Action. This PR modifies the category for CQ SARIF files to match CQ's expectations.


src/actions-util.ts
```diff
@@ -409,3 +410,44 @@ export function getPullRequestBranches(): PullRequestBranches | undefined {
 export function isAnalyzingPullRequest(): boolean {
   return getPullRequestBranches() !== undefined;
 }
+
+// A mapping from old categories to new ones.
```
henrymercer:
```suggestion
/**
 * A workaround for code quality to map category names from old default setup workflows
 * to ones that the code quality service expects.
 */
```

henrymercer (approved):
LGTM to unblock code quality — I don't love more hardcoding of information about languages, but this set shouldn't grow as we add new languages, and therefore there shouldn't be any additional work involved in adding a new language as a result of this PR.

lib/actions-util.js
```diff
@@ -392,4 +393,35 @@ function getPullRequestBranches() {
 function isAnalyzingPullRequest() {
     return getPullRequestBranches() !== undefined;
 }
+// A mapping from old categories to new ones.
+const qualityCategoryMapping = {
```
starcke, line 400: I think it is enough to convert the cases where result of the old mapping is different from the new mapping.

So I think it is enough to do the following three mappings:
```
cpp->c-cpp
java->java-kotlin
javascript->javascript-typescript
```

But it could be that I am missing something.

starcke (approved):
Seems pretty sensible and lightweight to me. But I'll defer to someone more knowledge on the action for the full review.

src/actions-util.ts
```diff
@@ -409,3 +411,47 @@ export function getPullRequestBranches(): PullRequestBranches | undefined {
...
+export function fixCodeQualityCategory(
+  logger: Logger,
+  category?: string,
+): string | undefined {
+  // The `category` should always be set by Default Setup. We perform this check
+  // to avoid potential issues if Code Quality supports Advanced Setup in the future
+  // and before this workaround is removed.
+  if (
+    category !== undefined &&
+    (isDefaultSetup() || isInTestMode()) &&
```
henrymercer: Could you explain the reasoning for enabling for all PR checks?  If we need this to be enabled in a PR check, would an explicit environment variable work better to keep tests as close to prod as possible?

mbg: It's impossible to test if `isDefaultSetup() == true` is a mandatory condition.

`fixCodeQualityCategory` is only called for Code Quality SARIF files / uploads, so it shouldn't affect any other PR checks.

I am open to adding a separate environment variable though.

henrymercer: From a unit test, I think we could mock `isDefaultSetup`, or set the `GITHUB_EVENT_NAME` environment variable.  From an integration test, I think a separate environment variable probably makes the most sense.

src/analyze.ts
```diff
@@ -691,31 +733,40 @@ export async function runQueries(
         undefined,
         sarifFile,
         config.debugMode,
+        automationDetailsId,
       );
+
+      let qualityAnalysisSummary: string | undefined = undefined;
```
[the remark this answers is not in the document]
henrymercer: I'm surprised the linter doesn't pick up on this, but think it is worth accepting the fix.

henrymercer (dismissed):

henrymercer (approved):

PR #3005, opened 2025-08-06
Make all errors on an unsupported platform `ConfigurationError`s

redsun82:


src/cli-errors.ts
```diff
@@ -156,6 +169,13 @@ export const cliErrorsConfig: Record<
   CliConfigErrorCategory,
   CliErrorConfiguration
 > = {
+  // if running on an unsupported platform, use this blanket category for all errors
+  [CliConfigErrorCategory.UnsupportedPlatform]: {
+    cliErrorMessageCandidates: [new RegExp("")],
+    precondition: isUnsupportedPlatform,
+    additionalErrorMessageToAppend:
```
henrymercer: We use this in a few other error messages to give advice about how to fix the error, but I think in this case we might want to swap the ordering, e.g. something like

"This platform/arch combination (${process.platform}/${process.arch}) is not supported by the CodeQL CLI. See https://codeql.github.com/docs/codeql-overview/system-requirements. Underlying error: <error>"

src/cli-errors.ts
```diff
@@ -298,6 +318,9 @@ export function getCliConfigCategoryIfExists(
   cliError: CliError,
 ): CliConfigErrorCategory | undefined {
   for (const [category, configuration] of Object.entries(cliErrorsConfig)) {
+    if (configuration.precondition && !configuration.precondition()) {
+      continue;
+    }
```
henrymercer: A potentially simpler option could be modifying `wrapCliConfigurationError` directly (`getCliConfigCategoryIfExists` is only used there and can be made private).  I don't have a strong opinion about it.

henrymercer (commented):
Thanks, looks good!  A couple of optional suggestions:

src/cli-errors.ts
```diff
@@ -156,6 +169,13 @@ export const cliErrorsConfig: Record<
   CliConfigErrorCategory,
   CliErrorConfiguration
 > = {
+  // if running on an unsupported platform, use this blanket category for all errors
+  [CliConfigErrorCategory.UnsupportedPlatform]: {
+    cliErrorMessageCandidates: [new RegExp("")],
+    precondition: isUnsupportedPlatform,
+    additionalErrorMessageToAppend:
+      `This platform/arch combination (${process.platform}/${process.arch}) is not supported by the CodeQL CLI. See https://codeql.github.com/docs/codeql-overview/system-requirements`,
```
mbg: Minor: "arch" seems a bit odd in this otherwise fairly well formatted message. I'd spell it out as "architecture".

src/cli-errors.ts
```diff
@@ -156,6 +169,13 @@ export const cliErrorsConfig: Record<
   CliConfigErrorCategory,
   CliErrorConfiguration
 > = {
+  // if running on an unsupported platform, use this blanket category for all errors
+  [CliConfigErrorCategory.UnsupportedPlatform]: {
+    cliErrorMessageCandidates: [new RegExp("")],
+    precondition: isUnsupportedPlatform,
```
mbg: It seems quite hacky to me to use the `cliErrorsConfig` mechanism and depend on the order + a new check that works differently from the the existing ones (the existing ones are all logically OR-ed, while `precondition` is AND-ed). 

I'd say that it would probably be better to handle this differently and insert a `isUnsupportedPlatform` check in `wrapCliConfigurationError` as @henrymercer suggests. I feel more strongly about this than him.

src/cli-errors.ts
redsun82: makes sense!

src/cli-errors.ts
redsun82: ok, I was hesitating, you're making me lean towards that too

redsun82:
@henrymercer @mbg do you think this warrants a changelog entry?

mbg:
@redsun82 I don't think a changelog entry is needed for this, since it shouldn't have any user-facing effects.

src/cli-errors.ts
```diff
@@ -317,11 +317,45 @@ export function getCliConfigCategoryIfExists(
...
+      ["darwin", "arm64"],
+    ].some(
+      ([platform, arch]) =>
+        platform === process.platform && arch === process.arch,
+    )
+  ) {
+    return new ConfigurationError(
+      "The CodeQL CLI does not support the platform/architecture combination of " +
+        `${process.platform}/${process.arch} ` +
+        "(see https://codeql.github.com/docs/codeql-overview/system-requirements). " +
```
mbg: Minor: Should we add this to the `DocUrl` enum?

src/cli-errors.ts
```diff
@@ -171,7 +171,7 @@ export const cliErrorsConfig: Record<
   },
   [CliConfigErrorCategory.GradleBuildFailed]: {
     cliErrorMessageCandidates: [
-      new RegExp("[autobuild] FAILURE: Build failed with an exception."),
+      new RegExp("\\[autobuild\\] FAILURE: Build failed with an exception."),
```
mbg, line 181: Good catch

src/cli-errors.ts
```diff
@@ -317,11 +317,45 @@ export function getCliConfigCategoryIfExists(
...
+function getUnsupportedPlatformError(
+  cliError: CliError,
+): ConfigurationError | undefined {
+  if (
+    ![
+      ["linux", "x64"],
+      ["win32", "x64"],
+      ["darwin", "x64"],
+      ["darwin", "arm64"],
+    ].some(
```
mbg: Minor: It might be nicer if this array was a top-level constant, rather than inlined here.

src/cli-errors.ts
```diff
@@ -317,11 +317,45 @@ export function getCliConfigCategoryIfExists(
...
+/**
+ * Changes an error received from the CLI to a ConfigurationError with the message
+ * optionally being transformed, if it is a known configuration error. Otherwise,
  * simply returns the original error.
  */
 export function wrapCliConfigurationError(cliError: CliError): Error {
+  const unsupportedPlatformError = getUnsupportedPlatformError(cliError);
+  if (unsupportedPlatformError !== undefined) {
+    return unsupportedPlatformError;
+  }
```
mbg: This seems a bit like an improvised try/catch block. Why not just return a bool from a function that checks if the platform is supported and then throw the error here?

src/cli-errors.test.ts
```diff
@@ -0,0 +1,421 @@
...
+// Helper function to create CommandInvocationError for testing
+function createCommandInvocationError(
+  cmd: string,
+  args: string[],
+  exitCode: number | undefined,
+  stderr: string,
+  stdout: string = "",
+): CommandInvocationError {
+  return new CommandInvocationError(cmd, args, exitCode, stderr, stdout);
+}
```
mbg: Minor: This seems a bit unnecessary given it's more or less the same as `CommandInvocationError`

src/cli-errors.test.ts
```diff
@@ -0,0 +1,421 @@
...
+});
+
+test("wrapCliConfigurationError - unsupported platform", (t) => {
+  const originalPlatform = process.platform;
+  const originalArch = process.arch;
+
+  try {
+    // Mock unsupported platform/arch
+    Object.defineProperty(process, "platform", { value: "unsupported" });
+    Object.defineProperty(process, "arch", { value: "unsupported" });
```
mbg: I believe we'd normally use `sinon` to stub a function which returns what we want for a test. I can't find any existing uses of `Object.defineProperty` in our codebase.

mbg (commented):
Thanks for making some changes and letting Copilot generate some tests. I have a few thoughts on the new version of this.

As already mentioned above, I don't think a changelog entry will be necessary, since we are just changing an error type and that doesn't affect users.

src/cli-errors.test.ts
redsun82: copilot's reasoning here was adding the default `stdout` value I think. I added that to the constructor directly

src/actions-util.ts
```diff
@@ -265,7 +265,7 @@ export class CommandInvocationError extends Error {
     public args: string[],
     public exitCode: number | undefined,
     public stderr: string,
-    public stdout: string,
+    public stdout: string = "",
```
mbg, line 268: Note: I don't actually see `stdout` being used here, only `stderr`, so this shouldn't break anything.

src/cli-errors.ts
```diff
@@ -317,11 +324,37 @@ export function getCliConfigCategoryIfExists(
...
+}
+
+/**
+ * Changes an error received from the CLI to a ConfigurationError with the message
+ * optionally being transformed, if it is a known configuration error. Otherwise,
  * simply returns the original error.
  */
 export function wrapCliConfigurationError(cliError: CliError): Error {
+  if (isUnsupportedPlatform()) {
+    return getUnsupportedPlatformError(cliError);
```
mbg, line 355: Minor: The `getUnsupportedPlatformError` function is probably not necessary and this could have been inlined here, but I don't feel strongly about it. Definitely not blocking for me.

mbg (approved):
Thanks! This looks good to me now. I added one minor comment, but feel free to ignore that.

PR #3009, opened 2025-08-07
Support auto-detecting Actions workflows

mbg:
This modifies `getRawLanguagesInRepo` to automatically detect Actions workflows. Linguist does not class those as a language, and so we failed to automatically detect them before.

I chose to implement this by checking for a non-empty `.github/workflows` directory, relative to `source-root`.


src/config-utils.ts
```diff
@@ -335,9 +359,18 @@ export async function getRawLanguagesInRepo(
...
 
   logger.debug(`Languages API response: ${JSON.stringify(response)}`);
-  return Object.keys(response.data as Record<string, number>).map((language) =>
-    language.trim().toLowerCase(),
+  const result = Object.keys(response.data as Record<string, number>).map(
+    (language) => language.trim().toLowerCase(),
   );
+
+  if (hasActionsWorkflows(sourceRoot)) {
+    logger.debug(`Found a .github/workflows directory`);
```
henrymercer, line 369:
```suggestion
    logger.debug(`Found a .github/workflows directory. Actions will be included as an autodetected language.`);
```

mbg: I was thinking about wording it like this, but then included the second `logger.debug` call that lists all languages detected by this function. I'd think the combination of the two debug messages would be enough, but I can change this if you think it's still useful.

henrymercer: On second thought, I think adding a "Autodetecting languages" log line at the start of the function would make the context clearer, then we can leave this line as is.

PR #3010, opened 2025-08-07
Clean up the database if it will be uploaded

henrymercer:
Even with `--expect-discarded-cache`, intermediate results may still be written to the database if there is high RAM pressure.  So clean up the database if we will upload it.

There are a couple of approaches to doing this.  I went with moving the cleanup into the upload stage to avoid needing to pull out separate logic about whether the upload should take place or not (and potentially needing to run this logic twice).  This does make the cleanup slightly less transparent — I added comments and don't think this is too problematic but feedback welcome.


src/codeql.ts
```diff
@@ -155,6 +155,10 @@ export interface CodeQL {
    * Run 'codeql database cleanup'.
    */
   databaseCleanup(databasePath: string, cleanupLevel: string): Promise<void>;
```
cklin: Shall we remove the `databaseCleanup()` function, since it is now replaced by `databaseCleanupCluster()` and no longer used?

PR #3011, opened 2025-08-07
Prefer providing CodeQL via dependency injection

henrymercer:
This makes it more obvious when CodeQL is being used, and therefore when we might need to provide stubs in tests.


src/analyze.test.ts
```diff
@@ -42,7 +42,7 @@ test("status report fields", async (t) => {
     sinon.stub(uploadLib, "validateSarifFileSchema");
 
     for (const language of Object.values(KnownLanguage)) {
```
mbg, line 44: Just an observation unrelated to this PR: `KnownLanguage` doesn't include "actions" anymore, so that doesn't get tested here.

mbg (approved):
Looks like a nice change, definitely makes it clearer when the CodeQL object is used.

It would be nice if we could get rid of `setCodeQL` entirely.

henrymercer:
> It would be nice if we could get rid of `setCodeQL` entirely.

Agreed, though this is trickier since we have a couple of tests that run the top-level Actions, namely `src/analyze-action-env.test.ts` and `src/analyze-action-input.test.ts`.  It might be worth seeing if we can rewrite these without losing coverage, but I'll leave that as future work.

PR #3023, opened 2025-08-11
Improve Rust analysis PR check

redsun82:
Also run the `rust` checks on "milestone" CLI releases, to ensure we remain backward compatible with those versions. This was prompted by https://github.com/github/codeql-action/pull/2960#pullrequestreview-3104730221

Running this on current `main` and then on that PR should improve our confidence we remain backward compatible.

It also turns out a probable `ruamel.yaml` update was changing a lot of generated workflows locally. It turns out the version was fixed in a couple of places but not in the `sync.sh` script I used. I've opted for writing the version exclusively in there and use it all over, which did mean updating the headers in all generated workflows to reflect the new instructions not specifying `ruamel.yaml`'s version.


pr-checks/checks/rust.yml
```diff
@@ -1,14 +1,17 @@
 name: "Rust analysis"
 description: "Tests creation of a Rust database"
-versions: ["linked", "default", "nightly-latest"]
+versions:
+  - stable-v2.19.3  # experimental rust support introduced, requires action to set `CODEQL_ENABLE_EXPERIMENTAL_FEATURES`
+  - stable-v2.22.1  # first public preview version
```
mbg: Assuming this is what primarily motived the the `yaml.width` change, I think you could just move the comments above the relevant list entries to avoid the problem?

.github/workflows/__packaging-inputs-js.yml
```diff
@@ -70,7 +70,8 @@ jobs:
         with:
           config-file: .github/codeql/codeql-config-packaging2.yml
           languages: javascript
-          packs: codeql-testing/codeql-pack1@1.0.0, codeql-testing/codeql-pack2, codeql-testing/codeql-pack3:other-query.ql
+          packs: codeql-testing/codeql-pack1@1.0.0, codeql-testing/codeql-pack2,
+            codeql-testing/codeql-pack3:other-query.ql
```
mbg: Is this even valid YAML?

pr-checks/sync.sh
```diff
@@ -4,6 +4,6 @@ set -e
 cd "$(dirname "$0")"
 python3 -m venv env
 source env/bin/activate
-pip3 install ruamel.yaml
+pip3 install ruamel.yaml==0.18.14
```
mbg: That would also need updating in `sync.py` which has `ruamel.yaml@0.17.31` hard-coded there.

pr-checks/checks/rust.yml
redsun82: no, the `yaml.width` changes the output workflows, does not take the inputs.

What I noticed is that there were a lot of new line-breaks all over the generated workflows. I thought that if I entered the "right" width I would get no changes, but couldn't find any such value. In the end I settled for a reasonable value that limited changes to only a couple of workflows.

.github/workflows/__packaging-inputs-js.yml
redsun82: uh, good question

redsun82: trying it https://eemeli.org/yaml-playground/, it is! and the newline gets replaced by a space (and dedented)

Yaml strings are still full of surprises :laughing:

pr-checks/checks/rust.yml
mbg: I think it would still be nicer to move the comments above the list elements. That's what we do elsewhere and it means you don't need such a wide code view / scrolling.

Do you have an example of what you mean by "a lot of new line-breaks all over the generated workflows"? Looking over a random sample didn't show anything odd.

.github/workflows/__packaging-inputs-js.yml
mbg: TIL. I had assumed you'd need `|`, `>`, or one of their variants.

pr-checks/checks/rust.yml
mbg: Or is that a result of updating `ruamel.yaml`? I had assumed that the update was to have `yaml.width`, but maybe I made the wrong assumption.

.github/workflows/__packaging-inputs-js.yml
redsun82: TIL indeed :laughing:

pr-checks/checks/rust.yml
redsun82: yeah, updating `ruamel.yaml` without specifying the width was breaking up lines more often than the previous version. In the end though I opted for using the version that was specified (elsewhere than `sync.sh`, which is what I was using), but also fix up that so that now the version is only written once. I will move the comments now.

mbg: Possibly a result of https://sourceforge.net/p/ruamel-yaml/tickets/529/. That's the only `width`-related change I saw in the changelog after 0.17.31

PR #3030, opened 2025-08-12
PR checks: support collections of workflows

mbg:
This is a sort-of experimental idea that I had in the context of https://github.com/github/codeql-action/pull/3028 where we bump the Go version in the workflows to a release candidate for testing purposes, but such PRs don't have to be merged anymore because the existing constraint of `>= 1.22` would include allow new version once it is fully released.

This PR updates the `sync.py` script to allow "collections" of PR checks that can be triggered manually, with specific inputs. So instead of creating a PR that bumps the Go version for testing purposes only, we could instead just dispatch the collection workflow with the desired Go version.

### Risk assessment

### DIFF
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -61,7 +61,7 @@ this_dir = pathlib.Path(__file__).resolve().parent
 
 allJobs = {}
 collections = {}
-for file in (this_dir / 'checks').glob('*.yml'):
+for file in sorted((this_dir / 'checks').glob('*.yml')):
     with open(file, 'r') as checkStream:
         checkSpecification = yaml.load(checkStream)
     matrix = []
--- a/.github/workflows/__go.yml
+++ b/.github/workflows/__go.yml
@@ -27,38 +27,6 @@ jobs:
     uses: ./.github/workflows/__go-custom-queries.yml
     with:
       go-version: ${{ inputs.go-version }}
-  go-tracing-legacy-workflow:
-    name: 'Go: tracing with legacy workflow'
-    permissions:
-      contents: read
-      security-events: read
-    uses: ./.github/workflows/__go-tracing-legacy-workflow.yml
-    with:
-      go-version: ${{ inputs.go-version }}
-  go-indirect-tracing-workaround:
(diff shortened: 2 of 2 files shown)


pr-checks/sync.py
```diff
@@ -113,11 +114,16 @@ def writeHeader(checkStream):
         installGo = True if checkSpecification['installGo'].lower() == "true" else False
 
     if installGo:
+        goVersionExpr = '>=1.21.0'
+
+        if 'inputs' in checkSpecification and 'go-version' in checkSpecification['inputs']:
+            goVersionExpr = '${{ inputs.go-version || \'>=1.21.0\' }}'
```
henrymercer: It's perhaps slightly non-ideal that we mention the `go-version` input in the workflow specifications, but `go-version` doesn't get used in any of the steps.

Another option here would be gating this on `installGo: "true"` — that way the workflow specification doesn't reference an input it doesn't use.

It's not a big deal though — I'm ok merging this.

mbg: > Another option here would be gating this on installGo: "true" — that way the workflow specification doesn't reference an input it doesn't use.

I am not sure I see what you mean here. It's already gated behind `installGo` (on L116) and doesn't appear in any (rendered) workflows where it isn't defined (at least for the `workflow_*` events, and for the other events it has the fallback to `>=1.21.0`).

henrymercer: I was thinking we could automatically add the `go-version` input to the generated collection workflow file when `installGo` is true, rather than needing to add it explicitly in `checkSpecification['inputs']`.

mbg: Thanks for clarifying! That would be nicer, I agree. 

It's probably not a bad thing that I didn't go for that initially though, since it's bought us the `inputs` processing and propagation for future use if we need it 😄

mbg:
I manually dispatched a run of `__go.yml` with a Go RC as input to check that all works as expected with the `workflow_dispatch` and `workflow_call` triggers: https://github.com/github/codeql-action/actions/runs/16963696649

PR #3031, opened 2025-08-12
Overlay: add database upload size limit

cklin:
This PR updates `uploadOverlayBaseDatabaseToCache()` to limit the maximum size of the overlay-base database that we upload to the Actions Cache.

### Risk assessment


src/overlay-database-utils.ts
```diff
@@ -18,6 +18,23 @@ export enum OverlayDatabaseMode {
...
+ * upload. Actions Cache has an overall capacity of 10 GB, and the Actions Cache
+ * client library uses zstd compression.
+ *
+ * Ideally we would apply a size limit to the compressed overlay-base database,
+ * but we cannot do so because compression is handled transparently by the
+ * Actions Cache client library. Instead we place a limit on the uncompressed
+ * size of the overlay-base database.
+ *
+ * Assuming 2.5:1 compression ratio, the 6 GB limit on uncompressed data would
+ * translate to a limit of around 2.4 GB after compression.
```
mbg, line 32: 2.4GB seems like a relatively large allowance, given that you may have multiple overlay databases, and the 10GB quota is shared between all caches and workflows in a repo. 

In the context of Advanced Setup, I am not too worried about this, because users can opt-out of overlay databases if they find that the cache interferes too much with other caches they depend on. (It would probably be nice to also give those users an option to customise the limit, but I appreciate that's unlikely to be a priority right now.)

In the context of Default Setup, this is more of a problem, because users can't opt-out but they may still have other workflows in the same repo that require the cache quota. 

That said, a limit is obviously better than no limit and I am happy to approve the PR for that reason. 

In the long run, we probably want to start working on a coherent strategy for the use of the cache quota across all three consumers we have in the CodeQL Action now (TRAP caching, dependency caching, and overlay databases). I don't think we have an internal issue to track that, and I'll open one on our team's end if it doesn't exist.

mbg (approved):
Thanks! I added some inline thoughts on the limit, but am happy to approve the PR.

You could probably also add a test for this that mocks various results of `tryGetFolderBytes` to check that the cache isn't uploaded when we get `undefined` or a result larger than the limit.

PR #3033, opened 2025-08-14
Add workflow for rolling back release

mbg:
### Risk assessment


.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+
+    uses: ./.github/workflows/prepare-release.yml
+
+  rollback:
+    name: "Create rollback release"
+    if: github.repository == 'github/codeql-action'
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    # Don't set the deployment environment for test runs
```
henrymercer, line 33:
```suggestion
    # Don't set the deployment environment for test runs
    # The Actions token does not have permissions to push changes to workflow files.
    # Since workflow files may change as part of a backport PR, we use the "Automation" environment for real runs to authenticate as a GitHub App and push these changes.
```

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+  prepare:
+    name: "Prepare release"
+    if: github.repository == 'github/codeql-action'
+
+    permissions:
+      contents: read
+
+    uses: ./.github/workflows/prepare-release.yml
+
+  rollback:
```
henrymercer, line 27: We'll soon be needing to release CodeQL Action v4 to run on Node 24.  I think the backport automation in `update-release-branch.yml` would create necessary backport releases, but I think we might want a way to more quickly rollback a change across all active release branches (without waiting for the full set of PR checks).  Perhaps it makes sense to matrix this job across all active release branches?

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+          ROLLBACK_TAG: ${{ inputs.rollback-tag || 'v0.0.0' }}
+          RELEASE_TAG: ${{ needs.prepare.outputs.version }}
+          MAJOR_VERSION_TAG: ${{ needs.prepare.outputs.major_version }}
+        run: |
+          git checkout "refs/tags/${ROLLBACK_TAG}"
+          git tag --annotate "${RELEASE_TAG}" --message "${RELEASE_TAG}"
+          git tag --annotate "${MAJOR_VERSION_TAG}" --message "${MAJOR_VERSION_TAG}" --force
+
+      - name: Push tags
+        if: github.event_name == 'workflow_dispatch'
```
henrymercer, line 113:
```suggestion
        # skip when testing
        if: github.event_name == 'workflow_dispatch'
```

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+        shell: bash
+        env:
+          RELEASE_TAG: ${{ needs.prepare.outputs.version }}
+          MAJOR_VERSION_TAG: ${{ needs.prepare.outputs.major_version }}
+        run: |
+          git push origin --atomic --force refs/tags/"${RELEASE_TAG}" refs/tags/"${MAJOR_VERSION_TAG}"
+
+      - name: Prepare partial Changelog
```
henrymercer, line 121: I think it could be clearer what we mean here by "partial changelog".  It's still not fully clear, but we could call it "release description" for instance?

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+          private-key: ${{ secrets.AUTOMATION_PRIVATE_KEY }}
+
+      - name: Create the rollback release
+        if: github.event_name == 'workflow_dispatch'
+        env:
+          PARTIAL_CHANGELOG: "${{ runner.temp }}/partial_changelog.md"
+          VERSION: "${{ needs.prepare.outputs.version }}"
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          # Do not mark this release as latest. The most recent CLI release must be marked as latest.
```
henrymercer:
```suggestion
          # Do not mark this release as latest. The most recent bundle release must be marked as latest.
```

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+          # Set as a draft to give us an opportunity to review the rollback release.
+          gh release create \
```
henrymercer, line 153: Will this print a link to the new release?  It might be nice to add a link to the job summary to make it easier to move forwards to the next step of reviewing and publishing the release.

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
+name: Rollback release
+on:
+  # You can trigger this workflow via workflow dispatch to start a rollback.
+  # This will create a draft release that mirrors the release for `rollback-tag`.
+  workflow_dispatch:
```
henrymercer, line 5: I assume this is intended to be triggered on `main`?

.github/workflows/rollback-release.yml
```diff
@@ -0,0 +1,170 @@
...
+          NEW_BRANCH: "${{ steps.mergeback-branch.outputs.new-branch }}"
+        run: |
+          git checkout "${NEW_BRANCH}"
+          mv ${NEW_CHANGELOG} CHANGELOG.md
+
+      - name: Create mergeback branch and PR
+        uses: ./.github/actions/prepare-mergeback-branch
+        with:
+          base: "main"
+          head: ""
```
henrymercer, line 175: Why is this blank?

.github/workflows/script/rollback_changelog.py
```diff
@@ -0,0 +1,62 @@
...
+  print(f"This release rolls back {rollback_version} due to issues with that release. It is identical to {target_version}.\n")
+  print(after_first_section)
+
+# We expect three version strings as input:
+#
+# - target_version: the version that we are re-releasing as `new_version`
+# - rollback_version: the version that we are rolling back, typically the one that followed `target_version`
+# - new_version: the new version that we are releasing `target_version` as, typically the one that follows `rollback_version`
+#
+# Example: python3 .github/workflows/script/rollback_changelog.py "1.2.3" "1.2.4" "1.2.5"
```
henrymercer: If it's an easy change to make (perhaps Copilot can do it) named arguments via something like argparse would make this easier to read

.github/workflows/rollback-release.yml
mbg: Yeah, that's the intention. I think it could probably work if triggered from a different branch, but I haven't verified that.

.github/workflows/rollback-release.yml
mbg: I used the existing terminology for this from the release workflow. Happy to change this to something else here, but we should probably update it there then as well to be consistent.

.github/workflows/rollback-release.yml
mbg: The `prepare-mergeback-branch` action only uses it as part of the mergeback PR title, where it is typically something like `refs/heads/releases/v3`. That didn't make sense here, because we are not touching the `v3` branch.

.github/workflows/rollback-release.yml
mbg: I'd propose doing this in a follow-up PR once we have the basic version of this workflow working.

.github/workflows/rollback-release.yml
mbg: I don't think it automatically does that, but I have pushed a change which computes the release URL and adds it to the job summary.

henrymercer (approved):

henrymercer (approved):
Nice work, thanks for the detailed testing!

PR #3036, opened 2025-08-15
Add workflow generator option for installing Java

mbg:
Gradle 9 is now the default version of Gradle on the GHA runners. This seems to cause failures because JDK 8 is the default and Gradle 9 requires JDK 17.

This PR adds a new `installJava` option to the `sync.py` script which behaves exactly the same as `installGo`, except for `setup-java`.

I have added `installJava: "true"` to the two workflows we have seen fail.

### Risk assessment


henrymercer (approved):
CCR is correct, otherwise LGTM!

PR #3014, opened 2025-08-07
Enable rebuilding PRs with conflicts

henrymercer:
Since we check in our compiled code, we frequently get merge conflicts in the compiled code.  We have a rebuild Action, but label triggers don't run when there are merge conflicts.  So this PR adds a `workflow_dispatch` trigger that lets us rebuild a PR with conflicts via `workflow_dispatch`.

Example run: https://github.com/github/codeql-action/actions/runs/16809277853

### Merge / deployment checklist

- [ ] Confirm this change is backwards compatible with existing workflows.
- [ ] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [ ] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #3054, opened 2025-08-21
Bundle Actions using esbuild

henrymercer:
This PR bundles the Action using esbuild.

### Reasoning

- Avoid risks of long paths in `node_modules`, specifically the risk that these paths could exceed `MAX_PATH` on Windows when checked out.
- Speed up the download of the Action, and every code scanning workflow as a result.

### Downsides we should be aware of

- Potential for more merge conflicts
  - Mitigation: I haven't enabled minification which I hope will reduce the likelihood of merge conflicts
  - Mitigation: You can now run the rebuild workflow on PRs even if they have merge conflicts (#3014)

### Risk assessment

This is a risky change.  Things I've done to minimise the risk:

- Avoid changing the module resolution system — this is tricky and requires us to change a lot of our unit tests.
- Keep the compiled Actions in the same place, so that users invoking them directly do not need to change their paths.
- Keep `defaults.json` copied over to `lib`.
- Minimise changes to unit testing setup by preserving the "compile to JS" workflow that's known to work reasonably well with ava.

### DIFF
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -36379,7 +36379,6 @@ var require_package = __commonJS({
         lint: "eslint --report-unused-disable-directives --max-warnings=0 .",
         "lint-ci": "SARIF_ESLINT_IGNORE_SUPPRESSED=true eslint --report-unused-disable-directives --max-warnings=0 . --format @microsoft/eslint-formatter-sarif --output-file=eslint.sarif",
         "lint-fix": "eslint --report-unused-disable-directives --max-warnings=0 . --fix",
-        removeNPMAbsolutePaths: "removeNPMAbsolutePaths . --force",
         test: "npm run transpile && ava src/**.test.ts --serial --verbose",
         "test-debug": "npm run test -- --timeout=20m",
         transpile: "tsc --build --verbose"
@@ -36454,7 +36453,6 @@ var require_package = __commonJS({
         "eslint-plugin-no-async-foreach": "^0.1.1",
         glob: "^11.0.3",
         nock: "^14.0.10",
-        removeNPMAbsolutePaths: "3.0.1",
         sinon: "^21.0.0",
         typescript: "^5.9.2"
       },
--- a/lib/autobuild-action.js
+++ b/lib/autobuild-action.js
@@ -36379,7 +36379,6 @@ var require_package = __commonJS({
         lint: "eslint --report-unused-disable-directives --max-warnings=0 .",
         "lint-ci": "SARIF_ESLINT_IGNORE_SUPPRESSED=true eslint --report-unused-disable-directives --max-warnings=0 . --format @microsoft/eslint-formatter-sarif --output-file=eslint.sarif",
         "lint-fix": "eslint --report-unused-disable-directives --max-warnings=0 . --fix",
-        removeNPMAbsolutePaths: "removeNPMAbsolutePaths . --force",
         test: "npm run transpile && ava src/**.test.ts --serial --verbose",
         "test-debug": "npm run test -- --timeout=20m",
         transpile: "tsc --build --verbose"
@@ -36454,7 +36453,6 @@ var require_package = __commonJS({
         "eslint-plugin-no-async-foreach": "^0.1.1",
         glob: "^11.0.3",
         nock: "^14.0.10",
-        removeNPMAbsolutePaths: "3.0.1",
         sinon: "^21.0.0",
         typescript: "^5.9.2"
       },
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -36379,7 +36379,6 @@ var require_package = __commonJS({
         lint: "eslint --report-unused-disable-directives --max-warnings=0 .",
         "lint-ci": "SARIF_ESLINT_IGNORE_SUPPRESSED=true eslint --report-unused-disable-directives --max-warnings=0 . --format @microsoft/eslint-formatter-sarif --output-file=eslint.sarif",
         "lint-fix": "eslint --report-unused-disable-directives --max-warnings=0 . --fix",
-        removeNPMAbsolutePaths: "removeNPMAbsolutePaths . --force",
         test: "npm run transpile && ava src/**.test.ts --serial --verbose",
         "test-debug": "npm run test -- --timeout=20m",
         transpile: "tsc --build --verbose"
@@ -36454,7 +36453,6 @@ var require_package = __commonJS({
         "eslint-plugin-no-async-foreach": "^0.1.1",
         glob: "^11.0.3",
         nock: "^14.0.10",
-        removeNPMAbsolutePaths: "3.0.1",
         sinon: "^21.0.0",
         typescript: "^5.9.2"
       },
--- a/lib/resolve-environment-action.js
+++ b/lib/resolve-environment-action.js
@@ -36379,7 +36379,6 @@ var require_package = __commonJS({
         lint: "eslint --report-unused-disable-directives --max-warnings=0 .",
         "lint-ci": "SARIF_ESLINT_IGNORE_SUPPRESSED=true eslint --report-unused-disable-directives --max-warnings=0 . --format @microsoft/eslint-formatter-sarif --output-file=eslint.sarif",
         "lint-fix": "eslint --report-unused-disable-directives --max-warnings=0 . --fix",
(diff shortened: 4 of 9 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -38,7 +38,6 @@ jobs:
         run: .github/workflows/script/check-js.sh
 
   check-file-contents:
-    if: github.event_name != 'push' || github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/releases/v')
```
aeisenberg, line 54: What's the reason for this (ande below)? Are we running this for PR events as well?

aeisenberg (commented):
Nice. I haven't tried this locally, but it looks good.

.github/workflows/pr-checks.yml
henrymercer: The `if` condition prevented running this on a push to a development branch, but it's actually a useful thing to run on push since it's a common mistake to forget to update the PR checks.  That way the person pushing the branch gets a notification and potentially fixes the issue before they open a PR.

henrymercer: Happy to move this to another PR, but it seemed like a fairly small change :)

aeisenberg: I'm fine with it if it's useful for you, but this seems to go against the general idea of trying to minimize the number of actions runs that the codeql-action uses. I'm no longer maintaining this project, so you can handle this in the way you feel is most appropriate.

henrymercer: I combined them all into a single job in a later commit.  I think a single job per push is fine, and these checks have been catching things early.

aeisenberg:
If I have time later, I'll try checking this PR out and making sure the dev flow still works for me.

.github/workflows/codescanning-config-cli.yml
```diff
@@ -55,6 +55,16 @@ jobs:
...
       uses: actions/checkout@v5
+
+    - name: Set up Node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20'
+        cache: 'npm'
+
+    - name: Install dependencies
+      run: npm install
```
aibaars: I think `npm ci` is better.
```suggestion
      run: npm ci
```

henrymercer:
We'll need to update the set of required checks before merging this PR.

aibaars (commented):
Let's run `npm ci` instead of `npm install`. Also don't forget to update poackage-lock.json`

build.mjs
```diff
@@ -0,0 +1,46 @@
...
+    to: ["defaults.json"],
+  },
+});
+
+const context = await esbuild.context({
+  entryPoints: globSync(["src/*-action.ts", "src/*-action-post.ts"]),
+  bundle: true,
+  format: "cjs",
+  outdir: OUT_DIR,
+  platform: "node",
```
navntoft, line 69: Consider adding [target](https://esbuild.github.io/api/#target) for good measure. If we happen to use syntax that whatever lowest version of Node we support doesn't handle, esbuild will complain early.

build.mjs
```diff
@@ -0,0 +1,46 @@
+import { rm } from "node:fs/promises";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+import * as esbuild from "esbuild";
+import { globSync } from "glob";
+import { copy } from "esbuild-plugin-copy";
+import { typecheckPlugin } from "@jgoz/esbuild-plugin-typecheck";
```
navntoft: Note that these plugins are relatively unknown, which adds risk both wrt. security and reliability.

I usually prefer to type-check separately. This speeds up turnaround time locally (not sure if applicable here), and it is pretty trivial to copy files (see [Prior](https://github.com/github/semmle-code/blob/main/buildutils-internal/prior/dev/build.js#L76)). Your call, though.

navntoft (commented):
> Avoid changing the module resolution system — this is tricky and requires us to change a lot of our unit tests.

Note that bundling CJS is less reliable than ESM due to its dynamic nature. ESBuild is designed with ESM in mind, and there has been (at least historically) some gotchas wrt. its CJS bundling.

build.mjs
henrymercer: The typechecking plugin means we can do the typechecking in parallel with the bundling.  On the other hand, we do currently keep a local compilation for the purposes of running the tests.  So we're probably not adding much time to the dev flow in practice by moving compilation from `npm test` to `npm run build`.

build.mjs
henrymercer: > Note that the target setting in tsconfig.json is only used by esbuild for determining the default value of useDefineForClassFields. It does not affect esbuild's own [target](https://esbuild.github.io/api/#target) setting, even though they have the same name.

TIL

henrymercer:
> Note that bundling CJS is less reliable than ESM due to its dynamic nature. ESBuild is designed with ESM in mind, and there has been (at least historically) some gotchas wrt. its CJS bundling.

Is there a different bundler you'd recommend for CJS?  I chose ESBuild since it's much faster than webpack and we're using it successfully in other projects, but we could explore something else.

aibaars:
> > Note that bundling CJS is less reliable than ESM due to its dynamic nature. ESBuild is designed with ESM in mind, and there has been (at least historically) some gotchas wrt. its CJS bundling.
> 
> Is there a different bundler you'd recommend for CJS? I chose ESBuild since it's much faster than webpack and we're using it successfully in other projects, but we could explore something else.

It looks like the typescript-action template repo uses `rollup`, no idea if it is better. https://github.com/actions/typescript-action/blob/main/package.json#L33 . They used to use `ncc` but switched not to long ago to rollup: https://github.com/actions/typescript-action/pull/969

henrymercer:
> It looks like the typescript-action template repo uses `rollup`, no idea if it is better. https://github.com/actions/typescript-action/blob/main/package.json#L33 . They used to use `ncc` but switched not to long ago to rollup: [actions/typescript-action#969](https://github.com/actions/typescript-action/pull/969)

I believe `rollup` was designed for ES modules perhaps more so than `esbuild` — you need to use [a plugin](https://github.com/rollup/plugins/tree/master/packages/commonjs) to use it with CommonJS modules.

navntoft:
> > It looks like the typescript-action template repo uses `rollup`, no idea if it is better. [actions/typescript-action@`main`/package.json#L33](https://github.com/actions/typescript-action/blob/main/package.json?rgh-link-date=2025-08-22T10%3A33%3A22Z#L33) . They used to use `ncc` but switched not to long ago to rollup: [actions/typescript-action#969](https://github.com/actions/typescript-action/pull/969)
> 
> I believe `rollup` was designed for ES modules perhaps more so than `esbuild` — you need to use [a plugin](https://github.com/rollup/plugins/tree/master/packages/commonjs?rgh-link-date=2025-08-22T10%3A33%3A22Z) to use it with CommonJS modules.

`esbuild` has been my favourite tool for bundling. It's easy to use, still fairly flexible, and fast. I have only used it for ESM bundling, though. I know that `vite` has used `esbuild` for transpilation and `rollup` for bundling (not sure if still true). Historically `webpack` has had the best support for CJS bundling, but it is slow and bloated. I haven't kept up on the bundling scene the past 2 years though, so a lot may have happened...

build.mjs
```diff
@@ -0,0 +1,74 @@
...
+      await rm(join(OUT_DIR, "defaults.json"), {
+        force: true,
+      });
+      await copyFile(
+        join(SRC_DIR, "defaults.json"),
+        join(OUT_DIR, "defaults.json"),
+      );
+    });
+  },
+}
```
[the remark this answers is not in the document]
henrymercer: We should try to lint this file when practical.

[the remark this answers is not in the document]
navntoft: The TS codebases from CS-IDX use prettier, FYI, with VS Code being set up to reformat on save. While I don't always like the way it formats things it removes the cognitive effort of having to think about semicolons (I don't type them, actually; prettier inserts them for me...)

navntoft (dismissed):
The esbuild setup looks sound. You will likely want 👀 from somebody else as well as I don't know of the complexities around distributing codeql-action.

As I said, there can be nasty surprises when bundling CJS, so be sure to test various setups.

PR #3055, opened 2025-08-21
Report which analysis kinds are enabled in status reports

mbg:
This adds a new field to `StatusReportBase` for a comma-separated list of analysis names that are active. 

### Risk assessment


src/config-utils.ts
```diff
@@ -1468,3 +1468,12 @@ export function generateCodeScanningConfig(
   }
   return augmentedConfig;
 }
+
+/**
+ * Returns `true` if Code Quality analysis is enabled, or `false` if not.
+ */
+export function isCodeQualityEnabled(config: Config): config is Config & {
```
henrymercer, line 1475: This could in future reference an `analysis-kinds` input.

mbg:
> My only question is would we rather wait for the analysis_kinds input so we can be more sure about when code scanning is enabled or not?

I think I am happy to merge this as-is (once I have made the other necessary backend changes), since it should be accurate for the current setup. I can then iterate on the details as we move forward and update this accordingly. That's also why I added the `isCodeQualityEnabled` helper so that this is easy to update with subsequent changes.

PR #3061, opened 2025-08-27
Add `analysis-kinds` input

mbg:
This PR follows up from #3055 and adds an `analysis-kinds` input to the `init` action. This is a required input, but defaults to `code-scanning` so that there's no change to existing workflows. The input is parsed and the resulting analysis kinds are stored in the configuration. 

The `isCodeQualityEnabled` function is updated to inspect that configuration, instead of the `quality-queries` input.

For backwards compatibility, we enable the `code-quality` analysis kind if there is an input to `quality-queries`.

Fundamentally, this PR shouldn't change the existing behaviour of the Action in any way.

### Risk assessment


init/action.yml
```diff
@@ -42,6 +42,19 @@ inputs:
...
+    description: >-
+      [Internal] A comma-separated list of analysis kinds to enable. This input is intended for
+      internal-use only at this time and the behaviour is subject to changes. Some features may
+      not be available depending on which analysis kinds are enabled.
+
+      Available options are:
+
+      - `code-scanning`: The default, security-focused analysis.
+      - `code-quality`: Analysis focused on code quality. This must be enabled in conjunction
+                        with `code-scanning`.
```
henrymercer, line 55: Out of interest, where's this restriction coming from?  We can address it in a follow up PR, but I think we want to be able to run code quality only analyses.

mbg: Yes, this restriction will be lifted in a follow-up PR. It's just representative of the current state of the implementation as of this PR.

henrymercer (dismissed):

PR #3063, opened 2025-08-28
Pin Xcode 16

mbg:
### Risk assessment

### DIFF
--- a/pr-checks/checks/export-file-baseline-information.yml
+++ b/pr-checks/checks/export-file-baseline-information.yml
@@ -5,10 +5,6 @@ installGo: "true"
 env:
   CODEQL_ACTION_SUBLANGUAGE_FILE_COVERAGE: true
 steps:
-  - name: Use Xcode 16
-    shell: bash
-    if: runner.os == 'macOS'
-    run: sudo xcode-select -s "/Applications/Xcode_16.app"
   - uses: ./../action/init
     id: init
     with:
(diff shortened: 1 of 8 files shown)


pr-checks/checks/export-file-baseline-information.yml
```diff
@@ -5,15 +5,15 @@ installGo: "true"
 env:
   CODEQL_ACTION_SUBLANGUAGE_FILE_COVERAGE: true
 steps:
+  - name: Use Xcode 16
+    shell: bash
+    if: runner.os == 'macOS'
+    run: sudo xcode-select -s "/Applications/Xcode_16.app"
```
redsun82: I wonder: do we want to downgrade Xcode also in a test that runs on `nightly-latest`, knowing that the latest CLI works with the latest Xcode on the runner, and we actually want to keep it that way?

pr-checks/checks/swift-autobuild.yml
```diff
@@ -3,15 +3,16 @@ description: "Tests creation of a Swift database using autobuild"
 versions: ["nightly-latest"]
 operatingSystems: ["macos"]
 steps:
+  - name: Use Xcode 16
+    shell: bash
+    if: runner.os == 'macOS'
+    run: sudo xcode-select -s "/Applications/Xcode_16.app"
```
redsun82: Same comment, maybe we want to keep using default Xcode with `nightly-latest`?

mbg:
@redsun82 I have removed the pinning step from the two workflows that only use `nightly-latest` and added an extra condition on the step for the other two workflows so that it doesn't run when `matrix.version == 'nightly-latest'`.

redsun82 (approved):

PR #3064, opened 2025-08-28
Allow Code Quality only analysis

mbg:
### Risk assessment


src/config-utils.ts
```diff
@@ -1111,10 +1118,45 @@ export async function initConfig(inputs: InitConfigInputs): Promise<Config> {
...
+  // Compute the full Code Scanning configuration that combines the configuration from the
+  // configuration file / `config` input with other inputs, such as `queries`.
+  const codeScanningConfig = generateCodeScanningConfig(
+    userConfig,
+    config.augmentationProperties,
+  );
+
+  // If Code Scanning analysis is disabled, then we initialise the database for Code Quality.
+  // That entails disabling the default queries and only running quality queries.
+  if (!isCodeScanningEnabled(config)) {
```
henrymercer: I like that we're looking at things that won't work proactively, but would consider changing the condition here.  It works since the analysis kinds are only code scanning and code quality, but I think in general the logic we want is more like `config.analysisKinds.length === 1 && isCodeQualityEnabled(config)`.

src/config-utils.ts
```diff
@@ -1111,10 +1118,45 @@ export async function initConfig(inputs: InitConfigInputs): Promise<Config> {
...
+    // Warn if any query customisations are present in the computed configuration.
+    if (hasQueryCustomisation(codeScanningConfig)) {
+      logger.warning(
+        "Query customizations will be ignored, because only `code-quality` analysis is enabled.",
+      );
```
henrymercer: Should we actually error in these cases?  It's easy to miss Actions warnings, and users could think they're analysing something that doesn't actually work.  That also avoids us the slight confusion of then having to override `config.originalUserInput`.

src/upload-lib.ts
```diff
@@ -943,7 +913,7 @@ function handleProcessingResultForUnsuccessfulExecution(
 
 export function validateUniqueCategory(
   sarif: SarifFile,
-  sentinelPrefix: string = CodeScanningTarget.sentinelPrefix,
+  sentinelPrefix: string = analyses.CodeScanning.sentinelPrefix,
```
henrymercer: Do we need a default value here?  Removing it might decrease the potential for errors.

src/config-utils.ts
```diff
@@ -1489,11 +1533,41 @@ export function generateCodeScanningConfig(
...
+    : AnalysisKind.CodeQuality;
+}
+
+/**
+ * Returns the analysis configuration for which the database is initialised. This is
+ * always `CodeScanning` unless `CodeScanning` is not enabled.
+ *
+ * @returns Returns `CodeScanning` if `AnalysisKind.CodeScanning` is enabled; otherwise `CodeQuality`.
+ */
+export function getDbAnalysisConfig(config: Config): AnalysisConfig {
```
henrymercer: This seems like a detail of how saving SARIF results currently works, in particular the fact that when both modes are on, we currently produce a code scanning file, then filter that to produce a code quality file.  I don't really think it's a property of the database, since we'll ask the database to analyse both code scanning and code quality results in that situation.

I'd consider removing this in favour of some more explicit logic concerning the SARIF extension in `analyze`, unless there are more situations where this notion of "primary consumer" matters.

src/config-utils.ts
mbg: I wasn't sure about this, but decided to go with a warning because in Default Setup we provide inputs to `queries` and `config`. We can probably change that there for when Code Quality is enabled, but Code Scanning isn't, but I didn't want to make this an error just yet in case it causes us problems during the transition there.

src/config-utils.ts
mbg: > This seems like a detail of how saving SARIF results currently works, in particular the fact that when both modes are on, we currently produce a code scanning file, then filter that to produce a code quality file. I don't really think it's a property of the database, since we'll ask the database to analyse both code scanning and code quality results in that situation.

I don't think this is true, since in CQ-only mode the database is specifically initialised with `code-quality` queries. What `getDbAnalysisConfig` tells us is whether the database is initialised with a Code Scanning configuration (in CS only or CS+CQ mode) or a Code Quality configuration (in CQ only mode). 

It's true though that this is currently only relevant when we are generating the "primary" SARIF file.

src/config-utils.ts
henrymercer: I appreciate wanting to make the transition smooth, but I have some concerns about modifying `config.originalUserInput`, particularly since we use this to generate telemetry.

My understanding is today we have analyses with a `quality-queries` input and without an `analysis-kinds` input.  This means we'll run code scanning, and therefore shouldn't hit this case.  Then the transition can move to an explicit `analysis-kinds` input with the `config` and other customization inputs removed.

Is there anything I'm missing here?

src/config-utils.ts
henrymercer: > I don't think this is true, since in CQ-only mode the database is specifically initialised with `code-quality` queries. What `getDbAnalysisConfig` tells us is whether the database is initialised with a Code Scanning configuration (in CS only or CS+CQ mode) or a Code Quality configuration (in CQ only mode).

I'm unsure how much meaning this has in practice.  For instance, if I really wanted to, I could create a code scanning configuration that runs the code quality queries only by disabling the default queries and running a custom query pack.  I'd expect the database to be identical to a database with `analysis-kinds: code-quality`.  The only difference I can see is the name of the SARIF file and the upload API.

src/config-utils.ts
mbg: I think there are two separate points here:

1. Whether we can swap over from using `quality-queries` to `analysis-kinds` and remove the `queries` input in CQ-only mode in one go. (Probably "yes" and this being a warning is just overly defensive.)
2. That modifying `config.originalUserInput` is an issue for telemetry.

Regarding 2., I agree that this isn't a very nice approach. I'll have a look at refactoring this tomorrow so that `originalUserInput` can remain as-is and we set up the configuration for a CQ-only database differently.

henrymercer (approved):

PR #3065, opened 2025-08-29
Use brace-expansion >2.0.1

redsun82:
### Risk assessment


package.json
```diff
@@ -52,7 +52,8 @@
     "path": "^0.12.7",
     "semver": "^7.7.2",
     "uuid": "^11.1.0",
-    "zlib": "^1.0.5"
+    "zlib": "^1.0.5",
+    "brace-expansion": ">=2.0.1"
```
mbg: Minor: I believe the dependencies are in alphabetical order, so it would be good if this addition maintained that order.

package-lock.json
```diff
@@ -7195,6 +7211,16 @@
         "node": "*"
       }
     },
+    "node_modules/minimatch/node_modules/brace-expansion": {
+      "version": "1.1.12",
```
henrymercer: Note that we're still pulling in v1.1.12 here — if we want to override this I think we need to modify the `overrides` property in `package.json`.

aibaars: Either that or update all the dependencies that rely on old versions of brace-expansion

henrymercer: Right, if we can update `minimatch`, that's better.  We've had situations before where we've wanted to bump a transient dependency before the direct dependency had bumped it.

redsun82: Aha, now I see why dependabot didn't manage to do this update. I'll get back to it on Monday.

package.json
aibaars: We don't depend on brace-expansion ourselves, right? Perhaps best to try to update the dependency that pulls in the outdated library version.

package-lock.json
redsun82: (notice I didn't tell which Monday :stuck_out_tongue_winking_eye: :sweat_smile: on it now)

package.json
redsun82: unfortunately, that dependency comes from https://www.npmjs.com/package/@actions/glob, which is stuck at 0.5.0 from a year ago

aibaars: Let's ask the actions folk to update their dependencies.

redsun82: looking better, that dependency points to brace-expansion < 2.0.0, which is not problematic. The only problematic ones are `2.0.0` (that we don't have) and `2.0.1`. I tried retracing where that comes from, but in the end just resorted to an explicit override of that specific version

redsun82:
> I had a look at the lockfile and we're still pulling in v1.1.11 of `brace-expansion` — does that need updating too?

No, I double checked and the versions that need updating are just 2.0.0 (which we don't use) and 2.0.1. Versions <2.0.0 are ok.

PR #3071, opened 2025-09-02
Fix missing/wrong properties in `createTestConfig`

mbg:
Extracted from #3064. Fixes some incorrect properties in `createTestConfig`.

### Risk assessment

### DIFF
--- a/src/testing-utils.ts
+++ b/src/testing-utils.ts
@@ -379,7 +379,7 @@ export function createTestConfig(overrides: Partial<Config>): Config {
       trapCaches: {},
       trapCacheDownloadTime: 0,
       dependencyCachingEnabled: CachingKind.None,
-    } as Config,
+    } satisfies Config,
     overrides,
   );
 }


src/testing-utils.ts
```diff
@@ -371,11 +373,13 @@ export function createTestConfig(overrides: Partial<Config>): Config {
...
         extraQueryExclusions: [],
+        overlayDatabaseMode: OverlayDatabaseMode.None,
+        useOverlayDatabaseCaching: false,
       },
       trapCaches: {},
       trapCacheDownloadTime: 0,
-      dependencyCachingEnabled: false,
-    },
+      dependencyCachingEnabled: CachingKind.None,
+    } as Config,
```
esbena:
```suggestion
    } satisfies Config,
```
Will this work?

esbena: (`as` is a "cast", `satisfies` is an assertion)

mbg: Thanks for spotting! Yes, that makes sense. The `as Config` was just something I had thrown in after noticing the problem while working on #3064 so I hadn't thought about it too much.

esbena (approved):

PR #3072, opened 2025-09-02
Overlay databases: Use `Config` instead of `AugmentationProperties`

mbg:
Extracted from #3064. Moves `extraQueryExclusions` and overlay database configuration out of `AugmentationProperties` and into `Config`. 

The reasoning for this change is that `AugmentationProperties` is intended for user inputs to the `init` Action that cannot be represented in the same format that is used by `UserConfig` (which is understood by the CodeQL CLI). Options stored in the `AugmentationProperties` are then intended to be combined with an existing `UserConfig` into a new `UserConfig`.

We will remove `AugmentationProperties` from `Config` (used to represent the CodeQL Action configuration / state that is shared across different steps in a CodeQL workflow) in #3075. This makes sense, because the purpose of `AugmentationProperties` (as explained above) is just to represent user inputs to the Action until they can be converted into a `UserConfig`.

### Risk assessment

### DIFF
--- a/src/config-utils.ts
+++ b/src/config-utils.ts
@@ -1489,13 +1489,14 @@ export function generateCodeScanningConfig(
 export function appendExtraQueryExclusions(
   extraQueryExclusions: ExcludeQueryFilter[],
   cliConfig: UserConfig,
-): UserConfig {
-  if (extraQueryExclusions.length === 0) {
-    return cliConfig;
-  }
-
-  // make a copy so we can modify it
+): Readonly<UserConfig> {
+  // make a copy so we can modify it and so that modifications to the input
+  // object do not affect the result that is marked as `Readonly`.
   const augmentedConfig = cloneObject(cliConfig);
+
+  if (extraQueryExclusions.length === 0) {
+    return augmentedConfig;
+  }
 
   augmentedConfig["query-filters"] = [
     // Ordering matters. If the first filter is an inclusion, it implicitly
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -91586,10 +91586,10 @@ async function getConfig(tempDir, logger) {
   return JSON.parse(configString);
 }
 function appendExtraQueryExclusions(extraQueryExclusions, cliConfig) {
-  if (extraQueryExclusions.length === 0) {
-    return cliConfig;
-  }
   const augmentedConfig = cloneObject(cliConfig);
+  if (extraQueryExclusions.length === 0) {
+    return augmentedConfig;
+  }
   augmentedConfig["query-filters"] = [
     // Ordering matters. If the first filter is an inclusion, it implicitly
     // excludes all queries that are not included. If it is an exclusion,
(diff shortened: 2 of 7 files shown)


src/config-utils.ts
```diff
@@ -1162,7 +1159,10 @@ export async function initConfig(inputs: InitConfigInputs): Promise<Config> {
       logger,
     ))
   ) {
-    augmentationProperties.extraQueryExclusions.push({
+    if (config.computedConfig["query-filters"] === undefined) {
+      config.computedConfig["query-filters"] = [];
+    }
+    config.computedConfig["query-filters"].push({
```
cklin: We need to be a careful here.  Ordering matters in `query-filters`, and there was a recent bug where we accidentally overrode the user-specified `query-filters` setting by putting our own exclusion at the top of the list.  See comments at https://github.com/github/codeql-action/blob/68d7fe3565231256a37124e5e4daf84c69e835ca/src/config-utils.ts#L1478-L1485

Saving the programmatic exclusions in `extraQueryExclusions` allows us to easily ensure that those exclusions always come after user-specified filters.  If we are mixing them all together in the same ``query-filters` field, we would need some other way to ensure proper ordering.

mbg: Yep, I am aware that the ordering here is critical and that you previously fixed an issue related to that. 

For context, the main motivation for this PR is that the overlay database settings don't really belong in `AugmentationProperties` and we are working on removing the `AugmentationProperties` in `Config`. 

In other words, the line you commented on does not produce a `UserConfig` that's later combined with other `AugmentationProperties` using `generateCodeScanningConfig`. The `UserConfig` that is mutated on the line you've commented on is the `UserConfig` that is given to the CLI. Since we are appending `exclude-from-incremental` here, it should have the same effect as what you previously accomplished by modifying `AugmentationProperties`.

cklin: First, thank you for the explanation!

There are two parts to my concern.
1. That this PR preserves the existing behavior of putting the `exclude-from-incremental` exclusion after user-specified query filters
2. That we help future developers (which could well be our future selves) avoid introducing this regression by accident

You made a strong case that you addressed the first part, and I agree with you on that.  What about the second part?

mbg: Thanks for separating those points. I think what 2. has made me realise here is that you had introduced `extraQueryExclusions` to `AugmentationProperties` as part of your work on the Action in the last few months, but that it also doesn't really belong into `AugmentationProperties`.

What I could propose then to address 2. is to move `extraQueryExclusions` to `Config` as well, add `exclude-from-incremental` to it here, and then apply `extraQueryExclusions` to the `query-filters` of the `UserConfig` that we pass on to the CLI as late as possible (i.e. just before invoking the CLI). That would then behave similarly to what we had previously, except we don't have the baggage of the entire `AugmentationProperties`.

There might be more we could do to protect against shooting ourselves in the foot, but I think that's largely outside the scope of this change.

mbg: I have pushed a commit which moves `extraQueryExclusions` out of `AugmentationProperties` into `Config`. The part of the code from `generateCodeScanningConfig` that appends the `extraQueryExclusions` to a `UserConfig` is now in its own function as well. I have added a bit more documentation at the point where we append the extra query exclusions that links to your previous PR as well.

henrymercer (approved):

src/config-utils.ts
cklin: Thank you!  These changes addressed all my concerns.

PR #3075, opened 2025-09-02
Remove `augmentationProperties` from `Config` type

mbg:
(#3072 should be merged first, and this PR is based on that.)

This PR updates `sendCompletedStatusReport` to use the computed packs from `computedConfig` rather than computing the packs from the input `UserConfig` and `AugmentationProperties` locally. 

As a result, `AugmentationProperties` is no longer needed in `Config` and is removed.

### Risk assessment


src/config-utils.ts
```diff
@@ -1104,14 +1100,14 @@ export async function initConfig(inputs: InitConfigInputs): Promise<Config> {
     );
   }
 
-  const config = await getDefaultConfig(inputs);
+  const { augmentationProperties, ...config } = await getDefaultConfig(inputs);
```
henrymercer: Given that we only use `getDefaultConfig` in this function, consider having `getDefaultConfig` populate `computedConfig` directly to simplify the flow.

mbg: I have made the `UserConfig` (from the inputs) an argument to `getDefaultConfig` (now `initActionState`), set it there, and generate `computedConfig` there. That avoids having to return `augmentationProperties` as well.

mbg:
@henrymercer I refactored creating the `InitWithConfigStatusReport` into a function and added some unit tests for that. That should be more robust than checking the debug logs, but let me know if you still want me to do that as well?

henrymercer (approved):
Great, even better!

PR #3080, opened 2025-09-04
Overlay: add automation ID to cache key

cklin:
This PR adds automation ID to the overlay-base database cache key so that we properly distinguish different analyses in the same repo for the same language.

Since I am changing the cache key format, I also moved the CodeQL bundle version to the end of the cache restore key, in case we want to remove it from the restore key sometime in the future.

Note that I chose to leave `CACHE_VERSION` unchanged because the old and the new cache keys are sufficiently different that there should be no risk of confusion.

Changes in this PR has been validated in an internal test repository.

### Risk assessment

### DIFF
--- a/src/overlay-database-utils.ts
+++ b/src/overlay-database-utils.ts
@@ -253,15 +253,19 @@ export async function uploadOverlayBaseDatabaseToCache(
 
   const codeQlVersion = (await codeql.getVersion()).version;
   const checkoutPath = getRequiredInput("checkout_path");
-  const cacheKey = await generateCacheKey(config, codeQlVersion, checkoutPath);
+  const cacheSaveKey = await getCacheSaveKey(
+    config,
+    codeQlVersion,
+    checkoutPath,
+  );
   logger.info(
-    `Uploading overlay-base database to Actions cache with key ${cacheKey}`,
+    `Uploading overlay-base database to Actions cache with key ${cacheSaveKey}`,
   );
 
   try {
     const cacheId = await withTimeout(
       MAX_CACHE_OPERATION_MS,
-      actionsCache.saveCache([dbLocation], cacheKey),
+      actionsCache.saveCache([dbLocation], cacheSaveKey),
       () => {},
     );
     if (cacheId === undefined) {
@@ -324,10 +328,14 @@ export async function downloadOverlayBaseDatabaseFromCache(
 
   const dbLocation = config.dbLocation;
   const codeQlVersion = (await codeql.getVersion()).version;
-  const restoreKey = await getCacheRestoreKey(config, codeQlVersion);
+  const cacheRestoreKeyPrefix = await getCacheRestoreKeyPrefix(
+    config,
+    codeQlVersion,
+  );
 
   logger.info(
-    `Looking in Actions cache for overlay-base database with restore key ${restoreKey}`,
+    "Looking in Actions cache for overlay-base database with " +
+      `restore key ${cacheRestoreKeyPrefix}`,
   );
 
   let databaseDownloadDurationMs = 0;
@@ -335,7 +343,7 @@ export async function downloadOverlayBaseDatabaseFromCache(
     const databaseDownloadStart = performance.now();
     const foundKey = await withTimeout(
       MAX_CACHE_OPERATION_MS,
-      actionsCache.restoreCache([dbLocation], restoreKey),
+      actionsCache.restoreCache([dbLocation], cacheRestoreKeyPrefix),
       () => {
         logger.info("Timed out downloading overlay-base database from cache");
       },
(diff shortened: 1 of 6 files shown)


src/overlay-database-utils.ts
```diff
@@ -407,5 +413,43 @@ function getCacheRestoreKey(config: Config, codeQlVersion: string): string {
...
+ * Creates a SHA-256 hash of the cache key components to ensure uniqueness
+ * while keeping the cache key length manageable.
+ *
+ * @param components Object containing all components that should influence cache key uniqueness
+ * @returns A short SHA-256 hash (first 16 characters) of the components
+ */
+function createCacheKeyHash(components: Record<string, any>): string {
+  const componentsJson = JSON.stringify(
+    components,
+    Object.keys(components).sort(),
```
mbg: Minor: This makes sense, but it's not very obvious from the documentation for `JSON.stringify` that the keys in the first argument will be stringified in the order given by the second argument, so I had to convince myself that this works as intended by actually trying it. Perhaps worth a small note here or a quick unit test that throws some permutations of an object at this function and checks that the resulting hash is the same?

src/overlay-database-utils.ts
```diff
@@ -407,5 +413,43 @@ function getCacheRestoreKey(config: Config, codeQlVersion: string): string {
...
+  //
+  // CACHE_PREFIX: distinguishes overlay-base databases from other cache objects
+  // CACHE_VERSION: cache format version
+  // componentsHash: hash of additional components (see above for details)
+  // languages: the languages included in the overlay-base database
+  // codeQlVersion: CodeQL bundle version
+  //
+  // Technically we can also include languages and codeQlVersion in the
+  // componentsHash, but including them explicitly in the cache key makes it
+  // easier to debug and understand the cache key structure.
```
mbg, line 457: Minor: I was initially going to ask whether `languages` should be hashed too (since it can potentially be a long list), but then I saw your comment here, which makes sense. 

For completeness, another option to consider here might be to only hash `languages` if there is more than one. That keeps the benefit of making it easier to understand for the "usual" case, while avoiding overly long keys.

src/overlay-database-utils.ts
```diff
@@ -407,5 +413,43 @@ function getCacheRestoreKey(config: Config, codeQlVersion: string): string {
...
-  return `${CACHE_PREFIX}-${CACHE_VERSION}-${languages}-${codeQlVersion}-`;
+
+  const cacheKeyComponents = {
+    automationID: await getAutomationID(),
+    // Add more components here as needed in the future
+  };
+  const componentsHash = createCacheKeyHash(cacheKeyComponents);
+
+  // For a cached overlay-base database to be considered compatible for overlay
+  // analysis, all components in the cache restore key must match:
```
mbg, line 447: I don't recall if we discussed the possibility of using partial restore keys in the PR that added the overlay caches (I skimmed over it again, but didn't see anything). You have probably considered it already (since there's a comment further up talking about partial restore keys), but it could be worth adding once you are happy with how stable the overlay database functionality is (e.g. to use an existing overlay database when a new CodeQL CLI version is released without dropping the CodeQL version from the cache key).

src/overlay-database-utils.ts
```diff
@@ -393,10 +395,14 @@ async function generateCacheKey(
   checkoutPath: string,
 ): Promise<string> {
   const sha = await getCommitOid(checkoutPath);
-  return `${getCacheRestoreKey(config, codeQlVersion)}${sha}`;
+  const restoreKey = await getCacheRestoreKey(config, codeQlVersion);
+  return `${restoreKey}${sha}`;
```
mbg: One thing I am not sure I realised until now is that `restoreCache` is happy to use the primary cache key as a prefix for restoring a cache. I think I was previously under the assumption that only the partial restore keys (if any) were prefix-matched. It might be worth adding a comment for this somewhere, since it might otherwise be non-obvious how restoring the cache can work if `sha` is included here, but not when restoring the cache.

src/overlay-database-utils.ts
```diff
@@ -393,10 +395,14 @@ async function generateCacheKey(
...
 }
 
-function getCacheRestoreKey(config: Config, codeQlVersion: string): string {
+async function getCacheRestoreKey(
+  config: Config,
+  codeQlVersion: string,
+): Promise<string> {
   // The restore key (prefix) specifies which cached overlay-base databases are
   // compatible with the current analysis: the cached database must have the
   // same cache version and the same CodeQL bundle version.
```
mbg: Should this comment be removed now that there's a more detailed one further down / merged with that one?

mbg (approved):
This is great, thank you for proactively tackling this problem! I agree with your reasoning for keeping the cache version the same as well.

I only have a few minor, non-blocking comments.

src/overlay-database-utils.ts
cklin: That is a good question.  I read [the JSON.stringify() documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify), and you are right that there is no mentioning of the `replacer` array (2nd argument here) affecting the ordering of object properties.

What I found was a promise that object property ordering is already guaranteed to be stable without the `replacer` array. So I copied that documentation into a code comment and removed the `replacer` arugment.

src/overlay-database-utils.ts
cklin: Thanks for the suggestion!

When I looked into it a while back, IIRC most analyses list 1 (most common) or 2 (much less common) languages. A handful have 3, and none have more. So I think I will keep things the way they are right now.

src/overlay-database-utils.ts
cklin: Thanks for the interesting comment! We will keep that in consideration.

src/overlay-database-utils.ts
cklin: You are right that there is a lot of overlap between the two comment blocks, but they serve somewhat different purposes.

I think of this comment as mainly about "why are we returning only one key when Actions Cache supports multiple restore keys".  The comment near the end of this function, in contrast, is about what goes into the restore key.

In any case, your point is well taken, and I updated the comments to remove the overlap.

src/overlay-database-utils.ts
cklin: I thought about it for quite a while. What is an effective way to highlight the prefix matching used for cache restore?

I ended up making three changes:
- Rename functions and variables to highlight the difference between "save key" and "restore key"
- Append "Prefix" to the restore key function and variable names to highlight the fact that they are key prefixes
- Add comment that the save key consists of the restore key prefix followed by the checkout SHA

Hopefully that will make things clearer!

src/overlay-database-utils.ts
mbg: Thanks for looking into that! I didn't realise either that this is actually just normal `JSON.stringify` behaviour - TIL! Your updated comment is excellent and more than sufficient in light of that. Thank you!

src/overlay-database-utils.ts
mbg: Thank you for taking the time to think about this and come up with a few changes to make this clearer! Out of those three changes, I think appending "Prefix" to the names has made the most difference, since it more clearly communicates that it is _intentionally_ a prefix of the cache key. 

I probably would have liked a comment (e.g. on `getCacheRestoreKeyPrefix`) that notes that the _primary_ key is prefix-matched and therefore omitting `sha` works fine. The subtlety here is that `restoreCache` has separate parameters for the "primary key" (a string; which you use with for the cache key prefix) and the "restore keys" (a string array; which you don't use right now). Although semantically they seem to work as if the primary key could just be the first element of restore keys, they are separated by both the `cache` library and Action. The existing comment for `getCacheRestoreKeyPrefix` talks about "restore keys" in "Actions cache supports using multiple restore keys" which can be interpreted as being about the "restore keys" parameter rather than the "primary key" one, so it's not clear from this that the "primary key" can also be a prefix. 

This is somewhat pedantic and definitely not blocking for this PR!

PR #3083, opened 2025-09-05
 Resolve supported languages using CodeQL CLI 

henrymercer:
This PR enables resolving supported languages using `codeql resolve languages --filter-to-languages-with-queries`, behind a feature flag.  This allows users to more easily start using new extractors with the CodeQL Action and simplifies adding new languages.

### Risk assessment

### DIFF
--- a/src/config-utils.ts
+++ b/src/config-utils.ts
@@ -305,6 +305,7 @@ export function getUnknownLanguagesError(languages: string[]): string {
 export async function getSupportedLanguageMap(
   codeql: CodeQL,
   features: FeatureEnablement,
+  logger: Logger,
 ): Promise<Record<string, string>> {
   const resolveSupportedLanguagesUsingCli = await features.getValue(
     Feature.ResolveSupportedLanguagesUsingCli,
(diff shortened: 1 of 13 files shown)


henrymercer:
Waiting on the next nightly to start integration testing this, but the code is ready for a first pass review.

src/tools-features.ts
```diff
@@ -5,10 +5,11 @@ import type { VersionInfo } from "./codeql";
 export enum ToolsFeature {
   AnalysisSummaryV2IsDefault = "analysisSummaryV2Default",
   DatabaseInterpretResultsSupportsSarifRunProperty = "databaseInterpretResultsSupportsSarifRunProperty",
+  ForceOverwrite = "forceOverwrite",
```
mbg, line 9: Minor: I assume you moved `ForceOverwrite` so that the elements are in alphabetical order, but the new `BuiltinExtractorsSpecifyDefaultQueries` entry is at the end. Should that come after `AnalysisSummaryV2IsDefault` for consistency or was there another reason you moved `ForceOverwrite`?

src/config-utils.ts
```diff
@@ -313,6 +314,11 @@ export async function getSupportedLanguageMap(
   const resolveResult = await codeql.betterResolveLanguages({
     filterToLanguagesWithQueries: resolveSupportedLanguagesUsingCli,
   });
+  if (resolveSupportedLanguagesUsingCli) {
+    logger.debug(
+      `The CodeQL CLI supports the following languages: ${Object.keys(resolveResult.extractors).join(", ")}`,
+    );
+  }
```
mbg, line 321: Minor: Since this is a `.debug` call, we can probably skip the `resolveSupportedLanguagesUsingCli` check since it wouldn't hurt having this in the debug output either way. Alternatively, I also wouldn't be opposed to making this an `info`-level message if the `resolveSupportedLanguagesUsingCli` check is kept.

mbg (commented):
I had an initial look over this and it looks good to me so far, thanks! There are just a couple of very minor points I thought of.

src/config-utils.ts
henrymercer: I think this info is only relevant if `resolveSupportedLanguagesUsingCli` is true — otherwise it contains the whole list of extractors including HTML etc

PR #3089, opened 2025-09-09
Implement dumping of processed sarif file

redsun82:
This introduces a `CODEQL_ACTION_SARIF_DUMP_DIR` environment variable for internal use that causes the processed SARIF file that `codeql-action` would upload to Code Scanning or Code Quality to be dumped locally. Crucially, this will also happen on `analyze` with `upload: never`.

This will allow us to test what is actually sent for upload after the SARIF processing that `upload-lib.ts` does (validating, merging, filtering, adding fingerprints etc.).

This SARIF file will be dumped in the directory specified by the environment variable, with `upload.sarif` or `upload.quality.sarif` as name. There is a risk of naming conflict if an output SARIF file has that name (which depends on the category), but I'm deeming it low risk (and not happening for our testing usage, where the file names are language names).

### Risk assessment

- Run analysis with this on a test repository


src/environment.ts
```diff
@@ -119,4 +119,10 @@ export enum EnvVar {
    * Whether to enable experimental extractors for CodeQL.
    */
   EXPERIMENTAL_FEATURES = "CODEQL_ENABLE_EXPERIMENTAL_FEATURES",
+
+  /**
+   * Whether to dump the processed SARIF file that would be uploaded, regardless of whether
+   * the upload is disabled. This is intended for testing and debugging purposes.
+   */
+  DUMP_SARIF = "CODEQL_ACTION_DUMP_SARIF",
```
esbena: Suggestion: make the caller manage the location the dump happens at.
```suggestion
  DUMP_SARIF_TO_DIR = "CODEQL_ACTION_DUMP_SARIF_TO_DIR",
```
This makes us less dependent on internal path choices.

mbg: I agree -- that's what I had in mind when I proposed the environment variable. I'd say, name it `CODEQL_ACTION_SARIF_DUMP_DIR` though to make it clearer that its value is expected to be a path. `DUMP_SARIF_TO_DIR` still makes me think it expects a boolean.

src/analyses.ts
```diff
@@ -85,4 +88,5 @@ export const CodeQuality: AnalysisConfig = {
   sarifExtension: ".quality.sarif",
   sarifPredicate: (name) => name.endsWith(CodeQuality.sarifExtension),
   sentinelPrefix: "CODEQL_UPLOAD_QUALITY_SARIF_",
+  sarifIdOutput: "quality-sarif-id",
```
mbg: Fun fact, I never actually added this as a possible output in https://github.com/github/codeql-action/blob/d7a501da0195366a45a932783bb4b1817372bf20/analyze/action.yml because I don't super love these separate outputs for security and code quality.

The [docs suggest that isn't required](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#outputs-for-docker-container-and-javascript-actions) so I left it out while I am still thinking about the best path forward. 

I think ideally, we'd return a list of SARIF files in a new output like `sarif-ids` and I may make that change as part of my ongoing work in this area later this week.

That's not super relevant to you, except it probably doesn't make sense to codify this name here and start building "infrastructure" around.

src/environment.ts
esbena: Agreed.

mbg (commented):
I have had an initial look over this. It looks like a more involved change than I had hoped -- I was hoping we could just check the environment variable once we have the post-processed SARIF, just before uploading it, and then dump it to a file/folder.

This part of the Action is evolving quite a bit at the moment due to the work on Code Quality and I was planning to make a bunch of changes here in the next few days. They probably wouldn't interact nicely with these changes, though.

redsun82:
> I have had an initial look over this. It looks like a more involved change than I had hoped -- I was hoping we could just check the environment variable once we have the post-processed SARIF, just before uploading it, and then dump it to a file/folder.
> 
> This part of the Action is evolving quite a bit at the moment due to the work on Code Quality and I was planning to make a bunch of changes here in the next few days. They probably wouldn't interact nicely with these changes, though.

I had hoped that as well, but:
* we want to do the processing and dump the sarif even if `upload: false`, so we need a way to arrive at the dump point but skip the uploading
* another unfortunately subtle point, is where to dump this file. At the moment we would do the dumping in the old version of the code we would only have a bunch of input sarif files. By unifying the access to the uploading to only one function with one parameter (so avoiding the `uploadFiles`/`uploadSpecifiedFiles` duality), we can decide a bit more clearly where to put the new file (in the directory, if the input is a directory, or as a sibling of the input, if it is a file). I could have turned `CODEQL_ACTION_DUMP_SARIF` into the desired output path, but decided against that because of the new Code Scanning/Code Quality duality (we would have needed two separate env variables...)

src/analyses.ts
redsun82: so would you rather have me revert this to setting that output explicitly in the two actions using this?

src/environment.ts
redsun82: yeah, makes sense actually

redsun82:
> * but decided against that because of the new Code Scanning/Code Quality duality (we would have needed two separate env variables...)

but in light of the other comments here, it makes sense to make that point to a directory instead

esbena:
>  but decided against that because of the new Code Scanning/Code Quality duality (we would have needed two separate env variables...)

We could still have `code-scanning`/`code-quality` folders inside our desired output dir. It's still a smaller dependency on internal path choices.

redsun82:
> > but decided against that because of the new Code Scanning/Code Quality duality (we would have needed two separate env variables...)
> 
> We could still have `code-scanning`/`code-quality` folders inside our desired output dir. It's still a smaller dependency on internal path choices.

we already have the `.sarif` / `.quality.sarif` discriminant for that. Pointing the env var to a directory is a good choice I think

redsun82:
@mbg I tried a more surgical approach now (I recommend hiding whitespace in the diff view). I did like the fact that I had only one function in the lib processing the sarifs (as opposed to `uploadFiles/uploadSpecifiedFiles`), but I can understand it's more pragmatic to limit the changes done in the area, especially as it is a moving target with the Code Quality work.

I still had to change the analyze action as well, because of the requirement to dump the SARIF file if requested even if `upload: false`.

src/analyses.ts
redsun82: that's what I ended up doing :slightly_smiling_face:

redsun82:
> * One PR which implements the environment variable controlling where the post-processed SARIF is written to, but which will only do this just before uploading (i.e. the simple case we hoped for). Hopefully that is an easy change that we can extract from this PR and quickly review and merge.

Fair enough, here's the PR :slightly_smiling_face: :
* https://github.com/github/codeql-action/pull/3097

redsun82:
@mbg should we maybe chat about this again?

I thought that maybe we could add a special `dry-run` value to the `upload` option to drive this, independently of the internal dump variable. Passing `dry-run` would mean the same thing as `always`, save for skipping the actual final download.

redsun82:
Obsoleted by:
* https://github.com/github/codeql-action/pull/3097
* https://github.com/github/codeql-action/pull/3180

PR #3091, opened 2025-09-09
Set `shell: bash` by default on all workflows

redsun82:
The default behaviour of steps in absence of a `shell` settings is error prone:
* if the steps runs on multiple platforms, it will use a different shell on POSIX and Windows, which can lead to surprising effects.
* if the step runs multiple lines, it won't fail if anything else than the last line fails, which can also lead to workflows hiding failures.

This change just sets `defaults: {run: {shell: bash}}` on _all_ workflows (generated and written by hand), and removes `shell: bash` from individual steps as no more needed. With this convention in place, we don't risk falling in that trap again.

To add up motivation to why we want to do this: this has uncovered some checks that we weren't really running on Windows, and that wereactually failing there because of CRLF vs LF issues. Luckily this was only concerning dev functionality rather than production one. In any case this PR fixes those issues as well.

### Risk assessment

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -99,7 +99,8 @@ jobs:
         tools: ${{ matrix.tools }}
     # confirm steps.init.outputs.codeql-path points to the codeql binary
     - name: Print CodeQL Version
-      run: "$CODEQL" version --format=json
+      run: >
+        "$CODEQL" version --format=json
       env:
         CODEQL: ${{steps.init.outputs.codeql-path}}
     - name: Perform CodeQL Analysis


pr-checks/sync.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 cd "$(dirname "$0")"
 python3 -m venv env
-source env/bin/activate
+source env/*/activate
```
mbg, line 6: Python 🤯 🤦🏻‍♂️

.github/workflows/pr-checks.yml
```diff
@@ -22,6 +26,10 @@ jobs:
     timeout-minutes: 45
 
     steps:
+      - name: Prepare git (Windows)
+        if: runner.os == 'Windows'
```
mbg, line 30: Minor: Checking `matrix.os == 'windows-latest'` might be a bit easier to work with here, since the string appears in the matrix and is therefore less of a magic value. I.e. without checking what possible values `runner.os` has, I wouldn't be sure that `'Windows'` is correct.

.github/workflows/pr-checks.yml
```diff
@@ -22,6 +26,10 @@ jobs:
...
+        run: git config --global core.autocrlf false
```
mbg, line 31: I'm mildly surprised that there isn't an option for `actions/checkout@v5` to set this / that `true` is the default on Windows.

.github/workflows/codeql.yml
```diff
@@ -95,7 +99,7 @@ jobs:
         tools: ${{ matrix.tools }}
     # confirm steps.init.outputs.codeql-path points to the codeql binary
     - name: Print CodeQL Version
-      run: ${{steps.init.outputs.codeql-path}} version --format=json
+      run: '${{steps.init.outputs.codeql-path}}' version --format=json
```
mbg: Minor: Should this be stored in an environment variable if we are already changing this line?

pr-checks/checks/cpp-deptrace-disabled.yml
```diff
@@ -18,7 +17,6 @@ steps:
       working-directory: autobuild-dir
     env:
       CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES: false
-  - shell: bash
```
henrymercer, line 21: Here, and in a couple of other places, the PR removes the `-` for the start of the new step as well as just the `shell` directive, so the workflow becomes invalid.  I think we want to start a new step with the `run`.

mbg, line 21: You removed the `-` here by accident. May be worth adding a `name` to make this more obvious.

pr-checks/checks/cpp-deptrace-enabled-on-macos.yml
```diff
@@ -18,7 +17,6 @@ steps:
       working-directory: autobuild-dir
     env:
       CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES: true
-  - shell: bash
```
mbg, line 21: Same here.

pr-checks/checks/cpp-deptrace-enabled.yml
```diff
@@ -18,7 +17,6 @@ steps:
       working-directory: autobuild-dir
     env:
       CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES: true
-  - shell: bash
```
mbg, line 21: And here.

pr-checks/checks/cpp-deptrace-disabled.yml
henrymercer: Unfortunately, we just get pending checks rather than failures for these invalid workflows.  Perhaps a tool like `actionlint` could surface these errors more visibly?

pr-checks/checks/go-indirect-tracing-workaround.yml
```diff
@@ -12,10 +12,8 @@ steps:
       languages: go
       tools: ${{ steps.prepare-test.outputs.tools-url }}
   - name: Build code
-    shell: bash
     run: go build main.go
   - uses: ./../action/analyze
-  - shell: bash
```
mbg, line 18: And here.

pr-checks/checks/go-tracing-autobuilder.yml
```diff
@@ -12,7 +12,6 @@ steps:
       tools: ${{ steps.prepare-test.outputs.tools-url }}
   - uses: ./../action/autobuild
   - uses: ./../action/analyze
-  - shell: bash
```
mbg, line 15: And here.

pr-checks/checks/go-tracing-custom-build-steps.yml
```diff
@@ -9,10 +9,8 @@ steps:
       languages: go
       tools: ${{ steps.prepare-test.outputs.tools-url }}
   - name: Build code
-    shell: bash
     run: go build main.go
   - uses: ./../action/analyze
-  - shell: bash
```
mbg, line 15: And here.

pr-checks/checks/go-tracing-legacy-workflow.yml
```diff
@@ -11,7 +11,6 @@ steps:
       languages: go
       tools: ${{ steps.prepare-test.outputs.tools-url }}
   - uses: ./../action/analyze
-  - shell: bash
```
mbg, line 14: And here.

pr-checks/checks/unset-environment.yml
```diff
@@ -15,13 +15,11 @@ steps:
       languages: cpp,csharp,go,java,javascript,python,ruby
       tools: ${{ steps.prepare-test.outputs.tools-url }}
   - name: Build code
-    shell: bash
     run: env -i PATH="$PATH" HOME="$HOME" ./build.sh
   - uses: ./../action/analyze
     id: analysis
     with:
       upload-database: false
-  - shell: bash
```
mbg, line 24: And here.

mbg (changes requested):
@henrymercer noticed that some of the workflows didn't start because you accidentally removed `-`s where `shell` was the first key in a step definition. I have marked all the places where I have spotted that problem, but it would be good if you could check as well to make sure I haven't missed any.

pr-checks/checks/cpp-deptrace-disabled.yml
mbg: I guess it might complain that the steps have both `uses` and `run` which isn't valid. At the same time, I feel like this could be a CodeQL `actions` query too.

.github/workflows/pr-checks.yml
redsun82: indeed, it has been [discussed](https://github.com/actions/checkout/issues/226) [before](https://github.com/actions/checkout/issues/135). FYI, we also do this on all `semmle-code` workflows

.github/workflows/codeql.yml
redsun82: makes sense

.github/workflows/pr-checks.yml
redsun82: on the other hand, checking `runner.os` will work even if we change the runners (for example pinning the runner version) or if we copy such a snippet in another workflow. All in all, I prefer the stability and independence of context of `runner.os`, it's just less prone to errors. One thing to note, all string comparisons on actions are case insensitive, so `runner.os == 'windows'` works as well. `runner.os` and `runner.arch` are good things to keep in mind when writing workflows.

pr-checks/checks/cpp-deptrace-disabled.yml
redsun82: aaah, hadn't noticed in my automatic search and remove, good catch!

.github/workflows/pr-checks.yml
mbg: My comment was less about whether string comparison is case-sensitive or what `runner` properties exist, but more that `runner.os` could theoretically have other plausible values like `'win'` or `'win11'` etc. -- we know as routine Actions users that this may not be the case, but it is not as obvious generally as the `matrix.os` comparison.

adityasharad: I think `runner.os` is fairly stable - it's always one of `Linux/Windows/macOS` according to https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context (and we've relied on that in the past). Agree it's not super obvious from your workflow though.

.github/workflows/codeql.yml
```diff
@@ -95,7 +99,9 @@ jobs:
         tools: ${{ matrix.tools }}
     # confirm steps.init.outputs.codeql-path points to the codeql binary
     - name: Print CodeQL Version
-      run: ${{steps.init.outputs.codeql-path}} version --format=json
+      run: "$CODEQL" version --format=json
```
henrymercer: This is a parse error since YAML treats `"$CODEQL"` as the string.  Something like:

```suggestion
      run: >
        "$CODEQL" version --format=json
```

should hopefully work.

PR #3093, opened 2025-09-09
Resolve pylint `no-else-return / R1705`

vil02:
This change resolves the [`no-else-return / R1705`](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html) warning. **No** further problems of this type are reported.
### Risk assessment

- ~**High risk:** Changes are not fully under feature flags, have limited visibility and/or cannot be tested outside of production.~

### DIFF
--- a/.github/update-release-branch.py
+++ b/.github/update-release-branch.py
@@ -145,8 +145,7 @@ def get_truncated_commit_message(commit):
   message = commit.commit.message.split('\n')[0]
   if len(message) > 60:
     return f'{message[:57]}...'
-  else:
-    return message
+  return message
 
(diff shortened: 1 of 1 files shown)


mbg:
Hi @vil02 👋🏻 

Thanks for opening this PR! We appreciate your interest and generally welcome contributions to this repo, but we don't currently use `pylint` for the handful of Python scripts that we have here. 

We're not keen on fixing a given linter issue, if we don't have anything in place to check that we generally don't violate a given linter's rules and that future changes to the codebase won't again violate the rule(s) being fixed.

We'd consider a PR that automates checking changes using e.g. `pylint`, but for now I'll go ahead and close this particular PR.

PR #3098, opened 2025-09-10
Overlay: Increase size limit for cached overlay base database

kaspersv:
This PR increases the size limit imposed on (uncompressed) overlay base databases for uploading to the Actions Cache from 6GB to 15GB. This is intended to unblock [staff-ship of Ruby overlay analysis](https://github.com/github/codeql-team/issues/4075) to the `github/github` repository which currently produces an (uncompressed) [overlay base database at 11GB](https://github.com/github/codeql-core/issues/5037). With an expected compression ratio of 2.5:1 the 15GB uncompressed limit translates into an effective 6GB limit. With a default Actions Cache quota of 10GB, this represents the majority of the Actions Cache quota. However, Actions Cache quotas are not enforced immediately. We are planning to [increase the size of the Actions Cache for `github/github`](https://github.slack.com/archives/C041NH2DFUJ/p1757489260601599) using the recently staff-shipped [larger action cache quotas](https://github.slack.com/archives/C0H15BV4K/p1755780627274089) as soon as [enforcement is staff-shipped](https://github.slack.com/archives/C07E2J3C1HD/p1757418080766049).  

Overlay analysis is [currently limited to the `github` and `dsp-testing` orgs](https://github.com/github/codeql-action/blob/31d3ae847e3e655d6e31918ac1d8af398338a360/src/config-utils.ts#L759) and controlled by feature flags (`codeql_action_overlay_analysis_*`). This size limit will be revisited before removing the restriction to the `github` and `dsp-testing` org. One possibility would be to make the limit relative to the actual Actions Cache size. 

Once this has been merged, we ideally need to release the `codeql-action`.

### Risk assessment

- **Low risk:** Changes are under feature flags and restricted to the `github` and `dsp-testing` orgs.


src/overlay-database-utils.ts
```diff
@@ -22,18 +22,23 @@ export const CODEQL_OVERLAY_MINIMUM_VERSION = "2.22.3";
...
  * but we cannot do so because compression is handled transparently by the
  * Actions Cache client library. Instead we place a limit on the uncompressed
  * size of the overlay-base database.
  *
- * Assuming 2.5:1 compression ratio, the 6 GB limit on uncompressed data would
- * translate to a limit of around 2.4 GB after compression.
+ * Assuming 2.5:1 compression ratio, the 15 GB limit on uncompressed data would
+ * translate to a limit of around 6 GB after compression. This is a high limit
+ * compared to the default 10GB Actions Cache capacity, but enforcement of Actions
+ * Cache quotas is not immediate.
```
mbg, line 36: I like the addition of a comment acknowledging that this is a large limit, but I don't think that justifying it based on the speed of the enforcement makes sense.

```suggestion
 * compared to the default 10GB Actions Cache capacity.
```

kaspersv: I do think the speed of enforcement makes quite a big difference in practice with a limit that is over 50% of the Actions Cache quota. If each compressed overlay base database was +5GB and existing cache entries were immediately evicted from the cache to make room before a new cache entry could be uploaded, then the cache wouldn't be able to fit two overlay base databases and other concurrent CodeQL action runs on PRs would effectively fall back to normal analysis while a new overlay base databases was being uploaded. That will not happen because enforcement of quotas is not immediate.

mbg: Note that the cache quota is not specific to CodeQL and is shared among all Actions caches for a repo. It may be that the speed of enforcement makes a difference for the overlay databases if those are the only caches for a repo and they are this large, but it would still hurt customers if we rapidly take up most of their quota and the rate of cache churn resulting from overlay caches competing with customer caches within the standard 10GB quota would plausibly also evict overlay databases before they can be useful.

PR #3099, opened 2025-09-10
Validate workflow to check that all `codeql-action` versions are the same

mbg:
This PR adds an extra check to `validateWorkflows` which collects all the refs used for `github/codeql-action` steps in the workflow and checks that they are all the same.

If there is a mismatch, then this results in a warning.

### Risk assessment

### DIFF
--- a/src/workflow.ts
+++ b/src/workflow.ts
@@ -168,10 +168,7 @@ export async function getWorkflowErrors(
   for (const job of Object.values(doc?.jobs || {})) {
     if (Array.isArray(job.steps)) {
       for (const step of job.steps) {
-        if (
-          step.uses !== undefined &&
-          step.uses.startsWith("github/codeql-action/")
-        ) {
+        if (step.uses?.startsWith("github/codeql-action/")) {
           const parts = step.uses.split("@");
           if (parts.length >= 2) {
             codeqlStepRefs.push(parts[parts.length - 1]);
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -90168,7 +90168,7 @@ async function getWorkflowErrors(doc, codeql) {
   for (const job of Object.values(doc?.jobs || {})) {
     if (Array.isArray(job.steps)) {
       for (const step of job.steps) {
-        if (step.uses !== void 0 && step.uses.startsWith("github/codeql-action/")) {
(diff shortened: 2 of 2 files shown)


src/workflow.ts
```diff
@@ -161,6 +162,32 @@ export async function getWorkflowErrors(
...
 
+  // Check that all `github/codeql-action` steps use the same ref, i.e. the same version.
+  // Mixing different versions of the actions can lead to unpredictable behaviour.
+  const codeqlStepRefs: string[] = [];
+  for (const job of Object.values(doc?.jobs || {})) {
+    if (Array.isArray(job.steps)) {
+      for (const step of job.steps) {
+        if (
+          step.uses !== undefined &&
+          step.uses.startsWith("github/codeql-action/")
```
esbena:
```suggestion          step.uses?.startsWith("github/codeql-action/")
```

src/workflow.ts
```diff
@@ -161,6 +162,32 @@ export async function getWorkflowErrors(
...
+        ) {
+          const parts = step.uses.split("@");
+          if (parts.length >= 2) {
```
esbena, line 173: Minor: don't we have this kind of parsing elsewhere already?

mbg: We have 
```typescript
  const format = new RegExp(
    "(?<owner>[^/]+)/(?<repo>[^/]+)/(?<path>[^@]+)@(?<ref>.*)",
  );
```
in `getRemoteConfig`, but I thought I'd keep this simple.

src/workflow.ts
mbg: Did you mean to suggest this in combination with another change?

On its own, this would get us the refs of all steps in the workflow, regardless of whether they're for `github/codeql-action` or not.

esbena: oh, I meant to do 
```suggestion
          step.uses?.startsWith("github/codeql-action/")
```

lorenzodavidacostaponce100-debug:
[utf8_extractor_20251001182843.txt](https://github.com/user-attachments/files/22652013/utf8_extractor_20251001182843.txt)
9s3

lorenzodavidacostaponce100-debug:
mbg/validate-action-version-in-workflow

PR #3100, opened 2025-09-10
Store and check action version in `Config`

mbg:
Stored the CodeQL Action version in `Config`. When loading the `Config`, we check that the `version` value matches the CodeQL Action version. If not, we throw a `ConfigurationError`.

This ensures that the version of the CodeQL Action used in a given workflow is consistent.

### Risk assessment


src/config-utils.ts
```diff
@@ -1308,7 +1313,21 @@ export async function getConfig(
...
+
+  if (config.version === undefined) {
+    throw new ConfigurationError(
+      `Loaded configuration file, but it does not contain the expected 'version' field.`,
+    );
+  }
+  if (config.version !== getActionVersion()) {
+    throw new ConfigurationError(
+      `Loaded a configuration file for version '${config.version}', but running version '${getActionVersion()}'`,
+    );
```
esbena, line 1327: This is a different message than the one produced by `/init` in the other PR. But this is acceptable, we don't really expect to hit this case once `/init` has been used in the wild for a while.

esbena (approved):

src/config-utils.ts
mbg: I think it makes sense for the messages to be different. In the other PR, it's a warning only and it is based on the ref used for the Action. Here, it's an error and based on the CodeQL Action version. In theory, the same CodeQL Action version can be associated with multiple different commits, so this isn't a perfect safeguard against mixing different commits in a workflow, and we may want to revisit that if we think it's enough of a problem.

In any case, since the conditions aren't the same, we probably want these to be different messages.

PR #3107, opened 2025-09-11
Add feature flag to roll out JAR minimization in the Java extractor

nickrolfe:
This adds a feature flag which, when enabled, will have two effects:
1. sets an environment variable to enable a feature in the Java extractor to rewrite the contents of dependency JARs so that they take up less space on disk.
2. modify the key used for Actions dependency caching. This means that, if we find any problems with the JAR-rewriting and need to disable the flag, we won't have to worry about clearing corrupted cached.

### Risk assessment


src/feature-flags.ts
```diff
@@ -262,6 +263,11 @@ export const featureConfig: Record<
     legacyApi: true,
     minimumVersion: undefined,
   },
+  [Feature.JavaMinimizeDependencyJars]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_JAVA_MINIMIZE_DEPENDENCY_JARS",
+    minimumVersion: "2.23.0",
```
mbg, line 276: Could this be a `ToolsFeature` returned by the CLI, rather than using `minimumVersion` here?

src/dependency-caching.ts
```diff
@@ -222,24 +252,30 @@ export async function uploadDependencyCaches(config: Config, logger: Logger) {
...
  * can be changed to invalidate old caches, the runner's operating system, and the specified language name.
  *
  * @param language The language being analyzed.
+ * @param minimizeJavaJars Whether the Java extractor should rewrite downloaded JARs to minimize their size.
  * @returns The prefix that identifies what a cache is for.
  */
-async function cachePrefix(language: Language): Promise<string> {
+async function cachePrefix(
+  language: Language,
+  minimizeJavaJars: boolean,
```
mbg, line 256: Minor: It might be worth passing the `features` object into `cachePrefix` and then performing the check here. That would avoid a little bit of code duplication in `init-action` and `analyze-action`, while also making it easier to depend on other FFs in the dependency caching implementation more generally.

src/dependency-caching.ts
```diff
@@ -38,6 +40,25 @@ export function getJavaTempDependencyDir(): string {
...
+  // JAR minimization is only relevant for Java and build-mode: none.
+  if (
+    config.buildMode === BuildMode.None &&
+    config.languages.includes(KnownLanguage.java)
+  ) {
+    return await features.getValue(Feature.JavaMinimizeDependencyJars, codeql);
+  }
+  return false;
+}
+
```
mbg: Minor: No strong feelings, but I am wondering if this is really the best place for this function? It only seems to be used to determine whether to perform the minimisation itself. While dependency caching is the main benefactor of that, it's not strictly an aspect of the dependency caching implementation.

Alternatively, there are a few checks in the other functions in this file that could possibly use this function (at the expense of threading `config`, `features`, and `codeql` throughout).

src/init-action.ts
```diff
@@ -596,6 +607,16 @@ async function run() {
         core.exportVariable("CODEQL_EXTRACTOR_PYTHON_EXTRACT_STDLIB", "true");
       }
     }
+    if (process.env["CODEQL_EXTRACTOR_JAVA_OPTION_MINIMIZE_DEPENDENCY_JARS"]) {
+      logger.debug(
+        "CODEQL_EXTRACTOR_JAVA_OPTION_MINIMIZE_DEPENDENCY_JARS is already set, so the Action will not override it.",
+      );
```
mbg: Minor: Do you see any value in including the existing value of `CODEQL_EXTRACTOR_JAVA_OPTION_MINIMIZE_DEPENDENCY_JARS` here?

src/init-action.ts
```diff
@@ -596,6 +607,16 @@ async function run() {
         core.exportVariable("CODEQL_EXTRACTOR_PYTHON_EXTRACT_STDLIB", "true");
       }
     }
+    if (process.env["CODEQL_EXTRACTOR_JAVA_OPTION_MINIMIZE_DEPENDENCY_JARS"]) {
```
mbg: Consider adding `CODEQL_EXTRACTOR_JAVA_OPTION_MINIMIZE_DEPENDENCY_JARS` to `EnvVar` in `environment.ts` and then use that everywhere.

src/init-action.ts
```diff
@@ -596,6 +607,16 @@ async function run() {
...
+      logger.debug(
+        "CODEQL_EXTRACTOR_JAVA_OPTION_MINIMIZE_DEPENDENCY_JARS is already set, so the Action will not override it.",
+      );
+    } else if (await minimizeJavaDependencyJars(config, features, codeql)) {
```
mbg: Another alternative with `minimizeJavaDependencyJars` would be to just inline it here, since it is the only place where it's used. You could re-use the result of querying `Feature.JavaMinimizeDependencyJars` from earlier in this function and combine it with the language + build-mode checks.

src/dependency-caching.ts
```diff
@@ -248,5 +284,10 @@ async function cachePrefix(language: Language): Promise<string> {
     prefix = `${prefix}-${customPrefix}`;
   }
 
+  // To ensure a safe rollout of JAR minimization, we change the key when the feature is enabled.
+  if (language === KnownLanguage.java && minimizeJavaJars) {
+    prefix = `${prefix}-minify`;
```
mbg: One concern I have is the following chain of events:

- The FF is enabled and we create a cache with a key like `codeql-dependencies-minify${hash}`.
- Something goes wrong and the cache is horribly broken.
- We decide to disable the FF again.
- `cachePrefix(..)` will return just `codeql-dependencies-` for the next run.
- Because we use the result of `cachePrefix(..)` as a possible restore key in the `actionsCache.restoreCache` call if we don't get an exact match for the primary cache key, we would restore the broken `codeql-dependencies-minify ${hash}` cache.
- The user now has to manually delete the problematic cache.

That sort of defeats the point of having a FF in the first place. We should ensure that we don't restore a minified cache if the FF is disabled. Maybe the easiest way of accomplishing this would be to add `minify` or whatever at the start of the cache key prefix? 

Alternatively, we could include different symbols after the main `prefix` depending on whether minimising is enabled or not.

nickrolfe: Good catch, thanks. I was vaguely aware that the Actions cache worked by matching a prefix, and not necessarily a full string, but I hadn't thought through the consequences. I think I'll just change it to `minify-${prefix}`.

src/feature-flags.ts
nickrolfe: I considered that, and thought it better to use a minimum version for two reasons:
1. A `ToolsFeature` would be declared by the CLI, but the CLI plays no role here; it's an extractor feature.
2. There were several earlier releases where the extractor supported this feature (i.e. would respond to the env var), but those were buggy (or caused problems by deleting private fields). So we want a version that not only supports the feature but is also known to be safe to use.

src/init-action.ts
nickrolfe: I went with this approach.

src/dependency-caching.ts
nickrolfe: I've left this one alone. Threading `features` and `codeql` through both to this functions and `cacheKey` didn't feel like an improvement to me. If we need to check other FFs in the dependency caching implementation in the future, then it won't be too hard to make the change at that point.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- We are rolling out a change in September 2025 that will reduce the size of Actions dependency caches for Java analyses. [#3107](https://github.com/github/codeql-action/pull/3107)
```
mbg: How about, rather than having a date here that might change, we could word this to be about the support for it. Something like:
```suggestion
- We added support for reducing the size of dependency caches for Java analyses, which will reduce cache usage and speed up workflows. This will be enabled automatically at a later time. [#3107](https://github.com/github/codeql-action/pull/3107)
```

src/init-action.ts
```diff
@@ -597,6 +606,24 @@ async function run() {
       }
     }
 
+    // If the feature flag to minimize Java dependency jars is enabled, and we are doing a Java
+    // `build-mode: none` analysis (i.e. the flag is relevant), then set the environment variable
+    // that enables the corresponding option int he Java extractor.
```
mbg:
```suggestion
    // that enables the corresponding option in the Java extractor.
```

src/init-action.ts
```diff
@@ -597,6 +606,24 @@ async function run() {
...
+    if (process.env[EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS]) {
+      logger.debug(
+        `${EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS} is already set to ${process.env[EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS]}, so the Action will not override it.`,
```
mbg: Just to make it clearer what the value is, in case it's set to something silly that blends into the rest of the sentence.
```suggestion
        `${EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS} is already set to '${process.env[EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS]}', so the Action will not override it.`,
```

src/feature-flags.ts
mbg: No strong feelings here, but isn't setting `minimumVersion` sort of equivalent to adding a `ToolsFeature` in that there's some CLI version that's going to add the `ToolsFeature` which is the minimum? I appreciate it's an extractor feature, but either way we don't check the extractors.

> There were several earlier releases where the extractor supported this feature (i.e. would respond to the env var), but those were buggy (or caused problems by deleting private fields). So we want a version that not only supports the feature but is also known to be safe to use.

This makes sense. Am I assuming that your implicit thinking here is that `minimumVersion` is easier to change, if we discover other bugs, than a `ToolsFeature`?

mbg:
I think you'll also need to rebuild the JS files (possibly requiring changes from `main`) and resolve the merge conflict in the changelog.

src/feature-flags.ts
nickrolfe: Right, if the flag rollout reveals problems that require a fix in the extractor and a new CLI release, we can just bump the `minimumVersion`.

If we used `ToolsFeature`, we'd have to change that feature name from something like `JavaMinimizeJars` to `JavaMinimizeJars2` or `JavaMinimizeJarsButActuallyWorking`.

I don't have particularly strong feelings about this, but using `ToolsFeature` would require waiting for another CLI release.

mbg: > If we used ToolsFeature, we'd have to change that feature name from something like JavaMinimizeJars to JavaMinimizeJars2 or JavaMinimizeJarsButActuallyWorking.

Yeah, we had to do that recently with something else.

> Right, if the flag rollout reveals problems that require a fix in the extractor and a new CLI release, we can just bump the minimumVersion.

The tricky aspect here is always understanding all the possible permutations in which the Action, the CLI, and the FFs interact. The `minimumVersion` here prevents the scenario where the buggy extractor implementation would get used if someone uses an older CLI, the newest Action, and the FF is on. A `ToolsFeature` would accomplish the same, except require another CLI release.

If things go wrong, then bumping `minimumVersion` probably isn't going to help with the case where: someone has a version of the Action (but not the latest) that includes the changes from this PR with the original `minimumVersion`, some CLI version below the new `minimumVersion`, and we re-enable the FF.

To prevent that, we'd either need a different `ToolFeature` or a different FF.

nickrolfe: Gotcha. Are you happy that we proceed using `minimumVersion`, which means we don't have to add a `ToolsFeature` and wait for another release? If there's a problem that requires a fix and a new release, then I would add a `ToolsFeature` at the same time and change the name of the feature flag so that folks using older versions of the Action (relying on `minimumVersion`) aren't at risk.

mbg: I am happy with that 👍🏻

PR #3109, opened 2025-09-11
init-action: save updated config

cklin:
This PR updates the init action to save the config again at the end of `run()`, so that config updates in `run()` are correctly propagated to the analyze action.

### Risk assessment

### DIFF
--- a/src/config-utils.test.ts
+++ b/src/config-utils.test.ts
@@ -229,7 +229,7 @@ test("load code quality config", async (t) => {
   });
 });
 
-test("loading config saves config", async (t) => {
+test("loading a saved config produces the same config", async (t) => {
   return await withTmpDir(async (tempDir) => {
     const logger = getRunnerLogger(true);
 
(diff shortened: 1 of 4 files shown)


src/config-utils.test.ts
```diff
@@ -229,51 +229,6 @@ test("load code quality config", async (t) => {
...
-        languagesInput: "javascript,python",
-        tempDir,
-        codeql,
-        workspacePath: tempDir,
-        logger,
-      }),
-    );
-
-    // The saved config file should now exist
-    t.true(fs.existsSync(configUtils.getPathToParsedConfigFile(tempDir)));
```
mbg, line 264: It might be possible to keep this test if you now manually call `saveConfig` before this check?

cklin: The stated purpose of the test is to verify that `initConfig()` saves the config file.  After the suggested change, `initConfig()` no longer saves the config file.

Can you clarify how you would like to see this test saved?

mbg: I think this test is more of a round-trip test: whether saving a config and then loading it results in the same object. It just so happened that `initConfig` was saving the config before, but not what was primarily being tested.

cklin: Done.

PR #3112, opened 2025-09-15
CI: Configure Python analysis

henrymercer:
While we're here, simplify the config a bit and add build modes.

### Risk assessment

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -107,7 +107,7 @@ jobs:
       uses: ./analyze
       with:
         category: "/language:javascript"
-        upload: ${{ matrix.os == 'ubuntu-24.04' && matrix.tools == '' && 'always' || 'never' }}
+        upload: ${{ (matrix.os == 'ubuntu-24.04' && !matrix.tools && 'always') || 'never' }}
 
   analyze-other:
     runs-on: ubuntu-latest


.github/workflows/codeql.yml
```diff
@@ -125,9 +131,12 @@ jobs:
     - name: Initialize CodeQL
       uses: ./init
       with:
-        languages: actions
-        config-file: ./.github/codeql/codeql-actions-config.yml
```
mbg, line 129: IMHO we should keep this and make it part of the `matrix`. Then we can have a Python config that filters out the test file. We could do the same with `config` in the `matrix` as well.

mbg (dismissed):
LGTM with one suggestion for filtering out the Python test file

.github/workflows/codeql.yml
henrymercer: I think since we don't yet have anything that needs to be language specific, it might be more readable to have a single `config` embedded into the workflow file.  I've pushed a commit to add that.

.github/workflows/codeql.yml
```diff
@@ -107,13 +107,17 @@ jobs:
       uses: ./analyze
       with:
         category: "/language:javascript"
+        upload: ${{ matrix.os == 'ubuntu-24.04' && matrix.tools == '' && 'always' || 'never' }}
```
mbg: Do you think it might be clearer to:

- Add parentheses around the part up to `||`?
- Rather than `matrix.tools == ''`, check the truthiness of `.tools` or that it is not `"linked"?

```suggestion
        upload: ${{ (matrix.os == 'ubuntu-24.04' && matrix.tools != 'linked' && 'always') || 'never' }}
```

henrymercer: In theory, we could start running nightlies as well, so I prefer the clarity of matching a single tools value.  We can definitely add brackets around the conjuncts though.

henrymercer:
```suggestion
        upload: ${{ (matrix.os == 'ubuntu-24.04' && !matrix.tools && 'always') || 'never' }}
```

PR #3117, opened 2025-09-17
Support non-lock files for C# cache key computation

mbg:
This PR modifies our implementation of dependency caching for C# to allow non-lock files to be used when computing a hash for the cache key.

The hash that's included in the cache key is derived from files which contain information about the project's dependency. The idea being that the hash changes only if the dependencies change. Currently, we only use `packages.lock.json` and `paket.lock` files for this. These files contain exact dependency information (including the exact versions that were resolved by the respective package manager). The resulting hashes we compute for the cache keys should therefore only change if the dependencies (including versions) change. This is the desired behaviour.

However, relatively few C# projects use lock files and fewer yet have them checked-in to the repository. This means that most C# projects do not benefit from our implementation of dependency caching at the moment.

This PR modifies the behaviour of our dependency caching implementation for C# as follows: 

- As before, we search for `packages.lock.json` or `paket.lock` files. If they exist, we use them to compute a hash for the cache key.
- If no such files exist, and the new `CsharpNewCacheKey` feature is enabled (via FF or environment variable), we search for additional files that typically contain dependency information for C# projects. If they exist, we use them to compute the hash for the cache key.

Using the additional files is less precise, because they may not contain exact version information (only version ranges or no versions at all) and they contain information unrelated to dependencies. This means that the hash we compute can change, even if the dependencies don't change. However, dependency caching will benefit more C# projects.

This PR is best reviewed commit-by-commit. The first few commits just add the new `Feature` and refactor the code a bit to facilitate the changes. 

Questions for reviewers:

- Right now, disabling the `Feature` wouldn't avoid restoring caches that were produced with the `Feature` enabled. We likely want to do something similar as for https://github.com/github/codeql-action/pull/3107 that marks caches produced with this `Feature` on.
- Should we use this opportunity to also extend the list of paths that are included in the cache for C#? The `setup-dotnet` Action considers more paths and computes them dynamically. Now that we have `getDependencyPaths` as a function, it would be easy to do this.
- If so, should that be gated behind the same FF?
- Are there any additional files we should include for C# hashes? 

### Risk assessment

### DIFF
--- a/src/dependency-caching.ts
+++ b/src/dependency-caching.ts
@@ -14,6 +14,14 @@ import { KnownLanguage, Language } from "./languages";
 import { Logger } from "./logging";
 import { getRequiredEnvParam } from "./util";
 
+class NoMatchingFilesError extends Error {
+  constructor(msg?: string) {
+    super(msg);
+
+    this.name = "NoMatchingFilesError";
+  }
+}
+
 /**
  * Caching configuration for a particular language.
  */
@@ -21,9 +29,12 @@ interface CacheConfig {
   /** Gets the paths of directories on the runner that should be included in the cache. */
   getDependencyPaths: () => string[];
   /**
-   * Gets patterns for the paths of files whose contents affect which dependencies are used
-   * by a project. We find all files which match these patterns, calculate a hash for
-   * their contents, and use that hash as part of the cache key.
+   * Gets an array of glob patterns for the paths of files whose contents affect which dependencies are used
+   * by a project. This function also checks whether there are any matching files and throws
+   * a `NoMatchingFilesError` error if no files match.
+   *
+   * The glob patterns are intended to be used for cache keys, where we find all files which match these
+   * patterns, calculate a hash for their contents, and use that hash as part of the cache key.
    */
   getHashPatterns: (codeql: CodeQL, features: Features) => Promise<string[]>;
 }
@@ -60,40 +71,90 @@ export function getJavaDependencyDirs(): string[] {
 }
 
 /**
+ * Checks that there are files which match `patterns`. If there are matching files for any of the patterns,
+ * this function returns all `patterns`. Otherwise, a `NoMatchingFilesError` is thrown.
+ *
+ * @param patterns The glob patterns to find matching files for.
+ * @returns The array of glob patterns if there are matching files.
+ */
+async function makePatternCheck(patterns: string[]): Promise<string[]> {
+  const globber = await makeGlobber(patterns);
+
+  if ((await globber.glob()).length === 0) {
+    throw new NoMatchingFilesError();
+  }
+
+  return patterns;
+}
+
+/**
  * Default caching configurations per language.
  */
 const defaultCacheConfigs: { [language: string]: CacheConfig } = {
   java: {
     getDependencyPaths: getJavaDependencyDirs,
-    getHashPatterns: async () => [
-      // Maven
-      "**/pom.xml",
-      // Gradle
-      "**/*.gradle*",
-      "**/gradle-wrapper.properties",
-      "buildSrc/**/Versions.kt",
-      "buildSrc/**/Dependencies.kt",
-      "gradle/*.versions.toml",
-      "**/versions.properties",
-    ],
+    getHashPatterns: async () =>
+      makePatternCheck([
+        // Maven
+        "**/pom.xml",
+        // Gradle
+        "**/*.gradle*",
+        "**/gradle-wrapper.properties",
+        "buildSrc/**/Versions.kt",
+        "buildSrc/**/Dependencies.kt",
+        "gradle/*.versions.toml",
+        "**/versions.properties",
+      ]),
   },
   csharp: {
     getDependencyPaths: () => [join(os.homedir(), ".nuget", "packages")],
-    getHashPatterns: async () => [
-      // NuGet
-      "**/packages.lock.json",
-      // Paket
-      "**/paket.lock",
-    ],
+    getHashPatterns: async () =>
+      makePatternCheck([
+        // NuGet
+        "**/packages.lock.json",
+        // Paket
+        "**/paket.lock",
+      ]),
   },
   go: {
     getDependencyPaths: () => [join(os.homedir(), "go", "pkg", "mod")],
-    getHashPatterns: async () => ["**/go.sum"],
+    getHashPatterns: async () => makePatternCheck(["**/go.sum"]),
   },
 };
 
 async function makeGlobber(patterns: string[]): Promise<glob.Globber> {
   return glob.create(patterns.join("\n"));
+}
+
+/**
+ * A wrapper around `cacheConfig.getHashPatterns` which catches `NoMatchingFilesError` errors,
+ * and logs that there are no files to calculate a hash for the cache key from.
+ *
+ * @param codeql The CodeQL instance to use.
+ * @param features Information about which FFs are enabled.
+ * @param language The language the `CacheConfig` is for. For use in the log message.
+ * @param cacheConfig The caching configuration to call `getHashPatterns` on.
+ * @param logger The logger to write the log message to if there is an error.
+ * @returns An array of glob patterns to use for hashing files, or `undefined` if there are no matching files.
+ */
+async function checkHashPatterns(
+  codeql: CodeQL,
+  features: Features,
+  language: Language,
+  cacheConfig: CacheConfig,
+  logger: Logger,
+): Promise<string[] | undefined> {
+  try {
+    return cacheConfig.getHashPatterns(codeql, features);
+  } catch (err) {
+    if (err instanceof NoMatchingFilesError) {
+      logger.info(
+        `Skipping download of dependency cache for ${language} as we cannot calculate a hash for the cache key.`,
+      );
+      return undefined;
+    }
+    throw err;
+  }
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/dependency-caching.ts
```diff
@@ -87,6 +87,54 @@ async function makePatternCheck(patterns: string[]): Promise<string[]> {
...
+      "**/*.csproj",
+      "**/packages.config",
+      "**/nuget.config",
+    ]);
+  }
+
+  // If we get to this point, the `basePatterns` didn't find any files,
+  // and `Feature.CsharpNewCacheKey` is either not enabled or we didn't
+  // find any files using those patterns either.
+  throw new NoMatchingFilesError();
```
henrymercer: I don't have all the context here, but perhaps it would be clearer to return `undefined` and have this path included in the types?

henrymercer (commented):
Looks reasonable from an Action perspective

src/dependency-caching.ts
```diff
@@ -87,6 +87,54 @@ async function makePatternCheck(patterns: string[]): Promise<string[]> {
...
+  if (await features.getValue(Feature.CsharpNewCacheKey, codeql)) {
+    // These are less accurate for use in cache key calculations, because they:
+    //
+    // - Don't contain the exact versions used. They may only contain version ranges or none at all.
+    // - They contain information unrelated to dependencies, which we don't care about.
+    //
+    // As a result, the hash we compute from these files may change, even if
+    // the dependencies haven't changed.
+    return makePatternCheck([
+      "**/*.csproj",
```
michaelnebel: @hvitved : Can you think of other files that are relevant to include i cache-key computations?

It is probably worth noting that we might risk using "bad" caches when version (or wildcard) ranges are used (in case the available version(s) of a dependency changes in a NuGet feed). Is it correctly understood that the only impact of this is that the "right" package version is just not available in the cache (and will be downloaded)?

Related question: For how long are caches stored?

michaelnebel (commented):
Looks plausible to me!

src/dependency-caching.ts
hvitved: Perhaps `**/global.json` and `**/*.sln`?

Actually, as far as I can tell, our dependency fetching logic will look in [_all_ small non-binary files](https://github.com/github/codeql/blob/2cd1d2fd2f95b51a1a4c3580a7d5fe58ab235003/csharp/extractor/Semmle.Extraction.CSharp.DependencyFetching/DependencyManager.cs#L89) for [lines like](https://github.com/github/codeql/blob/2cd1d2fd2f95b51a1a4c3580a7d5fe58ab235003/csharp/extractor/Semmle.Extraction.CSharp.DependencyFetching/FileContent.cs#L198C29-L198C48) `<PackageReference Include="..." />`, and then try to restore those dependencies. Obviously, we cannot meaningfully include all files in the hash, so perhaps we will just have to ignore this case.

mbg: Yes, there is a balance here between capturing all changes to dependencies and the ideal case, where the hash only changes if there are dependencies added or removed, or the exact versions that are in use change. If we hashed all "small non-binary" files, for example, we'd probably be uploading a new cache for every run. The main problem with that is how much of the cache quota we'd end up using.

The main goal here is to produce dependency caches for more C# projects in (primarily) Default Setup, where it's not possible to add custom caching steps. 

We could port the logic to check all "small non-binary" files for lines that look like they include dependency information into the Action to determine which files to hash for this, but I wouldn't be too keen on that duplication and it's probably beyond the scope of this change. For more advanced logic like that, it would probably make sense to revisit the parts of the original EDR for this that talk about having the CLI communicate the caching configuration to the Action.

hvitved: I agree with all of the above, I just wanted to point out this corner case in the dependency fetching logic.

mbg: Yes, that was helpful! Thank you ❤️ 

I only wanted to add the extra context for everyone's benefit.

For `**/global.json` and `**/*.sln`, do you know off the top of your head how common it is for them to contain dependency information?

For `global.json` files, I have only seen them used to indicate the .NET version - which we wouldn't want to cache (that should ideally be handled by the toolcache) and which I don't think affects the dependencies that are downloaded? Although, I imagine it would be possible to conditionally include dependencies based on the .NET version in use.

For `*.sln` files, I have only seen those include other project files.

hvitved: I think you are right; let's skip `global.json` files and `*.sln` files.

michaelnebel (commented):
The new pattern for deciding which C# files to use in the cache hash LGTM!

> Should we use this opportunity to also extend the list of paths that are included in the cache for C#? The setup-dotnet Action considers more paths and computes them dynamically. Now that we have getDependencyPaths as a function, it would be easy to do this.

Does this mean that we are "only" including "{working_dir}/.nuget/packages" in the cache?

The dependency manager for C# use the "scratch dir" for storing downloaded packages.

src/caching-utils.ts
```diff
@@ -71,6 +73,33 @@ export function getCachingKind(input: string | undefined): CachingKind {
...
+  //
+  // "Properties are visited using the same algorithm as Object.keys(), which
+  // has a well-defined order and is stable across implementations. For example,
+  // JSON.stringify on the same object will always produce the same string, and
+  // JSON.parse(JSON.stringify(obj)) would produce an object with the same key
+  // ordering as the original (assuming the object is completely
+  // JSON-serializable)."
+  const componentsJson = JSON.stringify(components);
+  return crypto
+    .createHash("sha256")
```
henrymercer, line 97: Non-blocking: We could use a non-cryptographic hash function here, but the data probably isn't big enough to make a noticeable difference.  It might be nice to time it to double check though.

(Edit: I see this code is mainly just being moved)

src/dependency-caching.ts
```diff
@@ -39,21 +47,100 @@ export function getJavaTempDependencyDir(): string {
...
+      "**/packages.config",
+      "**/nuget.config",
+    ]);
+  }
+
+  // If we get to this point, the `basePatterns` didn't find any files,
+  // and `Feature.CsharpNewCacheKey` is either not enabled or we didn't
+  // find any files using those patterns either.
+  return undefined;
```
henrymercer: The last bit isn't true since we'll return the result of `makePatternCheck`, even if it's `undefined`.  It might be clearer to have a helper function that checks whether the patterns match anything, and use that here instead.

src/dependency-caching.ts
```diff
@@ -107,23 +188,55 @@ export interface DependencyCacheRestoreStatus {
...
+  features: FeatureEnablement,
+  language: Language,
+  cacheConfig: CacheConfig,
+  logger: Logger,
+): Promise<string[] | undefined> {
+  const patterns = await cacheConfig.getHashPatterns(codeql, features);
+
+  if (patterns === undefined) {
+    logger.info(
+      `Skipping download of dependency cache for ${language} as we cannot calculate a hash for the cache key.`,
```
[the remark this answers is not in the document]
henrymercer: This seems valid, though "Skipping dependency caching" seems more natural.

src/dependency-caching.ts
```diff
@@ -333,12 +509,12 @@ async function cachePrefix(
...
-  if (language === KnownLanguage.java && minimizeJavaJars) {
-    prefix = `minify-${prefix}`;
-  }
+  // Calculate the feature prefix for the cache, if any. This is a hash that identifies
+  // experimental features that affect the cache contents.
+  const featurePrefix = await getFeaturePrefix(codeql, features, language);
 
-  return `${prefix}-${CODEQL_DEPENDENCY_CACHE_VERSION}-${runnerOs}-${language}-`;
+  // Assemble the cache key.
+  return `${featurePrefix}${prefix}-${CODEQL_DEPENDENCY_CACHE_VERSION}-${runnerOs}-${language}-`;
```
henrymercer: If users list cache entries using the API, I think it'd be clearer to them what those cache entries are for if we keep `codeql-dependencies-` at the front.

henrymercer (commented):
This generally looks good.  The only blocking comment from me is the naming of the cache entries.

For PRs like this in the future, consider splitting up into a core functionality part (e.g. here, handling the addition of feature flags to the cache key), and a language-specific part (e.g. here, changing the cache key for C#).

src/caching-utils.ts
mbg: Yeah, this is just moved code. That said, I think using `sha256` makes sense, because that's also what https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#hashfiles uses. This means that it's technically possible to compute the same hash in a workflow.

mbg:
> Does this mean that we are "only" including "{working_dir}/.nuget/packages" in the cache?
> 
> The dependency manager for C# use the "scratch dir" for storing downloaded packages.

@michaelnebel It is `$HOME/.nuget/packages`, but yes -- that is the only directory we are currently including in C# dependency caches. If you store dependencies elsewhere as well, then it might make sense to include that path (similarly to what we have done for Java as well).

src/dependency-caching.ts
mbg: For now, I have updated the comment to remove the erroneous part. I think that was leftover from a previous iteration of this function. 

I did notice that we could use `makePatternCheck` for the base patterns check as well though, so I have done that and added test coverage for `getCsharpHashPatterns`.

src/dependency-caching.ts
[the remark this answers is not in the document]
mbg: It was easy enough to restore the original log messages, so I have done that instead of using a generic log message.

src/dependency-caching.ts
mbg: As discussed elsewhere, I also thought this would be a good idea. Thanks for confirming!

I have pushed a commit which changes this, except for `minify-` so that we don't run into problems with that, since it's already rolled out.

Once the JAR minification experiment has concluded, we can remove all of the special handling for that along with that FF.

michaelnebel:
> > Does this mean that we are "only" including "{working_dir}/.nuget/packages" in the cache?
> > The dependency manager for C# use the "scratch dir" for storing downloaded packages.
> 
> @michaelnebel It is `$HOME/.nuget/packages`, but yes -- that is the only directory we are currently including in C# dependency caches. If you store dependencies elsewhere as well, then it might make sense to include that path (similarly to what we have done for Java as well).

We store dependencies in sub directories of the directory in `CODEQL_EXTRACTOR_CSHARP_SCRATCH_DIR`. (which should be set by the CLI). Could this be incorporated as well?

mbg:
@michaelnebel Thanks for pointing this out! I have thought about this a little, and I think it may be better to make that a separate change and put it behind a separate FF. That way we can see what impact each change has separately and get this change shipped faster.

henrymercer (approved):

PR #3119, opened 2025-09-17
Use `browser_download_url` instead of `url`

mbg:
Fixes a bug with https://github.com/github/codeql-action/pull/3110, where we used `url` instead of `browser_download_url`.

### Risk assessment

For internal use only. Please select the risk level of this change:

- **Low risk:** Changes are fully under feature flags, or have been fully tested and validated in pre-production environments and are highly observable, or are documentation or test only.

### Merge / deployment checklist

- Confirm this change is backwards compatible with existing workflows.
- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.
- Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary.

PR #3110, opened 2025-09-11
Fetch proxy binaries from `defaults.json` release

mbg:
Currently, the `start-proxy` action obtains the platform-specific `update-job-proxy` binary it needs from a hard-coded CodeQL CLI bundle release on the `github/codeql-action` repo, if it is not already in the runner's toolcache.

We will soon be including up-to-date versions of the `update-job-proxy` binaries with every CodeQL CLI bundle release.

This PR modifies the `start-proxy` action to search the release assets of the release pointed at by `defaults.json` for an appropriate `update-job-proxy` asset and downloads it, if it is not already in the runner's toolcache.

If the release pointed at by `defaults.json` doesn't contain the right asset for whatever reason, we revert to using the hard-coded release instead.

Because `update-job-proxy` isn't versioned, and we require a version for the toolcache, we use the CodeQL CLI version of the release the `update-job-proxy` is obtained from as its version.

I have added a few unit tests for this logic.

### Risk assessment

For internal use only. Please select the risk level of this change:

- **Low risk:** Changes are fully under feature flags, or have been fully tested and validated in pre-production environments and are highly observable, or are documentation or test only.
- **High risk:** Changes are not fully under feature flags, have limited visibility and/or cannot be tested outside of production.

### Merge / deployment checklist

- Confirm this change is backwards compatible with existing workflows.
- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.
- Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary.

PR #3120, opened 2025-09-17
Specify `Accept` header for `toolcache.downloadTool`

mbg:
Fixes a bug with https://github.com/github/codeql-action/pull/3110, where we missed setting this header for `toolcache.downloadTool` that's needed to get the actual artifact, rather than the JSON metadata.

Better solution than #3119, since `browser_download_url` is intended for use by browsers.

### Risk assessment


mbg:
Successfully tested that this also works with the fallback in https://github.com/github/codeql-action/pull/3121: https://github.com/github/codeql-action/actions/runs/17805925098?pr=3121

PR #3123, opened 2025-09-18
Fix `upload-sarif` Action failing if there are no Code Scanning SARIF files

mbg:
In https://github.com/github/codeql-action/pull/3064, we overlooked that `upload_lib.uploadFiles` is called in the `upload-sarif` Action, which will fail if it cannot find any SARIF files matching the predicate it is given.

For a Code Quality only analysis, there are no Code Scanning SARIF files and the call to `upload_lib.uploadFiles` therefore fails before the Code Quality SARIF file(s) can be uploaded.

This PR refactors the `upload-sarif` Action so that there's common code for Code Scanning and Code Quality to upload either one file or several matching files.

There are some optional, other changes in this PR as well, which I can remove if needed:

- ~The `UploadStatusReport` values for each SARIF upload are now combined for telemetry if both analysis kinds are enabled. Otherwise, we use the respective `UploadStatusReport`.~
- I have added a new `sarif-ids` output to the `upload-sarif` Action, which contains a stringified JSON object with details of the SARIF ids that we uploaded to different endpoints. Initially I went for just a comma-separated list of IDs, but I thought it would be useful to know which ID is for which service.

### Risk assessment

### DIFF
--- a/src/upload-sarif-action.ts
+++ b/src/upload-sarif-action.ts
@@ -32,6 +32,45 @@ interface UploadSarifStatusReport
   extends StatusReportBase,
     upload_lib.UploadStatusReport {}
 
+/**
+ * Searches for SARIF files for the given `analysis` in the given `sarifPath`.
+ * If any are found, then they are uploaded to the appropriate endpoint for the given `analysis`.
+ *
+ * @param logger The logger to use.
+ * @param features Information about FFs.
+ * @param sarifPath The path to a directory containing SARIF files.
+ * @param checkoutPath The checkout path.
+ * @param analysis The configuration of the analysis we should upload SARIF files for.
+ * @param category The SARIF category to use for the upload.
+ * @returns The result of uploading the SARIF file(s) or `undefined` if there are none.
+ */
+async function findAndUpload(
+  logger: Logger,
+  features: Features,
+  sarifPath: string,
+  checkoutPath: string,
+  analysis: analyses.AnalysisConfig,
+  category?: string,
+): Promise<upload_lib.UploadResult | undefined> {
+  const sarifFiles = upload_lib.findSarifFilesInDir(
+    sarifPath,
+    analysis.sarifPredicate,
+  );
+
+  if (sarifFiles.length !== 0) {
+    return await upload_lib.uploadSpecifiedFiles(
+      sarifFiles,
+      checkoutPath,
+      category,
+      features,
+      logger,
+      analysis,
+    );
+  }
+
+  return undefined;
+}
+
 async function sendSuccessStatusReport(
   startedAt: Date,
   uploadStats: upload_lib.UploadStatusReport,
@@ -86,6 +125,7 @@ async function run() {
   }
 
   try {
+    // `sarifPath` can either be a path to a single file, or a path to a directory.
     const sarifPath = actionsUtil.getRequiredInput("sarif_file");
     const checkoutPath = actionsUtil.getRequiredInput("checkout_path");
     const category = actionsUtil.getOptionalInput("category");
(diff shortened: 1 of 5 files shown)


src/upload-sarif-action.ts
```diff
@@ -86,54 +135,74 @@ async function run() {
...
       // The code quality service does not currently have an endpoint to wait for SARIF processing,
       // so we can't wait for that here.
     }
-    await sendSuccessStatusReport(startedAt, uploadResult.statusReport, logger);
+    await sendSuccessStatusReport(
+      startedAt,
+      upload_lib.combineSarifUploadResults([
+        uploadResult?.statusReport,
+        qualityUploadResult?.statusReport,
+      ]),
```
mbg: One note here: Since this is the `upload-sarif` action, we don't have a `Config` that tells us which `analysis-kinds` are enabled. Therefore, the telemetry will only include the data, but not which endpoint(s) were used. Perhaps it would be best for us to hold off this change until we add a field to distinguish which endpoints we uploaded to?

mbg: Alternatively, we could modify the telemetry here to at least include the `analysis-kinds` field, populated based on which endpoints we uploaded SARIF files to.

src/upload-sarif-action.ts
```diff
@@ -172,15 +175,21 @@ async function run() {
...
+          logger,
+        );
+      }
       // The code quality service does not currently have an endpoint to wait for SARIF processing,
       // so we can't wait for that here.
     }
-    await sendSuccessStatusReport(startedAt, uploadResult.statusReport, logger);
+    await sendSuccessStatusReport(
+      startedAt,
+      uploadResult?.statusReport || {},
```
esbena, line 200: Is it convention to add the empty object to these statuses by default?

src/upload-lib.ts
```diff
@@ -1039,3 +1039,42 @@ function filterAlertsByDiffRange(logger: Logger, sarif: SarifFile): SarifFile {
...
+      result.raw_upload_size_bytes = addOptional(
+        result.raw_upload_size_bytes,
+        report.raw_upload_size_bytes,
+      );
+      result.zipped_upload_size_bytes = addOptional(
+        result.zipped_upload_size_bytes,
+        report.zipped_upload_size_bytes,
+      );
+    }
+  }
```
esbena: This is too ad hoc for my taste. Once we add a new numeric field to the type, it will get dropped when going through this function

esbena: I.e. something like

```
  for (const report of reports) {
    if (!report) continue;
    for (const key of Object.keys(report) as (keyof UploadStatusReport)[]) {
      const value = report[key];
      if (typeof value !== "number") continue; // Skip non-numeric / undefined
      result[key] = ((result[key] ?? 0) as number) + value;
    }
  }
```
Conversely, we now risk adding things that can't be added...

Ultimately, we could do some typescript magic where we'd force ourselves to list every numerically typed property as being added or ignored, but that is complex.

esbena (commented):
Can you motivate the `Disable cpp in upload-quality-sarif check` commit?

mbg:
> Can you motivate the Disable cpp in upload-quality-sarif check commit?

@esbena C++ has no queries in the `code-quality` suite and the CLI fails as a result. See https://github.com/github/codeql-action/actions/runs/17831231939/job/50696560032?pr=3123#step:5:892

src/upload-sarif-action.ts
mbg: Note that this was an intermediate commit and the `{}` just made the type checker happy until I replaced it with the `statusReport` from the Code Quality upload or the combination of both uploads. 

---

The three properties of `UploadStatusReport` are all of type `number | undefined`, but the `UploadStatusReport` object itself is required here by `sendSuccessStatusReport`.

Previously, with only Code Scanning, we only would have reached `sendSuccessStatusReport` if the SARIF upload to Code Scanning is successful. In that case, the `UploadStatusReport` object always would have been populated.

Now we also have Code Quality and Code Scanning + Code Quality. So the status report should include the results of the respective upload(s).

src/upload-lib.ts
mbg: We could do something like that. Perhaps it is best if we leave sorting out the telemetry for a separate change so that we can think about this independently from the bug fix -- see my thoughts in the PR description and in https://github.com/github/codeql-action/pull/3123#discussion_r2359884335

esbena (approved):
> Perhaps it is best if we leave sorting out the telemetry for a separate change so that we can think about this independently from the bug fix

OK and conditional approve if we do that change elsewhere.

mbg:
I've removed the telemetry changes, other than that in the CQ-only case we send an empty `UploadStatusReport`. I think that's better than setting the fields to `0` since it allows us to tell apart status reports for CS/CS+CQ that have the data for CS from those that are for CQ-only and have no data.

PR #3125, opened 2025-09-18
Overlay: use restoreCache() timeout

cklin:
This PR changes overlay-base database download to pass the `segmentTimeoutInMs` option to `restoreCache()`, so that `restoreCache()` itself can properly abort slow downloads.

The `waitForResultWithTimeLimit()` wrapper around `restoreCache()` remains as a second line of defense, but with a higher 10-minute time limit, to guard against cache restore hangs outside segment downloads.

See linked issue for rationale and alternatives considered.

### Risk assessment

### DIFF
--- a/src/overlay-database-utils.ts
+++ b/src/overlay-database-utils.ts
@@ -158,7 +158,12 @@ function computeChangedFiles(
 // Constants for database caching
 const CACHE_VERSION = 1;
 const CACHE_PREFIX = "codeql-overlay-base-database";
-const MAX_CACHE_OPERATION_MS = 120_000; // Two minutes
+
+// The purpose of this ten-minute limit is to guard against the possibility
+// that the cache service is unresponsive, which would otherwise cause the
+// entire action to hang.  Normally we expect cache operations to complete
+// within two minutes.
+const MAX_CACHE_OPERATION_MS = 600_000;
 
 /**
  * Checks that the overlay-base database is valid by checking for the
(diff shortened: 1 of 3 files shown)


src/overlay-database-utils.ts
```diff
@@ -350,11 +350,16 @@ export async function downloadOverlayBaseDatabaseFromCache(
...
-      () => {
-        logger.info("Timed out downloading overlay-base database from cache");
+    const foundKey = await actionsCache.restoreCache(
+      [dbLocation],
+      cacheRestoreKeyPrefix,
+      undefined,
+      {
+        // Azure SDK download (which is the default) uses 128MB segments.
+        // Setting segmentTimeoutInMs to 3000 translates to segment download
+        // speed of about 40 MB/s, which should be achievable unless the
```
esbena: isn't 40MB/s in the high end for more residential connections?
What if this is somehow run from outside Actions? For instance for local test?

esbena (dismissed):
LGTM with one minor concern about download speeds outside the cloud.

src/overlay-database-utils.ts
mbg: I don't think you would generally run this locally. For tests, we'd mock the behaviour. Beyond that, I haven't checked, but I'd imagine that the API endpoints used by the Actions cache also are only accessible from hosted runners or at least require some suitable auth token. Setting that up locally is unlikely to ever be worth it. + even if we wanted to, it would be easy to adjust this value for local testing if it's too high.

src/overlay-database-utils.ts
```diff
@@ -350,11 +350,16 @@ export async function downloadOverlayBaseDatabaseFromCache(
   let databaseDownloadDurationMs = 0;
   try {
     const databaseDownloadStart = performance.now();
-    const foundKey = await waitForResultWithTimeLimit(
-      MAX_CACHE_OPERATION_MS,
-      actionsCache.restoreCache([dbLocation], cacheRestoreKeyPrefix),
-      () => {
-        logger.info("Timed out downloading overlay-base database from cache");
+    const foundKey = await actionsCache.restoreCache(
```
mbg: Is there any harm in keeping this wrapped in `waitForResultWithTimeLimit` with a suitable time limit?

mbg (dismissed):
This seems reasonable to me, if you're happy with it. 

In terms of the motivating issue, I suppose this is a change in objective? Rather than an overall timeout (which didn't work), there's now a timeout for individual segments (which does work, but there's no overall timeout)? So you might spend more time than `MAX_CACHE_OPERATION_MS` downloading the overlay cache if it's large, but no more than 3000ms per segment.

src/overlay-database-utils.ts
cklin: 40MB/s is indeed very high for residential connections.  I don't think I could reach this download speed from my connection at home.

That said, I am still inclined to keep this setting, at least for now.
- I agree with Michael that we are not very likely to run this code locally.
- Using overlay analysis is a performance optimization, and there is a point where download would take up so much time that the entire workflow ends up being slower end-to-end.  That break-even point will probably differ from one repository to another, but 40MB/s seems like a good figure to start with.

As we gain more experience, I am definitely open to adjusting this limit as appropriate.

src/overlay-database-utils.ts
cklin: No, there is no harm, and there would be benefits (such as protecting against hangs outside segment downloads).

I put the `waitForResultWithTimeLimit()` wrapper back, with a (longer) 10-minute time limit.

Thanks for the suggestion!

PR #3126, opened 2025-09-19
Add support for the repository properties API

mbg:
This PR has two main contributions:

1. We add general support for the new Repository Properties API.
2. We use the Repository Properties API to allow configuring queries in a similar way that the `queries` input can be used.

For the Repository Properties support, we fetch the list of repository properties at the start of the `init` Action. We filter this down to known properties (i.e. ones we care about) and store their values. The mapping of known property names to their values is then included in the Action state (i.e. the `Config` type) so that it remains accessible to other CodeQL Action steps in the same workflow and with consistent values.

For the query augmentation, we give the value obtained from the Repository Properties the highest precedence. This is because the intention here is to allow organisations to configure org-wide queries to include in analyses.

### Risk assessment

### DIFF
--- a/src/config/db-config.ts
+++ b/src/config/db-config.ts
@@ -2,7 +2,6 @@ import * as path from "path";
 
 import * as semver from "semver";
 
-import { isDefaultSetup } from "../actions-util";
 import * as errorMessages from "../error-messages";
 import {
   RepositoryProperties,
@@ -392,12 +391,10 @@ function combineQueries(
     // settings. If they don't allow combining with other query configurations, return just the
     // ones configured in the repository properties.
     if (!augmentationProperties.repoPropertyQueries.combines) {
-      if (!isDefaultSetup()) {
-        logger.info(
-          `Queries are configured in the repository properties and don't allow combining with other query settings. ` +
-            `Any queries configured elsewhere will be ignored.`,
-        );
-      }
+      logger.info(
+        `The queries configured in the repository properties don't allow combining with other query settings. ` +
+          `Any queries configured elsewhere will be ignored.`,
+      );
       return augmentationProperties.repoPropertyQueries.input;
     } else {
       // Otherwise, add them to the query array and continue.
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -86455,11 +86455,9 @@ function combineQueries(logger, config, augmentationProperties) {
       `Found query configuration in the repository properties (${"github-codeql-extra-queries" /* EXTRA_QUERIES */}): ${augmentationProperties.repoPropertyQueries.input.map((q) => q.uses).join(", ")}`
     );
     if (!augmentationProperties.repoPropertyQueries.combines) {
-      if (!isDefaultSetup()) {
-        logger.info(
-          `Queries are configured in the repository properties and don't allow combining with other query settings. Any queries configured elsewhere will be ignored.`
-        );
-      }
+      logger.info(
+        `The queries configured in the repository properties don't allow combining with other query settings. Any queries configured elsewhere will be ignored.`
+      );
       return augmentationProperties.repoPropertyQueries.input;
     } else {
       result.push(...augmentationProperties.repoPropertyQueries.input);


src/feature-flags/properties.ts
```diff
@@ -0,0 +1,86 @@
...
+    );
+    const remoteProperties = response.data as GitHubPropertiesResponse;
+
+    if (!Array.isArray(remoteProperties)) {
+      throw new Error(
+        `Expected repository properties API to return an array, but got: ${JSON.stringify(response.data)}`,
+      );
+    }
+
+    const knownProperties = new Set(Object.keys(RepositoryPropertyName));
```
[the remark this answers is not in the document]
mbg: Yes, that is the intended behaviour.

src/config/db-config.ts
```diff
@@ -364,25 +364,53 @@ function parseQueriesFromInput(
...
  */
 function combineQueries(
-  augmentedConfig: UserConfig,
+  config: UserConfig,
   augmentationProperties: AugmentationProperties,
-): QuerySpec[] | undefined {
+): QuerySpec[] {
+  const result: QuerySpec[] = [];
+
+  // Query settings obtained from the repository properties have the highest precedence.
```
henrymercer, line 380: This makes sense for default setup, but it could be surprising for advanced setup.  Can we add a log to explain what's going on here?

henrymercer (commented):
Looks like some additional config normalisation is needed in the tests, otherwise LGTM.

src/actions-util.ts
```diff
@@ -249,7 +249,12 @@ export function isSelfHostedRunner() {
 
 /** Determines whether we are running in default setup. */
 export function isDefaultSetup(): boolean {
-  return getWorkflowEventName() === "dynamic";
+  try {
+    return getWorkflowEventName() === "dynamic";
+  } catch {
+    // `getWorkflowEventName()` throws if `GITHUB_EVENT_NAME` is not set.
+    return false;
+  }
```
henrymercer: Something is very odd if `GITHUB_EVENT_NAME` is not set — I'd rather fail than make a guess that we're not in default setup.

src/feature-flags/properties.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @returns Returns a partial mapping from `RepositoryPropertyName` to values.
+ */
+export async function loadPropertiesFromApi(
+  gitHubVersion: GitHubVersion,
+  logger: Logger,
+  repositoryNwo: RepositoryNwo,
+): Promise<RepositoryProperties> {
+  // TODO: To be safe for now; later we should replace this with a version check once we know
+  // which version of GHES we expect this to be supported by.
+  if (gitHubVersion.type !== GitHubVariant.DOTCOM) {
```
henrymercer: This also disallows Proxima — is there a reason to do that?

Since the parent function feature flags, we could just use that feature flag.

src/config/db-config.ts
```diff
@@ -312,23 +363,83 @@ function parseQueriesFromInput(
...
+  ) {
+    // If there are queries configured as a repository property, these may be organisational
+    // settings. If they don't allow combining with other query configurations, return just the
+    // ones configured in the repository properties.
+    if (!augmentationProperties.repoPropertyQueries.combines) {
+      if (!isDefaultSetup()) {
+        logger.info(
+          `Queries are configured in the repository properties and don't allow combining with other query settings. ` +
+            `Any queries configured elsewhere will be ignored.`,
+        );
```
henrymercer: On reflection, maybe it would be helpful to always log when there is a repository property, since users might not be used to this way of configuring the Action.

src/feature-flags/properties.ts
mbg: No reason other than being defensive for when we remove the FF down the line. Happy to change this if we are happy that it won't cause problems.

src/config/db-config.ts
```diff
@@ -383,6 +383,11 @@ function combineQueries(
     augmentationProperties.repoPropertyQueries &&
     augmentationProperties.repoPropertyQueries.input
   ) {
+    logger.info(
+      `Found query configuration in the repository properties (${RepositoryPropertyName.EXTRA_QUERIES}): ` +
+        `${augmentationProperties.repoPropertyQueries.input.map((q) => q.uses).join(", ")}`,
+    );
```
henrymercer, line 388: This doesn't log whether the query configuration is combining or not, but that's probably OK.

mbg: Ah, I took your comment too literally and didn't think it was just about the line you commented on. I have removed the `isDefaultSetup()` check and adjusted the wording slightly so that there'll always be the second log message when combining queries is not permitted.

PR #3128, opened 2025-09-20
Add `concurrency` settings to PR checks

mbg:
This PR updates `sync.py` to add `concurrency` settings to all PR checks so that:

- There is one concurrency group per workflow per `ref`.
- For `pull_request` events, in-progress workflows in the same concurrency group are cancelled if a new workflow is starting.
- For other events, new workflows will wait until previous ones in the same concurrency group have finished.

This should reduce the number of concurrent workflows that we have running for the repo.
### Risk assessment

### DIFF
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -269,6 +269,17 @@ for file in sorted((this_dir / 'checks').glob('*.yml')):
                     'shell': 'bash',
                 },
             },
+            'concurrency': {
+                # Cancel in-progress workflows in the same 'group' for pull_request events,
+                # but other event types. This should have the effect that workflows on PRs
+                # get cancelled if there is a newer workflow in the same concurrency group.
+                # For other events, the new workflows should wait until earlier ones have finished.
+                # This should help reduce the number of concurrent workflows on the repo, and
+                # consequently the number of concurrent API requests.
+                'cancel-in-progress': "${{ github.event_name == 'pull_request' }}",
+                # The group is determined by the workflow name + the ref
+                'group': "${{ github.workflow }}-${{ github.ref }}"
+            },
             'jobs': {
                 checkName: checkJob
             }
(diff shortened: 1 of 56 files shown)


pr-checks/sync.py
```diff
@@ -269,6 +269,17 @@ def writeHeader(checkStream):
...
+            'concurrency': {
+                # Cancel in-progress workflows in the same 'group' for pull_request events,
+                # but not other event types. This should have the effect that workflows on PRs
+                # get cancelled if there is a newer workflow in the same concurrency group.
+                # For other events, the new workflows should wait until earlier ones have finished.
+                # This should help reduce the number of concurrent workflows on the repo, and
+                # consequently the number of concurrent API requests.
+                'cancel-in-progress': "${{ github.event_name == 'pull_request' }}",
+                # The group is determined by the workflow name + the ref
+                'group': "${{ github.workflow }}-${{ github.ref }}"
```
henrymercer, line 281: My understanding is this prevents duplicate checks on the same ref due to marking ready for review, for instance.  However couldn't we go further and base the concurrency group on the PR number, so that new pushes cancel runs for previous pushes?

mbg: According to the docs:

> For workflows triggered by `pull_request`, this is the pull request merge branch. For workflows triggered by `release`, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is `refs/heads/<branch_name>`. For pull requests events except `pull_request_target`, it is `refs/pull/<pr_number>/merge`. `pull_request_target` events have the `ref` from the base branch. For tags it is `refs/tags/<tag_name>`. For example, `refs/heads/feature-branch-1`.

So for `pull_request` events, we should get the fully-formed branch name (not the commit ref). You can see this in action on this PR: a bunch of in-progress workflows for https://github.com/github/codeql-action/pull/3128/commits/6fcf631e731b8bf359265aa9fa358edbd0635fa6 got cancelled because I pushed https://github.com/github/codeql-action/pull/3128/commits/2d8d6395ef9219dd6a9646ecd3681d1a65caf094.

henrymercer: Thanks for the clarification!

PR #3130, opened 2025-09-22
Support requesting latest nightly with `tools: nightly`

henrymercer:
To make it easier for us to dogfood CodeQL, add special values of the tools input `tools: nightly` and `tools: nightly-latest` that we can use to run the latest nightly.

### Risk assessment


src/setup-codeql.ts
```diff
@@ -277,6 +280,7 @@ export async function getCodeQLSource(
   if (
     toolsInput &&
     !CODEQL_BUNDLE_VERSION_ALIAS.includes(toolsInput) &&
+    !CODEQL_NIGHTLY_TOOLS_INPUTS.includes(toolsInput) &&
     !toolsInput.startsWith("http")
   ) {
     logger.info(`Using CodeQL CLI from local path ${toolsInput}`);
```
mbg, line 285: This check seems a bit error-prone, since it doesn't actually check what the body of the `if` statement cares about. Would it make sense to check whether `toolsInput` refers to something that exists in the filesystem here instead? Or at least refactor the conditions into a `isSpecialToolsValue` function?

src/setup-codeql.ts
```diff
@@ -33,8 +33,11 @@ export enum ToolsSource {
 }
 
 export const CODEQL_DEFAULT_ACTION_REPOSITORY = "github/codeql-action";
+const CODEQL_NIGHTLIES_REPOSITORY_OWNER = "dsp-testing";
+const CODEQL_NIGHTLIES_REPOSITORY_NAME = "codeql-cli-nightlies";
 
 const CODEQL_BUNDLE_VERSION_ALIAS: string[] = ["linked", "latest"];
+const CODEQL_NIGHTLY_TOOLS_INPUTS = ["nightly", "nightly-latest"];
```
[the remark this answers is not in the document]
henrymercer: This is for user convenience.  I don't think it's a maintenance issue.

src/setup-codeql.ts
```diff
@@ -771,3 +785,35 @@ async function useZstdBundle(
...
+    owner: CODEQL_NIGHTLIES_REPOSITORY_OWNER,
+    repo: CODEQL_NIGHTLIES_REPOSITORY_NAME,
+    per_page: 1,
+    page: 1,
+    prerelease: true,
+  });
+
+  const latestRelease = release.data[0];
+  if (!latestRelease) {
+    throw new Error("Could not find latest nightly release.");
```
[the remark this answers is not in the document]
mbg: This seems unnecessary.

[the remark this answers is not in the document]
henrymercer: I don't think this is helpful since the repo's under our control.

src/setup-codeql.ts
```diff
@@ -771,3 +785,35 @@ async function useZstdBundle(
...
+  // Since nightlies are prereleases, we can't just download the latest release
+  // on the repository. So instead we need to find the latest pre-release
+  // version and construct the download URL from that.
+  const release = await api.getApiClient().rest.repos.listReleases({
+    owner: CODEQL_NIGHTLIES_REPOSITORY_OWNER,
+    repo: CODEQL_NIGHTLIES_REPOSITORY_NAME,
+    per_page: 1,
+    page: 1,
+    prerelease: true,
+  });
```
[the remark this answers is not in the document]
henrymercer: This is reasonable, but we should throw a better error rather than logging.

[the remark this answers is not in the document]
mbg: I don't think this needs local error handling, but we may wish to be explicit in that:

- If we are happy for `getNightlyToolsUrl` to throw an exception, to indicate that in the JSDoc
- Verify that `release` contains `data` and that `data` is an array. If not, we should throw a suitable error

src/setup-codeql.ts
```diff
@@ -771,3 +782,35 @@ async function useZstdBundle(
...
+
+  const latestRelease = release.data[0];
+  if (!latestRelease) {
+    throw new Error("Could not find latest nightly release.");
+  }
+
+  return `https://github.com/${CODEQL_NIGHTLIES_REPOSITORY_OWNER}/${CODEQL_NIGHTLIES_REPOSITORY_NAME}/releases/download/${latestRelease.tag_name}/${getCodeQLBundleName(compressionMethod)}`;
```
mbg: Would it make sense to look through the release artifacts and search for the one we want, like we do for `update-job-proxy`?

.github/actions/prepare-test/action.yml
```diff
@@ -35,37 +35,31 @@ runs:
...
         elif [[ "$RUNNER_OS" == "Linux" ]]; then
-          artifact_name="codeql-bundle-linux64.$extension"
+          artifact_name="codeql-bundle-linux64.tar.gz"
         elif [[ "$RUNNER_OS" == "macOS" ]]; then
-          artifact_name="codeql-bundle-osx64.$extension"
+          artifact_name="codeql-bundle-osx64.tar.gz"
         elif [[ "$RUNNER_OS" == "Windows" ]]; then
-          artifact_name="codeql-bundle-win64.$extension"
+          artifact_name="codeql-bundle-win64.tar.gz"
         else
```
mbg, line 57: I am not sure I know off the top of my head what's prompted this change - could you explain?

mbg (commented):
Some initial comments - it looks like you've pushed some changes that relate to some of them already. I see we're also fighting random CI failures, so I'll hold off finishing my review until this PR has stabilised

.github/actions/prepare-test/action.yml
henrymercer: Nightly latest lookup is now handled by the Action, so the condition would never be true.  And if we're referring to a specific nightly, it might not have a zstd bundle.  We originally added this to improve our test coverage of zstd bundles, but now that these are shipped as latest, we have enough test coverage.  So it's not really worth refactoring this to determine whether the referenced nightly has a zstd bundle.

src/setup-codeql.ts
henrymercer: We could do but using the http URL is closer to how users would specify a bundle with the `tools` input.  Things might be less optimal if we used URLs returned by the API since we perform a few optimizations by inspecting the URL structure.

src/setup-codeql.ts
henrymercer: > Would it make sense to check whether `toolsInput` refers to something that exists in the filesystem here instead?

If `linked` or `nightly` was a valid path on the file system, we'd still want to prefer the "special" meaning.

I like the idea of the helper function.

mbg (approved):
Looks good to me. Should we document `nightly` as a valid value in the `description` for the `tools` input in `init/action.yml`?

henrymercer:
> Looks good to me. Should we document `nightly` as a valid value in the `description` for the `tools` input in `init/action.yml`?

Yes we should, thanks!

felickz:
Is there any future plans to differentiate `nightly` vs `nightly-latest` - they appear to behave the same ?

henrymercer:
> Is there any future plans to differentiate `nightly` vs `nightly-latest` - they appear to behave the same ?

No plans, they're just aliases.  In the future we might add support for referencing a particular nightly too via something like `nightly-yyyymmdd`, in which case I think it's useful to have that `nightly-latest` alias.

PR #3131, opened 2025-09-22
CI: Improve safety of update required checks script

henrymercer:
Fix an error condition that could cause the set of required checks to be emptied.

### Risk assessment


mbg (dismissed):
Copilot's suggestion is probably more robust, but I have no strong feelings about that this is a good improvement as-is.

PR #3135, opened 2025-09-23
Configure the API client to retry more often when in test mode

mbg:
Configures `createApiClientWithDetails` with a `retry` config that will attempt more retries, when in test mode. This is to try and work around rate limits on this repository when running a lot of PR checks.

### Risk assessment

### DIFF
--- a/src/api-client.ts
+++ b/src/api-client.ts
@@ -39,8 +39,18 @@ export interface GitHubApiExternalRepoDetails {
   apiURL: string | undefined;
 }
 
-export function getRetryConfig(): { retries: number } {
-  return isInTestMode() ? { retries: 10 } : { retries: 3 };
+export function getRetryConfig(): {
+  retries: number;
+  retryAfterBaseValue?: number;
+} {
+  // If we are in test mode, increase the allowed number of retries to 10
+  // and the base backoff from 1s to 10s. `plugin-retry` will wait
+  // `(failedAttempts ^ 2) * retryAfterBaseValue`-long (in ms) after
+  // each failed attempt.
+  // If we are not in test mode, we use the default configuration.
+  return isInTestMode()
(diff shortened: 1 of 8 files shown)


src/api-client.ts
```diff
@@ -38,6 +39,10 @@ export interface GitHubApiExternalRepoDetails {
   apiURL: string | undefined;
 }
 
+export function getRetryConfig(): { retries: number } {
+  return isInTestMode() ? { retries: 10 } : { retries: 3 };
```
esbena: This should have some explicit, and significant backoff numbers, or at least a comment about the backoff strategy. As is, we easily risk making the rate limit situation worse IMO.

mbg: Since the change here only matters for our CI, I figured we'd ship-to-learn to see how this goes with the default `retryAfterBaseValue`, rather than trying to nail down a good configuration.

However, I have now pushed a commit to increase the `retryAfterBaseValue` for test mode, made it explicit for both configurations, and added a comment explaining how the plugin uses it.

esbena: OK. Depending on +/-1 retry attempts, the total wait can span just over one hour, so this is guaranteed or at least likely to see a reset. 
Alternatively we could recognize the failure as being rate-limit related and act accordingly.

mbg: > Alternatively we could recognize the failure as being rate-limit related and act accordingly.

That seems like a more involved change. In the long run, we probably want a more robust solution for our CI here (e.g. a token with higher rate limits). 

At the same time, we need to be careful not to affect "normal" CodeQL runs, because we don't want them to be stuck retrying API calls for long.

mbg:
It doesn't seem like any of us are really convinced that this would really be a useful change to help with the rate limiting issues. I'll close this PR for now since we have merged other improvements, and still have ideas for other things we can try.

PR #3137, opened 2025-09-23
 Only run PR checks on Ubuntu by default 

henrymercer:
Continue running ones on macOS and Windows too where it's useful, but the default should be Linux only as most functionality is platform independent.

This should reduce unnecessary compute usage and help avoid rate limiting.

### Risk assessment


pr-checks/checks/all-platform-bundle.yml
```diff
@@ -1,7 +1,6 @@
 name: "All-platform bundle"
 description: "Tests using an all-platform CodeQL Bundle"
 versions: ["nightly-latest"]
-operatingSystems: ["ubuntu"]
```
mbg, line 4: Should this one maybe run on all platforms, since it's the all-platform bundle (to check that it actually works on all platforms)?

pr-checks/checks/autobuild-action.yml
```diff
@@ -1,5 +1,6 @@
 name: "autobuild-action"
 description: "Tests that the C# autobuild action works"
+operatingSystems: ["ubuntu", "macos", "windows"]
```
mbg, line 3: Should the other two `autobuild-` workflows also run on `macos`? Or was there a specific reason those only run on linux and windows?

pr-checks/checks/build-mode-autobuild.yml
mbg, line 1: Could this test be combined with one of `autobuild-direct-tracing*`? Seems similar.

Alternatively, maybe we could combine the three `build-mode` tests in one workflow with a matrix (and conditional steps based on the build-mode) / sequential jobs.

pr-checks/checks/cleanup-db-cluster-dir.yml
```diff
@@ -1,6 +1,5 @@
 name: "Clean up database cluster directory"
 description: "The database cluster directory is cleaned up if it is not empty."
-operatingSystems: ["ubuntu"]
```
mbg, line 3: This seems like it could be useful across all platforms because of difference in how IO behaves. That said, it feels like this could easily just be tacked on to another test that already runs on all platforms.

pr-checks/checks/job-run-uuid-sarif.yml
mbg, line 1: Feels like this could be added on to some other test.

pr-checks/checks/language-aliases.yml
mbg, line 1: This can probably be removed in favour of unit test coverage (which may already exist).

pr-checks/checks/resolve-environment-action.yml
```diff
@@ -1,5 +1,6 @@
 name: "Resolve environment"
 description: "Tests that the resolve-environment action works for Go and JavaScript/TypeScript"
+operatingSystems: ["ubuntu", "macos", "windows"]
```
mbg: I think running this on just `ubuntu` is fine since we only support `resolve build-environment` for Go and it only looks at Go versions. There isn't really anything platform-specific right now and we could always restore that if and when we have more platform-specific functionality.

pr-checks/checks/rust.yml
mbg, line 1: Do we need this specific one for Rust?

pr-checks/checks/global-proxy.yml
mbg, line 1: Maybe we should rename this file, given our `start-proxy` action which this test is unrelated to?

mbg (dismissed):
Makes sense. I have added a few comments for where we might be able to make further improvements, but they can all be handled separately.

pr-checks/checks/autobuild-action.yml
henrymercer: I think just because it's very unlikely to run Java analysis on macOS.

pr-checks/checks/build-mode-autobuild.yml
henrymercer: > Could this test be combined with one of `autobuild-direct-tracing*`? Seems similar.

I've combined with `autobuild-direct-tracing` now that we will use direct tracing by default.  But with the specific working directory it might still be useful to check `build-mode: autobuild` respects the working directory.

> Alternatively, maybe we could combine the three `build-mode` tests in one workflow with a matrix (and conditional steps based on the build-mode) / sequential jobs.

We could, but we could no longer generate the workflow, so won't do for now.

pr-checks/checks/autobuild-action.yml
mbg: If the tests are mainly for direct tracing, would it make sense to switch to a different traced language where we're more likely to run on all platforms (e.g. C#)?

Or alternatively, if the main goal is to test direct tracing, does it matter that it's unlikely to run Java analysis on macOS?

pr-checks/checks/cleanup-db-cluster-dir.yml
henrymercer: I'll preserve the existing behaviour here.  It seems unlikely that this would break.

pr-checks/checks/job-run-uuid-sarif.yml
henrymercer: True, but I'll leave this for future work.

pr-checks/checks/language-aliases.yml
henrymercer: I think a test is useful, particularly since default setup relies on language aliases.

pr-checks/checks/rust.yml
henrymercer: It provides some coverage against specific older versions.  I'll keep for now.

pr-checks/checks/autobuild-action.yml
henrymercer: Can you open an issue about this?  We want testing of both indirect and direct tracing, and I'd like to understand better why we disable the CLR tracer in this job.

pr-checks/checks/build-mode-autobuild.yml
```diff
@@ -18,6 +18,11 @@ steps:
       languages: java
       tools: ${{ steps.prepare-test.outputs.tools-url }}
 
+  - name: Install yq
+    if: runner.os == 'Windows'
+    run: |
+      choco install yq -y
+
```
mbg, line 25: Doesn't have to be done as part of this PR, but the `init-with-registries` workflow skips a check on Windows because `yq` is unavailable. We could probably install it in the same way there to run the check on Windows.

henrymercer: In the interests of getting this in, I'll leave this for the future.

mbg: Actually, we no longer run that workflow on Windows anyway, so this isn't needed in any case.

PR #3139, opened 2025-09-23
Fix `tools: linked` log message

henrymercer:
The CLI version number was wrong.

### Risk assessment


src/setup-codeql.ts
```diff
@@ -344,9 +319,33 @@ export async function getCodeQLSource(
     toolsInput = await getNightlyToolsUrl(logger);
```
mbg, line 319: Nit: Not something introduced by this PR, but I am not super keen on this rewriting of `toolsInput` from an alias to a URL. It makes it harder than necessary to follow what this function does. I think ideally this would be a function `resolveToolsAlias` that always returns a URL by either resolving the alias or returning the URL unchanged.

src/setup-codeql.ts
```diff
@@ -344,9 +319,33 @@ export async function getCodeQLSource(
...
+
+    logger.info(
+      `'tools: ${toolsInput}' was requested, so using CodeQL version ${cliVersion}, the version shipped with the Action.`,
+    );
+
+    if (toolsInput === "latest") {
+      logger.warning(
+        "`tools: latest` has been renamed to `tools: linked`, but the old name is still supported. No action is required.",
+      );
+    }
```
mbg, line 348: As discussed elsewhere, it was confusing for review purposes that this moved down here, which is because `cliVersion` was already defined here. I'd probably change this back and reference `defaults.cliVersion` in the log message, or import them differently. No strong feelings about this though.

PR #3141, opened 2025-09-23
Compute preliminary overlay database mode

cklin:
This PR updates the `init` action to compute a preliminary overlay database mode before the CodeQL CLI becomes available.  The preliminary overlay database mode will be used in a future PR.

### Risk assessment

### DIFF
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -90625,23 +90625,509 @@ var PACK_IDENTIFIER_PATTERN = (function() {
 })();
 
 // src/diff-informed-analysis-utils.ts
-var fs8 = __toESM(require("fs"));
-var path9 = __toESM(require("path"));
-
-// src/feature-flags.ts
 var fs7 = __toESM(require("fs"));
 var path8 = __toESM(require("path"));
+
+// src/feature-flags.ts
+var fs6 = __toESM(require("fs"));
+var path7 = __toESM(require("path"));
 var semver4 = __toESM(require_semver2());
 
 // src/defaults.json
 var bundleVersion = "codeql-bundle-v2.23.1";
 var cliVersion = "2.23.1";
 
-// src/overlay-database-utils.ts
-var crypto = __toESM(require("crypto"));
-var fs6 = __toESM(require("fs"));
-var path7 = __toESM(require("path"));
-var actionsCache = __toESM(require_cache3());
+// src/tools-features.ts
+var semver3 = __toESM(require_semver2());
+function isSupportedToolsFeature(versionInfo, feature) {
+  return !!versionInfo.features && versionInfo.features[feature];
+}
+
+// src/feature-flags.ts
+var DEFAULT_VERSION_FEATURE_FLAG_PREFIX = "default_codeql_version_";
+var DEFAULT_VERSION_FEATURE_FLAG_SUFFIX = "_enabled";
+var CODEQL_VERSION_ZSTD_BUNDLE = "2.19.0";
+var featureConfig = {
+  ["cleanup_trap_caches" /* CleanupTrapCaches */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_CLEANUP_TRAP_CACHES",
+    minimumVersion: void 0
+  },
+  ["cpp_dependency_installation_enabled" /* CppDependencyInstallation */]: {
+    defaultValue: false,
+    envVar: "CODEQL_EXTRACTOR_CPP_AUTOINSTALL_DEPENDENCIES",
+    legacyApi: true,
+    minimumVersion: "2.15.0"
+  },
+  ["diff_informed_queries" /* DiffInformedQueries */]: {
+    defaultValue: true,
+    envVar: "CODEQL_ACTION_DIFF_INFORMED_QUERIES",
+    minimumVersion: "2.21.0"
+  },
+  ["disable_csharp_buildless" /* DisableCsharpBuildless */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_DISABLE_CSHARP_BUILDLESS",
+    minimumVersion: void 0
+  },
+  ["disable_java_buildless_enabled" /* DisableJavaBuildlessEnabled */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_DISABLE_JAVA_BUILDLESS",
+    legacyApi: true,
+    minimumVersion: void 0
+  },
+  ["disable_kotlin_analysis_enabled" /* DisableKotlinAnalysisEnabled */]: {
+    defaultValue: false,
+    envVar: "CODEQL_DISABLE_KOTLIN_ANALYSIS",
+    legacyApi: true,
+    minimumVersion: void 0
+  },
+  ["export_diagnostics_enabled" /* ExportDiagnosticsEnabled */]: {
+    defaultValue: true,
+    envVar: "CODEQL_ACTION_EXPORT_DIAGNOSTICS",
+    legacyApi: true,
+    minimumVersion: void 0
+  },
+  ["resolve_supported_languages_using_cli" /* ResolveSupportedLanguagesUsingCli */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_RESOLVE_SUPPORTED_LANGUAGES_USING_CLI",
+    minimumVersion: void 0,
+    toolsFeature: "builtinExtractorsSpecifyDefaultQueries" /* BuiltinExtractorsSpecifyDefaultQueries */
+  },
+  ["overlay_analysis" /* OverlayAnalysis */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_OVERLAY_ANALYSIS",
+    minimumVersion: void 0
+  },
+  ["overlay_analysis_actions" /* OverlayAnalysisActions */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_OVERLAY_ANALYSIS_ACTIONS",
+    minimumVersion: void 0
+  },
+  ["overlay_analysis_code_scanning_actions" /* OverlayAnalysisCodeScanningActions */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_OVERLAY_ANALYSIS_CODE_SCANNING_ACTIONS",
+    minimumVersion: void 0
+  },
+  ["overlay_analysis_code_scanning_cpp" /* OverlayAnalysisCodeScanningCpp */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_OVERLAY_ANALYSIS_CODE_SCANNING_CPP",
+    minimumVersion: void 0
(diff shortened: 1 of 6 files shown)


src/config-utils.ts
```diff
@@ -734,7 +813,10 @@ export async function getOverlayDatabaseMode(
     );
     return nonOverlayAnalysis;
   }
-  if (!(await codeQlVersionAtLeast(codeql, CODEQL_OVERLAY_MINIMUM_VERSION))) {
+  if (
+    codeql !== undefined &&
+    !(await codeQlVersionAtLeast(codeql, CODEQL_OVERLAY_MINIMUM_VERSION))
+  ) {
```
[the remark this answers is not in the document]
cklin: The import statement for `CODEQL_OVERLAY_MINIMUM_VERSION` is definitely not missing.

src/feature-flags.ts
```diff
@@ -157,7 +156,7 @@ export const featureConfig: Record<
   [Feature.OverlayAnalysis]: {
     defaultValue: false,
     envVar: "CODEQL_ACTION_OVERLAY_ANALYSIS",
-    minimumVersion: CODEQL_OVERLAY_MINIMUM_VERSION,
+    minimumVersion: undefined,
```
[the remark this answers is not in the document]
cklin: The commit message already answered exactly this question:

> getOverlayDatabaseMode() already performs the same version check, so we can remove minimumVersion from Feature.OverlayAnalysis.

src/config-utils.ts
```diff
@@ -331,6 +332,36 @@ export async function getLanguages(
...
+ * before the CodeQL CLI is available.
+ */
+async function getUnverifiedLanguagesForOverlay(
+  languagesInput: string | undefined,
+  repository: RepositoryNwo,
+  sourceRoot: string,
+  logger: Logger,
+): Promise<string[]> {
+  // Obtain languages without filtering them.
+  const { rawLanguages } = await getRawLanguages(
```
mbg, line 350: `getRawLanguages` makes an API call as part of calling `getRawLanguagesInRepo` (if there's no `languages` input). Since `getRawLanguages` is currently only called at most once, the result of that API call is not cached. With this case, it could get called more than once. It might make sense to cache the response since it shouldn't change between calls.

src/config-utils.ts
```diff
@@ -331,6 +332,36 @@ export async function getLanguages(
...
 
+/**
+ * Get the (unverified) languages for overlay analysis.
+ *
+ * This is a simplified version of `getLanguages` that only resolves language
+ * aliases but does not check if the languages are actually supported by the
+ * CodeQL CLI. It is intended to be used for overlay analysis preparations
+ * before the CodeQL CLI is available.
+ */
+async function getUnverifiedLanguagesForOverlay(
```
mbg, line 343: Minor: It would be nice if this could live in `overlay-database-utils.ts` since it is specific to overlay databases. I imagine it's here instead because of the call to `getRawLanguages`. Do you think you could make `rawLanguages` a parameter of `getUnverifiedLanguagesForOverlay` and move it to `overlay-database-utils.ts`?

src/config-utils.ts
```diff
@@ -331,6 +332,36 @@ export async function getLanguages(
...
+  // Obtain languages without filtering them.
+  const { rawLanguages } = await getRawLanguages(
+    languagesInput,
+    repository,
+    sourceRoot,
+    logger,
+  );
+  const languageAliases = overlayLanguageAliases as Record<string, string>;
+
+  const languagesSet: string[] = [];
```
mbg, line 358: Minor: `languagesSet` suggests that this is a `Set` object, but it is actually an array. Also the name is misleading, because I don't think that `getRawLanguages` attempts any de-duplication, so in theory we could have multiple languages multiples times here if the `languages` input contains them more than once.

src/config-utils.ts
```diff
@@ -539,12 +569,48 @@ async function downloadCacheWithTime(
...
 }
 
+/**
+ * Amends the input config file if configInput is provided.
+ * If configInput is set, it takes precedence over configFile.
+ *
+ * This function should be called only once on any specific `InitConfigInputs`
+ * object. Otherwise it could emit a false warning.
+ */
+export function amendInputConfigFile(
```
mbg, line 579: This name is misleading because it suggests that we are somehow changing an existing configuration file, while we are actually just determining the one we wish to use (and writing the one provided as input to disk, if needed).

src/config-utils.ts
```diff
@@ -539,12 +569,48 @@ async function downloadCacheWithTime(
   return { trapCaches, trapCacheDownloadTime };
 }
 
+/**
+ * Amends the input config file if configInput is provided.
+ * If configInput is set, it takes precedence over configFile.
+ *
+ * This function should be called only once on any specific `InitConfigInputs`
+ * object. Otherwise it could emit a false warning.
```
mbg, line 577: It would be nice to enforce this, e.g. with some state capturing whether this has been called already for the given `inputs` or in the types.

src/config-utils.ts
```diff
@@ -539,12 +569,48 @@ async function downloadCacheWithTime(
...
+  logger: Logger,
+): void {
+  // if configInput is set, it takes precedence over configFile
+  if (inputs.configInput) {
+    if (inputs.configFile) {
+      logger.warning(
+        `Both a config file and config input were provided. Ignoring config file.`,
+      );
+    }
+    inputs.configFile = userConfigFromActionPath(inputs.tempDir);
```
mbg, line 590: Minor: I don't love that this relies on mutating `inputs` and that the return type of this function is `void`. Perhaps you could refactor this function so that it returns the appropriate value for `inputs.configFile` (i.e. either the initial `inputs.configFile` or the result of `userConfigFromActionPath(inputs.tempDir)`) and then update `inputs.configFile` to the result at the call site.

src/config-utils.ts
```diff
@@ -757,6 +839,62 @@ export async function getOverlayDatabaseMode(
...
+ * in InitConfigInputs, without depending on CodeQL.
+ *
+ * This is a simplified version of getOverlayDatabaseMode that can be called
+ * before the CodeQL CLI is available.
+ *
+ * @param inputs The initialization configuration inputs.
+ * @returns An object containing the overlay database mode and whether the
+ * action should perform overlay-base database caching.
+ */
+export async function getPreliminaryOverlayDatabaseMode(
```
mbg, line 853: I am concerned about the amount of logic duplication that is taking place here (particularly the involvement of `calculateAugmentation` and `generateCodeScanningConfig`). I understand that these are required for `getOverlayDatabaseMode`, but we risk turning this into a big source of errors if there is inconsistency in what this function determines vs what `initActionState` determines.

For example, this function does not account for `analysis-kinds` being only `code-quality`, in which case all query customisation should be disabled. Notably, that has different semantics for `repositoryProperties` (where additional configuration is an `info`-level log message) and for inputs/config file values (where it is a `ConfigurationError`).

I _assume_ that it is important that the `computedConfig` here mirrors the one in `initConfig` as accurately as possible, but correct me if I am wrong and this particular aspect doesn't matter.

If that does matter and viewing this issue in isolation, we would ideally have one function which figures `computedConfig` out given the necessary arguments for `calculateAugmentation` and `generateCodeScanningConfig` taking `analysis-kinds` into account, and then use that in both this function and `initConfig`.

src/init-action.ts
```diff
@@ -288,47 +327,14 @@ async function run() {
     }
 
     // Warn that `quality-queries` is deprecated if there is an argument for it.
-    const qualityQueriesInput = getOptionalInput("quality-queries");
-
-    if (qualityQueriesInput !== undefined) {
+    if (inputs.qualityQueriesInput !== undefined) {
```
mbg, line 330: Minor: is there any particular reason for removing this constant, since it's still used in two places as before?

src/config-utils.ts
```diff
@@ -787,35 +925,20 @@ function hasQueryCustomisation(userConfig: UserConfig): boolean {
...
-
-  const config = await initActionState(inputs, userConfig);
+  const userConfig = await loadUserConfig(
+    inputs.configFile,
+    inputs.workspacePath,
+    inputs.apiDetails,
+    tempDir,
+    logger,
+  );
+  const config = await initActionState(inputs, userConfig, codeql);
```
mbg, line 941: The interaction between this and what happens in `getPreliminaryOverlayDatabaseMode` is not obvious and possibly problematic. Specifically, `getPreliminaryOverlayDatabaseMode` suggests that it is related to only overlay databases, but it actually writes the configuration file that `loadUserConfig` ends up loading here. That configuration file already contains a CLI configuration that is a combination of a `config-file` or `config` input, with other Action inputs such as `queries` and `packs` as well as repository property values.

`initActionState` then performs essentially duplicate work calling `calculateAugmentation` (although I understand that it takes the `languages` as input and therefore could lead to different results than for the first call to it) which then gets fed to a second call to `generateCodeScanningConfig`. This could lead to some odd behaviours where assumptions that we make in `generateCodeScanningConfig` are invalidated because `userConfig` already contains merged results of various inputs from the first call to `generateCodeScanningConfig` that took place in `getPreliminaryOverlayDatabaseMode`. So we might end up with duplicate entires in the final configuration file, duplicate or confusing log messages, etc.

src/config-utils.ts
```diff
@@ -652,14 +724,18 @@ async function isOverlayAnalysisFeatureEnabled(
...
  * action should perform overlay-base database caching.
  */
 export async function getOverlayDatabaseMode(
-  codeql: CodeQL,
+  codeql: CodeQL | undefined,
   repository: RepositoryNwo,
   features: FeatureEnablement,
-  languages: Language[],
+  languages: Language[] | undefined,
+  languagesInput: string | undefined,
```
mbg, line 738: I am worried that having both (possibly disagreeing) values here could be a source of confusion or errors.

cklin:
Hi @mbg,

Thank you for your detailed comments.  Before we go into the details, can you confirm that I understand your main concerns correctly?

1. You are concerned that this change is high-risk because it is a fairly big change, not under feature flags, and there is insufficient test coverage.
2. You are concerned that there is too much code duplication between `getPreliminaryOverlayDatabaseMode()` and the rest of the `init` action, and that the behavior of `getPreliminaryOverlayDatabaseMode()` is inconsistent with the main `init` action logic.
3. You are concerned that there is insufficient logical separation between `getPreliminaryOverlayDatabaseMode()` and the rest of the `init` action (for example, with `initActionState()`), and that the hidden interactions will cause maintenance burden down the road.

Do I understand your main concerns correctly?  Am I missing anything?

henrymercer (commented):
Can we simplify our decision about whether to match the CLI version to the overlay version, and therefore reduce the amount of complexity here?  I think it would be OK if we matched the CLI version to the overlay version, even if once we obtained the CLI we decided later we aren't actually going to run overlay.  The main thing we need to ensure is that the decision is stable and doesn't flip on and off.

In particular, it would simplify things greatly if we didn't need to compute the full code scanning configuration, and instead made a decision based on the user's requested languages and build modes, and the feature flags.

PR #3146, opened 2025-09-24
Provide `Authorization` header when downloading `update-job-proxy`

mbg:
This changes the `start-proxy` action to provide an `Authorization` header for the `toolcache.downloadTool` call under the same circumstances where we provide one for the `toolcache.downloadTool` call when downloading the CLI bundle.

### Risk assessment

### DIFF
--- a/src/api-client.ts
+++ b/src/api-client.ts
@@ -4,6 +4,7 @@ import * as retry from "@octokit/plugin-retry";
 import consoleLogLevel from "console-log-level";
 
 import { getActionVersion, getRequiredInput } from "./actions-util";
+import { Logger } from "./logging";
 import { getRepositoryNwo, RepositoryNwo } from "./repository";
 import {
   ConfigurationError,
@@ -54,7 +55,7 @@ function createApiClientWithDetails(
   );
 }
 
-export function getApiDetails() {
+export function getApiDetails(): GitHubApiDetails {
   return {
     auth: getRequiredInput("token"),
     url: getRequiredEnvParam("GITHUB_SERVER_URL"),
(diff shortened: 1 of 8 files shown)


src/api-client.ts
```diff
@@ -72,6 +73,34 @@ export function getApiClientWithExternalAuth(
...
+): string | undefined {
+  if (
+    url.startsWith(`${apiDetails.url}/`) ||
+    (apiDetails.apiURL && url.startsWith(`${apiDetails.apiURL}/`))
+  ) {
+    logger.debug(`Providing an authorization token to download ${purpose}.`);
+    return `token ${apiDetails.auth}`;
+  }
+
+  logger.debug(`Downloading ${purpose} without an authorization token.`);
```
esbena: Nit: we are not downloading anything in this function.

src/start-proxy-action.ts
```diff
@@ -192,10 +193,20 @@ async function getProxyBinaryPath(logger: Logger): Promise<string> {
 
   let proxyBin = toolcache.find(proxyFileName, proxyInfo.version);
   if (!proxyBin) {
+    // We only want to provide an authorization header if we are downloading
+    // from the same GitHub instance the Action is running on.
+    // This avoids leaking Enterprise tokens to dotcom.
```
esbena: It feels like this comment belongs inside `getAuthorizationHeaderFor` rather than out here.
The code below specifically does not do any direct instance checking, despite the comment implying that.

esbena (approved):
LGTM, with some minor doc/log thoughts.

src/api-client.ts
mbg: Mainly a trade-off between keeping the log messages the same and not duplicating the conditions/logging to outside of `getAuthorizationHeaderFor`. I have no strong views here, and since these are `debug` messages, it probably doesn't matter if they change

esbena (dismissed):

PR #3148, opened 2025-09-24
build: use --serial in 'just test_file'

cklin:
Some tests require the `--serial` flag to pass.

### Risk assessment

### DIFF
--- a/justfile
+++ b/justfile
@@ -22,7 +22,7 @@ test: build
 
 # Run the tests for a single file
 test_file filename: build
-    npx ava --verbose {{filename}}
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -22,7 +22,7 @@ test: build
 
 # Run the tests for a single file
 test_file filename: build
-    npx ava --verbose {{filename}}
+    npx ava --serial --verbose {{filename}}
```
mbg, line 25: Suggestion: add a command for running an individual test to `package.json` and then call that from here?

PR #3153, opened 2025-09-25
Improve `pr-checks` workflow

mbg:
Improves the `pr-checks` workflow so that:

- All checks run, even if one fails.
- A diff of the changes to the JS bundles is added as job summary if there are changes
- A diff of the changes to the generated workflows is added as job summary if there are changes

### Risk assessment

### DIFF
--- a/.github/workflows/script/verify-pr-checks.sh
+++ b/.github/workflows/script/verify-pr-checks.sh
@@ -20,6 +20,14 @@ if [ ! -z "$(git status --porcelain)" ]; then
     git diff
     git status
     >&2 echo "Failed: PR checks are not up to date. Run 'cd pr-checks && python3 sync.py' to update"
+
+    echo "### Generated workflows diff" >> $GITHUB_STEP_SUMMARY
+    echo "" >> $GITHUB_STEP_SUMMARY
+    echo '```diff' >> $GITHUB_STEP_SUMMARY
+    git diff --output="$RUNNER_TEMP/workflows.diff"
+    cat "$RUNNER_TEMP/workflows.diff" >> $GITHUB_STEP_SUMMARY
+    echo '```' >> $GITHUB_STEP_SUMMARY
+
     exit 1
 fi
 echo "Success: PR checks are up to date"


.github/workflows/script/check-js.sh
```diff
@@ -16,6 +16,18 @@ if [ ! -z "$(git status --porcelain)" ]; then
     # If we get a fail here then the PR needs attention
     >&2 echo "Failed: JavaScript files are not up to date. Run 'rm -rf lib && npm run-script build' to update"
     git status
+
+    echo "### Transpiled JS diff" >> $GITHUB_STEP_SUMMARY
+    echo "" >> $GITHUB_STEP_SUMMARY
+    echo '```diff' >> $GITHUB_STEP_SUMMARY
+    git diff --output="$RUNNER_TEMP/js.diff"
+    cat "$RUNNER_TEMP/js.diff" >> $GITHUB_STEP_SUMMARY
```
henrymercer, line 24: Consider uploading the diff as an artifact

.github/workflows/script/verify-pr-checks.sh
```diff
@@ -20,6 +20,14 @@ if [ ! -z "$(git status --porcelain)" ]; then
     git diff
     git status
     >&2 echo "Failed: PR checks are not up to date. Run 'cd pr-checks && python3 sync.py' to update"
+
+    echo "### Generated workflows diff" >> $GITHUB_STEP_SUMMARY
+    echo "" >> $GITHUB_STEP_SUMMARY
+    echo '```diff' >> $GITHUB_STEP_SUMMARY
+    git diff --output="$RUNNER_TEMP/workflows.diff"
+    cat "$RUNNER_TEMP/workflows.diff" >> $GITHUB_STEP_SUMMARY
```
henrymercer, line 28: Ditto here

.github/workflows/script/check-js.sh
mbg: I initially had that, but thought this was nicer because you can see the syntax-highlighted diff without having to download a file, unpack it, and open it.

henrymercer: I think both are nice in case there's a big diff.  But this is a very minor comment, good to go as is.

mbg: Let's leave it for now then, but we can add it later if we find it's needed?

henrymercer:
A common gotcha is having out of date dependencies locally.  We could consider posting a comment on the PR with the command to run (e.g. `npm i && npm run build && npm test && npm run lint`).  We could also consider integrating a check that dependenices are up to date as part of the build process, if this is possible to do quickly.

henrymercer:
I'm not super keen on running `npm i` as part of `npm run build` just because of how slow it is, but this is also something we could consider.

mbg:
Agreed on both points. I don't think we should run `npm install` as part of anything we do frequently during normal development. 

I'll have a look at what we can do to detect whether `npm install` has to be run, but separately from this PR.

PR #3157, opened 2025-09-25
Fix `upload-sarif` not uploading non-`.sarif` files

### LINKED ISSUE #3156: upload-sarif 3.30.4 silently stopped uploading (2025-09-25)
The only difference between the below two runs is the upload-sarif version.

3.30.3:
Image

3.30.4:
Image

mbg:
Fixes #3156. 

In https://github.com/github/codeql-action/pull/3123, we inadvertently introduced a regression by using the same logic for uploading SARIF files as in the `analyze` Action: if a file ends in `.quality.sarif`, we upload it to Code Quality, and if it ends in `.sarif` (but not `.quality.sarif`), we upload it to Code Scanning.

However, that didn't account for the possibility that `sarif-path` is set to a file or directory containing files with other extensions (such as `.json`).

This PR restores this behaviour by using `!analyses.CodeQuality.sarifPredicate(name)` as the file predicate for Code Scanning in the `upload-sarif` Action.

I have also refactored `upload-sarif-action` slightly to move some parts to a new `upload-sarif` module that I have added some basic tests for.

### Risk assessment

### DIFF
--- a/src/upload-sarif.ts
+++ b/src/upload-sarif.ts
@@ -1,0 +1,55 @@
+import * as fs from "fs";
+
+import * as analyses from "./analyses";
+import { Features } from "./feature-flags";
+import { Logger } from "./logging";
+import * as upload_lib from "./upload-lib";
+
+/**
+ * Searches for SARIF files for the given `analysis` in the given `sarifPath`.
+ * If any are found, then they are uploaded to the appropriate endpoint for the given `analysis`.
+ *
+ * @param logger The logger to use.
+ * @param features Information about FFs.
+ * @param sarifPath The path to a SARIF file or directory containing SARIF files.
+ * @param pathStats Information about `sarifPath`.
+ * @param checkoutPath The checkout path.
+ * @param analysis The configuration of the analysis we should upload SARIF files for.
+ * @param category The SARIF category to use for the upload.
+ * @returns The result of uploading the SARIF file(s) or `undefined` if there are none.
+ */
+export async function findAndUpload(
+  logger: Logger,
+  features: Features,
+  sarifPath: string,
+  pathStats: fs.Stats,
+  checkoutPath: string,
+  analysis: analyses.AnalysisConfig,
+  category?: string,
+): Promise<upload_lib.UploadResult | undefined> {
+  let sarifFiles: string[] | undefined;
+
+  if (pathStats.isDirectory()) {
+    sarifFiles = upload_lib.findSarifFilesInDir(
+      sarifPath,
+      analysis.sarifPredicate,
+    );
+  } else if (pathStats.isFile() && analysis.sarifPredicate(sarifPath)) {
+    sarifFiles = [sarifPath];
+  } else {
+    return undefined;
+  }
+
+  if (sarifFiles.length !== 0) {
+    return await upload_lib.uploadSpecifiedFiles(
+      sarifFiles,
+      checkoutPath,
+      category,
+      features,
+      logger,
+      analysis,
+    );
+  }
+
+  return undefined;
+}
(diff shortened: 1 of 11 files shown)


src/upload-sarif.ts
```diff
@@ -0,0 +1,117 @@
...
+  const uploadResults: UploadSarifResults = {};
+  const uploadResult = await findAndUpload(
+    logger,
+    features,
+    sarifPath,
+    pathStats,
+    checkoutPath,
+    analyses.CodeScanning,
+    // Upload anything to Code Scanning that is not intended for Code Quality.
+    (name) => !analyses.CodeQuality.sarifPredicate(name),
```
esbena: I'm concerned about the addition of this argument (also, there's missing jsdoc for the new parameter).
Inside `findAndUpload` it is not clear that you should no longer use `analysis.sarifPredicate`, and that it only is subtly wrong to do so.
I'd prefer that we change the config object to have a different sarifPredicate:

```suggestion
    { 
      ... analyses.CodeScanning,
      // Upload anything to Code Scanning that is not intended for Code Quality.
      sarifPredicate: (name) => !analyses.CodeQuality.sarifPredicate(name)
    },
```

src/upload-sarif.ts
```diff
@@ -0,0 +1,117 @@
...
+  sarifPredicate: (name: string) => boolean,
+  category?: string,
+): Promise<upload_lib.UploadResult | undefined> {
+  let sarifFiles: string[] | undefined;
+
+  if (pathStats.isDirectory()) {
+    sarifFiles = upload_lib.findSarifFilesInDir(sarifPath, sarifPredicate);
+  } else if (pathStats.isFile() && sarifPredicate(sarifPath)) {
+    sarifFiles = [sarifPath];
+  } else {
```
esbena: This is subtly different from the sibling file finding logic in `upload-lib.ts`
https://github.com/github/codeql-action/blob/245e2e5cde27ee775820c30dec085797b344b9a0/src/upload-lib.ts#L448-L459. I wonder if we could unify them. I'm aware that upload-lib.ts was not modified in this PR, so this is an early concern about implementation divergence.

Concretely, the latter's lack of `sarifPredicate(sarifPath)` in the `else` is intentional as per your `.json` comment, but there's no jsdoc to document that the `isSarif` is unused when single files are provided..

esbena (commented):
This LGTM semantically, but I worry about erosion that the new predicate juggling introduces.

src/upload-sarif.ts
mbg: > I'd prefer that we change the config object to have a different sarifPredicate:

I agree. I wasn't particularly happy with this extra parameter when I implemented it last night either, but it was the easiest fix. Having thought about it more over night, I reached the same conclusion that a separate configuration where `sarifPredicate` is overriden is better.

But there is a further issue, which you haven't flagged up as such. Using `!analyses.CodeQuality.sarifPredicate(name)` as `sarifPredicate` for the call to `upload_lib.findSarifFilesInDir` in `findAndUpload` also subtly changes the behaviour to what we had before: it now uploads _all_ files (other than `.quality.sarif`) from the directory, rather than just `.sarif` ones (which was the previous behaviour, but isn't adequate anymore because we need to filter out the `.quality.sarif` ones). I think that may have been my original reasoning for the extra `sarifPredicate` parameter so that I could use one for the file case and another for the directory case (but then forgot and used it for both). 

I'll think about a better solution.

src/upload-sarif.ts
mbg: I agree that we should unify things and it's been on my radar to do that / get rid of `uploadFiles` since we worked on https://github.com/github/codeql-action/pull/3123. This probably is as good an opportunity to do that, in combination with coming up for a good solution for the other issue here.

mbg:
Closing this in favour of https://github.com/github/codeql-action/pull/3160, https://github.com/github/codeql-action/pull/3166, and https://github.com/github/codeql-action/pull/3167

PR #3160, opened 2025-09-26
Hotfix `upload-sarif` not uploading non-`.sarif` files

mbg:
Slimmed down #3157 

### Risk assessment

For internal use only. Please select the risk level of this change:

- **Low risk:** Changes are fully under feature flags, or have been fully tested and validated in pre-production environments and are highly observable, or are documentation or test only.

### Merge / deployment checklist

- Confirm this change is backwards compatible with existing workflows.
- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.
- Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary.

PR #3166, opened 2025-09-29
Add tests for `upload-sarif`

mbg:
This PR takes parts of #3157:

- Moves logic from `upload-sarif-action.ts` to `upload-sarif.ts`
- Adds unit test coverage for the functions in `upload-sarif.ts`
- Fixes `sarif-ids` output format to match documentation
- Throws a `ConfigurationError` if no SARIF files were uploaded, matching the behaviour of `upload-sarif` prior to https://github.com/github/codeql-action/pull/3123

This doesn't make any further changes to the actual upload logic beyond the hotfix we added in #3160. I am planning a follow-up PR to improve on that after having come up with an idea for a good approach over the weekend, but want to the tests, `sarif-ids` fix, and error for the case where no files were uploaded merged first.

### Risk assessment

### DIFF
--- a/src/upload-sarif.test.ts
+++ b/src/upload-sarif.test.ts
@@ -109,7 +109,7 @@ const uploadSarifMacro = test.macro({
         "uploadSpecifiedFiles",
       );
 
-      for (const analysisKind of Object.keys(expectedResult)) {
+      for (const analysisKind of Object.values(AnalysisKind)) {
         uploadSpecifiedFiles
           .withArgs(
             sinon.match.any,
@@ -134,8 +134,10 @@ const uploadSarifMacro = test.macro({
       for (const analysisKind of Object.values(AnalysisKind)) {
         const analyisKindResult = expectedResult[analysisKind];
         if (analyisKindResult) {
+          // We are expecting a result for this analysis kind, check that we have it.
           t.deepEqual(actual[analysisKind], analyisKindResult.uploadResult);
-
+          // Additionally, check that the mocked `uploadSpecifiedFiles` was called with only the file paths
+          // that we expected it to be called with.
           t.assert(
             uploadSpecifiedFiles.calledWith(
               analyisKindResult.expectedFiles?.map(toFullPath) ??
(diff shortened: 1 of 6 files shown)


src/upload-sarif.test.ts
```diff
@@ -113,16 +121,38 @@ const uploadSarifMacro = test.macro({
...
+            uploadSpecifiedFiles.calledWith(
+              analyisKindResult.expectedFiles?.map(toFullPath) ??
+                fullSarifPaths,
+              sinon.match.any,
+              sinon.match.any,
+              features,
+              logger,
+              analysisKind === AnalysisKind.CodeScanning
+                ? CodeScanning
+                : CodeQuality,
```
esbena, line 151: could this just be `analysisKind`? Or do we prefer the explicit casing?

esbena (approved):
LGTM. 

It's not blocking, but I'm getting a bit concerned about the complexity of the tests, especially
https://github.com/github/codeql-action/pull/3166/commits/2adc894410deceb53c0abc0883f9c32f1e809cdd. It's the classic tradeoff between what tests current behaviour/API and what is maintainable.

mbg:
> I'm getting a bit concerned about the complexity of the tests

I could modify `UploadResult` to include the array of files, but then we'd probably have to mock a bunch of other functions that `uploadSpecifiedFiles` calls or have more things for the tests to set up.

I'm inclined to accept the tests as they are for now, since I am planning to make further changes anyway and they might affect these tests and allow us to simplify this again later.

src/upload-sarif.test.ts
mbg: `analysisKind: AnalysisKind` while `CodeScanning: AnalysisConfig` and `CodeQuality: AnalysisConfig`. It maps the kind to the matching configuration. We could have this as a function for reusability instead of this ad-hoc mechanism, but we can't just use `analysisKind` here since we need an `AnalysisConfig` object for `uploadSpecifiedFiles`.

esbena: right. My eyes missed the `AnalysisKind.` prefix.

PR #2006, opened 2023-11-22
change to node20 for all actions

nickfyson:
This PR upgrades all the actions to use `node20`, and bumps the major version to `v3`.

It also adds a new PR check to confirm the consistency of node versions across all actions, and ensure that we do not change this version in backport PRs to older releases.

### Merge / deployment checklist

- [x] Confirm this change is backwards compatible with existing workflows.
- [x] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) has been updated if necessary.
- [x] Confirm the [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) has been updated if necessary.

PR #3169, opened 2025-09-30
[v4] Upgrade Node.js runtime from v20 to v24

mario-campos:
### Risk assessment

  - Changes should only apply to new `v4` tag, once it's released; `v3` should continue to work as-is.
  - All tests are passing.

- [x] Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.


README.md
```diff
@@ -62,7 +62,8 @@ For compiled languages:
 
 The following versions of the CodeQL Action are currently supported:
 
-- v3 (latest)
+- v3
+- v4 (latest)
```
henrymercer:
```suggestion
- v4 (latest)
- v3
```

henrymercer (approved):
Looks great!  I only have a very minor suggested change to the README.

Before merging this, we'll need to update the set of required checks — there's instructions on how to do that here: https://github.com/github/codeql-action/blob/main/CONTRIBUTING.md#keeping-the-pr-checks-up-to-date-admin-access-required

henrymercer (dismissed):

.github/workflows/script/check-js-20.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -eu
+
+# Change @types/node to v20 temporarily to check that the generated JS files are correct.
+contents=$(jq '.devDependencies."@types/node" = "^20.0.0"' package.json)
+echo "${contents}" > package.json
```
mbg: I am a bit confused by what's going on in this script. I can see that this downgrades `@types/node` and we then run `npm install` to re-generate `package-lock.json` and install the changed dependencies. 

But then we don't actually rebuild the JavaScript files before creating the `"Use @types/node 20"` commit. Instead we just run the build and check that it matches what is in the repo, which then obviously fails.

So, at a minimum, there should be a call to `npm run build` after `npm install`.

But if we do that, then it's not clear to me what this check accomplishes at all, because the build we will perform to regenerate the JavaScript files to accommodate the different version of `@types/node` would also include any other changes to the JS files. This would still get caught by the `check-js.sh` script for Node 24, but makes this one redundant. 

---

On a different note, I am not sure whether `@types/node` really has to be upgraded as part of this PR. It seems that we could leave it at the minimum version of Node that we support (i.e. `^20.0.0`) and only bump it once we drop support for `v3` of the CodeQL Action?

If that would work (which it seems to locally for me), then we could save ourselves the trouble here of having to rebuild the JS files for `v3` every time.

If it is necessary for us to align the `@types/node` version with the version of Node we are running with, then we'd also have to push a commit to switch to `v20` of `@types/node` when tagging a `v3` release. It also raises a question of: if the different library versions are needed for the different Node versions, then what implications does that have for developing against one and then switching to another for `v3`.

mario-campos: >But if we do that, then it's not clear to me what this check accomplishes at all, because the build we will perform to regenerate the JavaScript files to accommodate the different version of `@types/node` would also include any other changes to the JS files.

That's a valid point. I'm modeling this check after the one that was added [last time](https://github.com/github/codeql-action/pull/2006/files#diff-1da7a27709d11ac16eef3de8e46ed10672b7b1d63a790ffbcbd23a94b19efdcb) (v16 -> v20). My understanding is that it was there to ensure that the source files always match the generated files, regardless of the underlying Node.js runtime&mdash;i.e. to ensure backwards-compatibility with the older version of Node.js?

However, I do agree that it adds a bit of accidental complexity, which leads me to your next point...

---

>On a different note, I am not sure whether `@types/node` really has to be upgraded as part of this PR. It seems that we could leave it at the minimum version of Node that we support (i.e. `^20.0.0`) and only bump it once we drop support for `v3` of the CodeQL Action?
>
>If that would work (which it seems to locally for me), then we could save ourselves the trouble here of having to rebuild the JS files for `v3` every time.

That would be great if it works. [Copilot suggested that `@types/node` should match the runtime](https://github.com/copilot/share/c0711118-09e4-8462-a153-a448640a4958) to ensure that types match the new/updated APIs; hence the `@types/node` upgrade.

henrymercer: The thinking behind the check is to check that the Action typechecks against the older version of Node.  It doesn't really matter whether any files have changed or not, only that the types are compatible.  In this way, we avoid inadvertently depending on an API that doesn't exist and will only fail when we backport the release to `v3`.

However I don't see any issues using an older version of `@types/node` to also achieve this.

mario-campos: I've reverted the changes to `@types/node`, build.mjs, and the PR check.

mbg: > Copilot suggested that @types/node should match the runtime

Funnily enough, Copilot told me the opposite :)

mbg (dismissed):
LGTM, thanks for addressing my feedback and tackling this important work!

henrymercer (dismissed):
Let's hold off on merging this until Monday, and then kick off a release right after we merge.

henrymercer:
Also, it looks like there are a couple of merge conflicts to solve.  You'll probably need a re-approval after fixing them.

mario-campos:
Weirdly, I don't see those merge conflicts:

```shell
mario-campos@G49XGKM6FH ~/codeql-action (mario-campos/node24)> git status
On branch mario-campos/node24
Your branch is up to date with 'origin/mario-campos/node24'.

nothing to commit, working tree clean
```

mbg:
@mario-campos Have you run `git fetch origin` and then attempted `git merge origin/main`? I'd expect the merge conflicts to show up then.

henrymercer:
I think we're good to update the required checks (see [these instructions](https://github.com/github/codeql-action/blob/main/CONTRIBUTING.md#keeping-the-pr-checks-up-to-date-admin-access-required)) and kick off a release.

Once the v4 release is merged, the release automation should open a backport from v4 to v3.  That backport will bring in all the changes from v4, including the bump to `node24` in the `action.yml`s.  At that point we'll need to push a commit to the releases/v3 branch that moves the Actions back to `node20`.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ No user facing changes.
 - We added support for reducing the size of dependency caches for Java analyses, which will reduce cache usage and speed up workflows. This will be enabled automatically at a later time. [#3107](https://github.com/github/codeql-action/pull/3107)
 - You can now run the latest CodeQL nightly bundle by passing `tools: nightly` to the `init` action. In general, the nightly bundle is unstable and we only recommend running it when directed by GitHub staff. [#3130](https://github.com/github/codeql-action/pull/3130)
 - Update default CodeQL bundle version to 2.23.1. [#3118](https://github.com/github/codeql-action/pull/3118)
+- [v4+ only] The CodeQL Action now runs on Node.js v24. [#3169](https://github.com/github/codeql-action/pull/3169)
```
henrymercer: Just noticed that this has automerged into the wrong place.

mario-campos: Oh, yup. Weird. I'll bring it back up to `[UNRELEASED]`.

PR #3175, opened 2025-10-03
Support requesting latest version from toolcache with `tools: toolcache`

mbg:
Allows `toolcache` as a possible input for `tools`. Useful if you don't necessarily care about the a specific version of CodeQL and care more about having it readily available on the runner.

### Risk assessment


mbg:
https://github.com/github/codeql-action/actions/runs/18225669790/job/51896145459#step:6:28 is a run of the new PR check with `tools: toolcache` working as expected.

pr-checks/checks/bundle-from-toolcache.yml
```diff
@@ -0,0 +1,32 @@
...
+        const toolcache = require('@actions/tool-cache');
+        const allCodeqlVersions = toolcache.findAllVersions('CodeQL');
+        if (allCodeqlVersions.length === 0) {
+          throw new Error(`CodeQL could not be found in the toolcache`);
+        }
+  - id: init
+    uses: ./../action/init
+    with:
+      languages: javascript
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
```
mbg, line 22: I have set `continue-on-error: true` here since I didn't want to depend on the CLI always being available in the toolcache. In general, I am happy for this check to just exercise that the workflow generally works if `tools: toolcache`, but perhaps we could post a comment on the PR if no CLI was found in the toolcache?

henrymercer (changes requested):
I'm not against having the ability to request tools from the toolcache for internal testing purposes for instance.  However we need to make it very clear that this will cause alerts to open and close repeatedly while new runner images roll out (a period of about half a day or so every two weeks).  For this reason, this is an option we're unlikely to want to recommend.

Do you have a particular use case in mind?

pr-checks/checks/bundle-from-toolcache.yml
henrymercer: I think we can probably depend on CodeQL being available in the toolcache.

PR #3180, opened 2025-10-06
Introduce `CODEQL_ACTION_SKIP_SARIF_UPLOAD`

redsun82:
This triggers a subset of the behavior of `CODEQL_ACTION_TEST_MODE`, specifically just skipping the SARIF upload step. This is required for our internal testing where we want the SARIF file (via `CODEQL_ACTION_DUMP_SARIF_DIR`) but don't want to actually upload it, but we don't want the rest of the behaviour of `CODEQL_ACTION_TEST_MODE` that is specific for `codeql-action` own CI checks.

This is a fresh take at https://github.com/github/codeql-action/pull/3089, with less impact on the codebase.

### Risk assessment


src/environment.ts
```diff
@@ -128,4 +128,10 @@ export enum EnvVar {
    * whether the upload is disabled. This is intended for testing and debugging purposes.
    */
   SARIF_DUMP_DIR = "CODEQL_ACTION_SARIF_DUMP_DIR",
+
+  /**
+   * Whether to skip uploading SARIF results to GitHub. Intended for testing purposes.
+   * This setting is implied but is more specific than `CODEQL_ACTION_TEST_MODE`.
```
mbg: Minor: the wording here isn't great.

```suggestion
   * This setting is more specific than `CODEQL_ACTION_TEST_MODE`, which also implies this option.
```

src/upload-lib.ts
```diff
@@ -356,18 +356,16 @@ async function uploadPayload(
...
+  // If in test mode we don't want to upload the results,
+  if (util.shouldSkipSarifUpload()) {
     const payloadSaveFile = path.join(
       actionsUtil.getTemporaryDirectory(),
       "payload.json",
     );
-    logger.info(
-      `In test mode. Results are not uploaded. Saving to ${payloadSaveFile}`,
-    );
+    logger.info(`SARIF upload disabled. Saving to ${payloadSaveFile}`);
```
mbg: Minor: This no longer explains why the SARIF upload is disabled. A simple fix might be wording it like this:

```suggestion
    logger.info(`SARIF upload disabled with an environment variable. Saving to ${payloadSaveFile}`);
```

A better fix would be to include the responsible environment variable, perhaps with:

```suggestion
    logger.info(`SARIF upload disabled by `${ isInTestMode() ? EnvVar.TEST_MODE : EnvVar.SKIP_SARIF_UPLOAD }`. Saving to ${payloadSaveFile}`);
```

src/util.ts
```diff
@@ -764,12 +764,19 @@ export function isGoodVersion(versionSpec: string) {
 /**
  * Returns whether we are in test mode. This is used by CodeQL Action PR checks.
  *
- * In test mode, we don't upload SARIF results or status reports to the GitHub API.
+ * In test mode, we several uploads (SARIF results, status reports, DBs, ...).
```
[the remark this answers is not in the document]
mbg: Copilot is correct.

src/upload-lib.ts
```diff
@@ -356,18 +356,16 @@ async function uploadPayload(
 ): Promise<string> {
   logger.info("Uploading results");
 
-  // If in test mode we don't want to upload the results
-  if (util.isInTestMode()) {
+  // If in test mode we don't want to upload the results,
```
esbena: this comment is a bit stale now, ditto at the other `isInTestMode/shouldSkipSarifUpload` switch.

mbg (commented):
Thanks, this is generally much neater! I've added a few minor comments.

It is probably worth adding a test case for this.

redsun82:
> It is probably worth adding a test case for this

@mbg: how would you go on about this?

* the PR checks set `CODEQL_ACTION_TEST_MODE`, which "hides" this new behaviour. Maybe a PR check could unset that and set `CODEQL_ACTION_SKIP_SARIF_UPLOAD` instead? Wouldn't there be some other unwanted effects in the PR checks though?
* on the unit test side, there's currently no unit tests for `uploadSpecifiedFiles`, and `CODEQL_ACTION_TEST_MODE` is not covered at all in there. Adding tests for `uploadSpecifiedFiles` would probably be nice, but maybe a tad too much for the scope of this PR?

src/util.ts
```diff
@@ -764,12 +764,25 @@ export function isGoodVersion(versionSpec: string) {
...
+ */
+export function getSarifUploadSkipReason(): string | null {
+  if (isInTestMode()) {
+    return `SARIF upload is disabled via ${EnvVar.TEST_MODE}`;
+  }
+  if (process.env[EnvVar.SKIP_SARIF_UPLOAD] === "true") {
+    return `SARIF upload is disabled via ${EnvVar.SKIP_SARIF_UPLOAD}`;
+  }
+  return null;
+}
```
mbg: I had considered something like this when I wrote my comment, but wasn't too keen on it. I also don't love using the human-readable reason phrase to mean `true` for enablement purposes. The previous version with the `boolean` function was preferable.

Also, in general, we don't use `null` in the `codeql-action` codebase (unless it comes to us from outside the codebase) and would go for `undefined` here.

mbg:
> how would you go on about this?
> on the unit test side, there's currently no unit tests [..]

Hmm yes. Unit tests for `uploadPayload` would be best. I don't think it would be too hard to set this up with mocks for the IO operations (`fs.writeFileSync` and `client.request`) and checking which ones were called for given permutations of the environment variables. That said, if there aren't any unit tests here already, then don't feel obligated to set this up as part of this PR.

src/util.ts
redsun82: I was trying to avoid something like
```
    logger.info(`SARIF upload disabled by `${ isInTestMode() ? EnvVar.TEST_MODE : EnvVar.SKIP_SARIF_UPLOAD }`. Saving to ${payloadSaveFile}`);
```
as I didn't want to expose that `util` implementation detail. So I think I'll just go with your first proposal for a generic `with an environment variable`. The users of those variables (us) probably are savvy enough for that to be enough for them.

redsun82:
> > how would you go on about this?
> > on the unit test side, there's currently no unit tests [..]
> 
> Hmm yes. Unit tests for `uploadPayload` would be best. I don't think it would be too hard to set this up with mocks for the IO operations (`fs.writeFileSync` and `client.request`) and checking which ones were called for given permutations of the environment variables. That said, if there aren't any unit tests here already, then don't feel obligated to set this up as part of this PR.

Let's make that into a follow-up PR. I'll try to see what copilot can do there.

src/environment.ts
```diff
@@ -131,7 +131,7 @@ export enum EnvVar {
 
   /**
    * Whether to skip uploading SARIF results to GitHub. Intended for testing purposes.
-   * This setting is implied by `CODEQL_ACTION_TEST_MODE`, but is more specific.
+   * This setting is implied by but is more specific than `CODEQL_ACTION_TEST_MODE`.
```
mbg: Did you intend to revert this?

src/upload-lib.ts
```diff
@@ -356,18 +356,17 @@ async function uploadPayload(
...
-  // If in test mode we don't want to upload the results
-  if (util.isInTestMode()) {
+  if (util.shouldSkipSarifUpload()) {
     const payloadSaveFile = path.join(
       actionsUtil.getTemporaryDirectory(),
       "payload.json",
     );
     logger.info(
-      `In test mode. Results are not uploaded. Saving to ${payloadSaveFile}`,
+      `SARIF upload disabled via environment variable. Saving to ${payloadSaveFile}`,
```
mbg: Minor:
```suggestion
      `SARIF upload disabled by an environment variable. Saving to ${payloadSaveFile}`,
```

src/upload-sarif-action.ts
```diff
@@ -113,8 +113,10 @@ async function run() {
     core.setOutput("sarif-ids", JSON.stringify(uploadResults));
 
     // We don't upload results in test mode, so don't wait for processing
-    if (isInTestMode()) {
-      core.debug("In test mode. Waiting for processing is disabled.");
+    if (shouldSkipSarifUpload()) {
+      core.debug(
+        "SARIF upload disabled via environment variable. Waiting for processing is disabled.",
```
mbg:
```suggestion
        "SARIF upload disabled by an environment variable. Waiting for processing is disabled.",
```

src/environment.ts
redsun82: nope :sweat_smile:

redsun82:
@mbg thanks for the review! I think I addressed everything. I also opened a follow-up with unit tests: https://github.com/github/codeql-action/pull/3185 (with a little help from copilot)

mbg (approved):
Thanks for fixing those minor issues!

PR #3184, opened 2025-10-07
Overlays: allow any build mode for Go

nickrolfe:
We have a check that a traced language can only run overlay analysis with `build-mode: none`, but Go does not currently declare support for BMN, even though it has a similar autobuild mode that will work for overlay analysis.

This commit adds a hard-coded exception to that check, allowing any build mode for Go. This is intended as a short-term solution until Go declares BMN support.

For now, it should be safe to enable for all build modes:
- We can still control rollout, via the overlay feature flags, to repos we know will work correctly.
- It allows us to see what happens for manual builds.
- We should be able to ensure that the Go extractor/autobuilder fail for unsupported build configures in such a way that we fall back to full analysis.

### Risk assessment

- **None** - I am not validating these changes specifically, but I have been testing overlay analysis for Go in general.

### DIFF
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -87860,7 +87860,13 @@ async function getOverlayDatabaseMode(codeql, repository, features, languages, s
     return nonOverlayAnalysis;
   }
   if (buildMode !== "none" /* None */ && (await Promise.all(
-    languages.map(async (l) => await codeql.isTracedLanguage(l))
+    languages.map(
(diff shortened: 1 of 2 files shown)


nickrolfe:
I've updated the PR description and code comment to mention that the Go extractor/autobuilder will aim to ensure that we fall back to full analysis for build setups that won't work with overlays.

PR #3185, opened 2025-10-07
Add unit tests for `uploadPayload`

redsun82:
### Risk assessment

None, these are only tests.

- **CI**

### DIFF
--- a/src/upload-lib.test.ts
+++ b/src/upload-lib.test.ts
@@ -3,7 +3,7 @@ import * as path from "path";
 
 import * as github from "@actions/github";
 import { HTTPError } from "@actions/tool-cache";
-import test, { ExecutionContext } from "ava";
+import test from "ava";
 import * as sinon from "sinon";
 
 import * as analyses from "./analyses";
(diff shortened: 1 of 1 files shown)


src/upload-lib.test.ts
```diff
@@ -1,13 +1,30 @@
+// Node.js built-in modules
```
mbg: These comments are unnecessary.

src/upload-lib.test.ts
```diff
@@ -851,19 +868,414 @@ test("shouldConsiderInvalidRequest returns correct recognises processing errors"
   t.false(uploadLib.shouldConsiderInvalidRequest(error3));
 });
 
+// Helper function to set up common environment variables for upload tests
+function setupUploadEnvironment(
+  tmpDir: string,
+  extraVars?: Record<string, string>,
+) {
+  const originalEnv: Record<string, string | undefined> = {};
```
mbg: Storing and restoring the environment after each test is already handled by `setupTests`.

src/upload-lib.test.ts
```diff
@@ -851,19 +868,414 @@ test("shouldConsiderInvalidRequest returns correct recognises processing errors"
...
+    }
+  };
+}
+
+// Helper function to stub common external dependencies for upload tests
+function stubUploadDependencies() {
+  sinon.stub(api, "getGitHubVersion").resolves({
+    type: GitHubVariant.DOTCOM,
+  });
+  sinon.stub(api, "getAnalysisKey").resolves("test-key");
```
mbg: This isn't the format that `getAnalysisKey` returns, which may not matter for the tests, but is unexpected.

src/upload-lib.test.ts
```diff
@@ -851,19 +868,414 @@ test("shouldConsiderInvalidRequest returns correct recognises processing errors"
...
+      t.is(result.sarifID, "dummy-sarif-id");
+      t.truthy(result.statusReport.raw_upload_size_bytes);
+      t.truthy(result.statusReport.zipped_upload_size_bytes);
+      t.is(result.statusReport.num_results_in_sarif, 0);
+
+      // Verify external dependencies were called
+      t.true(addFingerprintsStub.calledOnce);
+    } finally {
+      sinon.restore();
+      cleanupEnv();
```
mbg: As noted above, this is unnecessary because `setupTests` already installs handlers into the test framework which do this.

src/upload-lib.test.ts
```diff
@@ -851,19 +868,414 @@ test("shouldConsiderInvalidRequest returns correct recognises processing errors"
...
+    .resolves({
+      status: 200,
+      data: { id: sarifId },
+    } as any);
+  sinon.stub(api, "getApiClient").value(() => mockApiClient);
+
+  return { requestStub };
+}
+
+test("uploadSpecifiedFiles - single SARIF file", async (t) => {
```
mbg: `uploadSpecifiedFiles` performs quite a bit of work (see my comment below about being confused that you've added tests for this, rather than `uploadPayload`). "single SARIF file" doesn't really describe what this test actually checks for, other than that it uses a single SARIF file.

src/upload-lib.test.ts
```diff
@@ -851,19 +868,414 @@ test("shouldConsiderInvalidRequest returns correct recognises processing errors"
...
+    const mockSarif = createMockSarif("test-id", "TestTool");
+    fs.writeFileSync(sarifPath, JSON.stringify(mockSarif));
+
+    const cleanupEnv = setupUploadEnvironment(tmpDir, {
+      CODEQL_ACTION_SKIP_SARIF_UPLOAD: "true",
+    });
+    const { addFingerprintsStub } = stubUploadDependencies();
+
+    try {
+      const result = await uploadLib.uploadSpecifiedFiles(
```
mbg: I am confused why these tests are for `uploadSpecifiedFiles`, rather than for `uploadPayload` (which was modified in #3180).

If the tests are intentionally for `uploadSpecifiedFiles`, rather than for `uploadPayload`, then we should make sure that they comprehensively cover all that we expect `uploadSpecifiedFiles` to do.

src/upload-lib.test.ts
```diff
@@ -851,19 +868,414 @@ test("shouldConsiderInvalidRequest returns correct recognises processing errors"
...
+      );
+
+      // Verify the result
+      t.is(result.sarifID, "dummy-sarif-id");
+      t.truthy(result.statusReport.raw_upload_size_bytes);
+      t.truthy(result.statusReport.zipped_upload_size_bytes);
+      t.is(result.statusReport.num_results_in_sarif, 0);
+
+      // Verify external dependencies were called
+      t.true(addFingerprintsStub.calledOnce);
```
mbg: It's not clear to me why this test checks that this specific stub was called, rather than a more comprehensive check of what `uploadSpecifiedFiles` did.

mbg (changes requested):
I appreciate the follow-up to #3180 to add tests. I started having a look, but stopped after the first couple of test cases:

- In #3180, the main change was to `uploadPayload`, and I was expecting tests to focus around the behaviour of `uploadPayload`. While it's not currently exported, we have numerous examples of functions that are exported for testing purposes only.
- Tests for `uploadSpecifiedFiles` would of course be welcome, but they would need to focus on the behaviour of `uploadSpecifiedFiles` itself, not primarily on the behaviour of `uploadPayload`.
- The `setupUploadEnvironment` function is largely superfluous; `setupTests` already does this.
- The tests would probably benefit from using a `macro` to set up what is needed for testing `uploadSpecifiedFiles`, rather than duplicating large chunks of setup code across every test case.
- A lot of the comments are nonsensical (and show a lack of understanding of what's going on) or superfluous

src/upload-lib.test.ts
redsun82: good to know!

src/upload-lib.test.ts
redsun82: I'm always a bit reluctant to export a function just for testing, as I feel that if it wasn't exported it was an implementation detail, and tying a test to it can make refactoring harder. On the other hand, it might be ok in this case. I guess there's always a balance to strike.

redsun82:
I probably let copilot a tad too loose on this :sweat_smile:

redsun82:
@mbg I've decicided to ditch all that copilot had done and (for the moment) only add `uploadPayload` tests like you suggested. Now that function should be completely covered (though I chose not to test the different `core.warning`s logic, as that is logging only and not functional).

src/upload-lib.test.ts
```diff
@@ -867,3 +872,123 @@ function createMockSarif(id?: string, tool?: string) {
...
+      );
+
+    await options.body(t, upload, requestStub, mockData);
+  },
+  title: (providedTitle = "", options: { analysis: analyses.AnalysisConfig }) =>
+    `uploadPayload - ${options.analysis.name} - ${providedTitle}`,
+});
+
+for (const analysis of [CodeScanning, CodeQuality]) {
+  test("uploads successfully", uploadPayloadMacro, {
```
mbg: Minor: I was going to comment that it would probably make sense to include the `analysis` name in the test names, but I then saw that this already happens via the macro. I have no strong feelings on which is better. In other test files where we have top-level `for`-loops, we splice the variables into the title provided to `test`, so it might be good for consistency to do that here as well, but I can see how doing it in the macro is less repetitive.

src/upload-lib.test.ts
```diff
@@ -867,3 +872,123 @@ function createMockSarif(id?: string, tool?: string) {
...
+          );
+          t.true(fs.existsSync(payloadFile));
+
+          const savedPayload = JSON.parse(fs.readFileSync(payloadFile, "utf8"));
+          t.deepEqual(savedPayload, mockData.payload);
+        }),
+    });
+  }
+
+  test("handles error", uploadPayloadMacro, {
```
mbg: Minor: maybe a better title would be
```suggestion
  test("wraps request errors using `wrapApiConfigurationError`", uploadPayloadMacro, {
```

src/upload-lib.test.ts
```diff
@@ -867,3 +872,123 @@ function createMockSarif(id?: string, tool?: string) {
...
+          owner: string;
+          repo: string;
+          response: {
+            status: number;
+            data: { id: string };
+            headers: any;
+            url: string;
+          };
+        },
+      ) => void | Promise<void>;
```
mbg: I don't think we have this pattern with a continuation (here: `body`) for the macro to call elsewhere, but I can see why it could be useful. Were there any particular reasons why you chose to do it this way over the established pattern of having inputs/input modifiers and an expected result? (All tests but the last are expected to return a `string` so you could have a `expectedId` parameter, and you could just not use the macro for the test where you expect failure.)

src/upload-lib.test.ts
```diff
@@ -867,3 +872,123 @@ function createMockSarif(id?: string, tool?: string) {
     ],
   };
 }
+
+const uploadPayloadMacro = test.macro({
+  exec: async (
+    t: ExecutionContext<unknown>,
+    options: {
+      analysis: analyses.AnalysisConfig;
+      body: (
```
mbg: Did you have any particular reason for having this `options` object rather than having `analysis` and `body` be parameters of `exec`?

src/upload-lib.test.ts
mbg: I think it's OK, since we have done this elsewhere and `codeql-action` isn't a library used elsewhere. If we refactor the implementation, we'd probably need to adjust the tests one way or another anyway.

src/upload-lib.test.ts
redsun82: initially I had more flags that profited from being named explicitly, but I agree that at this stage it's not really needed

src/upload-lib.test.ts
redsun82: fundamentally I wanted to have the setup that the macro does in common for all the tests. I had a version of this where all the test behaviour was being driven by flags in `options` (courtesy of copilot), but then the macro code was a mess of conditionals and the tests weren't as clear. The upload skipping tests too do some different checking that I find would be a tad bad to shove in the macro code.

redsun82: Also, maybe that could be covered better by a setup function, rather than a macro?

redsun82: maybe, considering the test macro is not doing any checks after calling `body`

redsun82: yeah, I like that better :slightly_smiling_face:

PR #3203, opened 2025-10-12
Handle user errors for invalid `UserConfig`s and missing query files

mbg:
Classifies two additional types of errors as `ConfigurationError`s:

- Invalid config files
- Missing query files

### Risk assessment

### DIFF
--- a/src/config/db-config.test.ts
+++ b/src/config/db-config.test.ts
@@ -442,6 +442,25 @@ test("parseUserConfig - throws a ConfigurationError if the file is not valid YAM
   );
 });
 
+test("parseUserConfig - validation isn't picky about `query-filters`", (t) => {
+  const loggedMessages: LoggedMessage[] = [];
+  const logger = getRecordingLogger(loggedMessages);
+
+  t.notThrows(() =>
+    dbConfig.parseUserConfig(
+      logger,
+      "test",
+      `
+        query-filters:
+          - something
+          - include: foo
+          - exclude: bar
+        `,
+      true,
+    ),
+  );
+});
+
 test("parseUserConfig - throws a ConfigurationError if validation fails", (t) => {
   const loggedMessages: LoggedMessage[] = [];
   const logger = getRecordingLogger(loggedMessages);
--- a/src/db-config-schema.json
+++ b/src/db-config-schema.json
@@ -87,7 +87,8 @@
         },
         {
           "$ref": "#/definitions/IncludeQueryFilter"
-        }
+        },
+        {}
       ]
     },
     "ExcludeQueryFilter": {
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -35375,7 +35375,8 @@ var require_db_config_schema = __commonJS({
             },
             {
               $ref: "#/definitions/IncludeQueryFilter"
-            }
+            },
+            {}
           ]
         },
         ExcludeQueryFilter: {


src/config/db-config.ts
```diff
@@ -474,3 +492,43 @@ export function generateCodeScanningConfig(
...
+    const doc = yaml.load(contents);
+    const result = new jsonschema.Validator().validate(doc, schema);
+
+    if (result.errors.length > 0) {
+      throw new ConfigurationError(
+        errorMessages.getInvalidConfigFileMessage(
+          pathInput,
+          `The configuration file contained ${result.errors.length} error(s)`,
+        ),
+      );
```
henrymercer: Could we list the errors?

src/config/db-config.ts
```diff
@@ -26,13 +28,29 @@ export interface QuerySpec {
...
+}
+
 /**
  * Format of the config file supplied by the user.
  */
 export interface UserConfig {
   name?: string;
   "disable-default-queries"?: boolean;
-  queries?: QuerySpec[];
+  queries?: QueryInput[];
```
henrymercer: The config file only accepts query specs.  I double checked this here: https://github.com/github/codeql-action/actions/runs/18463233951

src/db-config-schema.json
```diff
@@ -0,0 +1,156 @@
...
+    },
+    "disable-default-queries": {
+      "type": "boolean",
+      "description": "Whether to disable default queries"
+    },
+    "queries": {
+      "type": "array",
+      "description": "List of additional queries to run",
+      "items": {
+        "$ref": "#/definitions/QueryInput"
```
henrymercer: This should be just QuerySpec

src/db-config-schema.json
```diff
@@ -0,0 +1,156 @@
...
+      }
+    },
+    "paths-ignore": {
+      "type": "array",
+      "description": "Paths to ignore during analysis",
```
henrymercer, line 24:
```suggestion
      "description": "Paths to exclude from analysis",
```

src/db-config-schema.json
```diff
@@ -0,0 +1,156 @@
...
+    "QueryFilter": {
+      "description": "Query filter that can either include or exclude queries",
+      "oneOf": [
+        {
+          "$ref": "#/definitions/ExcludeQueryFilter"
+        },
+        {
+          "$ref": "#/definitions/IncludeQueryFilter"
+        }
+      ]
```
henrymercer, line 92: It's possible to specify neither `include` nor `exclude`, and it's allowable, although unusual, to specify both an `include` and an `exclude`.  I think we should just specify that the `include` and `exclude` properties are both optional.

src/config/db-config.ts
mbg: Do you think we should include them in the exception (perhaps limited to `n`-many) or just log them?

henrymercer: I would include the first 10 say in the exception.  We cut off exceptions in telemetry so it's unlikely to cause a problem.

src/config/db-config.test.ts
```diff
@@ -391,3 +397,111 @@ test(
...
+        `,
+        true,
+      ),
+    {
+      instanceOf: ConfigurationError,
+    },
+  );
+});
+
+test("parseUserConfig - validation isn't picky about `query-filters`", (t) => {
```
henrymercer, line 445: Can you test a query filter that has both `include` and `exclude` properties?

src/db-config-schema.json
```diff
@@ -0,0 +1,145 @@
...
+      "oneOf": [
+        {
+          "$ref": "#/definitions/ExcludeQueryFilter"
+        },
+        {
+          "$ref": "#/definitions/IncludeQueryFilter"
+        },
+        {}
+      ]
+    },
```
henrymercer, line 93: I'd suggest changing this to an object with optional `include` and `exclude` properties, and updating the types in `db-config.ts` to reflect the fact we might have a filter with both `include` and `exclude` or neither `include` nor `exclude`.

PR #3204, opened 2025-10-12
Add `setup-codeql` action

mbg:
This PR takes a stab at adding a `setup-codeql` action, which just installs the CodeQL CLI. Unlike `init`, it does not initialise a database.

This doesn't change the `init` action in any way yet. I.e. the goal of this PR is not to support running `setup-codeql` followed by `init`, but just `setup-codeql` on its own with no subsequent `codeql-action` steps.

### Risk assessment

Currently, none.

### DIFF
--- a/.github/workflows/__analyze-ref-input.yml
+++ b/.github/workflows/__analyze-ref-input.yml
@@ -27,6 +27,11 @@ on:
         description: The version of Go to install
         required: false
         default: '>=1.21.0'
+      python-version:
+        type: string
+        description: The version of Python to install
+        required: false
+        default: '3.13'
   workflow_call:
     inputs:
       go-version:
@@ -34,6 +39,11 @@ on:
         description: The version of Go to install
         required: false
         default: '>=1.21.0'
+      python-version:
+        type: string
+        description: The version of Python to install
+        required: false
+        default: '3.13'
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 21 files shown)


src/setup-codeql-action.ts
```diff
@@ -0,0 +1,203 @@
...
+  getRequiredEnvParam,
+  initializeEnvironment,
+  ConfigurationError,
+  wrapError,
+  checkActionVersion,
+  getErrorMessage,
+} from "./util";
+
+/** Fields of the init status report populated when the tools source is `download`. */
+interface InitToolsDownloadFields {
```
henrymercer: AFAICT this is duplicated from `init-action.ts` — can we share the definition?

src/setup-codeql-action.ts
```diff
@@ -0,0 +1,203 @@
...
+    toolsSource = initCodeQLResult.toolsSource;
+
+    core.setOutput("codeql-path", codeql.getPath());
+    core.setOutput("codeql-version", (await codeql.getVersion()).version);
+  } catch (unwrappedError) {
+    const error = wrapError(unwrappedError);
+    core.setFailed(error.message);
+    const statusReportBase = await createStatusReportBase(
+      ActionName.SetupCodeQL,
+      error instanceof ConfigurationError ? "user-error" : "aborted",
```
henrymercer: We should use "failure" instead.  "aborted" was used for `init` to be able to distinguish between failures that happened before/after CLI setup, but here there is no need.

henrymercer (approved):

PR #3206, opened 2025-10-14
Use `uploadSarif` rather than `uploadFiles` in `analyze` action

mbg:
This replaces the two calls to `uploadFiles` in the `analyze` action with one call to `uploadSarif`. We introduced `uploadSarif` in https://github.com/github/codeql-action/pull/3167 for the `upload-sarif` action. Using `uploadSarif` here means that we no longer use different implementations of the same logic in `analyze` and `upload-sarif`.

### Risk assessment

- **Low risk:** 

- **FF**

### DIFF
--- a/src/analyze-action.ts
+++ b/src/analyze-action.ts
@@ -385,14 +385,20 @@ async function run() {
         }
       }
 
-      core.setOutput(
-        "sarif-id",
-        uploadResults[analyses.AnalysisKind.CodeScanning]?.sarifID,
-      );
-      core.setOutput(
-        "quality-sarif-id",
-        uploadResults[analyses.AnalysisKind.CodeQuality]?.sarifID,
-      );
+      // Set the SARIF id outputs only if we have results for them, to avoid
+      // having keys with empty values in the action output.
+      if (uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined) {
+        core.setOutput(
+          "sarif-id",
+          uploadResults[analyses.AnalysisKind.CodeScanning].sarifID,
+        );
+      }
+      if (uploadResults[analyses.AnalysisKind.CodeQuality] !== undefined) {
+        core.setOutput(
+          "quality-sarif-id",
+          uploadResults[analyses.AnalysisKind.CodeQuality].sarifID,
+        );
+      }
     } else {
       logger.info("Not uploading results");
     }
@@ -434,8 +440,7 @@ async function run() {
     if (util.isInTestMode()) {
       logger.debug("In test mode. Waiting for processing is disabled.");
     } else if (
-      uploadResults !== undefined &&
-      uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined &&
+      uploadResults?.[analyses.AnalysisKind.CodeScanning] !== undefined &&
       actionsUtil.getRequiredInput("wait-for-processing") === "true"
     ) {
       await uploadLib.waitForProcessing(
@@ -479,8 +484,7 @@ async function run() {
 
   if (
     runStats !== undefined &&
-    uploadResults !== undefined &&
-    uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined
+    uploadResults?.[analyses.AnalysisKind.CodeScanning] !== undefined
   ) {
     await sendStatusReport(
       startedAt,
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -96411,14 +96411,18 @@ async function run() {
           );
         }
       }
-      core14.setOutput(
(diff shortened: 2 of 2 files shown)


src/analyze-action.ts
```diff
@@ -450,13 +444,17 @@ async function run() {
     return;
   }
 
-  if (runStats && uploadResult) {
+  if (
+    runStats &&
+    uploadResults &&
+    uploadResults[analyses.AnalysisKind.CodeScanning]
+  ) {
```
esbena, line 488: Minor consistency.
Above we have:
```ts
      uploadResults !== undefined &&
      uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined &&
      actionsUtil.getRequiredInput("wait-for-processing") === "true"	      actionsUtil.getRequiredInput("wait-for-processing") === "true"
```
where we check explicitly against `undefined`, here we just use the truthiness. Either is fine, but since both styles were added in this PR I figured it could be called out.

esbena (dismissed):
LGTM, with one optional code style comment.

esbena (dismissed):

src/analyze-action.ts
```diff
@@ -450,13 +444,17 @@ async function run() {
     return;
   }
 
-  if (runStats && uploadResult) {
+  if (
+    runStats !== undefined &&
+    uploadResults !== undefined &&
+    uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined
```
henrymercer:
```suggestion
    uploadResults?[analyses.AnalysisKind.CodeScanning] !== undefined
```

src/analyze-action.ts
```diff
@@ -408,12 +401,13 @@ async function run() {
     if (util.isInTestMode()) {
       logger.debug("In test mode. Waiting for processing is disabled.");
     } else if (
-      uploadResult !== undefined &&
+      uploadResults !== undefined &&
+      uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined &&
```
henrymercer:
```suggestion
      uploadResults[analyses.AnalysisKind.CodeScanning] !== undefined &&
```

src/analyze-action.ts
```diff
@@ -343,30 +341,25 @@ async function run() {
...
+      );
+
+      core.setOutput(
+        "sarif-id",
+        uploadResults[analyses.AnalysisKind.CodeScanning]?.sarifID,
+      );
+      core.setOutput(
+        "quality-sarif-id",
+        uploadResults[analyses.AnalysisKind.CodeQuality]?.sarifID,
+      );
```
henrymercer: Perhaps only call `setOutput` if the value is not `undefined`?

src/analyze-action.ts
mbg: TypeScript doesn't like that

src/analyze-action.ts
mbg: TypeScript doesn't like this either

src/analyze-action.ts
mbg: `setOutput` converts both `undefined` and `null` to `''`, which I think is probably fine here since it should be equivalent to not calling `setOutput`, but I can change it if there's a reason for not doing that I am missing.

henrymercer: Presumably `outputs` would be `{ "quality-sarif-id": "" }` as opposed to `{}`?  Or are empty string outputs disallowed?

src/analyze-action.ts
henrymercer: Sorry my mistake trying to write correct code in the comment box :)

```suggestion
    uploadResults?.[analyses.AnalysisKind.CodeScanning] !== undefined
```

src/analyze-action.ts
henrymercer:
```suggestion
      uploadResults?.[analyses.AnalysisKind.CodeScanning] !== undefined &&
```

src/analyze-action.ts
mbg: Ah I thought I had tried that (and corrected for your typo), but I guess not. This does work, but I am not sure I find the `?.` in front of the indexing particularly readable here.

henrymercer: I prefer it, but the linter doesn't have an opinion about it, so feel free to take it or leave it.

src/analyze-action.ts
mbg: I tried this out in a test repo and that does seem to be what's happening in that case. I will change this. Thanks for persisting on this point.

src/analyze-action.ts
mbg: I have changed it, now that the linter has an opinion about it 😄

PR #3211, opened 2025-10-17
Make analysis kinds available for `starting` status report

mbg:
This PR changes the `init` action so that the `analysis-kinds` input is processed before the `starting` status report is sent. This allows us to include information about the enabled analysis kinds in that report.

One aspect of this that we may want to think about is that, if the `analysis-kinds` input is invalid, we now throw a `ConfigurationError` for that before the `starting` report is sent. This will still get caught and result in a `user-error` status report, but without a matching `starting` status report.

Alternatives: 

- We could catch `ConfigurationError`s thrown by `initAnalysisKinds` and re-throw them after the `starting` status report has been sent to maintain the normal order of operations.
- We could ignore the exception and then re-run `initAnalysisKinds` at the normal time.

### Risk assessment


henrymercer (approved):
Your alternative suggestion of ignoring the exception and then re-running initAnalysisKinds at the normal time seems more robust, but given that incorrect analysis kinds should be pretty rare, this seems OK too.

henrymercer (approved):
Looks reasonable

PR #3218, opened 2025-10-20
Add experimental functionality for labelling PRs by their size

henrymercer:
Label PRs by their size to give a quick assessment of how difficult they might be to review.

### Risk assessment

- Test only

### DIFF
--- a/.github/workflows/label-pr-size.yml
+++ b/.github/workflows/label-pr-size.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Run sizeup
-        uses: lerebear/sizeup-action@b7beb3dd273e36039e16e48e7bc690c189e61951
+        uses: lerebear/sizeup-action@b7beb3dd273e36039e16e48e7bc690c189e61951 # 0.8.12
         with:
           token: "${{ secrets.GITHUB_TOKEN }}"
           configuration-file-path: ".github/sizeup.yml"


.github/workflows/label-pr-size.yml
```diff
@@ -0,0 +1,26 @@
+name: Label PR with size
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - edited
+      - ready_for_review
```
mbg, line 10: We have other workflows with these (or mostly these) triggers. Could we add this as a job to e.g. `pr-checks.yml`?

.github/workflows/label-pr-size.yml
```diff
@@ -0,0 +1,26 @@
...
+  pull-requests: write
+
+jobs:
+  sizeup:
+    name: Label PR with size
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Run sizeup
+        uses: lerebear/sizeup-action@b7beb3dd273e36039e16e48e7bc690c189e61951
```
mbg: Should we document what this commit pins?
```suggestion
        uses: lerebear/sizeup-action@b7beb3dd273e36039e16e48e7bc690c189e61951 # 0.8.12
```

.github/workflows/label-pr-size.yml
henrymercer: The PR checks also trigger on `pull_request` and `workflow_dispatch`.  We could filter out those triggers with an if but I'm not sure of the benefit.

mbg (approved):

PR #3097, opened 2025-09-10
Dump soon to be uploaded SARIF on request

redsun82:
This introduces a new internal environment variable flag (`CODEQL_ACTION_SARIF_DUMP_DIR`) that, when set to `true`, causes the SARIF file that will be uploaded to be dumped to the specified directory. The filename will be `upload.sarif` or `upload.quality.sarif` depending on the upload target.

### Risk assessment

For internal use only. Please select the risk level of this change:

- **Low risk:** Changes are fully under feature flags, or have been fully tested and validated in pre-production environments and are highly observable, or are documentation or test only.

### Merge / deployment checklist

- Confirm this change is backwards compatible with existing workflows.
- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.
- Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary.

PR #3222, opened 2025-10-21
Perform SARIF post-processing independently of upload

mbg:
This PR is currently based on #3206, which should be merged first.

Conceptually, this PR follows on from https://github.com/github/codeql-action/pull/3097 to allow processed SARIF files to be written to disk, if required.

This is accomplished by:

- Extracting the post-processing and uploading parts of `uploadSpecifiedFiles` into separate functions.
- Replacing the call to `uploadSpecifiedFiles` in `uploadSarif` with calls to the new `postProcessSarifFiles` and `uploadProcessedFiles`.
- Moving the `UploadKind` check into `uploadSarif` and making the call to `uploadProcessedFiles` conditional on it. **Important**: This is a change in behaviour, which means that we now always post-process SARIF files, even if `upload` is not `always`.
- Adding a new `post-process-output` input to the `analyze` action which, if set, causes `writeProcessedFiles` to write the processed SARIF files to disk between the calls to `postProcessSarifFiles` and `uploadProcessedFiles`.

### Risk assessment

### DIFF
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -776,8 +776,8 @@ export async function writePostProcessedFiles(
   // If there's an explicit input, use that. Otherwise, use the value from the environment variable.
   const outputPath = pathInput || process.env[EnvVar.SARIF_DUMP_DIR];
 
-  // If we have an output path, write the SARIF file to it.
-  if (outputPath !== undefined) {
+  // If we have a non-empty output path, write the SARIF file to it.
+  if (outputPath !== undefined && outputPath.trim() !== "") {
     dumpSarifFile(
       JSON.stringify(postProcessingResults.sarif),
       outputPath,
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -95936,7 +95936,7 @@ async function postProcessSarifFiles(logger, features, checkoutPath, sarifPaths,
 }
 async function writePostProcessedFiles(logger, pathInput, uploadTarget, postProcessingResults) {
   const outputPath = pathInput || process.env["CODEQL_ACTION_SARIF_DUMP_DIR" /* SARIF_DUMP_DIR */];
-  if (outputPath !== void 0) {
+  if (outputPath !== void 0 && outputPath.trim() !== "") {
     dumpSarifFile(
       JSON.stringify(postProcessingResults.sarif),
       outputPath,
--- a/lib/upload-lib.js
+++ b/lib/upload-lib.js
@@ -92750,7 +92750,7 @@ async function postProcessSarifFiles(logger, features, checkoutPath, sarifPaths,
 }
 async function writePostProcessedFiles(logger, pathInput, uploadTarget, postProcessingResults) {
   const outputPath = pathInput || process.env["CODEQL_ACTION_SARIF_DUMP_DIR" /* SARIF_DUMP_DIR */];
-  if (outputPath !== void 0) {
+  if (outputPath !== void 0 && outputPath.trim() !== "") {
     dumpSarifFile(
       JSON.stringify(postProcessingResults.sarif),
       outputPath,
--- a/lib/upload-sarif-action.js
+++ b/lib/upload-sarif-action.js
@@ -93406,7 +93406,7 @@ async function postProcessSarifFiles(logger, features, checkoutPath, sarifPaths,
 }
 async function writePostProcessedFiles(logger, pathInput, uploadTarget, postProcessingResults) {
   const outputPath = pathInput || process.env["CODEQL_ACTION_SARIF_DUMP_DIR" /* SARIF_DUMP_DIR */];
-  if (outputPath !== void 0) {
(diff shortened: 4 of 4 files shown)


src/upload-lib.ts
```diff
@@ -807,6 +805,33 @@ export async function uploadSpecifiedFiles(
...
+ * Uploads the
+ *
+ * @param logger The logger to use.
+ * @param checkoutPath The path at which the repository was checked out.
+ * @param uploadTarget The analysis configuration.
+ * @param processingResults The results of post-processing SARIF files.
+ *
+ * @returns The results of uploading the `processingResults` to `uploadTarget`.
+ */
+export async function uploadProcessedFiles(
```
esbena: "Post-processed" and "processed" are used interchangeably here. Should we converge on one or the other throughout the entire PR?

Samples:
- `post-processing`
- `uploadProcessedFiles`
- `processingResults`

analyze/action.yml
```diff
@@ -70,6 +70,12 @@ inputs:
     description: Whether to upload the resulting CodeQL database
     required: false
     default: "true"
+  post-process-output:
```
esbena: Is the order of these input declarations chosen strongly based on discoverability for the user? If not, then I'd like this to be right next to the `output` input declared near the top in the name of coherence.

Second, and I'm sorry if this naming has been discussed at length in related PRs earlier: `post-process-output` reads like a boolean toggle, not a path. `post-processed-output` sounds more like a path. Additionally we always do the post-processing regardless of wether we save them to disk..

henrymercer: > `post-processed-output` sounds more like a path.

I agree this would be a better name.

src/analyze-action.ts
```diff
@@ -339,17 +339,21 @@ async function run() {
...
-    const uploadInput = actionsUtil.getOptionalInput("upload");
-    if (runStats && actionsUtil.getUploadValue(uploadInput) === "always") {
+    const uploadKind = actionsUtil.getUploadValue(
+      actionsUtil.getOptionalInput("upload"),
+    );
+    if (runStats) {
       const checkoutPath = actionsUtil.getRequiredInput("checkout_path");
       const category = actionsUtil.getOptionalInput("category");
 
       uploadResults = await uploadSarif(
```
henrymercer: Now that this function doesn't necessarily upload the files, should we rename it to `processAndUploadSarif`?

src/upload-lib.ts
```diff
@@ -768,32 +756,135 @@ export async function uploadSpecifiedFiles(
...
+  return uploadProcessedFiles(
+    logger,
+    checkoutPath,
+    uploadTarget,
+    processingResults,
+  );
+}
+
+/**
+ * Uploads the
```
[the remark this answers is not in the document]
henrymercer: Copilot is correct.

src/upload-sarif.test.ts
```diff
@@ -183,3 +202,56 @@ test(
     },
   },
 );
+
+test("uploadSarif doesn't upload if `upload` != `always`", async (t) => {
```
henrymercer: Minor: Can we rename this test?  The test tests `upload == "never"`, not `upload != "always"`.

analyze/action.yml
mbg: > Second, and I'm sorry if this naming has been discussed at length in related PRs earlier: `post-process-output` reads like a boolean toggle, not a path. post-processed-output sounds more like a path. Additionally we always do the post-processing regardless of wether we save them to disk..

Ironically, I think this kind of bad naming is actually something I complained about on `@redsun82`'s PR previously, which he then changed to something better. I agree that this isn't a good name and have changed it to `processed-sarif-path` for now. That makes it consistent with other `path` inputs (other than `output` here). I don't think think that `post-processed-output` is really any better than `post-process-output`. 

> Is the order of these input declarations chosen strongly based on discoverability for the user? If not, then I'd like this to be right next to the output input declared near the top in the name of coherence.

Generally, I think the order is roughly based on how relevant the input is to a user. (See e.g. that we have the ones that are automatically populated or used internally only at the bottom.)

src/upload-lib.ts
mbg: I have changed everything to be consistently based on "post-processing" rather than "processing".

src/upload-lib.ts
```diff
@@ -768,32 +756,134 @@ export async function uploadSpecifiedFiles(
...
+  logger: Logger,
+  pathInput: string | undefined,
+  uploadTarget: analyses.AnalysisConfig,
+  postProcessingResults: PostProcessingResults,
+) {
+  // If there's an explicit input, use that. Otherwise, use the value from the environment variable.
+  const outputPath = pathInput || process.env[EnvVar.SARIF_DUMP_DIR];
+
+  // If we have an output path, write the SARIF file to it.
+  if (outputPath !== undefined) {
```
henrymercer: Perhaps weaken this check — if the environment variable is set to the empty string, we probably want to avoid producing an output.

PR #3232, opened 2025-10-24
Ensure uniqueness of overlay-base database cache keys

kaspersv:
This PR adds a workflow run id and workflow attempt id to the overlay base database cache key postfix to ensure uniqueness of the cache keys. The cache key prefix used for restoring overlay-base databases remains unchanged. 

The motivation for the change is that currently if a successful overlay analysis run on the default branch is retried or a repo contains a scheduled CodeQL workflow on the default branch that runs twice without any changes on the default branch in between, then uploading of the overlay-base database will fail due to a cache key collision. 

The first commit which adds a unit test for overlay-base cache keys was cherry-picked from https://github.com/github/codeql-action/pull/3158.

### Risk assessment

This individual change is not under feature flag, but it only affects overlay analysis, which is under feature flag and can be disabled in case of unexpected negative impact.

### DIFF
--- a/src/overlay-database-utils.ts
+++ b/src/overlay-database-utils.ts
@@ -16,6 +16,7 @@ import { type Config } from "./config-utils";
 import { getCommitOid, getFileOidsUnderPath } from "./git-utils";
 import { Logger, withGroupAsync } from "./logging";
 import {
+  getErrorMessage,
   isInTestMode,
   tryGetFolderBytes,
   waitForResultWithTimeLimit,
@@ -276,6 +277,7 @@ export async function uploadOverlayBaseDatabaseToCache(
     config,
     codeQlVersion,
     checkoutPath,
+    logger,
   );
   logger.info(
     `Uploading overlay-base database to Actions cache with key ${cacheSaveKey}`,
@@ -457,9 +459,18 @@ export async function getCacheSaveKey(
   config: Config,
   codeQlVersion: string,
   checkoutPath: string,
+  logger: Logger,
 ): Promise<string> {
-  const runId = getWorkflowRunID();
-  const attemptId = getWorkflowRunAttempt();
+  let runId = 1;
+  let attemptId = 1;
+  try {
+    runId = getWorkflowRunID();
+    attemptId = getWorkflowRunAttempt();
+  } catch (e) {
+    logger.warning(
+      `Failed to get workflow run ID or attempt ID. Reason: ${getErrorMessage(e)}`,
+    );
+  }
   const sha = await getCommitOid(checkoutPath);
   const restoreKeyPrefix = await getCacheRestoreKeyPrefix(
     config,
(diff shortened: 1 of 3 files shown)


src/overlay-database-utils.ts
```diff
@@ -448,17 +453,19 @@ export async function downloadOverlayBaseDatabaseFromCache(
...
  * commit SHA) and the commit SHA of the current checkout.
  */
-async function getCacheSaveKey(
+export async function getCacheSaveKey(
   config: Config,
   codeQlVersion: string,
   checkoutPath: string,
 ): Promise<string> {
+  const runId = getWorkflowRunID();
+  const attemptId = getWorkflowRunAttempt();
```
mbg: These two functions may throw errors if the corresponding environment variables are not defined. I don't see any reason why that would be a concern since they should always be set in Actions, but I thought I'd mention it.

src/overlay-database-utils.ts
```diff
@@ -448,17 +453,19 @@ export async function downloadOverlayBaseDatabaseFromCache(
...
   const sha = await getCommitOid(checkoutPath);
   const restoreKeyPrefix = await getCacheRestoreKeyPrefix(
     config,
     codeQlVersion,
   );
-  return `${restoreKeyPrefix}${sha}`;
+  return `${restoreKeyPrefix}${runId}-${attemptId}-${sha}`;
```
mbg: `sha` will be stable for a given commit, but `runId` and `attemptId` may vary. Would you be happy with a `${sha}-${runId}-${attemptId}` order for this to maintain the typical order from least specific to most specific?

mbg (approved):
This looks good to me. I added two comments: one is a very minor note about about the behaviour of `getWorkflowRunID` and `getWorkflowRunAttempt` if it matters to you, otherwise you can ignore it. The other is about the order of the elements in the cache key. Both points are not blocking.

src/overlay-database-utils.ts
kaspersv: Yes, let us follow the typical order. Fixed in commit 3.

src/overlay-database-utils.ts
kaspersv: It is a corner-case of a corner-case, but it doesn't add much code to handle it. I've added error handling in commit 4, such that we default to 1 in the unlikely case these functions fail.

src/overlay-database-utils.ts
```diff
@@ -448,17 +455,28 @@ export async function downloadOverlayBaseDatabaseFromCache(
...
+  let runId = 1;
+  let attemptId = 1;
+  try {
+    runId = getWorkflowRunID();
+    attemptId = getWorkflowRunAttempt();
+  } catch (e) {
+    logger.warning(
+      `Failed to get workflow run ID or attempt ID. Reason: ${getErrorMessage(e)}`,
+    );
+  }
```
mbg, line 473: Again, I still don't think this would ever be an issue, but I think I'd prefer throwing an error rather than defaulting to `1`. Alternatively omit the `runId` and `attemptId` from the cache key.

Not blocking though - I am happy to approve this as is. This can always be tweaked later.

PR #3233, opened 2025-10-24
Add `getOptionalEnvVar` helper

mbg:
We have a helper function for inputs which gets the corresponding value, if it is set and not empty. We don't have this for environment variables. This PR adds `getOptionalEnvVar` to address that.

### Risk assessment

### DIFF
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -774,10 +774,10 @@ export async function writePostProcessedFiles(
   postProcessingResults: PostProcessingResults,
 ) {
   // If there's an explicit input, use that. Otherwise, use the value from the environment variable.
-  const outputPath = pathInput || process.env[EnvVar.SARIF_DUMP_DIR];
+  const outputPath = pathInput || util.getOptionalEnvVar(EnvVar.SARIF_DUMP_DIR);
 
   // If we have a non-empty output path, write the SARIF file to it.
-  if (outputPath !== undefined && outputPath.trim() !== "") {
+  if (outputPath !== undefined) {
     dumpSarifFile(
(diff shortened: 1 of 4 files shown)


src/upload-lib.ts
```diff
@@ -774,10 +774,11 @@ export async function writePostProcessedFiles(
   postProcessingResults: PostProcessingResults,
 ) {
   // If there's an explicit input, use that. Otherwise, use the value from the environment variable.
-  const outputPath = pathInput || process.env[EnvVar.SARIF_DUMP_DIR];
+  const outputPath: string | undefined =
+    pathInput || util.getOptionalEnvVar[EnvVar.SARIF_DUMP_DIR];
```
henrymercer:
```suggestion
    pathInput || util.getOptionalEnvVar(EnvVar.SARIF_DUMP_DIR);
```

src/util.ts
```diff
@@ -673,6 +673,17 @@ export function getRequiredEnvParam(paramName: string): string {
   return value;
 }
 
+/**
+ * Get an environment variable, but return `undefined` if it is not set or empty.
+ */
+export function getOptionalEnvVar(paramName: string): string | undefined {
+  const value = process.env[paramName];
+  if (value?.trim().length === 0) {
```
henrymercer, line 681: Minor: this trims but `getRequiredEnvParam` doesn't — should we be consistent?

mbg: Let's be inconsistent for now. Mainly because this would be a stricter requirement for `getOptionalEnvVar` than what we currently have and `getOptionalEnvVar` is only used in one (new) place for now where we don't risk breaking anything.

PR #3238, opened 2025-10-27
Move diff-range computation into utils

kaspersv:
This PR moves the diff-range computations and associated tests into `diff-informed-analysis-utils` as preparation for a future PR where the diff-range computation will be performed during the `init` action instead of the `analyze` action. The PR is intended to be a pure refactoring, without any observable effects. 

The two commits are derived from splitting the first commit from https://github.com/github/codeql-action/pull/3192. 

Diff-informed analysis is guarded by the `diff_informed_queries` feature flag. The `diff_informed_queries` flag is already fully rolled out. 

### Risk assessment


src/diff-informed-analysis-utils.ts
```diff
@@ -93,3 +106,174 @@ export function readDiffRangesJsonFile(
...
+
+function getDiffRanges(
+  fileDiff: FileDiff,
+  logger: Logger,
+): DiffThunkRange[] | undefined {
+  // Diff-informed queries expect the file path to be absolute. CodeQL always
+  // uses forward slashes as the path separator, so on Windows we need to
+  // replace any backslashes with forward slashes.
+  const filename = path
+    .join(actionsUtil.getRequiredInput("checkout_path"), fileDiff.filename)
```
mbg, line 198: I note that this input is specific to the `analyze` and `upload-sarif` actions, so this will need some kind of change if you are planning to call it from the `init` action in a follow-up change.

kaspersv: Thanks for the pointer. I will keep that in mind for the next changes.

kaspersv:
> I am wondering if it would be better to move this to somewhere like api-client.ts though, since there isn't really anything specific to diff-informed queries.

I think an argument can be made both ways. It is somewhat specific to diff-informed analysis in that `getDiffRanges` (and consequently `getPullRequestEditedDiffRanges`) is designed to return the information needed specifically for diff-informed analysis rather than all the information returned by `compareCommitsWithBasehead`. In particular, diff-informed analysis currently works at a file-level granularity and therefore the functions don't return any column information even though it could be extracted.

I am happy to move the functions to `api-client.ts` if you prefer.

mbg:
No strong feelings. I am happy that some parts are specific to diff-informed analysis and, while there might be parts we could extract for more general usage, we can always do that if/when we actually need them for something else. Feel free to merge this as-is.

PR #3240, opened 2025-10-27
Support skipping workflow validation

mbg:
For internal use, we have CodeQL workflows that we know are valid and don't require validation on every run. This PR changes workflow validation so that it is always skipped for `dynamic` workflows. It also adds an environment variable which allows us to skip workflow validation. 

This PR also slightly refactors the check into a function in `workflow.ts` and adds test coverage.

### Risk assessment

### DIFF
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -92218,7 +92218,7 @@ async function checkWorkflow(logger, codeql) {
     if (validateWorkflowResult === void 0) {
       logger.info("Detected no issues with the code scanning workflow.");
     } else {
-      logger.warning(
+      logger.debug(
         `Unable to validate code scanning workflow: ${validateWorkflowResult}`
       );
     }
--- a/src/workflow.ts
+++ b/src/workflow.ts
@@ -486,7 +486,7 @@ export async function checkWorkflow(logger: Logger, codeql: CodeQL) {
     if (validateWorkflowResult === undefined) {
       logger.info("Detected no issues with the code scanning workflow.");
     } else {
-      logger.warning(
+      logger.debug(
         `Unable to validate code scanning workflow: ${validateWorkflowResult}`,
       );
     }


src/init.ts
```diff
@@ -9,13 +10,37 @@ import { getOptionalInput, isSelfHostedRunner } from "./actions-util";
...
 import * as util from "./util";
+import { validateWorkflow } from "./workflow";
+
+/**
+ * A wrapper around `validateWorkflow` which reports the outcome.
+ *
+ * @param logger The logger to use.
+ * @param codeql The CodeQL instance.
+ */
+export async function checkWorkflow(logger: Logger, codeql: CodeQL) {
```
henrymercer: Minor: It would be clearer to have one exported function for validating the workflow.  For instance, we could move this to `workflow.ts` and make `validateWorkflow` private.

src/init.ts
```diff
@@ -9,13 +10,37 @@ import { getOptionalInput, isSelfHostedRunner } from "./actions-util";
...
+  // Check the workflow for problems, unless `SKIP_WORKFLOW_VALIDATION` is `true`.
+  if (process.env[EnvVar.SKIP_WORKFLOW_VALIDATION] !== "true") {
```
henrymercer: This is fine, but we could also just not validate workflows triggered on `dynamic` events — these must have originated from an internally maintained workflow.

src/init.ts
```diff
@@ -1,21 +1,54 @@
...
+    process.env[EnvVar.SKIP_WORKFLOW_VALIDATION] !== "true"
+  ) {
+    core.startGroup("Validating workflow");
+    const validateWorkflowResult = await validateWorkflow(codeql, logger);
+    if (validateWorkflowResult === undefined) {
+      logger.info("Detected no issues with the code scanning workflow.");
+    } else {
+      logger.warning(
+        `Unable to validate code scanning workflow: ${validateWorkflowResult}`,
+      );
```
henrymercer: Perhaps this should be an info or debug log?  Getting the workflow is a best effort operation, and it's not a huge problem to skip validation.

src/init.ts
mbg: I had initially planned to do that, but then reconsidered in order not drag more dependencies into `workflow.ts`, but I think it already depends on all we need, so there's probably no reason not to.

src/init.ts
mbg: This was a warning-level message already before my changes. I am not opposed to changing it as part of this PR though.

henrymercer: Yeah, I'm advocating for changing it.  But it's minor.

PR #3247, opened 2025-10-28
Check disk usage using Node.js API

henrymercer:
This API was introduced in Node.js 18.  Let's use it instead of the `check-disk-space` package.

### Risk assessment


src/util.ts
```diff
@@ -1108,15 +1108,17 @@ export async function checkDiskUsage(
...
       getRequiredEnvParam("GITHUB_WORKSPACE"),
     );
-    const mbInBytes = 1024 * 1024;
-    const gbInBytes = 1024 * 1024 * 1024;
-    if (diskUsage.free < 2 * gbInBytes) {
+
+    const blockSizeInBytes = diskUsage.bsize;
+    const numBlocksPerMb = (1024 * 1024) / blockSizeInBytes;
+    const numBlocksPerGb = (1024 * 1024 * 1024) / blockSizeInBytes;
+    if (diskUsage.bavail < 2 * numBlocksPerGb) {
```
esbena, line 1109: I'm surprised that we have to go through the block computation, and that there isn't a more direct measure in the API. But 🤷🏻

src/util.ts
```diff
@@ -1125,8 +1127,8 @@ export async function checkDiskUsage(
       core.exportVariable(EnvVar.HAS_WARNED_ABOUT_DISK_SPACE, "true");
     }
     return {
-      numAvailableBytes: diskUsage.free,
-      numTotalBytes: diskUsage.size,
+      numAvailableBytes: diskUsage.bavail * blockSizeInBytes,
+      numTotalBytes: diskUsage.blocks * blockSizeInBytes,
     };
   } catch (error) {
     logger.warning(
```
mbg, line 1125: In a similar spirit to the log-level change we made in #3240, should we change this from a warning to a debug message? This is only really a best effort, so it may be better to avoid the workflow annotation that results from this.

src/util.ts
```diff
@@ -1108,15 +1108,17 @@ export async function checkDiskUsage(
       return undefined;
     }
```
mbg: As discussed elsewhere, we could consider removing the SIP check now.

henrymercer:
I compared https://github.com/github/codeql-action/actions/runs/18873645058/job/53859006836 and https://github.com/github/codeql-action/actions/runs/18844897855, and the numbers look similar.

src/util.ts
henrymercer: I think while we expect the workflow validation to fail in a bunch of circumstances, this disk space check shouldn't fail.  So I'd suggest we keep it as a warning.

src/util.ts
henrymercer: Done just before you reviewed :)

src/util.ts
mbg: No strong feelings, but I could imagine that this could also fail with weird filesystem / permission setups, especially on self-hosted runners.

src/util.test.ts
```diff
@@ -534,3 +534,12 @@ test("getCgroupCpuCountFromCpus returns undefined if the CPU file exists but is
...
   });
 });
+
+test("checkDiskUsage succeeds and produces positive numbers", async (t) => {
+  const diskUsage = await util.checkDiskUsage(getRunnerLogger(true));
+  t.true(
+    diskUsage?.numAvailableBytes !== undefined &&
+      diskUsage.numAvailableBytes > 0,
+  );
+  t.true(diskUsage?.numTotalBytes !== undefined && diskUsage.numTotalBytes > 0);
```
mbg: Minor, but I quite like `t.truthy` in an `if` condition that guards the check on the property. The `&&` makes it harder to debug which sub-condition failed.

mbg (dismissed):
Thanks for adding a unit test!

src/util.test.ts
henrymercer: It's hard to write because we can't compare `undefined` with 0.  I'll make two separate assertions.

src/util.test.ts
```diff
@@ -534,3 +534,11 @@ test("getCgroupCpuCountFromCpus returns undefined if the CPU file exists but is
     );
   });
 });
+
+test("checkDiskUsage succeeds and produces positive numbers", async (t) => {
+  process.env["GITHUB_WORKSPACE"] = os.tmpdir();
+  const diskUsage = await util.checkDiskUsage(getRunnerLogger(true));
+  t.true(diskUsage !== undefined);
+  t.true(diskUsage!.numAvailableBytes > 0);
+  t.true(diskUsage!.numTotalBytes > 0);
```
mbg: This is fine, but I had this in mind:
```suggestion
  if (t.truthy(diskUsage)) {
    t.true(diskUsage.numAvailableBytes > 0);
    t.true(diskUsage.numTotalBytes > 0);
  }
```
Since the condition should assert that `diskUsage` is not `undefined`.

PR #3250, opened 2025-10-28
Use Node `fs` APIs instead of `del`

henrymercer:
### Risk assessment


src/util.ts
```diff
@@ -1263,19 +1262,13 @@ export async function checkSipEnablement(
   }
 }
 
-export async function cleanUpGlob(glob: string, name: string, logger: Logger) {
+export async function cleanUpPath(file: string, name: string, logger: Logger) {
```
esbena, line 1258: On the change from `Glob` to `Path`: Have you somehow confirmed that the callers of this renamed function do not have trailing `*` or something breaking like that? (from the diff, it sounds that `del` supports that)

esbena (commented):
LGTM with a very minor semantic concern.

src/util.ts
henrymercer: Yes, we always pass a specific file or directory.

PR #3251, opened 2025-10-28
Turn enablement errors into configuration errors

mbg:
This PR adds more cases to `wrapApiConfigurationError` where we want specific errors to be treated as `ConfigurationError`s.

### Risk assessment

### DIFF
--- a/src/api-client.ts
+++ b/src/api-client.ts
@@ -318,7 +318,7 @@ export function wrapApiConfigurationError(e: unknown) {
         "Please check that your token is valid and has the required permissions: contents: read, security-events: write",
       );
     }
-    if (isEnablementError(httpError.message)) {
+    if (httpError.status === 403 && isEnablementError(httpError.message)) {
       return new ConfigurationError(
         getFeatureEnablementError(httpError.message),
       );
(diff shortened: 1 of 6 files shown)


src/api-client.ts
```diff
@@ -304,6 +312,11 @@ export function wrapApiConfigurationError(e: unknown) {
         "Please check that your token is valid and has the required permissions: contents: read, security-events: write",
       );
     }
+    if (isEnablementError(httpError.message)) {
```
esbena: do the errors have a fixed `httpError.status` code?

src/api-client.test.ts
```diff
@@ -169,4 +169,39 @@ test("wrapApiConfigurationError correctly wraps specific configuration errors",
     res,
     new util.ConfigurationError("Resource not accessible by integration"),
   );
+
+  // Enablement errors.
```
esbena, line 173: I find the test messages a bit brittle since they do exact matching on the user-facing texts, but I find it unlikely to result in significant churn, so 👍🏻

src/api-client.ts
mbg: These should always be 403 errors. I have added a check for that to this condition.

src/api-client.test.ts
mbg: I have wrapped this up in a function to avoid the duplication in case we want to change it down the line. Ideally, I would have put it in `error-messages.ts` but that currently (transitively) depends on `api-client.ts` and I didn't want to refactor that as part of this PR so I have left a TODO comment for it.

PR #3279, opened 2025-11-09
Linter: check JSDoc parameter names exist

mbg:
Adds a linter rule (using the `eslint-plugin-jsdoc` plugin for `eslint`) that checks JSDoc comment parameter names to make sure that the parameters exist.

This doesn't enforce that JSDoc comments must exist or that they must be up-to-date.

I have fixed the few instances where we had outdated JSDoc comments and this rule was violated.

### Risk assessment

Dev only.

I ran the linter and it will run as part of CI.

N/A.


henrymercer (approved):
❤️

PR #3286, opened 2025-11-13
C#: Cache temporary dependency directory for BMN

mbg:
Follow-up from #3117. This PR extends the paths we cache for C# BMN with a similar solution as for Java BMN by configuring a particular, stable path for the extractor to store dependencies in.

There are four key parts to this:

- If the FF for this feature is enabled, we include an extra path of our choosing in the cache for C#. Note that Actions considers these paths when restoring caches, so no existing caches are compatible with the feature enabled. 
- If the FF is enabled, we include it in the feature hash that is included in the cache key.
- If we are analysing C# in BMN and the FF is enabled, we set a suitable environment variable for the C# extractor. Setting the environment variable for versions of the extractor which do not support it has no effect, so we don't need to check the CLI features / version.
- We clean up the directory at the end of an analysis, after we may have stored it in a cache.

This change depends on support for the `CODEQL_EXTRACTOR_CSHARP_OPTION_BUILDLESS_DEPENDENCY_DIR` environment variable in the C# extractor, but is designed to be safe to merge as-is before that is available.

### Risk assessment


michaelnebel:
@mbg : If the action only requires being able to communicate the dependency directory via an environment variable - then maybe we shouldn't expose the functionality as an extractor option?

mbg:
> @mbg : If the action only requires being able to communicate the dependency directory via an environment variable - then maybe we shouldn't expose the functionality as an extractor option?

I have no strong feelings about this and just mirrored what we did for the Java extractor. IIRC the team there explicitly suggested making it an extractor option.

Personally, I see no harm in it being an extractor option, but I am happy for you to make that decision and just let me know what environment variable we should set here

michaelnebel:
> > @mbg : If the action only requires being able to communicate the dependency directory via an environment variable - then maybe we shouldn't expose the functionality as an extractor option?
> 
> I have no strong feelings about this and just mirrored what we did for the Java extractor. IIRC the team there explicitly suggested making it an extractor option.
> 
> Personally, I see no harm in it being an extractor option, but I am happy for you to make that decision and just let me know what environment variable we should set here

Let's do the same as for Java for consistency.

michaelnebel (dismissed):
Looks plausible to me, but it is probably a good idea to a get a review from someone that knows more about the actions code.

src/analyze-action-env.test.ts
```diff
@@ -74,11 +74,20 @@ test("analyze action with RAM & threads from environment variables", async (t) =
...
+        "--ram=4992",
+      ),
+    );
+    t.assert(
+      runQueriesStub.calledOnceWith(
+        sinon.match.any,
+        "--ram=4992",
+        "--threads=-1",
+      ),
+    );
```
henrymercer, line 91: Nice!  Is it worth also worth putting the RAM and threads arguments in the same order?

src/dependency-caching.test.ts
```diff
@@ -387,3 +411,28 @@ test("getFeaturePrefix - non-C# - returns '' if CsharpNewCacheKey is enabled", a
     t.deepEqual(result, "", `Expected no feature prefix for ${knownLanguage}`);
   }
 });
+
+test("getFeaturePrefix - C# - returns prefix if CsharpCacheBuildModeNone is enabled", async (t) => {
+  const codeql = createStubCodeQL({});
+  const features = createFeatures([Feature.CsharpCacheBuildModeNone]);
+
+  const result = await getFeaturePrefix(codeql, features, KnownLanguage.csharp);
+  t.notDeepEqual(result, "");
```
henrymercer, line 658: Minor: why deep equality here?

src/dependency-caching.test.ts
```diff
@@ -387,3 +411,28 @@ test("getFeaturePrefix - non-C# - returns '' if CsharpNewCacheKey is enabled", a
...
+  const codeql = createStubCodeQL({});
+  const features = createFeatures([Feature.CsharpCacheBuildModeNone]);
+
+  for (const knownLanguage of Object.values(KnownLanguage)) {
+    // Skip C# since we expect a result for it, which is tested in the previous test.
+    if (knownLanguage === KnownLanguage.csharp) {
+      continue;
+    }
+    const result = await getFeaturePrefix(codeql, features, knownLanguage);
+    t.deepEqual(result, "", `Expected no feature prefix for ${knownLanguage}`);
```
henrymercer, line 674: Also here

src/dependency-caching.ts
```diff
@@ -66,6 +69,38 @@ export function getJavaDependencyDirs(): string[] {
...
+export function getCsharpTempDependencyDir(): string {
+  return join(getTemporaryDirectory(), "codeql_csharp", "repository");
+}
+
+/**
+ * Returns an array of paths of directories on the runner that should be included in a dependency cache
+ * for a C# analysis.
+ *
+ * @returns The paths of directories on the runner that should be included in a dependency cache
+ * for a C# analysis.
```
henrymercer, line 86: Nit, here and elsewhere: avoid duplicating doc comments by just keeping the `@returns` paragraph.

henrymercer (dismissed):

michaelnebel:
@mbg : The extractor part has been merged. I suppose it needs to be released before being testable in conjunction with this change?

mbg:
> I suppose it needs to be released before being testable in conjunction with this change?

@michaelnebel We can test this with a nightly (or a custom CLI build) in a test repo. 

To actually be more widely usable, we do have to release the CLI and this change.

src/dependency-caching.ts
mbg: Probably not worth the effort of changing here, but happy to reduce duplication going forward.

src/dependency-caching.test.ts
mbg: IIRC I just followed what other tests did for asserting string equality. I didn't check/try whether `is` would be enough.

src/analyze-action-env.test.ts
mbg: Probably -- I would lie if I said I didn't get this wrong after initially copying it from one to the other. Not sure if this PR is the best place to do it though.

PR #3287, opened 2025-11-13
Open mergeback PR last

henrymercer:
This reduces the likelihood of publishing a tag but not a release.

### Risk assessment

Dev only

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,14 +18,25 @@ For internal use only. Please select the risk level of this change:
 
 #### Which use cases does this change impact?
 
-<!-- Delete options that don't apply. -->
+<!-- Delete options that don't apply. If in doubt, do not delete an option. -->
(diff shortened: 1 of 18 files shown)


.github/workflows/post-release-mergeback.yml
```diff
@@ -161,3 +151,13 @@ jobs:
             --latest=false \
             --title "$VERSION" \
             --notes-file "$PARTIAL_CHANGELOG"
+
+      - name: Create mergeback branch and PR
+        if: ${{ endsWith(github.ref_name, steps.getVersion.outputs.latest_release_branch) }}
```
henrymercer, line 156: I also changed this so that we will try to open the mergeback even if the tag exists.  This means we can retry generating the mergeback if this workflow fails, even if the tag is already published.

esbena (approved):

PR #3292, opened 2025-11-13
Remove push trigger from manual checks

henrymercer:
To avoid issues with concurrency groups.


henrymercer:
Never mind, this [isn't sufficient](https://github.com/github/codeql-action/actions/runs/19346203594).  We may need to manually encode the name of the workflow in the concurrency group rather than using `${{ github.workflow }}`.

PR #3310, opened 2025-11-18
Overlay: Fall back to full analysis if runner disk space is low

kaspersv:
This PR introduces a fallback from overlay analysis to normal analysis if available runner disk space is low when deciding whether to perform overlay analysis during the `init` action. The fallback only applies when overlay analysis is enabled by feature flag and is ignored if overlay analysis is enabled by environment variable. The hardcoded limit of 20GB available disk space has been chosen based on telemetry from the `codeql-action` and will allow most +3m analysis runs to run in overlay mode.  

### Risk assessment

### DIFF
--- a/src/config-utils.ts
+++ b/src/config-utils.ts
@@ -693,7 +693,9 @@ export async function getOverlayDatabaseMode(
       diskUsage.numAvailableBytes < OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_BYTES
     ) {
       const diskSpaceMb =
-        diskUsage === undefined ? 0 : diskUsage.numAvailableBytes / 1_000_000;
+        diskUsage === undefined
+          ? 0
+          : Math.round(diskUsage.numAvailableBytes / 1_000_000);
       overlayDatabaseMode = OverlayDatabaseMode.None;
       useOverlayDatabaseCaching = false;
       logger.info(
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -86928,7 +86928,7 @@ async function getOverlayDatabaseMode(codeql, features, languages, sourceRoot, b
   )) {
     const diskUsage = await checkDiskUsage(logger);
     if (diskUsage === void 0 || diskUsage.numAvailableBytes < OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_BYTES) {
-      const diskSpaceMb = diskUsage === void 0 ? 0 : diskUsage.numAvailableBytes / 1e6;
+      const diskSpaceMb = diskUsage === void 0 ? 0 : Math.round(diskUsage.numAvailableBytes / 1e6);
       overlayDatabaseMode = "none" /* None */;
       useOverlayDatabaseCaching = false;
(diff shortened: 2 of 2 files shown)


src/config-utils.ts
```diff
@@ -667,28 +679,43 @@ export async function getOverlayDatabaseMode(
...
+    ) {
+      const diskSpaceMb =
+        diskUsage === undefined ? 0 : diskUsage.numAvailableBytes / 1_000_000;
+      overlayDatabaseMode = OverlayDatabaseMode.None;
+      useOverlayDatabaseCaching = false;
       logger.info(
         `Setting overlay database mode to ${overlayDatabaseMode} ` +
-          "with caching because we are analyzing the default branch.",
+          `due to insufficient disk space (${diskSpaceMb} MB).`,
       );
```
mbg: The call to `getOverlayDatabaseMode` isn't gated behind the FF, so this code path is always run unless the `CODEQL_OVERLAY_DATABASE_MODE` environment variable is set.

The log message for "Setting overlay database mode to none due to insufficient disk space" may get logged even if the overlay analysis would not be enabled anyway because the FFs aren't enabled. 

It might be better to perform this check guarded by `isOverlayAnalysisFeatureEnabled` so that it is only logged when it would make sense.

src/config-utils.ts
```diff
@@ -667,28 +679,43 @@ export async function getOverlayDatabaseMode(
...
-      overlayDatabaseMode = OverlayDatabaseMode.OverlayBase;
-      useOverlayDatabaseCaching = true;
+  } else {
+    const diskUsage = await checkDiskUsage(logger);
+    if (
+      diskUsage === undefined ||
+      diskUsage.numAvailableBytes < OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_BYTES
+    ) {
+      const diskSpaceMb =
+        diskUsage === undefined ? 0 : diskUsage.numAvailableBytes / 1_000_000;
```
mbg: Minor: Maybe the result of the division should be rounded, so we don't end up with excessive precision in the log message.

src/config-utils.ts
kaspersv: Good point. Fixed in commit 3.

src/config-utils.ts
kaspersv: Fixed in commit 4.

src/config-utils.ts
```diff
@@ -679,20 +687,37 @@ export async function getOverlayDatabaseMode(
...
+            "with caching because we are analyzing a pull request.",
+        );
+      } else if (await isAnalyzingDefaultBranch()) {
+        overlayDatabaseMode = OverlayDatabaseMode.OverlayBase;
+        useOverlayDatabaseCaching = true;
+        logger.info(
+          `Setting overlay database mode to ${overlayDatabaseMode} ` +
+            "with caching because we are analyzing the default branch.",
+        );
+      }
```
mbg, line 720: Minor: This could now just be two `else if` blocks rather than an `if/else` inside an `else` block.

PR #3318, opened 2025-11-19
Support ignoring generated files

mbg:
This PR adds experimental support for excluding files that are marked as `linguist-generated=true` in a `.gitattributes` file from analysis.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/git-utils.test.ts
+++ b/src/git-utils.test.ts
@@ -439,7 +439,7 @@ test("GitVersionInfo.isAtLeast correctly compares versions", async (t) => {
 test("listFiles returns array of file paths", async (t) => {
   sinon
     .stub(gitUtils, "runGitCommand")
-    .resolves(["dir/file.txt", "README.txt"].join(os.EOL));
+    .resolves(["dir/file.txt", "README.txt", ""].join(os.EOL));
 
   await t.notThrowsAsync(async () => {
     const result = await gitUtils.listFiles("/some/path");
@@ -453,7 +453,7 @@ test("getGeneratedFiles returns generated files only", async (t) => {
 
   runGitCommandStub
     .onFirstCall()
-    .resolves(["dir/file.txt", "test.json", "README.txt"].join(os.EOL));
+    .resolves(["dir/file.txt", "test.json", "README.txt", ""].join(os.EOL));
   runGitCommandStub
     .onSecondCall()
     .resolves(
@@ -461,6 +461,7 @@ test("getGeneratedFiles returns generated files only", async (t) => {
         "dir/file.txt: linguist-generated: unspecified",
         "test.json: linguist-generated: true",
         "README.txt: linguist-generated: false",
+        "",
       ].join(os.EOL),
     );
 


src/actions-util.ts
```diff
@@ -254,7 +254,18 @@ export function isDynamicWorkflow(): boolean {
 
 /** Determines whether we are running in default setup. */
 export function isDefaultSetup(): boolean {
-  return isDynamicWorkflow();
+  return isDynamicWorkflow() && !isCCR();
+}
+
+/** Determines whether we are running in CCR. */
+export function isCCR(): boolean {
```
henrymercer, line 265: What about introducing an environment variable we set in CCR, rather than relying on the analysis key?

src/git-utils.ts
```diff
@@ -408,3 +410,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+    ["ls-files"],
+    "Unable to list tracked files.",
+  );
+  return stdout.split(os.EOL);
+}
+
+export async function getGeneratedFiles(
+  workingDirectory: string,
+): Promise<string[]> {
+  const files = await listFiles(workingDirectory);
```
henrymercer, line 425: This could potentially be a very large number of files, too many to pass on the command line.

If we're mainly interested in CCR, could we filter down to just the diff here?

Alternatively, we could parse globs from the `.gitattributes` file rather than finding all files that match.  That would be more likely to contain one entry for a large directory rather than potentially hundreds.

Or we could add a limit on the number of files on which we'll run `check-attr`.

src/git-utils.ts
```diff
@@ -408,3 +410,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
+    workingDirectory,
+    ["ls-files"],
+    "Unable to list tracked files.",
+  );
+  return stdout.split(os.EOL);
+}
+
+export async function getGeneratedFiles(
+  workingDirectory: string,
+): Promise<string[]> {
```
henrymercer, line 424: Have you measured how long this operation takes overall on a large mono-repo?

src/config-utils.ts
```diff
@@ -845,6 +854,19 @@ export async function initConfig(
 
   const config = await initActionState(inputs, userConfig);
 
+  // If we are in CCR or the corresponding FF is enabled, try to determine
+  // which files in the repository are marked as generated and add them to
+  // the `paths-ignore` configuration.
+  if ((await features.getValue(Feature.IgnoreGeneratedFiles)) || isCCR()) {
```
henrymercer: I'd advise rolling this out behind a feature flag first, even in CCR.

mbg: This was actually an `&&`, but I changed it for testing purposes 😅

src/git-utils.ts
mbg: I believe it should be relatively quick, since Git caches this information IIRC. That said, I haven't explicitly tested it on a large repo, but I can do that.

src/git-utils.ts
mbg: > This could potentially be a very large number of files, too many to pass on the command line.

True. The files could be passed in via `stdin` as well I think, which might be a more robust solution.

> If we're mainly interested in CCR, could we filter down to just the diff here?

Potentially. We should first look at whether there is actually enough of a performance hit for large repos for this to make sense though.

> Alternatively, we could parse globs from the .gitattributes file rather than finding all files that match. That would be more likely to contain one entry for a large directory rather than potentially hundreds.

I considered this, but I'd like to avoid it if we can get the information from `git` itself so we don't have to try and duplicate what it does.

src/actions-util.ts
mbg: As discussed elsewhere, that is sort of what we are doing here already (and we decide on whether are in CCR in other places in the same way currently).

A better solution would be to add a new analysis kind (or similar) for CCR, but that would be more work and currently the same as `code-quality` in essentially everything but name. 

I'd suggest we should stick with this for the moment (since we also use the same approach elsewhere) and look at improving it longer-term.

src/api-client.ts
```diff
@@ -189,7 +190,7 @@ export async function getWorkflowRelativePath(): Promise<string> {
  * the GitHub API, but after that the result will be cached.
  */
 export async function getAnalysisKey(): Promise<string> {
-  const analysisKeyEnvVar = "CODEQL_ACTION_ANALYSIS_KEY";
+  const analysisKeyEnvVar = EnvVar.ANALYSIS_KEY;
```
henrymercer: Nit: inline this?

src/config-utils.ts
```diff
@@ -948,6 +953,29 @@ export async function initConfig(
...
+  // the `paths-ignore` configuration.
+  if ((await features.getValue(Feature.IgnoreGeneratedFiles)) && isCCR()) {
+    try {
+      const generatedFiles = await getGeneratedFiles(inputs.sourceRoot);
+
+      if (generatedFiles.length > 0) {
+        config.computedConfig["paths-ignore"] ??= [];
+        config.computedConfig["paths-ignore"].push(...generatedFiles);
+        logger.info(
+          `Detected ${generatedFiles.length} generated file(s), which will be excluded from analysis: ${generatedFiles.join(", ")}`,
```
henrymercer: This might be a very long list — consider logging 10 and saying "(and n-10 more)"

src/git-utils.test.ts
```diff
@@ -435,3 +435,41 @@ test("GitVersionInfo.isAtLeast correctly compares versions", async (t) => {
...
+    .resolves(["dir/file.txt", "README.txt"].join(os.EOL));
+
+  await t.notThrowsAsync(async () => {
+    const result = await gitUtils.listFiles("/some/path");
+    t.is(result.length, 2);
+    t.is(result[0], "dir/file.txt");
+  });
+});
+
+test("getGeneratedFiles returns generated files only", async (t) => {
```
henrymercer, line 451: Add a test for correct behaviour when there's a blank line at the end of the file (i.e., not specifying an empty string path)?

src/git-utils.ts
```diff
@@ -408,3 +410,34 @@ export async function isAnalyzingDefaultBranch(): Promise<boolean> {
...
   return currentRef === defaultBranch;
 }
+
+export async function listFiles(workingDirectory: string): Promise<string[]> {
+  const stdout = await runGitCommand(
+    workingDirectory,
+    ["ls-files"],
+    "Unable to list tracked files.",
+  );
+  return stdout.split(os.EOL);
```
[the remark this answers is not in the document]
henrymercer: Agreed, it aligns better with the docstring if this function doesn't output empty string files.

src/config-utils.ts
```diff
@@ -948,6 +953,29 @@ export async function initConfig(
     }
   }
 
+  // If we are in CCR or the corresponding FF is enabled, try to determine
+  // which files in the repository are marked as generated and add them to
+  // the `paths-ignore` configuration.
+  if ((await features.getValue(Feature.IgnoreGeneratedFiles)) && isCCR()) {
+    try {
+      const generatedFiles = await getGeneratedFiles(inputs.sourceRoot);
```
henrymercer, line 963: I'd like to see a telemetry diagnostic so we we know how long this operation takes, including in the p99 case.

src/config-utils.ts
```diff
@@ -948,6 +953,29 @@ export async function initConfig(
     }
   }
 
+  // If we are in CCR or the corresponding FF is enabled, try to determine
```
henrymercer, line 957: Or just check the FF.

```suggestion
  // If we are in CCR and the corresponding FF is enabled, try to determine
```

mbg:
> We could add a PR check as an integration test. We'd need to drop the CCR-only restriction though.

@henrymercer I have opened https://github.com/github/codeql-action/pull/3408 to add a PR check for a basic CCR-like setup. I don't think we need to drop the CCR-only restriction for us to be able to test this.

henrymercer:
> @henrymercer I have opened #3408 to add a PR check for a basic CCR-like setup. I don't think we need to drop the CCR-only restriction for us to be able to test this.

Sure, I was thinking about testing this out in code security mode on our repo, since we check in generated code.  But the CCR PR check works too.

src/config-utils.ts
```diff
@@ -1413,3 +1424,32 @@ async function logGitVersionTelemetry(
...
+  generatedFilesCount: number,
+): Promise<void> {
+  if (config.languages.length < 1) {
+    return;
+  }
+
+  addDiagnostic(
+    config,
+    // Arbitrarily choose the first language. We could also choose all languages, but that
+    // increases the risk of misinterpreting the data.
```
henrymercer, line 1444: This is no longer arbitrary ;)

PR #3351, opened 2025-12-10
 Determine CodeQL version from feature flags on GHEC-DR 

henrymercer:
This lets us roll back a bad version of CodeQL more quickly in GitHub Enterprise Cloud with data residency.

### Risk assessment

Workflow types:

Products:

Environments:

- **GHEC-DR**

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -556,10 +556,14 @@ const CODEQL_ACTION_WARNED_ABOUT_VERSION_ENV_VAR =
 let hasBeenWarnedAboutVersion = false;
 
 export enum GitHubVariant {
+  /** [GitHub.com](https://github.com) */
   DOTCOM = "GitHub.com",
-  GHES = "GHES",
-  GHEC_DR = "GHEC-DR",
+  /** [GitHub Enterprise Server](https://docs.github.com/en/enterprise-server@latest/admin/overview/about-github-enterprise-server) */
+  GHES = "GitHub Enterprise Server",
+  /** [GitHub Enterprise Cloud with data residency](https://docs.github.com/en/enterprise-cloud@latest/admin/data-residency/about-github-enterprise-cloud-with-data-residency) */
+  GHEC_DR = "GitHub Enterprise Cloud with data residency",
 }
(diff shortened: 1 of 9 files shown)


src/util.ts
```diff
@@ -556,13 +556,13 @@ const CODEQL_ACTION_WARNED_ABOUT_VERSION_ENV_VAR =
 let hasBeenWarnedAboutVersion = false;
 
 export enum GitHubVariant {
-  DOTCOM,
-  GHES,
-  GHE_DOTCOM,
+  DOTCOM = "GitHub.com",
+  GHES = "GHES",
+  GHEC_DR = "GHEC-DR",
```
mbg: Minor: Should the string value be spelled out as e.g. `GHEC (with Data Residency)` or similar?

src/util.test.ts
```diff
@@ -433,8 +433,8 @@ function formatGitHubVersion(version: util.GitHubVersion): string {
   switch (version.type) {
     case util.GitHubVariant.DOTCOM:
       return "dotcom";
-    case util.GitHubVariant.GHE_DOTCOM:
-      return "GHE dotcom";
+    case util.GitHubVariant.GHEC_DR:
+      return "GHEC-DR";
     case util.GitHubVariant.GHES:
       return `GHES ${version.version}`;
     default:
```
mbg, line 440: Minor: How about something like this now that `GitHubVariant` has string values associated with the elements?

```suggestion
  switch (version.type) {
    case util.GitHubVariant.DOTCOM:
    case util.GitHubVariant.GHEC_DR:
      return `${version.type}`;
    case util.GitHubVariant.GHES:
      return `GHES ${version.version}`;
    default:
```

src/util.ts
henrymercer: 👍 I've expanded this.

src/util.test.ts
henrymercer: Now that we're using the full names above, I think it's preferable to keep the abbreviated names here to keep test titles concise.

PR #3352, opened 2025-12-11
Clean up `JavaMinimizeDependencyJars` feature flag

nickrolfe:
The feature flag has been fully rolled out for some time, so we can simplify the code.

At the same time, stop using the `minify-` prefix for dependency cache keys.

There is still some Java-specific code to set the corresponding extractor option. I've left a TODO comment and will open an internal issue for a more principled way of handling that.

<hr/>

### Risk assessment

The feature flag being removed is fully rolled out, so this PR should have no observable effects.

Workflow types:

Products:

    - I don't know about this one.

Environments:

### DIFF
--- a/src/init-action.ts
+++ b/src/init-action.ts
@@ -89,6 +89,13 @@ import {
 import { checkWorkflow } from "./workflow";
 
 /**
+ * First version of CodeQL where the Java extractor safely supports the option to minimize
+ * dependency jars. Note: some earlier versions of the extractor will respond to the corresponding
+ * option, but may rewrite jars in ways that lead to extraction errors.
+ */
+export const CODEQL_VERSION_JAR_MINIMIZATION = "2.23.0";
+
+/**
  * Sends a status report indicating that the `init` Action is starting.
  *
  * @param startedAt
@@ -651,7 +658,7 @@ async function run() {
         `${EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS} is already set to '${process.env[EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS]}', so the Action will not override it.`,
       );
     } else if (
-      (await codeQlVersionAtLeast(codeql, "2.23.0")) && // First version of the extractor to safely support this option
+      (await codeQlVersionAtLeast(codeql, CODEQL_VERSION_JAR_MINIMIZATION)) &&
       config.dependencyCachingEnabled &&
       config.buildMode === BuildMode.None &&
       config.languages.includes(KnownLanguage.java)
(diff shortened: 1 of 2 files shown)


src/init-action.ts
```diff
@@ -638,18 +638,20 @@ async function run() {
...
+    // TODO: Remove this language-specific mechanism and replace it with a more general one that
+    // tells extractors when dependency caching is enabled, and then the Java extractor can make its
+    // own decision about whether to rewrite jars.
     if (process.env[EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS]) {
       logger.debug(
         `${EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS} is already set to '${process.env[EnvVar.JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS]}', so the Action will not override it.`,
       );
     } else if (
-      (await features.getValue(Feature.JavaMinimizeDependencyJars, codeql)) &&
+      (await codeQlVersionAtLeast(codeql, "2.23.0")) && // First version of the extractor to safely support this option
```
henrymercer: Nit, non-blocking: this might be slightly clearer as a constant at the top of the file.

henrymercer (dismissed):

src/init-action.ts
```diff
@@ -88,6 +88,9 @@ import {
 } from "./util";
 import { checkWorkflow } from "./workflow";
 
+/** First version of CodeQL where the Java extractor safely supports the option to minimize dependency jars. */
+export const CODEQL_VERSION_JAR_MINIMIZATION = "2.23.0";
```
mbg: Minor: Thanks for extracting this into a constant. Some extra thoughts:

- It may be good to extend the comment to highlight that some CLI versions before `2.23.0` support `JAVA_EXTRACTOR_MINIMIZE_DEPENDENCY_JARS` but that there was a subsequent fix/improvement in `2.23.0`. 
- No strong feelings, but I wonder if it would be cleaner to have this in a different file / keep the FF around but enable it by default.

nickrolfe: I've updated the comment, as you suggested.

For your other point, I'll defer to you and Henry to make a decision.

henrymercer: We have most of the CLI versions in `codeql.ts`, although there is one exception `CODEQL_VERSION_ZSTD_BUNDLE` which lives in `setup-codeql.ts` to avoid circular imports.  It would be slightly better to move this to `codeql.ts` for consistency but I think there are few practical implications: if we're going to deprecate a CodeQL version we're going to search the codebase for `CODEQL_VERSION_`.

henrymercer (approved):

src/init-action.ts
nickrolfe: OK, I'll merge what's here, but feel free to follow up and rearrange the code as you see fit.

PR #3358, opened 2025-12-15
Add status report for uploading databases to API

henrymercer:
This PR adds a status report for uploading databases.  This contains:

- The DB language
- The size of the zipped DB
- Whether the DB is an overlay base database
- How long the upload took
- Error information, if there was an error uploading the database.

### Risk assessment

Workflow types:

Products:

Environments:

See https://github.com/henrymercer-test-org/overlay-testing/actions/runs/20233483858/job/58082003115#step:5:5392


src/database-upload.ts
```diff
@@ -90,6 +107,7 @@ export async function cleanupAndUploadDatabases(
         actionsUtil.getRequiredInput("checkout_path"),
       );
       try {
+        const startTime = Date.now();
```
mbg: Any particular reason for using `Date.now()` for the measurement here rather than `performance.now()` that we use elsewhere?

src/database-upload.ts
```diff
@@ -107,13 +125,25 @@ export async function cleanupAndUploadDatabases(
...
         logger.debug(`Successfully uploaded database for ${language}`);
       } finally {
         bundledDbReadStream.close();
       }
     } catch (e) {
       // Log a warning but don't fail the workflow
       logger.warning(`Failed to upload database for ${language}: ${e}`);
+      reports.push({
+        language,
+        error: util.getErrorMessage(e),
```
mbg, line 146: If `getErrorMessage(e)` is useful here, would it also make sense in the log message above rather than just `e`?

src/database-upload.ts
```diff
@@ -44,20 +58,22 @@ export async function cleanupAndUploadDatabases(
...
   if (!(await gitUtils.isAnalyzingDefaultBranch())) {
     // We only want to upload a database if we are analyzing the default branch.
     logger.debug("Not analyzing default branch. Skipping upload.");
-    return;
+    return [];
   }
 
-  const cleanupLevel =
+  // If config.overlayDatabaseMode is OverlayBase, then we have overlay base databases for all languages.
+  const isOverlayBase =
```
mbg: I think this name is slightly misleading, because it also depends on the FF. Perhaps `uploadOverlayDb` would be a better name?

src/database-upload.ts
henrymercer: `performance.now()` would be better 👍

mbg (dismissed):
Thanks for addressing all the suggestions!

henrymercer:
Fixed a merge conflict.

PR #3374, opened 2025-12-17
CI: Perform a best-effort scan of the debug artifacts during release validation

henrymercer:
Produce a best effort warning if sensitive environment variables are detected in debug artifacts.  This is not reliable enough for production but the intent is that the test will run as part of release validation when the bundle update PR is opened.

### Risk assessment

Workflow types: n/a
Products: n/a

Environments:

The CI check should fail if the artifact scan starts failing.

### DIFF
--- a/src/artifact-scanner.ts
+++ b/src/artifact-scanner.ts
@@ -127,6 +127,26 @@ async function scanArchiveFile(
       await exec.exec("tar", ["-xzf", archivePath, "-C", tempExtractDir], {
         silent: true,
       });
+    } else if (fileName.endsWith(".tar.zst")) {
+      // Extract tar.zst files
+      logger.debug(`Extracting tar.zst file: ${archivePath}`);
+      await exec.exec(
+        "tar",
+        ["--zstd", "-xf", archivePath, "-C", tempExtractDir],
+        {
+          silent: true,
+        },
+      );
+    } else if (fileName.endsWith(".zst")) {
+      // Extract .zst files (single file compression)
+      logger.debug(`Extracting zst file: ${archivePath}`);
+      const outputFile = path.join(
+        tempExtractDir,
+        path.basename(archivePath, ".zst"),
+      );
+      await exec.exec("zstd", ["-d", archivePath, "-o", outputFile], {
+        silent: true,
+      });
     } else if (fileName.endsWith(".gz")) {
       // Extract .gz files (single file compression)
       logger.debug(`Extracting gz file: ${archivePath}`);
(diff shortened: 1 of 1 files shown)


src/artifact-scanner.ts
```diff
@@ -0,0 +1,357 @@
...
+  depth: number = 0,
+): Promise<ScanResult> {
+  const result: ScanResult = {
+    scannedFiles: 1,
+    findings: [],
+  };
+
+  // Check if it's an archive file and recursively scan it
+  const fileName = path.basename(fullPath).toLowerCase();
+  const isArchive =
```
esbena, line 218: have you looked at all extensions in our assets in practice?
Haven't we recently played around with something like `.xz` for instance?

henrymercer:
> I hoped more of this could have been offloaded to a library.

For extracting archives, I wanted to use system utilities since this is only going to be run in CI.  I'm open to suggestions if you have a library in mind to do the job here, but I think the implementation is fairly slim.

> I worry a bit about the extract/scan/.../delete sequence and limited disk space, but since we are the creators of the asset, we are probably fine in practice. A streaming scan would be preferable.

If we start seeing failures, it would probably be a good idea to convert to a streaming scan, or ignore things like the source archive.  I think since this is CI-only it doesn't necessarily make sense front-loading that work.

> Hacky solution proposal: pipe all assets through `<source> | tar x...... | grep <patterns>`. But if we already have something that works, then I can approve without it.

There are some features like tracking the file in which the token was found, what type of token it was, and not relying on having bsdtar to extract zips.  Once we add these, I think we're at the point where I'd rather maintain a JS program than a shell script :)
 
> The changes to the generated files seem benign, but they are unexpected. Is that the norm due to non-determinism these days?

Unfortunately we are bundling this artifact scanner in order to run it before we upload debug artifacts in CI.  It's behind the `isInTestMode()` check, which requires a specific environment variable to be set, but it is being included in the production artifacts.

src/artifact-scanner.ts
henrymercer:
```
find <debug artifacts> -type f | sed 's/.*\.//' | sort | uniq -c | sort -rn
3284 gz
2872 dep
 222 js
 111 log
  52 h
  49 yml
  39 json
   9 qls
   8 zst
   8 zip
   4 trap
   4 set
   4 odep
   3 properties
   3 orig
   3 env
   3 args
   2 yaml
   2 s
   2 rb
   2 py
   2 manifest
   2 list
   2 link
   2 kt
   2 jsonl
   2 java
   2 go
   2 codeql-lock
   2 c
```

I pushed a commit to extract zstd archives too.

esbena (approved):
👍🏻 Reasonable tradeoff between problem and implementation cost for now.

PR #3375, opened 2025-12-17
Require tools feature for uploading overlay DBs

henrymercer:
The `database bundle` command does not currently include a critical overlay database metadata file.  Support is being added in conjunction with the `bundleSupportsOverlay` tools feature.  This PR only uploads overlay DBs when using a CLI with that tool feature.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/database-upload.ts
+++ b/src/database-upload.ts
@@ -145,7 +145,9 @@ export async function cleanupAndUploadDatabases(
       reports.push({
         language,
         error: util.getErrorMessage(e),
-        ...(bundledDbSize ? { zipped_upload_size_bytes: bundledDbSize } : {}),
+        ...(bundledDbSize !== undefined
+          ? { zipped_upload_size_bytes: bundledDbSize }
+          : {}),
       });
     }
   }
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -97585,7 +97585,7 @@ async function cleanupAndUploadDatabases(repositoryNwo, codeql, config, apiDetai
       reports.push({
         language,
         error: getErrorMessage(e),
-        ...bundledDbSize ? { zipped_upload_size_bytes: bundledDbSize } : {}
+        ...bundledDbSize !== void 0 ? { zipped_upload_size_bytes: bundledDbSize } : {}
       });
     }
   }


src/database-upload.ts
```diff
@@ -144,6 +145,7 @@ export async function cleanupAndUploadDatabases(
       reports.push({
         language,
         error: util.getErrorMessage(e),
+        ...(bundledDbSize ? { zipped_upload_size_bytes: bundledDbSize } : {}),
```
mbg: Is using the truthiness of `bundledDbSize` a problem here? E.g. do we care that this would be `false` if `bundledDbSize` was `0` rather than `undefined`?

mbg (dismissed):
One question, otherwise this looks reasonable.

src/database-upload.ts
henrymercer: It shouldn't be 0, but worth correcting 👍

src/database-upload.ts
```diff
@@ -145,7 +145,9 @@ export async function cleanupAndUploadDatabases(
       reports.push({
         language,
         error: util.getErrorMessage(e),
-        ...(bundledDbSize ? { zipped_upload_size_bytes: bundledDbSize } : {}),
+        ...(bundledDbSize !== undefined
+          ? { zipped_upload_size_bytes: bundledDbSize }
+          : {}),
```
mbg, line 150: Is this not equivalent to just

```suggestion
        zipped_upload_size_bytes: bundledDbSize
```

henrymercer:
```
➜  ~ node
Welcome to Node.js v25.2.1.
> let x = {}
undefined
> x["a"] = undefined
undefined
> x
{ a: undefined }
```

mbg: Yeah, but practically speaking there is not typically a difference in using an object with a key that has `undefined` as value vs a key that is absent. Asking for the value of the key gives `undefined` in both cases. The question is whether we care about the distinction that the key itself exists or not.

henrymercer: That's fair.  I'd prefer to leave this to the future and address it across the codebase.

PR #3386, opened 2026-01-05
Introduce a type-level check that CodeQL is passed during feature flag lookup if it is needed

henrymercer:
Introduce a type-level check that CodeQL is passed during feature flag lookup if it is needed.

- Use `satisfies` rather than a type annotation to maintain a more precise type for `featureConfig`
- Add `FeatureWithoutCLI` type to record features whose enablement don't depend on the version of CodeQL
- Update `FeatureEnablement` interface to add a call for `FeatureWithoutCLI` that doesn't require CodeQL, and make the call for `Feature` require CodeQL
- Update places that inspect `featureConfig` to do a safe cast to the former `FeatureConfig` type using `satisfies FeatureConfig as FeatureConfig`

### Risk assessment

Dev change only.

### DIFF
--- a/src/feature-flags.test.ts
+++ b/src/feature-flags.test.ts
@@ -257,8 +257,11 @@ for (const feature of Object.keys(featureConfig)) {
         const expectedFeatureEnablement = initializeFeatures(true);
         mockFeatureFlagApiEndpoint(200, expectedFeatureEnablement);
 
+        // The type system should prevent this happening, but test that if we
+        // bypass it we get the expected error.
         await t.throwsAsync(
-          async () => features.getValue(feature as FeatureWithoutCLI),
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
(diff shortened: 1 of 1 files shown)


src/testing-utils.ts
```diff
@@ -316,7 +316,7 @@ export function createFeatures(enabledFeatures: Feature[]): FeatureEnablement {
       throw new Error("not implemented");
     },
     getValue: async (feature) => {
-      return enabledFeatures.includes(feature);
+      return enabledFeatures.includes(feature as Feature);
```
mbg, line 319: Minor: could this be a type annotation on `feature`?

mbg (approved):
LGTM, other than one minor point. Very similar to what I came up with :)

src/testing-utils.ts
henrymercer: I agree that's slightly better, but I don't think it's worth going through another round of approval :)

PR #3404, opened 2026-01-19
Include base database OIDs when bundling database

henrymercer:
Include the base database OIDs file when bundling the database since this file is required to compute the set of changed files and perform an overlay analysis.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -755,7 +755,11 @@ export async function bundleDb(
   if (fs.existsSync(databaseBundlePath)) {
     await fs.promises.rm(databaseBundlePath, { force: true });
   }
-  // Copy the base database OIDs file into the database location if it exists
+  // When overlay is enabled, the base database OIDs file is included at the
+  // root of the database cluster. However when we bundle a database, we only
+  // include the per-language database. So, to ensure the base database OIDs
+  // file is included in the database bundle, we copy it from the cluster into
+  // the individual database location before bundling.
   const baseDatabaseOidsFilePath = getBaseDatabaseOidsFilePath(config);
   const additionalFiles: string[] = [];
   if (fs.existsSync(baseDatabaseOidsFilePath)) {


src/util.ts
```diff
@@ -745,7 +755,23 @@ export async function bundleDb(
...
-  await codeql.databaseBundle(databasePath, databaseBundlePath, dbName);
+  // Copy the base database OIDs file into the database location if it exists
+  const baseDatabaseOidsFilePath = getBaseDatabaseOidsFilePath(config);
+  const additionalFiles: string[] = [];
+  if (fs.existsSync(baseDatabaseOidsFilePath)) {
+    await fsPromises.copyFile(
+      baseDatabaseOidsFilePath,
+      path.join(databasePath, BASE_DATABASE_OIDS_FILE_NAME),
+    );
+    additionalFiles.push(BASE_DATABASE_OIDS_FILE_NAME);
```
esbena, line 770: for reference: we also do this elsewhere

esbena (dismissed):

src/codeql.ts
```diff
@@ -147,11 +147,19 @@ export interface CodeQL {
...
+   * These paths are relative to the database root.
+   *
+   * Older versions of the CodeQL CLI do not support including additional paths in the bundle.
+   * In those cases, this parameter will be ignored.
    */
   databaseBundle(
     databasePath: string,
     outputFilePath: string,
     dbName: string,
+    tryAlsoIncludeRelativePaths: string[],
```
mbg: Does the name start with `try` because of the feature check? If so, we should probably just call this `includeRelativePaths` (or `extraPathsToInclude`).

src/codeql.ts
```diff
@@ -920,6 +929,16 @@ async function getCodeQLForCmd(
...
         ...getExtraOptionsFromEnv(["database", "bundle"]),
       ];
+      if (
+        await this.supportsFeature(ToolsFeature.BundleSupportsIncludeOption)
+      ) {
+        args.push(
+          ...tryAlsoIncludeRelativePaths.flatMap((relativePath) => [
+            "--include",
+            relativePath,
+          ]),
```
mbg: This could just be `map` instead of `flatMap` I think.

src/util.ts
```diff
@@ -745,7 +755,23 @@ export async function bundleDb(
   if (fs.existsSync(databaseBundlePath)) {
     await fs.promises.rm(databaseBundlePath, { force: true });
   }
-  await codeql.databaseBundle(databasePath, databaseBundlePath, dbName);
+  // Copy the base database OIDs file into the database location if it exists
```
mbg: I think this comment could be improved: right now, it is not super obvious what the difference between `databasePath` here and `config.dbLocation` is, and why the file has to be copied from the latter to the former.

src/codeql.ts
henrymercer: We could do that, and perform the feature check in the caller.  I don't have a strong opinion.

src/codeql.ts
henrymercer: `tryAlsoIncludeRelativePaths` is an array so this needs to be a `flatMap`.

src/codeql.ts
mbg: No, keep the feature check here. My point was that the feature check is essentially a FF check to stop us from trying to use a feature that is not yet supported by the CLI. Our expectation would be that eventually this will always happen. In general, we don't name things `try[...]` if they are only used if a FF is enabled that we eventually expect to always be true.

src/codeql.ts
mbg: > `tryAlsoIncludeRelativePaths` is an array so this needs to be a `flatMap`.

I think that's the wrong justification: `map` would be fine in that case. The problem is actually that the anonymous function is returning an array and that you want to flatten what would be an array of a arrays into a single array. So `flatMap` is a sensible choice, but for that reason. (I had previously missed that the function was returning an array, rather than a string.)

Incidentally, you could have something like `tryAlsoIncludeRelativePaths.map((relativePath) => /* same as before */).flat()` for the same result, but it's better to stick with `flatMap` in this case.

mbg (approved):
Thanks! The changes made in response to my comments look good 👍🏻

PR #3409, opened 2026-01-20
Ensure that proxy log artifacts have unique names

mbg:
The post action for `start-proxy` uploads the authentication proxy's logs in debug mode. However, it currently uses a fixed name for this (always `proxy-log-file`). This is a problem in e.g. matrix'ed workflows where multiple CodeQL jobs may want to upload the respective proxy logs.

This PR lightly refactors our existing code for uploading debug artifacts, where we already address this problem for other debug artifacts, so that the same functions can then be used for the `start-proxy` post action.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/debug-artifacts.test.ts
+++ b/src/debug-artifacts.test.ts
@@ -36,17 +36,22 @@ test("getArtifactSuffix", (t) => {
 
   // Suffixes for non-empty, valid `matrix` inputs.
   const testMatrices = [
-    { language: "go" },
-    { language: "javascript", "build-mode": "none" },
+    { matrix: { language: "go" }, expected: "-go" },
+    {
+      matrix: { language: "javascript", "build-mode": "none" },
+      expected: "-none-javascript",
+    },
+    {
+      matrix: { "build-mode": "none", language: "javascript" },
+      expected: "-none-javascript",
+    },
   ];
 
   for (const testMatrix of testMatrices) {
-    const suffix = debugArtifacts.getArtifactSuffix(JSON.stringify(testMatrix));
-    t.not(suffix, "");
-
-    for (const key of Object.keys(testMatrix)) {
-      t.assert(suffix.includes(testMatrix[key] as string));
-    }
+    const suffix = debugArtifacts.getArtifactSuffix(
+      JSON.stringify(testMatrix.matrix),
+    );
+    t.is(suffix, testMatrix.expected);
   }
 });
 


src/debug-artifacts.ts
```diff
@@ -246,22 +246,47 @@ export async function tryUploadAllAvailableDebugArtifacts(
...
+ * @returns A suffix that uniquely identifies the `matrix` value for the current job, or `""` if there
+ * is no matrix value.
+ */
+export function getArtifactSuffix(matrix: string | undefined): string {
+  let suffix = "";
+  if (matrix) {
+    try {
+      for (const [, matrixVal] of Object.entries(
+        JSON.parse(matrix) as any[][],
+      ).sort())
```
esbena:
```
Object.entries(
        JSON.parse(matrix) as any[][],
      ).sort()
```
The sorting here is slightly dubious. As the sorting will happen on the stringification of each array (which is the same as `array.join(',')`), rather than the individual elements of entry pairs. 
But it works out in practice, since I don't think the matrix names have commas in them. And even if they did, the sorting would still be valid for our use case.

mbg:
I re-ran the "Start proxy" PR check in debug mode to exercise the new logic. This worked as expected: https://github.com/github/codeql-action/actions/runs/21172332497/job/60892639862?pr=3409#step:12:79

src/debug-artifacts.ts
```diff
@@ -246,22 +246,47 @@ export async function tryUploadAllAvailableDebugArtifacts(
...
+        suffix += `-${matrixVal}`;
+    } catch {
+      core.info(
```
esbena:
```suggestion
      core.warn(
```

esbena (approved):
LGTM with some comments on the existing code that I made before I realised it was just moving around..

src/debug-artifacts.ts
mbg: Yeah, this was already in the code I refactored and it probably works OK for our purposes. Nonetheless, it's easy to improve while we are here, so I have pushed a commit which hopefully does that.

src/debug-artifacts.ts
mbg: I initially thought I'd leave this as is to avoid this resulting in workflow annotations, but given that this only runs when in debug mode anyway, it's probably better to highlight it.

src/debug-artifacts.test.ts
```diff
@@ -24,6 +24,29 @@ test("sanitizeArtifactName", (t) => {
...
+  ];
+
+  for (const testMatrix of testMatrices) {
+    const suffix = debugArtifacts.getArtifactSuffix(JSON.stringify(testMatrix));
+    t.not(suffix, "");
+
+    for (const key of Object.keys(testMatrix)) {
+      t.assert(suffix.includes(testMatrix[key] as string));
+    }
+  }
```
henrymercer, line 55: Consider mapping each test matrix to the expected suffix, so it's easier to see exactly what suffix we'd get.

src/debug-artifacts.ts
```diff
@@ -246,22 +246,46 @@ export async function tryUploadAllAvailableDebugArtifacts(
...
+    }
+  }
+  return suffix;
+}
+
+// Enumerates different, possible outcomes for artifact uploads.
+export type UploadArtifactsResult =
+  | "no-artifacts-to-upload"
+  | "upload-successful"
+  | "upload-failed";
```
henrymercer, line 283: How about using an enum?

src/debug-artifacts.ts
```diff
@@ -246,22 +246,46 @@ export async function tryUploadAllAvailableDebugArtifacts(
...
+ *
+ * @param matrix A stringified JSON value, usually the value of the `matrix` input.
+ * @returns A suffix that uniquely identifies the `matrix` value for the current job, or `""` if there
+ * is no matrix value.
+ */
+export function getArtifactSuffix(matrix: string | undefined): string {
+  let suffix = "";
+  if (matrix) {
+    try {
+      const matrixObject = JSON.parse(matrix) as any[][];
```
henrymercer: For improved type safety, should this be `unknown[][]`?

src/debug-artifacts.ts
mbg: I probably would've gone for an `enum` if this was new, but I stuck with what we already had here (just named it for reusability).

src/debug-artifacts.ts
mbg: I changed this to not assume anything about what `JSON.parse` might return, since for user-specified input it could be any valid JSON value. Instead, I added a check that `matrixObject` is actually a JS object and then just assume that it is an `object` after that.

esbena (approved):

PR #3415, opened 2026-01-22
Address missing telemetry at the start of Actions

henrymercer:
The start of some of the Actions' entrypoints was not covered by a try/catch statement, resulting in missed telemetry if any of the code there errored.

Address it by pulling almost all the code into a try/catch statement for each of the Actions.

We could potentially also move this try/catch into the wrapper, but it can be useful to include other variables if they are set (like entries from the config) in the error telemetry.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/status-report.ts
+++ b/src/status-report.ts
@@ -625,19 +625,15 @@ export async function sendUnhandledErrorStatusReport(
       undefined,
       undefined,
       logger,
-      `Unhandled error: ${getErrorMessage(error)}`,
+      `Unhandled CodeQL Action error: ${getErrorMessage(error)}`,
       error instanceof Error ? error.stack : undefined,
     );
     if (statusReport !== undefined) {
       await sendStatusReport(statusReport);
-    } else if (isInTestMode()) {
-      throw new Error(
-        "Failed to create status report for unhandled error in test mode.",
-      );
     }
   } catch (e) {
     logger.warning(
-      `Failed to send the error status report: ${getErrorMessage(e)}.`,
+      `Failed to send the unhandled error status report: ${getErrorMessage(e)}.`,
     );
     if (isInTestMode()) {
       throw e;
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -94421,19 +94421,15 @@ async function sendUnhandledErrorStatusReport(actionName, actionStartedAt, error
       void 0,
       void 0,
       logger,
-      `Unhandled error: ${getErrorMessage(error3)}`,
+      `Unhandled CodeQL Action error: ${getErrorMessage(error3)}`,
       error3 instanceof Error ? error3.stack : void 0
     );
     if (statusReport !== void 0) {
       await sendStatusReport(statusReport);
-    } else if (isInTestMode()) {
-      throw new Error(
-        "Failed to create status report for unhandled error in test mode."
-      );
     }
   } catch (e) {
     logger.warning(
-      `Failed to send the error status report: ${getErrorMessage(e)}.`
+      `Failed to send the unhandled error status report: ${getErrorMessage(e)}.`
     );
     if (isInTestMode()) {
       throw e;
--- a/lib/autobuild-action.js
+++ b/lib/autobuild-action.js
@@ -88893,19 +88893,15 @@ async function sendUnhandledErrorStatusReport(actionName, actionStartedAt, error
       void 0,
       void 0,
       logger,
-      `Unhandled error: ${getErrorMessage(error3)}`,
+      `Unhandled CodeQL Action error: ${getErrorMessage(error3)}`,
       error3 instanceof Error ? error3.stack : void 0
     );
     if (statusReport !== void 0) {
       await sendStatusReport(statusReport);
-    } else if (isInTestMode()) {
-      throw new Error(
-        "Failed to create status report for unhandled error in test mode."
-      );
     }
   } catch (e) {
     logger.warning(
-      `Failed to send the error status report: ${getErrorMessage(e)}.`
+      `Failed to send the unhandled error status report: ${getErrorMessage(e)}.`
     );
     if (isInTestMode()) {
       throw e;
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -92258,19 +92258,15 @@ async function sendUnhandledErrorStatusReport(actionName, actionStartedAt, error
       void 0,
       void 0,
       logger,
-      `Unhandled error: ${getErrorMessage(error3)}`,
+      `Unhandled CodeQL Action error: ${getErrorMessage(error3)}`,
       error3 instanceof Error ? error3.stack : void 0
     );
     if (statusReport !== void 0) {
       await sendStatusReport(statusReport);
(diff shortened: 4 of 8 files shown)


src/init-action.ts
```diff
@@ -194,65 +194,70 @@ async function sendCompletedStatusReport(
...
+    persistInputs();
 
-  const gitHubVersion = await getGitHubVersion();
-  checkGitHubVersionInRange(gitHubVersion, logger);
-  checkActionVersion(getActionVersion(), gitHubVersion);
+    apiDetails = {
+      auth: getRequiredInput("token"),
+      externalRepoAuth: getOptionalInput("external-repository-token"),
+      url: getRequiredEnvParam("GITHUB_SERVER_URL"),
+      apiURL: getRequiredEnvParam("GITHUB_API_URL"),
```
mbg, line 223: `getRequiredInput` throws a `ConfigurationError` when the input is not supplied. That makes sense.

`getRequiredEnvParam` throws an `Error` rather than a `ConfigurationError`. We'd probably always expect these environment variables to be available, and if they're not it's probably nothing a user can do anything about, so `Error` probably makes sense for that. 

Worth thinking about in general though, whether we want one or the other for a given input/env variable.

Observation: since this was previously outside the `try`/`catch` block, we also didn't get telemetry for failures here before.

mbg (approved):
Thank you for putting this together! This generally looks good. I have done some checking here to verify that the exceptions we get from the functions that are now in the `try`/`catch` blocks are of the types we'd want (see also my comment there). 

Two higher-level questions:

1. Should we add comments at the start of the `run` functions to remind us that only variables which are shared between different scopes below should be declared outside the `try` block and should be initialised only using safe functions / ideally inside the `try` block? 
2. Would it make sense to add `try`/`catch` blocks around the main `run` functions, where the `catch` block sends some status report if an exception does leak out?

src/init-action.ts
henrymercer: I'd be surprised if these environment variables weren't set within the Actions environment, but agreed in general.

henrymercer:
> Would it make sense to add `try`/`catch` blocks around the main `run` functions, where the `catch` block sends some status report if an exception does leak out?

It's defensive, but I'd rather not just rely on the comments.  Implemented.

src/status-report.ts
```diff
@@ -606,3 +606,30 @@ export async function createInitWithConfigStatusReport(
...
+export async function sendUnexpectedErrorStatusReport(
+  actionName: ActionName,
+  actionStartedAt: Date,
+  error: unknown,
+  logger: Logger,
+): Promise<void> {
+  try {
+    const statusReport = await createStatusReportBase(
+      actionName,
+      "failure",
```
mbg, line 623: Minor: It would probably be useful to add new status for this (`hard-failure`, `uncaught`, ...) or tag these status reports in some other, easily identifiable way. We likely want to easily be able to tell if this happens, since it's not really supposed to happen at all.

src/status-report.ts
```diff
@@ -606,3 +606,30 @@ export async function createInitWithConfigStatusReport(
...
+      logger,
+      undefined,
+      getErrorMessage(error),
+    );
+    if (statusReport !== undefined) {
+      await sendStatusReport(statusReport);
+    }
+  } catch (e) {
+    logger.warning(
+      `Caught an exception while sending the error status report: ${e}.`,
```
mbg: Few points here:

- Should `e` be wrapped in `getErrorMessage`?
- Change "error status report" to "unhandled error status report" or similar (possibly also in the function name)?
- Re-throw the exception if `isInTestMode()` like in `createStatusReportBase`?

src/start-proxy-action.ts
```diff
@@ -203,6 +202,23 @@ async function runWrapper() {
...
+  const logger = getActionsLogger();
+
+  try {
+    await run(startedAt);
+  } catch (error) {
+    core.setFailed(`start-proxy action failed: ${util.getErrorMessage(error)}`);
+    await sendUnexpectedErrorStatusReport(
+      ActionName.StartProxy,
+      startedAt,
+      error,
```
mbg: For `start-proxy`, we deliberately don't include the error to avoid leaking any sensitive information in the main handler in the (now) `run` function. We should probably omit the error here too.

src/status-report.ts
henrymercer: I think a new tag is probably best for this — IMO we already have too many statuses (`aborted` comes to mind), and we have made mistakes in the past where we have inadvertently excluded statuses from monitors.

Since adding a new tag and tracking it through the status reports requires a few coordinated changes, I'll add a prefix to the error message so we have some way of easily finding these errors in the meantime.

src/status-report.ts
henrymercer: - Yes
- "unhandled" sounds better than "unexpected" 👍
- Good idea

src/status-report.ts
```diff
@@ -630,10 +630,17 @@ export async function sendUnhandledErrorStatusReport(
     );
     if (statusReport !== undefined) {
       await sendStatusReport(statusReport);
+    } else if (isInTestMode()) {
+      throw new Error(
+        "Failed to create status report for unhandled error in test mode.",
+      );
```
mbg: Intuitively this is unnecessary, because `createStatusReportBase` throws if `isInTestMode()` is `true`, so we should never have the `statusReport === undefined` case there. 

Did you add this deliberately to catch a potential scenario where `createStatusReportBase` doesn't throw in test mode, but `statusReport` is `undefined` anyway?

src/status-report.ts
```diff
@@ -630,10 +630,17 @@ export async function sendUnhandledErrorStatusReport(
...
     }
   } catch (e) {
     logger.warning(
       `Failed to send the error status report: ${getErrorMessage(e)}.`,
```
mbg: Minor: Perhaps add "unhandled" before "error status report", so it's more obvious what this log message relates to.

src/status-report.ts
```diff
@@ -616,15 +616,17 @@ export async function sendUnhandledErrorStatusReport(
...
     const statusReport = await createStatusReportBase(
       actionName,
       "failure",
       actionStartedAt,
       undefined,
       undefined,
       logger,
-      undefined,
-      getErrorMessage(error),
+      `Unhandled error: ${getErrorMessage(error)}`,
```
mbg: Minor: Maybe change this to `"Unhandled CodeQL Action error"` or `"Unhandled ${actionName} error"` so it's a less generic prefix / less likely to clash with some error message we might get from elsewhere.

src/status-report.ts
henrymercer: > Did you add this deliberately to catch a potential scenario where `createStatusReportBase` doesn't throw in test mode, but `statusReport` is `undefined` anyway?

Yes, but this is probably overly defensive.  I'll remove it.

src/status-report.ts
henrymercer: I'd originally held off since this is customer facing, but it's probably not confusing in retrospect.

PR #3420, opened 2026-01-23
Simplify computation of job status

henrymercer:
Refactor how we compute job status so this happens in a single place, and whether or not we have a CodeQL config.  Also add comments to clarify what happens in the case that the CodeQL config is absent.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,10 +8,7 @@ updates:
       - Rebuild
     # Ignore incompatible dependency updates
     ignore:
-      # There is a type incompatibility issue between v0.0.9 and our other dependencies.
-      - dependency-name: "@octokit/plugin-retry"
-        versions: ["~6.0.0"]
-      # This is broken due to the way configuration files have changed. 
+      # This is broken due to the way configuration files have changed.
       # This might be fixed when we move to eslint v9.
       - dependency-name: "eslint-plugin-import"
         versions: [">=2.30.0"]
--- a/.github/workflows/__build-mode-autobuild.yml
+++ b/.github/workflows/__build-mode-autobuild.yml
@@ -76,6 +76,14 @@ jobs:
         with:
           java-version: ${{ inputs.java-version || '17' }}
           distribution: temurin
+      - name: Install yq
+        if: runner.os == 'Windows'
+        env:
+          YQ_PATH: ${{ runner.temp }}/yq
+          YQ_VERSION: v4.50.1
+        run: |-
+          gh release download --repo mikefarah/yq --pattern "yq_windows_amd64.exe" "$YQ_VERSION" -O "$YQ_PATH/yq.exe"
+          echo "$YQ_PATH" >> "$GITHUB_PATH"
       - name: Set up Java test repo configuration
         run: |
           mv * .github ../action/tests/multi-language-repo/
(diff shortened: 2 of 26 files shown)


src/init-action-post.ts
```diff
@@ -139,6 +141,51 @@ async function run(startedAt: Date) {
   }
 }
 
+function getFinalJobStatus(config: Config | undefined): JobStatus {
```
mbg, line 151: Minor: The previous `getFinalJobStatus` had some documentation, but this has disappeared here. It would be good to restore/update this if possible.

Alternatively/additionally, this function performs to separate tasks:

1. It checks whether `EnvVar.JOB_STATUS` is set and, if so, converts its value to a `JobStatus` value.
2. If not, it tries to determine the job status.

It might make sense to break these tasks up into two functions (with one calling the other at the end) with corresponding names. That could make it easier to name, document, and test as well.

src/init-action-post.ts
```diff
@@ -139,6 +141,51 @@ async function run(startedAt: Date) {
...
+    // Validate the job status from the environment. If it is invalid, return unknown.
+    if (
+      Object.values(JobStatus).includes(jobStatusFromEnvironment as JobStatus)
+    ) {
+      return jobStatusFromEnvironment as JobStatus;
+    }
+    return JobStatus.UnknownStatus;
+  }
+
+  let jobStatus: JobStatus;
```
mbg, line 157: It would be nice to have a unit test for the part of this function from here onwards.

src/init-action-post-helper.ts
```diff
@@ -129,48 +128,32 @@ export async function tryUploadSarifIfRunFailed(
...
+    );
+  } catch (e) {
+    logger.debug(
+      `Failed to upload a SARIF file for this failed CodeQL code scanning run. ${e}`,
+    );
+    return createFailedUploadFailedSarifResult(e);
+  }
 }
 
 export async function run(
```
mbg, line 158: Not part of this PR, but why do we have a `run` function here? It's a bit confusing that we typically have `run` functions in the `-action[-post].ts` entry points, and here we have one in a module that's not that. We also have a `run` function call another `run` function.

src/init-action-post-helper.ts
```diff
@@ -129,48 +128,32 @@ export async function tryUploadSarifIfRunFailed(
...
-      );
-      return createFailedUploadFailedSarifResult(e);
-    }
-  } else {
-    core.exportVariable(
-      EnvVar.JOB_STATUS,
-      process.env[EnvVar.JOB_STATUS] ?? JobStatus.SuccessStatus,
-    );
+  // If the only enabled analysis kind is `code-quality`, then we shouldn't
+  // upload the failed SARIF to Code Scanning.
```
mbg: Minor: Not a change in this PR, but this comment should probably be reworded to say "If `code-scanning` is not an enabled analysis kind, [..]". We probably made a change at some point to have more robust logic here, but didn't update the comment.

src/init-action-post.ts
```diff
@@ -85,7 +87,7 @@ async function run(startedAt: Date) {
         logger,
       );
 
-      // If we are analysing the default branch and some kind of caching is enabled,
+      // If we are analyzing the default branch and some kind of caching is enabled,
```
mbg, line 90: Minor: This seems like an unnecessary and unrelated change 😅

henrymercer: I have a spell checker in my IDE that was complaining, nothing personal 😆

src/init-action-post-helper.ts
henrymercer: I agree this is suboptimal, but will leave this as is for now to avoid scope creep.

src/init-action-post.ts
henrymercer: I'll leave this as is for now.

PR #3421, opened 2026-01-23
Tolerate errors loading repository properties

henrymercer:
Repository properties can be used to configure the analysis.  Therefore in general we probably don't want to tolerate errors loading repository properties, because this could mean that we run a different set of queries to what was requested, for instance.

However in the first stages of the rollout, tolerate errors.

Also, don't attempt to load repository properties for repositories owned by users, as these cannot have custom properties.

### Risk assessment

Workflow types:

Products:

Environments:

I've tested it on this repository, and will run another test on a repo owned by a user.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1307,12 +1307,12 @@ export class Result<T, E> {
     public readonly value: T | E,
   ) {}
 
-  /** A success result. */
+  /** Creates a success result. */
   static success<T>(value: T): Success<T> {
     return new Result(true, value) as Success<T>;
   }
 
-  /** A failure result. */
+  /** Creates a failure result. */
   static failure<E>(value: E): Failure<E> {
     return new Result(false, value) as Failure<E>;
   }
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -86563,11 +86563,11 @@ var Result = class _Result {
     this._ok = _ok;
     this.value = value;
   }
-  /** A success result. */
+  /** Creates a success result. */
   static success(value) {
     return new _Result(true, value);
   }
-  /** A failure result. */
+  /** Creates a failure result. */
   static failure(value) {
     return new _Result(false, value);
   }


init/action.yml
```diff
@@ -159,6 +159,9 @@ inputs:
     description: >-
       Explicitly enable or disable caching of project build dependencies.
     required: false
+  repository-owner-type:
+    default: ${{ github.event.repository.owner.type }}
```
mbg: This seems like a reasonable solution that avoids an (additional) API call.

To check: whether `repository` is available for all event types we care about (especially `dynamic`).

Alternative ideas:

- Query the API, but we'd want to avoid that if it incurs an extra request. For this option, ideally we'd get this out of a call we have to make anyway.
- Check the status code of the response. If we get a 404, then it's probably because it is a user-owned repo.

src/init-action.ts
```diff
@@ -775,6 +796,41 @@ async function run(startedAt: Date) {
   );
 }
 
+async function loadRepositoryProperties(
```
mbg, line 805: Move this to `init.ts` or `properties.ts`, add a JSDoc, and some unit tests?

src/init-action.ts
```diff
@@ -237,12 +242,12 @@ async function run(startedAt: Date) {
     );
 
     // Fetch the values of known repository properties that affect us.
-    const enableRepoProps = await features.getValue(
-      Feature.UseRepositoryProperties,
+    const repositoryProperties = await loadRepositoryProperties(
```
mbg: Minor: Name this to indicate that it isn't a repository properties object itself, but a `Result` that may contain one? E.g. `repositoryPropertiesResult`.

src/init-action.ts
```diff
@@ -363,10 +368,26 @@ async function run(startedAt: Date) {
...
+      repositoryProperties: repositoryProperties.orElse({}),
       logger,
     });
 
+    if (repositoryProperties.isError()) {
+      addDiagnostic(
+        config,
+        // Arbitrarily choose the first language. We could also choose all languages, but that
+        // increases the risk of misinterpreting the data.
+        config.languages[0],
```
mbg: Minor: We keep using this pattern; time to add a function for it so we don't have to repeat it / the comment every time?

src/util.ts
```diff
@@ -1292,3 +1292,33 @@ export function joinAtMost(
...
+  isOk(): this is Success<T> {
+    return this._ok;
+  }
+
+  isError(): this is Failure<E> {
+    return !this._ok;
+  }
+
+  orElse(defaultValue: T): T {
+    return this._ok ? (this.value as T) : defaultValue;
```
mbg: Would TS accept `isOk()` instead of `this._ok` and narrow `this.value` to `T` so we don't need the `as T`?

src/util.ts
```diff
@@ -1292,3 +1292,33 @@ export function joinAtMost(
 
   return array.join(separator);
 }
+
+type Success<T> = Result<T, never>;
+type Failure<E> = Result<never, E>;
+
+export class Result<T, E> {
```
mbg, line 1304: I like this -- it's a good idea to start including errors in the types.

Would you mind adding docs and unit tests for this?

init/action.yml
mbg: I just noticed that, in e.g. `getPullRequestBranches()`, we seem to just be able to access the `github` context in the action. So we might be able to do that here too without the extra input?

src/init-action.ts
```diff
@@ -802,7 +803,7 @@ async function loadRepositoryProperties(
   features: FeatureEnablement,
   logger: Logger,
 ): Promise<Result<RepositoryProperties, unknown>> {
-  const repositoryOwnerType = getOptionalInput("repository-owner-type");
+  const repositoryOwnerType = github.context.payload.repository?.owner.type;
```
mbg, line 814: Nice to see that this works! Have you checked whether this is available for `dynamic` workflows or if we'd need a different solution there? 

I suppose the logic here is that we attempt the API call by default, unless we can show that the repo is owned by a user, so it will work one way or another in e.g. Default Setup. It would still be nice to know whether we can skip the API call for user repos there or rely on the brute-force attempt.

src/util.ts
```diff
@@ -1293,32 +1293,42 @@ export function joinAtMost(
...
+ * A simple result type representing either a success or a failure.
+ */
 export class Result<T, E> {
   private constructor(
     private readonly _ok: boolean,
     public readonly value: T | E,
   ) {}
 
-  static ok<T>(value: T): Success<T> {
+  /** A success result. */
```
mbg: Minor: These are the same comments as for the types, but these are actually functions constructing values of those types. How about "Constructs a success result", and similarly for the failure case below?

mbg (dismissed):
Thank you for addressing all of my feedback! 🏆 

This looks good to me. I added a couple of minor comments. Not blocking, but one to think about with respect to `dynamic` workflows if you haven't checked yet.

src/init-action.ts
henrymercer: I have on a test repo — I'll DM you the link.

PR #3422, opened 2026-01-24
Warn if a private registry configuration uses a PAT, but has no username

mbg:
To authenticate using a PAT, the corresponding username is required. Using a PAT on its own, without the corresponding username, does not work. 

This PR modifies the `start-proxy` action to check whether a configuration seems to use a GitHub PAT without a username and issue a warning in that case.

In the process, I also taught the artifact scanner about fine-grained tokens (start with `github_pat_` and are longer than classic PATs) and extended the tests there.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/artifact-scanner.ts
+++ b/src/artifact-scanner.ts
@@ -88,7 +88,7 @@ export function isAuthToken(
   patterns: TokenPattern[] = GITHUB_TOKEN_PATTERNS,
 ) {
   for (const { type, pattern } of patterns) {
-    if (pattern.test(value)) {
+    if (value.match(pattern)) {
       return type;
     }
   }
--- a/lib/start-proxy-action.js
+++ b/lib/start-proxy-action.js
@@ -103998,7 +103998,7 @@ var GITHUB_TOKEN_PATTERNS = [
 ];
 function isAuthToken(value, patterns = GITHUB_TOKEN_PATTERNS) {
   for (const { type: type2, pattern } of patterns) {
-    if (pattern.test(value)) {
+    if (value.match(pattern)) {
       return type2;
     }
   }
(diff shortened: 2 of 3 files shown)


src/artifact-scanner.ts
```diff
@@ -48,6 +77,24 @@ interface ScanResult {
...
+ * @param value The value to match against.
+ * @param patterns The patterns to check.
+ * @returns The type of the first matching pattern, or `undefined` if none match.
+ */
+export function isAuthToken(
+  value: string,
+  patterns: TokenPattern[] = GITHUB_TOKEN_PATTERNS,
+) {
+  for (const { type, pattern } of patterns) {
+    if (pattern.test(value)) {
```
henrymercer: It would be safer to use `value.match(pattern)`, so that the thing that makes successive `exec`s return successive matches, `lastIndex`, is reset.  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#using_test_on_a_regex_with_the_global_flag and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/Symbol.match#description

src/start-proxy.ts
```diff
@@ -161,6 +169,19 @@ export function getCredentials(
...
     }
 
+    // If the password or token looks like a GitHub PAT, warn if no username is configured.
+    if (
+      !isDefined(e.username) &&
+      ((isDefined(e.password) && isPAT(e.password)) ||
+        (isDefined(e.token) && isPAT(e.token)))
+    ) {
+      logger.warning(
+        `A ${e.type} private registry is configured for ${e.host || e.url} using a GitHub Personal Access Token (PAT), but no username was provided. ` +
```
henrymercer, line 179: Can we be sure that the URL will not contain any credentials?

mbg: Only the password/token fields are encrypted in the backend, so users shouldn't be storing any credentials in the host/url fields. (I believe the UI says something to that effect as well when configuring a registry.)

In any case, we have always been logging the host/url values in `credentialToStr`, which is always called by:

```typescript
    logger.info(
      `Credentials loaded for the following registries:\n ${credentials
        .map((c) => credentialToStr(c))
        .join("\n")}`,
    );
```

PR #3423, opened 2026-01-24
Add `installYq` option to `sync.py` and install `yq` directly from GitHub release

mbg:
On Windows, `choco install yq` routinely fails. For some reason, this doesn't cause the step itself to fail, but does cause obvious failures later on in the workflow.

Under the hood, `choco install` just queries a feed at `https://community.chocolatey.org/api/v2/` for where to download a suitable binary for `yq` from, which turns out to be a GitHub release. 

This PR changes the workflow to sidestep `choco` entirely and just download `yq` directly from the release that's currently used by `choco`.

Note for reviewers: in the first commit, I attempted to cache the data for `choco install` instead, but I discovered that `choco` doesn't really support caching to the point where it doesn't have to query the feed / redownloads the files.

### Risk assessment

Environments:

- **Other** - Please provide details. CI outcome.

### DIFF
--- a/pr-checks/sync.py
+++ b/pr-checks/sync.py
@@ -230,10 +230,14 @@ for file in sorted((this_dir / 'checks').glob('*.yml')):
             'name': 'Install yq',
             'if': "runner.os == 'Windows'",
             'env': {
-                'YQ_PATH': '${{ runner.temp }}/yq'
+                'YQ_PATH': '${{ runner.temp }}/yq',
+                # This is essentially an arbitrary version of `yq`, which happened to be the one that
+                # `choco` fetched when we moved away from using that here.
+                # See https://github.com/github/codeql-action/pull/3423
+                'YQ_VERSION': 'v4.50.1'
             },
             'run': LiteralScalarString(
-                'gh release download --repo mikefarah/yq --pattern "yq_windows_amd64.exe" v4.50.1 -O "$YQ_PATH/yq.exe"\n'
+                'gh release download --repo mikefarah/yq --pattern "yq_windows_amd64.exe" "$YQ_VERSION" -O "$YQ_PATH/yq.exe"\n'
                 'echo "$YQ_PATH" >> "$GITHUB_PATH"'
             ),
         })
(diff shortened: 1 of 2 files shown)


pr-checks/sync.py
```diff
@@ -223,6 +223,21 @@ def writeHeader(checkStream):
...
+
+    if installYq:
+        steps.append({
+            'name': 'Install yq',
+            'if': "runner.os == 'Windows'",
+            'env': {
+                'YQ_PATH': '${{ runner.temp }}/yq'
+            },
+            'run': LiteralScalarString(
+                'gh release download --repo mikefarah/yq --pattern "yq_windows_amd64.exe" v4.50.1 -O "$YQ_PATH/yq.exe"\n'
```
esbena:
```suggestion
                'gh release download --repo mikefarah/yq --pattern "yq_windows_amd64.exe" "$YQ_VERSION" -O "$YQ_PATH/yq.exe"'
```
I kind of agree with copilot. 
Let's move the version out to a variable, and add and add an explanation and a direct link to the release as a comment.

Also, the trailing `\n` is confusing to read. Could we do something cleaner syntactically, or is this a workflow builder limitation?

mbg: > Let's move the version out to a variable, and add and add an explanation and a direct link to the release as a comment.

Sure, I can do that.

> Also, the trailing `\n` is confusing to read. Could we do something cleaner syntactically, or is this a workflow builder limitation?

There might be something that's cleaner syntactically in Python, but this was the sanest option I came up with since Python multi-line strings mostly led to results with weird spacing. I am open to suggestions.

mbg:
> Do you happen to know why this isn't done through https://github.com/frenck/action-setup-yq (or similar) 

I suppose it's to avoid adding extra layers of dependencies, particularly on third-party actions. 

> so dependabot would have a theoretical chance?

Dependabot would be able to update the (new) third-party action, but not the version of `yq` we use. If we don't add an extra dependency here, then there's nothing for Dependabot to do.

Maybe if we stuck with `choco` and pinned a particular version of `yq` in a config file, Dependabot would be able to update that for us, but we previously let `choco` just pull whatever (latest) version it wanted anyway.

---

I think the bottom line here is that this is a minor tool dependency that (ideally) would come pre-installed on Windows runners. This is the only workflow that currently runs on Windows and needs it, and we only have a very basic use for it. I don't think we'd want to add extra maintenance overhead or a particularly complex solution to maintain this. The main goal of this PR is just to avoid the intermittent workflow failures.

PR #3424, opened 2026-01-26
Add feature flag to skip computing baseline file coverage information on PRs

henrymercer:
- Right now, this just skips computing baseline file coverage, but as follow up, we'll change this to also skip running the CodeQL queries to determine actual file coverage.
- We don't plan to roll this out externally for now.  `codeql database interpret-results` currently expects baseline information to always be present, and displays a log message "The database provided was made with a CLI version before 2.11.2 which did not record file baseline information, so file baseline information will be absent." when it is missing.  We'll address this and put this feature behind a tools feature flag.  At this point we'll also add a changelog entry.  I'm proposing merging this now to make more incremental code changes and to allow us to test this internally.
- There are two items of drive-by cleanup: removing an unused CLI function, and updating the version of the `caniuse-lite` package to avoid a pesky warning during builds.

### Risk assessment

Workflow types:

Products:

Environments:

See https://github.com/github/codeql-action/actions/runs/21356271329

### DIFF
--- a/src/init-action.ts
+++ b/src/init-action.ts
@@ -13,7 +13,6 @@ import {
   getOptionalInput,
   getRequiredInput,
   getTemporaryDirectory,
-  isAnalyzingPullRequest,
   persistInputs,
 } from "./actions-util";
 import { AnalysisKind, getAnalysisKinds } from "./analyses";
@@ -37,12 +36,13 @@ import {
   makeTelemetryDiagnostic,
 } from "./diagnostics";
 import { EnvVar } from "./environment";
-import { Feature, FeatureEnablement, Features } from "./feature-flags";
+import { Feature, Features } from "./feature-flags";
 import { loadPropertiesFromApi } from "./feature-flags/properties";
 import {
   checkInstallPython311,
   checkPacksForOverlayCompatibility,
   cleanupDatabaseClusterDirectory,
+  getFileCoverageInformationEnabled,
   initCodeQL,
   initConfig,
   runDatabaseInitCluster,
@@ -54,7 +54,7 @@ import {
   OverlayBaseDatabaseDownloadStats,
   OverlayDatabaseMode,
 } from "./overlay-database-utils";
-import { getRepositoryNwo, RepositoryNwo } from "./repository";
+import { getRepositoryNwo } from "./repository";
 import { ToolsSource } from "./setup-codeql";
 import {
   ActionName,
@@ -792,24 +792,6 @@ function getTrapCachingEnabled(): boolean {
 
   // On hosted runners, enable TRAP caching by default
   return true;
-}
-
-async function getFileCoverageInformationEnabled(
-  debugMode: boolean,
-  repositoryNwo: RepositoryNwo,
-  features: FeatureEnablement,
-): Promise<boolean> {
-  return (
-    // Always enable file coverage information in debug mode
-    debugMode ||
-    // We're most interested in speeding up PRs, and we want to keep
-    // submitting file coverage information for the default branch since
-    // it is used to populate the status page.
-    !isAnalyzingPullRequest() ||
-    // For now, restrict this feature to the GitHub org
-    repositoryNwo.owner !== "github" ||
-    !(await features.getValue(Feature.SkipFileCoverageOnPrs))
-  );
 }
 
 async function recordZstdAvailability(
--- a/lib/init-action.js
+++ b/lib/init-action.js
@@ -92028,6 +92028,16 @@ function cleanupDatabaseClusterDirectory(config, logger, options = {}, rmSync2 =
     }
   }
 }
+async function getFileCoverageInformationEnabled(debugMode, repositoryNwo, features) {
+  return (
+    // Always enable file coverage information in debug mode
+    debugMode || // We're most interested in speeding up PRs, and we want to keep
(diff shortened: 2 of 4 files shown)


src/init-action.ts
```diff
@@ -364,6 +366,16 @@ async function run(startedAt: Date) {
...
+      enableFileCoverageInformation:
+        // Always enable file coverage information in debug mode
+        debugMode ||
+        // We're most interested in speeding up PRs, and we want to keep
+        // submitting file coverage information for the default branch since
+        // it is used to populate the status page.
+        !isAnalyzingPullRequest() ||
+        // For now, restrict this feature to the GitHub org
+        repositoryNwo.owner !== "github" ||
+        !(await features.getValue(Feature.SkipFileCoverageOnPrs)),
```
mbg: Could you move this to a separate function and add some unit tests for it?

src/analyze.ts
```diff
@@ -495,10 +495,18 @@ export async function runQueries(
...
-      if (qualityAnalysisSummary) {
+      if (analysisSummary.trim()) {
+        logger.info(analysisSummary);
+      }
+      if (qualityAnalysisSummary?.trim()) {
         logger.info(qualityAnalysisSummary);
       }
+      if (!config.enableFileCoverageInformation) {
+        logger.info(
+          "File coverage information is disabled for this PR analysis for performance reasons. " +
```
mbg: Minor: the wording of this could be interpreted as being specific to a given PR, rather than PRs in general. It could lead a user to ask why it is disabled for their PR specifically. Can we word this differently?

For example, if we (eventually) plan to do this for all pull requests:

```suggestion
          "File coverage information is disabled for pull requests. " +
```

src/analyze.ts
```diff
@@ -495,10 +495,18 @@ export async function runQueries(
...
+            "It will still be enabled for analyses triggered by a push or a schedule.",
```
mbg: The future tense here suggests that this is something we will implement in the future, but that wouldn't happen right now. 

Additionally, "by a push" is misleading, because file coverage information would not be generated if a user _pushed_ a commit to a PR branch. You're talking about `push` _events_. Talking about the events may be confusing for `dynamic` workflows though, where these rules don't apply in the same way.

```suggestion
            "It is available for analyses on the main branch.",
```

src/codeql.ts
```diff
@@ -628,6 +620,13 @@ async function getCodeQLForCmd(
         extraArgs.push("--overlay-base");
       }
 
+      const baselineFilesOptions = config.enableFileCoverageInformation
+        ? [
+            "--calculate-language-specific-baseline",
+            "--sublanguage-file-coverage",
```
mbg, line 626: I think this is supported by all versions of GHES that we support now (`GitHub Enterprise Server v3.12.0+` according to the CLI). A few questions about this:

- I suppose that once we have a tools feature for `enableFileCoverageInformation`, that it implies the availability of `--sublanguage-file-coverage` so we don't need any other checks here.
- Are there any downstream implications of this option that we need to / should account for?
- Any tests that need updating, or should be added?

src/codeql.ts
```diff
@@ -639,12 +638,11 @@ async function getCodeQLForCmd(
...
           config.dbLocation,
           `--source-root=${sourceRoot}`,
-          "--calculate-language-specific-baseline",
+          ...baselineFilesOptions,
           "--extractor-include-aliases",
-          "--sublanguage-file-coverage",
           ...extraArgs,
           ...getExtraOptionsFromEnv(["database", "init"], {
-            ignoringOptions: ["--overwrite"],
+            ignoringOptions: ["--overwrite", ...baselineFilesOptions],
```
mbg, line 648: Could we add a comment here for why these options in particular cannot be overridden?

src/init-action.ts
henrymercer: Moved to separate function 👍 Since it's just a logical expression, I don't think we can unit test this in a meaningful way.

src/analyze.ts
henrymercer: Good feedback, reworded.

src/analyze.ts
henrymercer: Good feedback, reworded.

src/codeql.ts
henrymercer: For the avoidance of doubt, these are not new options and the only change we are making is to stop passing these behind a feature flag when we want to disable file coverage information.

I've updated the file baseline PR check to disable the FF so this continues working when we flip the FF for this repo.

src/codeql.ts
henrymercer: 👍 Done

henrymercer:
> Should we also add some telemetry here to measure what impact this has?

Agreed.  I think it'd make sense to do that in the CLI.

src/codeql.ts
mbg: I totally missed that `--sublanguage-file-coverage` was on L644 before, sorry! In that case, nevermind :)

src/init-action.ts
mbg: My reasoning here was that, while reviewing, I found it time-consuming to check that this expression produces the correct result in all cases. A short unit test which does that instead would be less error-prone and robust when making changes here down the line.

PR #3427, opened 2026-01-26
Bump the Action minor version number on new CodeQL minor version series

henrymercer:
- Each CodeQL minor version series corresponds to a GHES version
- Each Action ships with a pinned version of the CodeQL bundle that is used on GHES
- Each GHES version has an associated Action version that pins the last CodeQL bundle for that minor version series

That handles the normal case, however what happens if we want to make a backport to GHES after we've already started the minor version series?

We need to release a new version of the Action based off that Action version pinning the CodeQL bundle associated with that GHES release.

This PR ensures we can assign a sensible version number to that version, by bumping the minor version number of the Action on each new CodeQL minor version series.

Suppose the version that shipped with CLI v2.23.9 was v4.31.11.  Then this change makes sure the version that ships with CLI v2.24.0 will be v4.32.0 (or later).  This means we can create a GHES backport that continues to pin CLI v2.23.9 and number this v4.31.12. 

### Risk assessment

Workflow types:

Products:

Environments:

We'll find out because the bundle upgrade workflow will fail or the PR will be wrong.

### DIFF
--- a/.github/workflows/update-bundle.yml
+++ b/.github/workflows/update-bundle.yml
@@ -57,17 +57,21 @@ jobs:
       - name: Update bundle
         uses: ./.github/actions/update-bundle
 
-      - name: Bump action minor version for new CodeQL minor version series
+      - name: Bump Action minor version if new CodeQL minor version series
         id: bump-action-version
         run: |
+          prior_cli_version=$(jq -r '.priorCliVersion' src/defaults.json)
           cli_version=$(jq -r '.cliVersion' src/defaults.json)
-          # Check if this is a new minor version series (patch version is 0)
-          if [[ "$cli_version" =~ ^[0-9]+\.[0-9]+\.0$ ]]; then
-            echo "New CodeQL minor version series detected ($cli_version), bumping action minor version"
+
+          prior_minor=$(echo "$prior_cli_version" | cut -d. -f2)
+          current_minor=$(echo "$cli_version" | cut -d. -f2)
+
+          if [[ "$current_minor" != "$prior_minor" ]]; then
+            echo "New CodeQL minor version series ($prior_cli_version -> $cli_version), bumping Action minor version"
             npm version minor --no-git-tag-version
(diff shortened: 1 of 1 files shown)


.github/workflows/update-bundle.yml
```diff
@@ -57,6 +57,20 @@ jobs:
       - name: Update bundle
         uses: ./.github/actions/update-bundle
 
+      - name: Bump action minor version for new CodeQL minor version series
+        id: bump-action-version
+        run: |
+          cli_version=$(jq -r '.cliVersion' src/defaults.json)
+          # Check if this is a new minor version series (patch version is 0)
+          if [[ "$cli_version" =~ ^[0-9]+\.[0-9]+\.0$ ]]; then
```
mbg: This logic breaks if we ever need to skip a `.0` CLI release and go straight to `.1` or higher. I think ideally we'd extract the minor CLI version from `src/defaults.json` before the `./.github/actions/update-bundle` step and then compare it to the one here. Not urgent, so feel free to do that in a follow-up PR.

.github/workflows/update-bundle.yml
```diff
@@ -71,11 +85,21 @@ jobs:
...
+          ACTION_VERSION_BUMPED: ${{ steps.bump-action-version.outputs.bumped }}
         run: |
           cli_version=$(jq -r '.cliVersion' src/defaults.json)
+          action_version=$(jq -r '.version' package.json)
+
+          pr_body="This pull request updates the default CodeQL bundle, as used with \`tools: linked\` and on GHES, to $cli_version."
+          if [[ "$ACTION_VERSION_BUMPED" == "true" ]]; then
+            pr_body="$pr_body
+
+          Since this is a new CodeQL minor version series, this PR also bumps the Action version to $action_version."
```
mbg: Minor: This is a bit horrible. Perhaps in a follow-up, see if you can make this nicer

PR #3437, opened 2026-01-28
Report queried feature flags in status reports

mbg:
It is useful to know which feature flags are affecting a particular CodeQL workflow. Although this is typically determined by querying the `/features` API, it can also be influenced by environment variables, CLI versions, etc.

This PR modifies the `Features` class to track which feature flags are queried during a particular step and what they evaluated to. This data is then included in status reports when possible.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/lib/analyze-action.js
+++ b/lib/analyze-action.js
@@ -93747,9 +93747,12 @@ async function createStatusReportBase(actionName, status, actionStartedAt, confi
       core12.exportVariable("CODEQL_ACTION_TESTING_ENVIRONMENT" /* TESTING_ENVIRONMENT */, testingEnvironment);
     }
     const isSteadyStateDefaultSetupRun = process.env["CODE_SCANNING_IS_STEADY_STATE_DEFAULT_SETUP"] === "true";
-    const featureFlags = Object.entries(
-      queriedFeatures || {}
-    ).map(([feature, outcome]) => ({ feature, ...outcome }));
+    let featureFlags = void 0;
+    if (queriedFeatures) {
+      featureFlags = Object.entries(queriedFeatures).map(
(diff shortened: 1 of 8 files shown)


mbg:
I think those are essentially two separate points:

1. Reporting all FFs.
    - Not inherently useful, because we know that a FF that isn't reported wasn't relevant for a particular step.
    - This would entail some overhead in calling `getValue` for all FFs. (Also means we require a `CodeQL` instance to query them all.)
    - Possibly some misleading results if a FF is set for only a particular step using an env var, but not others. 
1. Understanding whether a FF was queried in an earlier step.
    - This is useful to know and would already be possible with the approach taken in this PR, although somewhat complicated to query since we'd have to find the matching earlier step that we care about.
    - Alternatively, we could save the queried FFs between steps and report them all.
    - That might lead to misleading results again if a FF is set for only a particular step using an env var, but not others. 
    
The approach I went with here is the most conservative in terms of not really performing any extra work and only reporting what was current for a given step, while getting the actual value of a FF for a given step.

henrymercer:
> 1. Not inherently useful, because we know that a FF that isn't reported wasn't relevant for a particular step.

I don't think this is true.  There are a bunch of flags that we get the value of in `init`, save to the config or database, and use in `analyze`.

> 1. This would entail some overhead in calling `getValue` for all FFs. (Also means we require a `CodeQL` instance to query them all.)

We save the feature flags API response to disk.  If there's overhead in getting the value of environment variables, we could cache those too since feature flags should be stable across a job.

> 1. Possibly some misleading results if a FF is set for only a particular step using an env var, but not others.

That's fair — while this is a bad pattern, it's very possible that there are workflows out there that do this.

PR #3444, opened 2026-01-29
Report some types of errors in `start-proxy` status reports

mbg:
This PR allows the `start-proxy` action to report some types of errors in status reports. The basic idea here is that we have a `StartProxyErrorType` enum to represent types of failures we care about, with associated, fixed error messages. We then have a `StartProxyError` type for exceptions which only accepts members of the `StartProxyErrorType` enum as messages.

Most of the size of this PR comes from refactoring functions from `start-proxy-action` to `start-proxy` to make it easier to handle the individual error cases, and to add better unit test coverage.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/start-proxy.test.ts
+++ b/src/start-proxy.test.ts
@@ -391,20 +391,35 @@ test("getDownloadUrl returns matching release asset", async (t) => {
   t.is(info.url, "url-we-want");
 });
 
-test("credentialToStr - hides passwords/tokens", (t) => {
+test("credentialToStr - hides passwords", (t) => {
   const secret = "password123";
   const credential = {
     type: "maven_credential",
+    password: secret,
   };
-  t.false(
-    startProxyExports
-      .credentialToStr({ password: secret, ...credential })
-      .includes(secret),
+
+  const str = startProxyExports.credentialToStr(credential);
+
+  t.false(str.includes(secret));
+  t.is(
+    "Type: maven_credential; Host: undefined; Url: undefined Username: undefined; Password: true; Token: false",
+    str,
   );
-  t.false(
-    startProxyExports
-      .credentialToStr({ token: secret, ...credential })
-      .includes(secret),
+});
+
+test("credentialToStr - hides tokens", (t) => {
+  const secret = "password123";
+  const credential = {
+    type: "maven_credential",
+    token: secret,
+  };
+
+  const str = startProxyExports.credentialToStr(credential);
+
+  t.false(str.includes(secret));
+  t.is(
+    "Type: maven_credential; Host: undefined; Url: undefined Username: undefined; Password: false; Token: true",
+    str,
   );
 });
 


src/start-proxy.test.ts
```diff
@@ -1,21 +1,72 @@
...
   checkExpectedLogMessages,
   getRecordingLogger,
   makeTestToken,
   setupTests,
+  withRecordingLoggerAsync,
 } from "./testing-utils";
 
 setupTests(test);
 
+test("sendFailedStatusReport - does not report arbitrary error messages", async (t) => {
```
henrymercer: From what I can tell, the idea behind this test is that while the `DownloadFailed` error message would normally be regarded as safe, here the error object is not an instance of `StartProxyError` and it is therefore rejected.

I think it would be clearer to have a test that throws a different error message, and verifies the message that is sent to `createStatusReportBase` precisely (i.e. checking for "Error from start-proxy Action omitted" rather than the full original error message not being present).

Also, it might be worth adding a similar test for the top-level `runWrapper`.  Note we'd have to follow the pattern of `analyze-action` in having one file per test case.

src/start-proxy.test.ts
```diff
@@ -301,3 +352,187 @@ test("getDownloadUrl returns matching release asset", async (t) => {
...
+  t.false(
+    startProxyExports
+      .credentialToStr({ password: secret, ...credential })
+      .includes(secret),
+  );
+  t.false(
+    startProxyExports
+      .credentialToStr({ token: secret, ...credential })
+      .includes(secret),
+  );
```
henrymercer: I would personally feel more confident about checking what the output is rather than what it isn't, as this avoids situations where for whatever reason we include a substring of the sensitive data.

src/start-proxy.ts
```diff
@@ -1,12 +1,141 @@
...
+ * i.e. that does not contain sensitive information.
+ *
+ * @param error The error for which to get an error message.
+ */
+export function getSafeErrorMessage(error: Error): string {
+  // If the error is a `StartProxyError`, the constructor ensures that the
+  // message comes from `StartProxyErrorType`.
+  if (error instanceof StartProxyError) {
+    return error.message;
+  }
```
henrymercer: This check relies on type safety.  While we try to make sure our code is type safe, that isn't something TypeScript can guarantee, and it would be safer to check the allowed values in `StartProxyErrorType` explicitly.

src/start-proxy.ts
```diff
@@ -1,12 +1,141 @@
...
+  language: KnownLanguage | undefined,
+  unwrappedError: unknown,
+) {
+  const error = util.wrapError(unwrappedError);
+  core.setFailed(`start-proxy action failed: ${error.message}`);
+
+  const statusReportMessage = getSafeErrorMessage(error);
+
+  // We skip sending the error message and stack trace here to avoid the possibility
+  // of leaking any sensitive information into the telemetry.
```
henrymercer: Nit: this needs a small update now that we report known safe error messages.

src/start-proxy.ts
```diff
@@ -277,3 +406,124 @@ export async function getDownloadUrl(
...
+
+/**
+ * Pretty-prints a `Credential` value to a string, but hides the actual password or token values.
+ *
+ * @param c The credential to convert to a string.
+ */
+export function credentialToStr(c: Credential): string {
+  return `Type: ${c.type}; Host: ${c.host}; Url: ${c.url} Username: ${
+    c.username
+  }; Password: ${c.password !== undefined}; Token: ${c.token !== undefined}`;
```
henrymercer, line 438: Nit: existing code, but as a drive by improvement we could say something like "password present, token absent" rather than "password: true, token: false"

src/start-proxy.ts
```diff
@@ -277,3 +406,124 @@ export async function getDownloadUrl(
...
+ * @param authorization The authorization information to use.
+ * @returns If successful, the path to the downloaded file.
+ */
+export async function downloadProxy(
+  logger: Logger,
+  url: string,
+  authorization: string | undefined,
+) {
+  try {
+    return toolcache.downloadTool(url, undefined, authorization, {
```
henrymercer: Nit: again, existing code I think, but comment the name of the `undefined` input so this code is easier to read outside of an IDE

src/start-proxy.ts
```diff
@@ -277,3 +406,124 @@ export async function getDownloadUrl(
...
+    const extracted = await extractProxy(logger, temp);
+    proxyBin = await cacheProxy(
+      logger,
+      extracted,
+      proxyFileName,
+      proxyInfo.version,
+    );
+  }
+  proxyBin = path.join(proxyBin, proxyFileName);
+  return proxyBin;
```
henrymercer:
```suggestion
  return path.join(proxyBin, proxyFileName);
```

src/start-proxy.ts
mbg: Yeah, this is existing code. The log message format is currently mentioned in [some docs](https://docs.github.com/en/code-security/how-tos/view-and-interpret-data/viewing-code-scanning-logs#determining-whether-code-scanning-default-setup-used-any-private-registries), so I'd rather not change it here now. 

I have a few PRs in-flight which will mostly make that section of the docs obsolete, since each relevant extractor will produce a TSP diagnostic if the private registry config was picked up successfully by it. Once those have shipped, I'll edit that page anyway, and then we can change the format here.

src/start-proxy.test.ts
mbg: > I think it would be clearer to have a test that throws a different error message, and verifies the message that is sent to createStatusReportBase precisely (i.e. checking for "Error from start-proxy Action omitted" rather than the full original error message not being present).

I have added more tests here to cover various cases, and named them differently. 

> Also, it might be worth adding a similar test for the top-level runWrapper. Note we'd have to follow the pattern of analyze-action in having one file per test case.

I have some thoughts on this, but they are outside of the scope of this PR I think.

src/start-proxy.ts
mbg: As discussed, I have changed the approach here slightly so that the `enum` doesn't have the associated string values anymore (and we don't store any string values in the `StartProxyError` objects), but instead store the (numeric) key and then resolve it to an error message in `getSafeErrorMessage`.

henrymercer (approved):

PR #3466, opened 2026-02-10
Test connections to private registries in `start-proxy`

mbg:
Modifies the `start-proxy` action to perform a basic connectivity test for each configured private registry. The outcomes of these tests are logged. The implementation is designed to be tolerant to failures and is gated behind a new FF.

Note that this PR includes the approved changes from #3438 in the first commit.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/start-proxy/reachability.test.ts
+++ b/src/start-proxy/reachability.test.ts
@@ -16,7 +16,7 @@ import { ProxyInfo, Registry } from "./types";
 setupTests(test);
 
 class MockReachabilityBackend implements ReachabilityBackend {
-  public async checkConnection(_registry: Registry): Promise<number> {
+  public async checkConnection(_url: URL): Promise<number> {
     return 200;
   }
 }
@@ -91,6 +91,30 @@ test("checkConnections - handles other exceptions", async (t) => {
     `Successfully tested connection to ${mavenRegistry.url}`,
     `Testing connection to ${nugetFeed.url}`,
     `Connection test to ${nugetFeed.url} failed: Some generic error`,
+    `Finished testing connections`,
+  ]);
+});
+
+test("checkConnections - handles invalid URLs", async (t) => {
+  const backend = new MockReachabilityBackend();
+  const messages = await withRecordingLoggerAsync(async (logger) => {
+    const reachable = await checkConnections(
+      logger,
+      {
+        ...proxyInfo,
+        registries: [
+          {
+            type: "nuget_feed",
+            url: "localhost",
+          },
+        ],
+      },
+      backend,
+    );
+    t.is(reachable.size, 0);
+  });
+  checkExpectedLogMessages(t, messages, [
+    `Skipping check for localhost since it is not a valid URL.`,
     `Finished testing connections`,
   ]);
 });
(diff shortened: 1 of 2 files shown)


src/start-proxy-action.ts
```diff
@@ -133,7 +152,17 @@ async function run(startedAt: Date) {
...
+      proxyBin,
+      proxyConfig,
+      proxyLogFilePath,
+      logger,
+    );
+
+    // Check that the private registries are reachable.
+    if (await features.getValue(Feature.StartProxyConnectionChecks)) {
+      await checkConnections(logger, proxyInfo);
+    }
```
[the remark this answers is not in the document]
mbg: The intent is not to be "disk-only". The FFs are typically loaded by `init` before `start-proxy` is run.

src/start-proxy/reachability.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  const backend = new MockReachabilityBackend();
+  sinon
+    .stub(backend, "checkConnection")
+    .onSecondCall()
+    .throws(new ReachabilityError(400));
+  const messages = await withRecordingLoggerAsync(async (logger) => {
+    const reachable = await checkConnections(logger, proxyInfo, backend);
+    t.is(reachable.size, 1);
+    t.true(reachable.has(mavenRegistry));
+    t.false(reachable.has(nugetFeed));
```
henrymercer: Nit: This check is a bit surprising since we already check the size is 1

src/start-proxy/types.ts
```diff
@@ -0,0 +1,50 @@
...
+export function getAddressString(address: Address): string {
+  if (address.url === undefined) {
+    return address.host;
+  } else {
+    return address.url;
+  }
+}
+
+export type ValidRegistry<T extends Registry = Registry> = T & Address;
+export type ValidCredential = ValidRegistry<Credential>;
```
henrymercer: Minor: what about `RegistryConfig` / `RawRegistry` instead of `Registry`, and `Registry` instead of `ValidRegistry`?  Just so we're calling out when something's invalid rather than calling out when it's valid.

src/start-proxy-action.ts
```diff
@@ -103,9 +110,21 @@ async function run(startedAt: Date) {
     const proxyLogFilePath = path.resolve(tempDir, "proxy.log");
     core.saveState("proxy-log-file", proxyLogFilePath);
 
-    // Get the configuration options
+    // Initialise FFs, but only load them from disk if they are already available.
```
henrymercer: Minor: This could be parenthesised in one of two ways:

- Load them from disk if they're already there, otherwise load them from the API
- Load them from disk if they're already there, otherwise don't load them at all

I think it's the former, but would be nice to make it clearer.

src/start-proxy/types.ts
mbg: I'd prefer that too and wanted to do it that way around, but it was more complicated. I'll have another look at whether I can make that work.

src/start-proxy-action.ts
mbg: I think this comment is accidental and left over from an earlier version of the `start-proxy` FF work. Initially I had an implementation there which would only load the FFs from disk in `start-proxy` and never fetch them, but then I decided that it didn't really make sense to add that complexity (especially since `start-proxy` runs after `init` and so the FFs should already be cached anyway).

I'll update this comment.

PR #3474, opened 2026-02-11
Add `csra` analysis kind

mbg:
First stab at adding this new analysis kind. See the linked internal issues for more details. The new analysis kind largely mirrors `code-scanning`. 

We do not plan this in combination with other analysis kinds, so I added some logic to prevent that.

The SARIF upload for this analysis kind is a bit different than the other two, so I added a `transformPayload` function to the `AnalysisConfig` interface that can change what's included in the payload. I preferred that over having a conditional block near the upload, since that's harder to see and less reusable.

I extended the existing tests for this new analysis kind.

**Notes for reviewers**

This is ready for an initial review to examine the high-level approach here. There's almost certainly room for polishing this and adding some more tests.

### Risk assessment

This change should not directly impact any of the below, but it may indirectly as a result of changes to code paths that are also hit by those.

Workflow types:

Products:

Environments:

### DIFF
--- a/pr-checks/checks/analysis-kinds.yml
+++ b/pr-checks/checks/analysis-kinds.yml
@@ -3,7 +3,7 @@ description: "Tests basic functionality for different `analysis-kinds` inputs."
 versions: ["linked", "nightly-latest"]
 analysisKinds: ["code-scanning", "code-quality", "code-scanning,code-quality", "risk-assessment"]
 env:
-  CODEQL_ACTION_CSRA_ASSESSMENT_ID: 1
+  CODEQL_ACTION_RISK_ASSESSMENT_ID: 1
   CHECK_SCRIPT: |
     const fs = require('fs');
 
--- a/src/analyses.test.ts
+++ b/src/analyses.test.ts
@@ -8,7 +8,7 @@ import {
   AnalysisKind,
   CodeScanning,
   compatibilityMatrix,
-  CSRA,
+  RiskAssessment,
   getAnalysisConfig,
   getAnalysisKinds,
   parseAnalysisKinds,
@@ -122,9 +122,9 @@ test("Code Scanning configuration does not accept other SARIF extensions", (t) =
   }
 });
 
-test("CSRA configuration transforms SARIF upload payload", (t) => {
-  process.env[EnvVar.CSRA_ASSESSMENT_ID] = "1";
-  const payload = CSRA.transformPayload({
+test("Risk Assessment configuration transforms SARIF upload payload", (t) => {
+  process.env[EnvVar.RISK_ASSESSMENT_ID] = "1";
+  const payload = RiskAssessment.transformPayload({
     commit_oid: "abc",
     sarif: "sarif",
     ref: "ref",
@@ -138,11 +138,11 @@ test("CSRA configuration transforms SARIF upload payload", (t) => {
   t.deepEqual(expected, payload);
 });
 
-test("CSRA configuration throws for negative assessment IDs", (t) => {
-  process.env[EnvVar.CSRA_ASSESSMENT_ID] = "-1";
+test("Risk Assessment configuration throws for negative assessment IDs", (t) => {
+  process.env[EnvVar.RISK_ASSESSMENT_ID] = "-1";
   t.throws(
     () =>
-      CSRA.transformPayload({
+      RiskAssessment.transformPayload({
         commit_oid: "abc",
         sarif: "sarif",
         ref: "ref",
@@ -154,16 +154,16 @@ test("CSRA configuration throws for negative assessment IDs", (t) => {
     {
       instanceOf: Error,
       message: (msg) =>
-        msg.startsWith(`${EnvVar.CSRA_ASSESSMENT_ID} must not be negative: `),
+        msg.startsWith(`${EnvVar.RISK_ASSESSMENT_ID} must not be negative: `),
     },
   );
 });
 
-test("CSRA configuration throws for invalid IDs", (t) => {
-  process.env[EnvVar.CSRA_ASSESSMENT_ID] = "foo";
+test("Risk Assessment configuration throws for invalid IDs", (t) => {
+  process.env[EnvVar.RISK_ASSESSMENT_ID] = "foo";
   t.throws(
     () =>
-      CSRA.transformPayload({
+      RiskAssessment.transformPayload({
         commit_oid: "abc",
         sarif: "sarif",
         ref: "ref",
@@ -175,7 +175,7 @@ test("CSRA configuration throws for invalid IDs", (t) => {
     {
       instanceOf: Error,
       message: (msg) =>
-        msg.startsWith(`${EnvVar.CSRA_ASSESSMENT_ID} must not be NaN: `),
+        msg.startsWith(`${EnvVar.RISK_ASSESSMENT_ID} must not be NaN: `),
     },
   );
 });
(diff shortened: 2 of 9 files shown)


pr-checks/checks/analysis-kinds.yml
```diff
@@ -77,3 +72,11 @@ steps:
...
     with:
       script: ${{ env.CHECK_SCRIPT }}
+  - name: Check quality query does not appear in CSRA SARIF
+    if: contains(matrix.analysis-kinds, 'csra')
+    uses: actions/github-script@v8
+    env:
+      SARIF_PATH: "${{ runner.temp }}/results/javascript.csra.sarif"
+      EXPECT_PRESENT: "false"
+    with:
+      script: ${{ env.CHECK_SCRIPT }}
```
henrymercer: Why do we implement this behaviour when CSRA and code quality analysis kinds can't be run together?

src/analyses.ts
```diff
@@ -144,6 +179,30 @@ export const CodeQuality: AnalysisConfig = {
...
   sentinelPrefix: "CODEQL_UPLOAD_QUALITY_SARIF_",
+  transformPayload: (payload) => payload,
+};
+
+/**
+ * Retrieves the CSRA assessment id from an environment variable and adds it to the payload.
+ * @param payload The base payload.
+ */
+function addAssessmentId(payload: UploadPayload): AssessmentPayload {
+  const assessmentId = parseInt(
```
henrymercer: Agree with Copilot that we should validate that this is not NaN

src/analyses.ts
```diff
@@ -144,6 +179,30 @@ export const CodeQuality: AnalysisConfig = {
...
+  return { sarif: payload.sarif, assessment_id: assessmentId };
+}
+
+export const CSRA: AnalysisConfig = {
+  kind: AnalysisKind.CSRA,
+  name: "csra",
+  target: SARIF_UPLOAD_ENDPOINT.CSRA,
+  sarifExtension: ".csra.sarif",
+  sarifPredicate: (name) => name.endsWith(CSRA.sarifExtension),
+  fixCategory: fixCodeQualityCategory,
```
henrymercer: Was this a copy/paste error or are we sharing this fix category logic?  If so, let's rename and update docs.

src/analyses.ts
```diff
@@ -4,13 +4,28 @@ import {
...
+  AssessmentPayload,
+  BasePayload,
+  UploadPayload,
+} from "./upload-lib/types";
+import { ConfigurationError, getRequiredEnvParam } from "./util";
 
 export enum AnalysisKind {
   CodeScanning = "code-scanning",
   CodeQuality = "code-quality",
+  CSRA = "csra",
```
henrymercer: Since it's probably going to be a pain to change this in the future, let's bike shed — could we avoid a FLA here and use something like `risk-assessment`?

src/analyses.ts
mbg: It was intentionally conservative. I have since discussed this with the team and it's not needed for this, so I will change it to the identity function.

src/testing-utils.ts
```diff
@@ -147,11 +147,64 @@ export function setupActionsVars(tempDir: string, toolsDir: string) {
...
+    this.currentGroup = name;
+    this.unfinishedGroups.add(name);
+  }
+
+  endGroup() {
+    if (this.currentGroup !== undefined) {
+      this.unfinishedGroups.delete(this.currentGroup);
+    }
+    this.currentGroup = undefined;
+  }
```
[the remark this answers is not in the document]
mbg: Nested groups aren't really supported in Actions logs and the formatting breaks. Therefore, knowing that there is an unfinished group (whether it is a result of starting another group or not calling `endGroup`) is generally speaking all that matters.

src/upload-lib/types.ts
```diff
@@ -0,0 +1,22 @@
+export interface BasePayload {
+  sarif: string;
```
henrymercer, line 6: Minor: we could add a doc comment that this is gzipped.

src/analyses.test.ts
```diff
@@ -67,3 +75,107 @@ test("getAnalysisKinds - throws if `analysis-kinds` input is invalid", async (t)
...
+    workflow_run_id: 1,
+    checkout_uri: "uri",
+    tool_names: [],
+  }) as AssessmentPayload;
+
+  const expected: AssessmentPayload = { sarif: "sarif", assessment_id: 1 };
+  t.deepEqual(expected, payload);
+});
+
+test("CSRA configuration throws for negative assessment IDs", (t) => {
```
henrymercer: Minor: is 0 OK?

src/analyses.ts
```diff
@@ -3,14 +3,30 @@ import {
...
+  AssessmentPayload,
+  BasePayload,
+  UploadPayload,
+} from "./upload-lib/types";
+import { ConfigurationError, getRequiredEnvParam } from "./util";
 
 export enum AnalysisKind {
   CodeScanning = "code-scanning",
   CodeQuality = "code-quality",
+  CSRA = "risk-assessment",
```
henrymercer: Minor: consider renaming the key to `RiskAssessment`

src/testing-utils.ts
```diff
@@ -147,11 +147,64 @@ export function setupActionsVars(tempDir: string, toolsDir: string) {
...
 
+type LogLevel = "debug" | "info" | "warning" | "error";
+
 export interface LoggedMessage {
-  type: "debug" | "info" | "warning" | "error";
+  type: LogLevel;
   message: string | Error;
 }
 
+export class RecordingLogger implements Logger {
```
henrymercer, line 157: Should this replace the implementation in `getRecordingLogger`?

src/analyses.ts
```diff
@@ -85,12 +101,27 @@ export async function getAnalysisKinds(
...
+  }
+
+  // Check that all enabled analysis kinds are compatible with each other.
+  for (const analysisKind of analysisKinds) {
+    for (const otherAnalysisKind of analysisKinds) {
+      if (analysisKind === otherAnalysisKind) continue;
+
+      if (!compatibilityMatrix[analysisKind].has(otherAnalysisKind)) {
+        throw new ConfigurationError(
+          `${otherAnalysisKind} cannot be enabled at the same time as ${analysisKind}`,
```
henrymercer: Optional, nit: since this is symmetric, we could instead say `${analysisKind} and ${otherAnalysisKind} cannot be enabled at the same time`

src/analyses.ts
```diff
@@ -144,6 +180,38 @@ export const CodeQuality: AnalysisConfig = {
...
+  transformPayload: (payload) => payload,
+};
+
+/**
+ * Retrieves the CSRA assessment id from an environment variable and adds it to the payload.
+ * @param payload The base payload.
+ */
+function addAssessmentId(payload: UploadPayload): AssessmentPayload {
+  const rawAssessmentId = getRequiredEnvParam(EnvVar.CSRA_ASSESSMENT_ID);
+  const assessmentId = parseInt(rawAssessmentId, 10);
```
henrymercer, line 192: Technically this will truncate floats, but I think that's fine in practice.

src/testing-utils.ts
mbg: Eventually perhaps, but adding this was already a tangential change in this PR since I wanted a way to track the log groups. I'll update the tests to use this new implementation in a follow-up PR as a low priority.

src/analyses.test.ts
mbg: I am not sure it matters too much. We don't expect this to ever be used other than with an automatically populated value, and the value is just passed back to the upload endpoint. 

I'd suggest we leave it like this for now and only tighten the constraints later if we have a requirement for it.

src/analyses.ts
mbg: Similar as with https://github.com/github/codeql-action/pull/3474#discussion_r2817409309, I don't think this is a practical concern for now.

PR #3482, opened 2026-02-13
Release notes: Use author if they are GitHub staff

mbg:
Modifies `update-release-branch.py` to use the PR author if they are GitHub staff, or the merger as before otherwise.

### Risk assessment

Environments:

Release process will fail.

### DIFF
--- a/.github/update-release-branch.py
+++ b/.github/update-release-branch.py
@@ -173,11 +173,8 @@ def get_merger_of_pr(repo, pr):
 def get_pr_author_if_staff(pr):
   if pr.user is None:
     return None
-  try:
-    if getattr(pr.user, 'site_admin', False):
-      return pr.user.login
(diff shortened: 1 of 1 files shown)


.github/update-release-branch.py
```diff
@@ -168,6 +169,17 @@ def get_pr_for_commit(commit):
 def get_merger_of_pr(repo, pr):
   return repo.get_commit(pr.merge_commit_sha).author.login
 
+# Get the PR author if they are GitHub staff, otherwise None.
+def get_pr_author_if_staff(pr):
+  if pr.user is None:
+    return None
+  try:
+    if getattr(pr.user, 'site_admin', False):
```
henrymercer: I don't think this would throw if we're providing a default value, so I think we can remove the try/catch.

PR #3484, opened 2026-02-15
Add feature for forcing the `nightly` bundle in `dynamic` workflows

mbg:
Adds a `Feature` which, when enabled, and the workflow was triggered by a `dynamic` event forces `getCodeQLSource` to pick the latest, nightly release. 

Some drive-by improvements and observations:

- I added a couple of comments to better document `getCodeQLSource`.
- While writing those, I noticed that the `!toolsInput.startsWith("http")` check would probably cause problems if a file happened to have a path that starts with `http`. Probably not a very likely problem, but could be tricky to troubleshoot if it happens. We could improve this by performing more explicit checks for `http://` and `https://` or whether the suspected file exists locally or not. Outside of the scope of this PR though.
- I added a unit test for `tools: nightly` which didn't seem to exist.
- While doing that, I noticed that `getNightlyToolsUrl` seems to assume that the nightly tag is `codeql-bundle-` followed by a semver, but this is no longer the case for nightly releases. There's a fallback logic which handles this fine, but we should probably update this to expect the date-based tags.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/.github/workflows/__bundle-from-nightly.yml
+++ b/.github/workflows/__bundle-from-nightly.yml
@@ -55,14 +55,15 @@ jobs:
           version: ${{ matrix.version }}
           use-all-platform-bundle: 'false'
           setup-kotlin: 'true'
-      - id: setup-codeql
-        uses: ./../action/setup-codeql
+      - id: init
+        uses: ./../action/init
         env:
           CODEQL_ACTION_FORCE_NIGHTLY: true
         with:
           tools: ${{ steps.prepare-test.outputs.tools-url }}
+          languages: javascript
       - name: Fail if the CodeQL version is not a nightly
-        if: "!contains(steps.setup-codeql.outputs.codeql-version, '+')"
+        if: "!contains(steps.init.outputs.codeql-version, '+')"
         run: exit 1
     env:
       CODEQL_ACTION_TEST_MODE: true
--- a/pr-checks/checks/bundle-from-nightly.yml
+++ b/pr-checks/checks/bundle-from-nightly.yml
@@ -3,12 +3,13 @@ description: "The nightly CodeQL bundle should be used when forced"
 versions:
   - linked # overruled by the FF set below
 steps:
-  - id: setup-codeql
-    uses: ./../action/setup-codeql
+  - id: init
+    uses: ./../action/init
     env:
       CODEQL_ACTION_FORCE_NIGHTLY: true
     with:
       tools: ${{ steps.prepare-test.outputs.tools-url }}
+      languages: javascript
   - name: Fail if the CodeQL version is not a nightly
-    if: "!contains(steps.setup-codeql.outputs.codeql-version, '+')"
+    if: "!contains(steps.init.outputs.codeql-version, '+')"
     run: exit 1


src/setup-codeql.ts
```diff
@@ -302,13 +321,26 @@ export async function getCodeQLSource(
...
+  const forceNightly =
+    forceNightlyValueFF && (isDynamicWorkflow() || util.isInTestMode());
+
   if (
-    toolsInput !== undefined &&
-    CODEQL_NIGHTLY_TOOLS_INPUTS.includes(toolsInput)
+    forceNightly ||
+    (toolsInput !== undefined &&
+      CODEQL_NIGHTLY_TOOLS_INPUTS.includes(toolsInput))
   ) {
```
esbena: I agree with that the comment and implementation agrees, but I do have a hard time following all of the dis- and conjuncts here. 

Perhaps a structure like:
```
const canBeControlledByFF = isDynamicWorkflow() || util.isInTestMode();
const canBeControlledByToolsInput = toolsInput !== undefined;
...
```

src/setup-codeql.ts
```diff
@@ -302,13 +321,26 @@ export async function getCodeQLSource(
...
-    logger.info(
-      `Using the latest CodeQL CLI nightly, as requested by 'tools: ${toolsInput}'.`,
-    );
+    if (forceNightly) {
+      logger.info(
+        `Using the latest CodeQL CLI nightly, as forced by the ${Feature.ForceNightly} feature flag.`,
```
esbena, line 340: I'm tempted to put this in the job summary outputs, or the annotations, to make it absolutely clear that overrides are happening here.

src/setup-codeql.ts
```diff
@@ -302,13 +321,26 @@ export async function getCodeQLSource(
    */
   let url: string | undefined;
 
+  // We only allow forcing the nightly CLI via the FF for `dynamic` events (or in test mode).
+  // For advanced workflows, a value from `CODEQL_NIGHTLY_TOOLS_INPUTS` can be specified.
+  const forceNightlyValueFF = await features.getValue(Feature.ForceNightly);
+  const forceNightly =
+    forceNightlyValueFF && (isDynamicWorkflow() || util.isInTestMode());
```
esbena: For telemetry, I suppose we can just check these values to reconstruct the runtime choices without needing to enhance the status reports?

esbena (approved):
LGTM with some superficial logging observability/logging concerns.

src/setup-codeql.ts
mbg: Not quite until we ship something like https://github.com/github/codeql-action/pull/3437

src/setup-codeql.ts
esbena: (discussed on slack)

mbg: I have added a couple of commits which should enable this.

esbena (dismissed):
The recent changes LGTM.

henrymercer (dismissed):
I had a quick look over the last two commits.

- I like the solution with storing unwritten nolang diagnostics.
- I would suggest changing the severity of the diagnostic to info, since the UI will complain relatively loudly about warnings.
- The changes look correct but it is hard to say for sure that they will work without a PR check or example run — consider running this on a test repo before merging.

mbg:
@henrymercer I have made those changes and ran a test in a test repository. See the references in the PR timeline.

esbena (approved):
Latest (trivial) changes LGTM, and the sample run looks as advertised.

PR #3486, opened 2026-02-17
Log information about the runner which may affect the private registry proxy

mbg:
This PR modifies the `start-proxy` action to perform a best-effort inspection of some environment variables and language-specific settings which may affect the operation of the authentication proxy. Relevant values are logged so that they are visible in the workflow log.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/testing-utils.ts
+++ b/src/testing-utils.ts
@@ -271,6 +271,8 @@ export function checkExpectedLogMessages(
     t.fail(
       `Expected\n\n${listify(missingMessages)}\n\nin the logger output, but didn't find it in:\n\n${messages.map((m) => ` - '${m.message}'`).join("\n")}`,
     );
+  } else {
+    t.pass();
   }
 }
 


src/start-proxy/environment.ts
```diff
@@ -0,0 +1,157 @@
...
+  HTTP_PROXY = "HTTP_PROXY",
+  HTTPS_PROXY = "HTTPS_PROXY",
+  ALL_PROXY = "ALL_PROXY",
+}
+
+/**
+ * Checks whether any proxy-related environment variables are set and logs their values if so.
+ */
+export function checkProxyEnvVars(logger: Logger) {
+  // Both upper-case and lower-case variants of these environment variables are used.
```
esbena, line 175: (Technically, the "upper-case" could be mixed case.)

src/start-proxy-action.ts
```diff
@@ -76,6 +77,17 @@ async function run(startedAt: Date) {
         .join("\n")}`,
     );
 
+    // Check the environment for any configurations which may affect the proxy.
+    // This is a best effort process to give us insights into potential factors
+    // which may affect the operation of our proxy.
+    try {
+      checkProxyEnvironment(logger, language);
```
esbena: This logging is on by default? Do we want a FF for this, or run when the actions debug variable is set?

src/start-proxy/environment.ts
mbg: I thought it was only the fully-upper case or fully-lower case variants that were generally understood. Do you think it's worth making these checks generally case-insensitive?

src/start-proxy-action.ts
mbg: Gating it by debug mode makes sense to me, which I have done (+ for test mode) in https://github.com/github/codeql-action/pull/3486/commits/11c6c18818b35b709193c85f83d1183538f7414c

src/start-proxy/environment.ts
```diff
@@ -0,0 +1,157 @@
...
+ * Checks whether an environment variable named `name` is set and logs its value if set.
+ *
+ * @param logger The logger to use.
+ * @param name The name of the environment variable.
+ * @returns True if set or false otherwise.
+ */
+function checkEnvVar(logger: Logger, name: string): boolean {
+  const value = process.env[name];
+  if (isDefined(value)) {
+    logger.info(`Environment variable '${name}' is set to '${value}'.`);
```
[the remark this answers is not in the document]
mbg: I have added some protection against this in https://github.com/github/codeql-action/pull/3486/commits/99fcc7b2a12a3895509ff680d8b72d80b2e210ab by trying to parse the value as a URL in `checkEnvVar` and, if successful, clearing the `username` and `password`. Broadly, this should only be a concern for the members of `ProxyEnvVars` since, for the Java properties, there are separate `http(s).proxyUser` and `http(s).proxyPassword` which we don't try to find and therefore don't log.

src/start-proxy/environment.ts
esbena: no, I think the implementation is fine. We can't cater to every non-standard use case here. It is just the comment that is subtly wrong.

src/start-proxy/environment.ts
```diff
@@ -128,6 +131,36 @@ export function checkJdkSettings(logger: Logger, jdkHome: string) {
   }
 }
 
+/** Invokes `java` to get it to show us the active configuration. */
+async function showJavaSettings(logger: Logger): Promise<void> {
+  try {
+    const java = await io.which("java", true);
```
esbena, line 137: I'm probably grasping at straws here. But what if it is not our Java version that is problematic? But some other Java version chosen at a different point in time. I suppose this is somewhat addressed by the fact that we log File contents explicitly in addition to this effective settings logging

mbg: I think this is a valid question and one I have thought about. Other than what you have already said about us inspecting some of the configuration files, my thinking is that there are essentially two scenarios since this is default setup:

1. The runner is a GH-hosted runner, and the configuration of all available JDKs is the standard one we can observe ourselves.
2. The runner is self-hosted and the active Java version at the time the action runs is probably the one a customer would intend to run their build with.

Based on that, I don't think this is unreasonable.

PR #3487, opened 2026-02-17
Cache first failure building an overlay base DB to avoid repeated failures

henrymercer:
When overlay analysis (improved incremental analysis) fails on a runner — typically due to insufficient disk space — this PR records that failure in the Actions cache so that subsequent runs will skip overlay analysis automatically until something changes (e.g. a larger runner is provisioned or a new CodeQL version is released).

See the backlinked internal issue for more information.

I recommend reviewing the first commit separately from the rest as this moves the overlay utilities into their own directory.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/init-action-post-helper.ts
+++ b/src/init-action-post-helper.ts
@@ -295,23 +295,20 @@ async function recordOverlayStatus(
     "This job attempted to run with improved incremental analysis but it did not complete successfully. " +
     "This may have been due to disk space constraints: using improved incremental analysis can " +
     "require a significant amount of disk space for some repositories.";
-  const outro =
-    "If you want to enable improved incremental analysis, increase the disk space available to the runner. " +
-    "If that doesn't help, contact GitHub Support for further assistance.";
 
   if (saved) {
     logger.error(
       `${blurb} ` +
         "This failure has been recorded in the Actions cache, so the next CodeQL analysis will run " +
-        "without improved incremental analysis. " +
-        `${outro}`,
+        "without improved incremental analysis. If you want to enable improved incremental analysis, " +
+        "increase the disk space available to the runner. " +
+        "If that doesn't help, contact GitHub Support for further assistance.",
     );
   } else {
     logger.error(
       `${blurb} ` +
         "The attempt to save this failure status to the Actions cache failed. The Action will attempt to " +
-        "save this failure status again on the next run, so future runs will skip improved incremental analysis. " +
-        `${outro}`,
+        "run with improved incremental analysis again.",
     );
   }
 }


src/overlay/index.ts
mbg, line 1: I like this reorganisation, especially using `index.ts` to allow `import ... from "./overlay"`

src/overlay/status.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * We perform enablement checks for overlay analysis to avoid using it on runners that are too small
+ * to support it. However these checks cannot avoid every potential issue without being overly
+ * conservative. Therefore, if our enablement checks enable overlay analysis for a runner that is
+ * too small, we want to remember that, so that we will not try to use overlay analysis until
+ * something changes (e.g. a larger runner is provisioned, or a new CodeQL version is released).
+ *
+ * We use the Actions cache as a lightweight way of providing this functionality.
+ */
```
mbg, line 9: This sort of top-of-file comment capturing the motivation / design is great. We should do more of this.

src/overlay/status.ts
```diff
@@ -0,0 +1,31 @@
...
+  diskUsage: DiskUsage,
+): Promise<string> {
+  // Total disk space, rounded to the nearest 10 GB. This is included in the cache key so that if a
+  // customer upgrades their runner, we will try again to use overlay analysis, even if the CodeQL
+  // version has not changed. We round to the nearest 10 GB to work around small differences in disk
+  // space.
+  //
+  // Limitation: this can still flip from "too small" to "large enough" and back again if the disk
+  // space fluctuates above and below a multiple of 10 GB.
+  const diskSpaceToNearest10Gb = `${10 * Math.floor(diskUsage.numTotalBytes / (10 * 1024 * 1024 * 1024))}GB`;
```
mbg, line 185: Design question: this fundamentally assumes that the CodeQL analysis typically runs on comparable runners. I.e. the assumption is that unless the amount of total disk space is increased deliberately, the runner specs are the same. Practically speaking, I'd expect that to be the case as well. However, I am not sure whether it is necessarily the case or this is an assumption we have made previously.

My concern is that, if a customer has a runner group for CodeQL containing runners with different specs, we might flip-flop on this -- I think you express that in the "Limitation" part of the comment. That wouldn't be a great experience for a customer if it happened. Is there a way we can mitigate this?

src/overlay/status.test.ts
```diff
@@ -0,0 +1,51 @@
...
+import { mockCodeQLVersion, setupTests } from "../testing-utils";
+import { DiskUsage } from "../util";
+
+import { getCacheKey } from "./status";
+
+setupTests(test);
+
+function makeDiskUsage(totalGiB: number): DiskUsage {
+  return {
+    numTotalBytes: totalGiB * 1024 * 1024 * 1024,
```
mbg, line 22: Minor: could we define `1024 * 1024 * 1024` as a constant somewhere since it's used in this test and also in the code being tested, or take advantage of some more strongly typed library for units?

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+    "overlay-status",
+    language,
+    STATUS_FILE_NAME,
+  );
+  await fs.promises.mkdir(path.dirname(statusFile), { recursive: true });
+
+  try {
+    const foundKey = await waitForResultWithTimeLimit(
+      MAX_CACHE_OPERATION_MS,
+      actionsCache.restoreCache([statusFile], cacheKey),
```
mbg, line 96: The paths to restore are an implicit component of the cache key. In this case, if `statusFile` is different between store and restore, then the cache won't get restored here. Since the path depends on `getTemporaryDirectory()`, we are dependent on that returning the same path every time.

I see that our dependency caching implementation also relies on `getTemporaryDirectory()` returning the same path for Java/C#. We should probably check that this doesn't cause any issues and ideally move to something that's more reliably stable.

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+      () => {
+        logger.info("Timed out restoring overlay status from cache");
```
mbg: Minor: Should this be a warning?

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+
+    if (!fs.existsSync(statusFile)) {
+      logger.debug(
+        "Overlay status cache entry found but status file is missing",
+      );
+      return undefined;
+    }
+
+    const contents = await fs.promises.readFile(statusFile, "utf-8");
+    return JSON.parse(contents) as OverlayStatus;
```
mbg: The `JSON.parse(contents) as T` pattern is something we are trying to get away from. Elsewhere, we have been starting to make an effort to validate that the result of `JSON.parse` matches what we expect (e.g. in `start-proxy`). Without any validation, we could have valid JSON, but something that looks nothing like an `OverlayStatus` object. I'd recommend checking that it is an object and then either checking that `builtOverlayBaseDatabase` is present, or casting it to `Partial<OverlayStatus>`

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+  status: OverlayStatus,
+  logger: Logger,
+): Promise<boolean> {
+  const cacheKey = await getCacheKey(codeql, language, diskUsage);
+  const statusFile = path.join(
+    getTemporaryDirectory(),
+    "overlay-status",
+    language,
+    STATUS_FILE_NAME,
+  );
```
mbg: Minor: this is duplicated between `getOverlayStatus` and `saveOverlayStatus`. It would be good to de-duplicate this.

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+  await fs.promises.mkdir(path.dirname(statusFile), { recursive: true });
+  await fs.promises.writeFile(statusFile, JSON.stringify(status));
```
mbg: Should these be inside the `try`/`catch`?

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+  logger: Logger,
+): Promise<OverlayStatus | undefined> {
+  const cacheKey = await getCacheKey(codeql, language, diskUsage);
+  const statusFile = path.join(
+    getTemporaryDirectory(),
+    "overlay-status",
+    language,
+    STATUS_FILE_NAME,
+  );
+  await fs.promises.mkdir(path.dirname(statusFile), { recursive: true });
```
mbg: Should this be inside the `try`/`catch`?

src/overlay/status.ts
```diff
@@ -8,8 +8,124 @@
...
+  await fs.promises.writeFile(statusFile, JSON.stringify(status));
+
+  try {
+    const cacheId = await waitForResultWithTimeLimit(
+      MAX_CACHE_OPERATION_MS,
+      actionsCache.saveCache([statusFile], cacheKey),
+      () => {},
```
mbg: Minor: Log this? I see that we check `cacheId === undefined` below, but it's a bit inconsistent between here and the restore function.

src/overlay/status.ts
```diff
@@ -28,7 +28,7 @@ const MAX_CACHE_OPERATION_MS = 30_000;
 /** File name for the serialized overlay status. */
 const STATUS_FILE_NAME = "overlay-status.json";
 
-/** Status of an overlay analysis for a particular language. */
+/** Status of an overlay analysis for a set of languages. */
```
mbg: Minor: `array` instead of `set`, since the latter suggests a particular data structure / invariant that is not obviously enforced? Alternatively: use `Set<string>`?

src/overlay/status.ts
```diff
@@ -34,6 +34,32 @@ export interface OverlayStatus {
...
+ */
+export async function shouldSkipOverlayAnalysis(
+  codeql: CodeQL,
+  languages: string[],
+  diskUsage: DiskUsage,
+  logger: Logger,
+): Promise<boolean> {
+  const status = await getOverlayStatus(codeql, languages, diskUsage, logger);
+  if (status === undefined) {
+    logger.debug("No cached overlay status found.");
```
mbg: Minor: all cases that return `undefined` in `getOverlayStatus` already log the cause, so this just logs something less specific afterwards.

src/overlay/status.ts
```diff
@@ -34,6 +34,32 @@ export interface OverlayStatus {
...
+    return false;
+  }
+  if (!status.builtOverlayBaseDatabase) {
```
mbg: This would throw if `status` is not an object (see my comment about missing validation after `JSON.parse`). This would also evaluate to `true` if `builtOverlayBaseDatabase` is anything false-y, so even if it's not explicitly `false` we'd skip overlay analysis. That's probably not what we want.

src/config-utils.ts
```diff
@@ -762,13 +764,25 @@ export async function getOverlayDatabaseMode(
       codeScanningConfig,
     )
   ) {
+    const diskUsage = await checkDiskUsage(logger);
```
mbg: Observation: This moves the disk usage calculation out of the part of the code that can be gated by `Feature.OverlayAnalysisSkipResourceChecks`.

src/config-utils.ts
```diff
@@ -762,13 +764,25 @@ export async function getOverlayDatabaseMode(
...
     const performResourceChecks = !(await features.getValue(
       Feature.OverlayAnalysisSkipResourceChecks,
       codeql,
     ));
     if (
+      diskUsage &&
+      (await shouldSkipOverlayAnalysis(codeql, languages, diskUsage, logger))
```
mbg, line 817: I understand that the disk capacity is included in the cache key and therefore needed to try and restore it, but it's a bit odd that a side effect of that is that we don't check `shouldSkipOverlayAnalysis` here. It might be worth adding a comment to explain that relationship, and that it is a deliberate choice to continue with overlay analysis if we can't obtain `diskUsage` information here (at least until we get to `runnerSupportsOverlayAnalysis` which will disable overlay analysis if `diskUsage` is `undefined`).

src/config-utils.ts
```diff
@@ -762,13 +764,25 @@ export async function getOverlayDatabaseMode(
...
+    ) {
+      logger.info(
+        `Setting overlay database mode to ${OverlayDatabaseMode.None} ` +
+          "because overlay analysis previously failed with this combination of languages, " +
+          "disk space, and CodeQL version. " +
+          "Consider running CodeQL analysis on a larger runner.",
+      );
```
mbg, line 823: Minor: Should this be a diagnostic? Probably a more general question for several of the "overlay analysis is disabled because of X" messages.

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,46 @@ export async function run(
   return uploadFailedSarifResult;
 }
 
+async function recordOverlayStatus(
```
mbg, line 259: Minor: JSDoc?

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,46 @@ export async function run(
...
+  config: Config,
+  logger: Logger,
+) {
+  if (
+    config.overlayDatabaseMode === OverlayDatabaseMode.OverlayBase &&
+    process.env[EnvVar.ANALYZE_DID_COMPLETE_SUCCESSFULLY] !== "true"
+  ) {
+    const overlayStatus = {
+      builtOverlayBaseDatabase: false,
+    } satisfies OverlayStatus;
```
mbg: Why `satisfies` and not just a type annotation on `overlayStatus`?

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,46 @@ export async function run(
...
+      codeql,
+      config.languages,
+      diskUsage,
+      overlayStatus,
+      logger,
+    );
+    if (saved) {
+      logger.debug(
+        `Saved overlay status to the Actions cache: ${JSON.stringify(overlayStatus)}`,
+      );
```
mbg: Minor: The "message" part here duplicates what `saveOverlayStatus` logs at info level. Consider changing it to just the `JSON.stringify`-ed result and logging it regardless of whether the cache was saved or not.

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,46 @@ export async function run(
...
+    } else {
+      logger.warning(
+        `Failed to save overlay status to the Actions cache. Status was: ${JSON.stringify(overlayStatus)}`,
+      );
```
mbg: Aside from logging the `JSON.stringify`-ed status, this just duplicates a log message from `saveOverlayStatus`

src/overlay/status.ts
```diff
@@ -0,0 +1,178 @@
...
+  // version has not changed. We round to the nearest 10 GB to work around small differences in disk
+  // space.
+  //
+  // Limitation: this can still flip from "too small" to "large enough" and back again if the disk
+  // space fluctuates above and below a multiple of 10 GB.
+  const diskSpaceToNearest10Gb = `${10 * Math.floor(diskUsage.numTotalBytes / (10 * 1024 * 1024 * 1024))}GB`;
+
+  // Include the CodeQL version in the cache key so we will try again to use overlay analysis when
+  // new queries and libraries that may be more efficient are released.
+  return `codeql-overlay-status-${[...languages].sort().join("+")}-${(await codeql.getVersion()).version}-runner-${diskSpaceToNearest10Gb}`;
```
mbg, line 189: I suspect that this cache key is probably too simplistic, but want to check:

- Do we expect this to run for Advanced Setup workflows? If so, then we should probably include at least the analysis key since there could be different analyses with different `runs-on` settings.
    - Right now, I don't see any part of the overlay implementation that's gated on `isDefaultSetup` 
    - If we are considering Advanced Setup, we'd also potentially have to consider a much space of relevant settings. We could hash them together with `createCacheKeyHash`.
- The [runner context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context) includes the name of the runner. We could potentially include that here as well so we don't just identify the runner based on the available disk space.
- It might be useful to include some kind of version for this implementation in the cache key, to avoid running into issues down the line if we need to make changes to any of this.

Minor: Should this use the same character (_) for joining the languages as the main overlay db caches?

src/overlay/status.ts
```diff
@@ -0,0 +1,178 @@
...
+  const cacheKey = await getCacheKey(codeql, languages, diskUsage);
+  const statusFile = getStatusFilePath(languages);
+  await fs.promises.mkdir(path.dirname(statusFile), { recursive: true });
+
+  try {
+    const foundKey = await waitForResultWithTimeLimit(
+      MAX_CACHE_OPERATION_MS,
+      actionsCache.restoreCache([statusFile], cacheKey),
+      () => {
+        logger.info("Timed out restoring overlay status from cache.");
```
mbg: Minor: Should this be a warning?

src/overlay/status.ts
```diff
@@ -0,0 +1,178 @@
...
+): Promise<boolean> {
+  const status = await getOverlayStatus(codeql, languages, diskUsage, logger);
+  if (status === undefined) {
+    logger.debug("No cached overlay status found.");
+    return false;
+  }
+  if (
+    status.attemptedToBuildOverlayBaseDatabase &&
+    !status.builtOverlayBaseDatabase
+  ) {
```
mbg, line 65: This would throw if `status` is not an object (see my comment about missing validation after `JSON.parse`). `!status.builtOverlayBaseDatabase` would also evaluate to `true` if `builtOverlayBaseDatabase` is anything false-y, so even if it's not explicitly `false` we'd skip overlay analysis. Conversely with `status.attemptedToBuildOverlayBaseDatabase`. That's probably not what we want.

src/overlay/status.ts
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * @returns `undefined` if no status was found in the cache (e.g. first run with
+ * this cache key) or if the cache operation fails.
+ */
+export async function getOverlayStatus(
+  codeql: CodeQL,
+  languages: string[],
+  diskUsage: DiskUsage,
+  logger: Logger,
+): Promise<OverlayStatus | undefined> {
```
mbg, line 88: Rather than returning `OverlayStatus | undefined`, how about always returning an `OverlayStatus` object here? Based on the commits, I see you have gone back-and-forth between always storing the status vs. only storing it on failure. 

You could construct an `OverlayStatus` value here in the branches that currently return `undefined`. This would be easier if one of the fields was negated, i.e. if `OverlayStatus` contained `overlayBaseBuildFailed` rather than `builtOverlayBaseDatabase` and `attemptedToBuildOverlayBaseDatabase` was optional.

I think an approach like that might simplify the logic and make it clearer what's going on.

src/config-utils.ts
```diff
@@ -1001,25 +1019,56 @@ export async function initConfig(
...
+          attributes: {
+            languages: config.languages,
+          },
+          markdownMessage:
+            `Improved incremental analysis was skipped because it failed previously on this runner. ` +
+            "Improved incremental analysis may require a significant amount of disk space on some repositories. " +
+            "If you want to enable improved incremental analysis, increase the disk space available " +
+            "to the runner. If that doesn't help, contact GitHub Support for further assistance.\n\n" +
+            "Improved incremental analysis will be automatically retried when the next version of CodeQL is released. " +
+            `You can also manually trigger a retry by [removing](${DocUrl.DELETE_ACTIONS_CACHE_ENTRIES}) \`codeql-overlay-status-*\` entries from the Actions cache.`,
```
mbg: Few thoughts here:

- Consider including the CodeQL CLI version somewhere?
- "this runner" might be misleading someone into thinking that we have defined the exact runner (based on name, IP, ...).

```suggestion
            `Improved incremental analysis was skipped because it previously failed for this repository with CodeQL version A.B.C on a runner with the same hardware resources. ` +
            "Improved incremental analysis may require a significant amount of disk space for some repositories. " +
            "If you want to enable improved incremental analysis, increase the disk space available " +
            "to the runner. If that doesn't help, contact GitHub Support for further assistance.\n\n" +
            "Improved incremental analysis will be automatically retried when the next version of CodeQL is released. " +
            `You can also manually trigger a retry by [removing](${DocUrl.DELETE_ACTIONS_CACHE_ENTRIES}) \`codeql-overlay-status-*\` entries from the Actions cache.`,
```

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,60 @@ export async function run(
...
+  features: FeatureEnablement,
+  logger: Logger,
+) {
+  // Currently it is only important to store overlay status if the analysis attempted but failed
+  // to build an overlay base database.
+  if (
+    config.overlayDatabaseMode === OverlayDatabaseMode.OverlayBase &&
+    process.env[EnvVar.ANALYZE_DID_COMPLETE_SUCCESSFULLY] !== "true" &&
+    (await features.getValue(Feature.OverlayAnalysisStatusSave))
+  ) {
```
mbg, line 269: Minor: The rest of this function only does something if these conditions are true. Consider changing this so that the function returns early if the conditions are false, so that the entire rest of the function isn't in the `if` body.

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,60 @@ export async function run(
...
+    const overlayStatus = {
+      attemptedToBuildOverlayBaseDatabase: true,
+      builtOverlayBaseDatabase: false,
+    } satisfies OverlayStatus;
+
+    const diskUsage = await checkDiskUsage(logger);
+    if (diskUsage === undefined) {
+      logger.warning(
+        "Failed to determine disk usage, so unable to save overlay status to the Actions cache.",
```
mbg: Minor: avoid using "so" as a connecting word instead of an inverted "because": 

```suggestion
        "Unable to save overlay status to the Actions cache, because the available disk space could not be determined.",
```

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,60 @@ export async function run(
...
+      logger,
+    );
+    if (saved) {
+      logger.debug(
+        `Saved overlay status to the Actions cache: ${JSON.stringify(overlayStatus)}`,
+      );
+      logger.error(
+        "This job attempted to run with improved incremental analysis but it did not complete successfully. " +
+          "This may have been due to disk space constraints: using improved incremental analysis can " +
+          "require a significant amount of disk space for some repositories. " +
```
mbg: This part would probably be useful to log whether or not the state is successfully uploaded.

src/init-action-post-helper.ts
```diff
@@ -246,6 +251,60 @@ export async function run(
...
+          "This failure has been recorded in the Actions cache, so " +
+          "rerunning this job will run a new CodeQL analysis without improved incremental analysis. " +
```
mbg: Not an option for default setup.

mbg (commented):
Thanks for putting this together! There's a lot going on here, including some things I like a lot. I have left a bunch of detailed comments, and there are also some high-level points:

- From previous experience, we know that we need to be careful about the construction of cache keys. We should consider more thoroughly what we need to include in the keys here to not shoot ourselves in the foot. In particular, I'd like us to better identify the runner, the analysis (thinking about Advanced Setup), and guarding against changes we need to make to the implementation.
- Caches can interact poorly with feature flags, if the feature flags affect what might be in a cache. We currently have a number of FFs which affect the overlay analysis behaviour and analysis in general. We might want to include these in the cache key so that e.g. we don't roll out a feature flag that breaks all overlay base database builds, then roll back the feature flag, but are stuck with caches that indicate failure for two weeks.
- The decision whether a base database build failed is currently local to a single workflow run. Consider a scenario where we successfully built an overlay base database with CodeQL version X and uploaded it. Now we are running again for a new commit, but building the overlay base database fails with the same CodeQL version -- perhaps due to an intermittent failure. We upload the status file and block all future base overlay db builds for this CodeQL version and all PR runs from even trying to download the existing base db, which is still in the cache. Perhaps it would be worth checking whether there is an existing base db for the same CodeQL version in the cache?

src/overlay/status.ts
henrymercer: Good point, this might be something that's worth checking in telemetry before rolling this out.

src/overlay/status.ts
henrymercer: Do you have any ideas of how to work around this?  It seems desirable to store temporary files under `RUNNER_TEMP` where possible.

src/config-utils.ts
henrymercer: Addressed

src/config-utils.ts
henrymercer: Agreed, updated such that when the FF is enabled we will only continue with overlay if we can check disk space.

src/config-utils.ts
henrymercer: This is added as a diagnostic later, to avoid refactoring `getOverlayDatabaseMode` to depend on `config` (which it would then need to mutate)

src/init-action-post-helper.ts
henrymercer: `satisfies` preserves more type information, but we're not using that here.  I've changed.

src/init-action-post-helper.ts
henrymercer: Removed this log

src/overlay/status.ts
henrymercer: We discussed this offline, but in summary:

- We will run on advanced setup, but the cache key here matches the design for overlay that only depends on languages.
- The runner label doesn't seem to be available.  The runner name will vary for runner labels that are associated with multiple runners.
- We can add a version number when we change the implementation of the cache key.  There's no need to state v1 explicitly.
- Having the languages joined with `+` makes it easier to read the different parts of the cache key, particularly when some but not all groups analyze multiple languages.  But I don't feel strongly about this so happy to change.

src/overlay/status.ts
henrymercer: I'm going to keep the `OverlayStatus | undefined` — it feels more extensible to force callers to deal with undefined, rather than needing to invent a zero value.

src/init-action-post-helper.ts
henrymercer: I've changed to just mention the "next run"

henrymercer:
@mbg Thanks for the review, I believe I've addressed all your comments, and we discussed the ones in your overall comment offline.

src/init-action-post-helper.ts
```diff
@@ -251,57 +251,68 @@ export async function run(
...
+      `${blurb} ` +
+        "This failure has been recorded in the Actions cache, so the next CodeQL analysis will run " +
+        "without improved incremental analysis. " +
+        `${outro}`,
+    );
+  } else {
+    logger.error(
+      `${blurb} ` +
+        "The attempt to save this failure status to the Actions cache failed. The Action will attempt to " +
+        "save this failure status again on the next run, so future runs will skip improved incremental analysis. " +
```
mbg: Minor: Simplify the second sentence to "The Action will attempt to run with improved incremental analysis again." since it may not _necessarily_ fail again.

mbg (dismissed):
Happy to approve now. The changes you've made in response to my previous feedback are good and we have discussed the other points in depth elsewhere.

Just one minor comment, don't feel obliged to address it as part of this PR.

PR #3497, opened 2026-02-20
Update eslint to v9

henrymercer:
This PR updates eslint to v9.  It is too early to update to v10 since some of our linter plugins don't yet have support.

The eslint config file has changed substantially due to converting to flat config and changes to upstream plugins.  I have attempted to ensure that the rules we are running are the same, and note that there is only one change to disable directives, which is to remove a `while (true)` statement that is no longer flagged by the no constant condition linting rule.  However this PR would benefit from a careful look at this config file.

I checked `import/no-cycle` manually, since this has caught user-visible issues for us in the past.  Strangely I could not get `import/no-cycle` to work with eslint v9.  It ran, but did not detect an example import cycle.  I have replaced it by `import-x/no-cycle`, but it's still odd to me that this didn't work initially.

### Risk assessment

Environments:

Going through linter plugins to check rules, confirming other disable directives are still required.

n/a

### DIFF
--- a/.github/workflows/label-pr-size.yml
+++ b/.github/workflows/label-pr-size.yml
@@ -17,6 +17,7 @@ jobs:
   sizeup:
     name: Label PR with size
     runs-on: ubuntu-slim
+    if: github.event.pull_request.merged != true
 
     steps:
       - name: Run sizeup
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 No user facing changes.
 
+## 4.32.4 - 20 Feb 2026
+
+- Update default CodeQL bundle version to [2.24.2](https://github.com/github/codeql-action/releases/tag/codeql-bundle-v2.24.2). [#3493](https://github.com/github/codeql-action/pull/3493)
+- Added an experimental change which improves how certificates are generated for the authentication proxy that is used by the CodeQL Action in Default Setup when [private package registries are configured](https://docs.github.com/en/code-security/how-tos/secure-at-scale/configure-organization-security/manage-usage-and-access/giving-org-access-private-registries). This is expected to generate more widely compatible certificates and should have no impact on analyses which are working correctly already. We expect to roll this change out to everyone in February. [#3473](https://github.com/github/codeql-action/pull/3473)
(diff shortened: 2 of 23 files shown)


mbg:
@henrymercer High-level question here: would we be better off upgrading to an earlier/older minor version of `v9` first? Or is the upgrade path from `v8`->`v9` always going to be this flaky?

sam-robson:
Dug a little and I think this is missing a few rules. `compat.extends("eslint:recommended")` pulled in all of the `eslint:recommended` ruleset, whereas the new `githubFlatConfigs.recommended` grabs from a [static list ](https://github.com/github/eslint-plugin-github/blob/main/lib/configs/flat/recommended.js#L26-L140) which looks to be based on an older snapshot of `eslint:recommended`.

These are the missing rules
```
for-direction
getter-return
no-async-promise-executor
no-constant-binary-expression
no-dupe-else-if
no-empty-static-block
no-import-assign
no-loss-of-precision
no-misleading-character-class
no-new-native-nonconstructor
no-nonoctal-decimal-escape
no-prototype-builtins
no-shadow-restricted-names
no-unsafe-optional-chaining
no-unused-private-class-members
no-useless-backreference
no-useless-catch
```

I've got a [draft PR ready upstream](https://github.com/github/eslint-plugin-github/pull/683), if you could verify that I'm onto something here!

henrymercer:
> @henrymercer High-level question here: would we be better off upgrading to an earlier/older minor version of `v9` first? Or is the upgrade path from `v8`->`v9` always going to be this flaky?

The most confusing aspect of this for me has been the `import/no-cycle` regression.  I still don't know the root cause of that.  We could bisect to see if an earlier version solves it, but I'm not sure it's a good use of time now that we have `import-x/no-cycle` which works.

> Dug a little and I think this is missing a few rules. `compat.extends("eslint:recommended")` pulled in all of the `eslint:recommended` ruleset, whereas the new `githubFlatConfigs.recommended` grabs from a [static list ](https://github.com/github/eslint-plugin-github/blob/main/lib/configs/flat/recommended.js#L26-L140) which looks to be based on an older snapshot of `eslint:recommended`.

Thanks @sam-robson, this is a great point.  I thought that the GitHub ruleset extends eslint/recommended but that's not quite true.

typescript-eslint disables some of the standard eslint rules due to a couple of reasons, e.g. the rule is already checked by the TypeScript compiler, or the rule is replaced by a more precise TypeScript-specific one.  So I don't think we necessarily want to have all the eslint recommended rules enabled.  But we can enable both `js.configs.recommended` and `tseslint.configs.recommendedTypeChecked` to make sure we're getting all the eslint recommended rules that aren't explicitly disabled by typescript-eslint.

I'm not sure whether the upstream `githubFlatConfigs.recommended` plugin has gone through a process of disabling certain rules for similar reasons.  However it seems valid that there's a risk that new rules slip through the cracks if the pattern is to enable each rule explicitly.  So some rules we would want may indeed be missing.

PR #3498, opened 2026-02-20
Add feature flag for more lenient overlay resource checks

henrymercer:
We would like to start rolling out overlay to standard Actions runners.  This PR adds a feature flag to lower the disk space requirement to 14 GB, which is small enough to include the set of standard Actions runners.

Based on https://github.com/github/codeql-action/pull/3487 to avoid conflicts.  We don't plan to roll this out before https://github.com/github/codeql-action/pull/3487 is merged.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/config-utils.ts
+++ b/src/config-utils.ts
@@ -80,6 +80,15 @@ const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_BYTES =
   OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_MB * 1_000_000;
 
 /**
+ * The v2 minimum available disk space (in MB) required to perform overlay
+ * analysis. This is a lower threshold than the v1 limit, allowing overlay
+ * analysis to run on runners with less available disk space.
+ */
+const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_MB = 14000;
+const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_BYTES =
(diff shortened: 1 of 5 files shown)


src/config-utils.ts
```diff
@@ -79,6 +79,15 @@ const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_MB = 20000;
...
 
+/**
+ * The v2 minimum available disk space (in MB) required to perform overlay
+ * analysis. This is a lower threshold than the v1 limit, allowing overlay
+ * analysis to run on runners with less available disk space.
+ */
+const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_MB = 14000;
+const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_BYTES =
+  OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_MB * 1_000_000;
+
```
esbena, line 90: > This is a lower threshold than the v1 limit, allowing overlay
> * analysis to run on runners with less available disk space

 I'm wondering if the naming should reflect that instead. Or do we envision this to be the new default in a moment?

esbena: Also, I do dislike the fact that we are adding significant jsdoc to a variable that is used on the line below, which in turn then does not have any jsdoc. But I do see the precedent just above.

esbena (approved):

src/config-utils.ts
henrymercer: I thought I'd go with v2 so if we need to change the disk space again, we don't need to name something "lowest" or "medium" etc :)

PR #3502, opened 2026-02-24
Remove all CCR-specific code and tests

mbg:
This is no longer needed.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -23,13 +23,13 @@ For internal use only. Please select the risk level of this change:
 Workflow types:
 
 - **Advanced setup** - Impacts users who have custom CodeQL workflows.
-- **Managed** - Impacts users with `dynamic` workflows (Default Setup, CCR, ...).
+- **Managed** - Impacts users with `dynamic` workflows (Default Setup, Code Quality, ...).
 
 Products:
 
 - **Code Scanning** - The changes impact analyses when `analysis-kinds: code-scanning`.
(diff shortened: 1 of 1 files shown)


src/config-utils.ts
```diff
@@ -964,10 +964,13 @@ export async function initConfig(
     }
   }
 
-  // If we are in CCR or the corresponding FF is enabled, try to determine
+  // If we are in a dynamic workflow or the corresponding FF is enabled, try to determine
   // which files in the repository are marked as generated and add them to
   // the `paths-ignore` configuration.
-  if ((await features.getValue(Feature.IgnoreGeneratedFiles)) && isCCR()) {
+  if (
+    (await features.getValue(Feature.IgnoreGeneratedFiles)) &&
```
henrymercer, line 971: Change of behaviour, but should be safe since this FF isn't rolled out yet.

henrymercer (approved):
🫡

src/config-utils.ts
```diff
@@ -964,10 +964,13 @@ export async function initConfig(
...
+    isDynamicWorkflow()
+  ) {
```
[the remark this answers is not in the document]
mbg: This is true, and I am pretty sure we had this on the PR that introduced it, and just like then I am not sure I want to re-run all CI for it 😆

src/config-utils.ts
mbg: Yeah, it's not rolled out. We also know since merging that feature that it may be more widely useful, so it makes sense to widen the condition here while we are at it.

PR #3506, opened 2026-02-24
Improve type inference of `Result<T, E>`

henrymercer:
Thanks @mbg for pointing out this drawback.

By defining `Result<T, E>` as a sum type `Success<T> | Failure<E>`, we can now infer that a `Result` is a failure if it is not a success.  For example, if we have:

```ts
const a = new Success(2) as Result<number, string>;
if (a.isSuccess()) {
  logger.info(`Success with value ${a.value}`);
} else {
  logger.info(`Failure with error ${a.value}`);
}
```

we can now infer in the else branch that `a.value` must be a `string`.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1295,7 +1295,7 @@ export function joinAtMost(
   return array.join(separator);
 }
 
-/** A simple result type representing either a success or a failure. */
+/** An interface representing something that is either a success or a failure. */
 interface ResultLike<T, E> {
   /** The value of the result, which can be either a success value or a failure value. */
   value: T | E;


src/util.ts
```diff
@@ -1295,42 +1295,51 @@ export function joinAtMost(
   return array.join(separator);
 }
 
-/** A success result. */
-type Success<T> = Result<T, never>;
-/** A failure result. */
-type Failure<E> = Result<never, E>;
+/** A simple result type representing either a success or a failure. */
```
mbg: Minor: this has the same description as `Result`

```suggestion
/** An interface representing something that is either a success or a failure. */
```

mbg (approved):

henrymercer:
(Discussed offline: the class-based approach lets us use a fluent API for functions like `orElse`, which is slightly more readable)

PR #3507, opened 2026-02-24
Add a repository property for disabling overlay 

henrymercer:
Add a repository property that can be used to disable overlay for repositories that are owned by an organisation.

This means repositories that are owned by an organisation can disable overlay without needing to disable default setup too.

### Risk assessment

Workflow types:

Products:

Environments:

- **Test repository** - This change will be tested on a test repository before merging. (backlinked)

### DIFF
--- a/lib/autobuild-action.js
+++ b/lib/autobuild-action.js
@@ -104154,6 +104154,11 @@ var featureConfig = {
     // cannot be found when interpreting results.
     minimumVersion: void 0
   },
+  ["start_proxy_use_features_release" /* StartProxyUseFeaturesRelease */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_START_PROXY_USE_FEATURES_RELEASE",
+    minimumVersion: void 0
+  },
   ["upload_overlay_db_to_api" /* UploadOverlayDbToApi */]: {
     defaultValue: false,
     envVar: "CODEQL_ACTION_UPLOAD_OVERLAY_DB_TO_API",
--- a/lib/resolve-environment-action.js
+++ b/lib/resolve-environment-action.js
@@ -104145,6 +104145,11 @@ var featureConfig = {
     // cannot be found when interpreting results.
     minimumVersion: void 0
   },
+  ["start_proxy_use_features_release" /* StartProxyUseFeaturesRelease */]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_START_PROXY_USE_FEATURES_RELEASE",
+    minimumVersion: void 0
+  },
   ["upload_overlay_db_to_api" /* UploadOverlayDbToApi */]: {
     defaultValue: false,
     envVar: "CODEQL_ACTION_UPLOAD_OVERLAY_DB_TO_API",
--- a/src/feature-flags.ts
+++ b/src/feature-flags.ts
@@ -77,6 +77,7 @@ export enum Feature {
   QaTelemetryEnabled = "qa_telemetry_enabled",
   /** Note that this currently only disables baseline file coverage information. */
   SkipFileCoverageOnPrs = "skip_file_coverage_on_prs",
+  StartProxyUseFeaturesRelease = "start_proxy_use_features_release",
   UploadOverlayDbToApi = "upload_overlay_db_to_api",
   UseRepositoryProperties = "use_repository_properties_v2",
   ValidateDbConfig = "validate_db_config",
@@ -325,6 +326,11 @@ export const featureConfig = {
     // before rolling this out externally, we should set a minimum version here
     // since current versions of the CodeQL CLI will log if baseline information
     // cannot be found when interpreting results.
+    minimumVersion: undefined,
+  },
+  [Feature.StartProxyUseFeaturesRelease]: {
+    defaultValue: false,
+    envVar: "CODEQL_ACTION_START_PROXY_USE_FEATURES_RELEASE",
     minimumVersion: undefined,
   },
   [Feature.UploadOverlayDbToApi]: {
--- a/src/start-proxy-action.ts
+++ b/src/start-proxy-action.ts
@@ -5,7 +5,7 @@ import * as core from "@actions/core";
 
 import * as actionsUtil from "./actions-util";
 import { getGitHubVersion } from "./api-client";
-import { Feature, FeatureEnablement, initFeatures } from "./feature-flags";
+import { FeatureEnablement, initFeatures } from "./feature-flags";
 import { KnownLanguage } from "./languages";
 import { getActionsLogger, Logger } from "./logging";
 import { getRepositoryNwo } from "./repository";
@@ -98,7 +98,7 @@ async function run(startedAt: Date) {
     };
 
     // Start the Proxy
-    const proxyBin = await getProxyBinaryPath(logger);
+    const proxyBin = await getProxyBinaryPath(logger, features);
     const proxyInfo = await startProxy(
       proxyBin,
       proxyConfig,
--- a/src/start-proxy.test.ts
+++ b/src/start-proxy.test.ts
@@ -7,6 +7,7 @@ import sinon from "sinon";
 
 import * as apiClient from "./api-client";
 import * as defaults from "./defaults.json";
+import { setUpFeatureFlagTests } from "./feature-flags/testing-util";
 import { KnownLanguage } from "./languages";
 import { getRunnerLogger, Logger } from "./logging";
 import * as startProxyExports from "./start-proxy";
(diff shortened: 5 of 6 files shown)


src/feature-flags/properties.ts
```diff
@@ -7,28 +7,50 @@ import { GitHubVariant, GitHubVersion } from "../util";
...
+  value: string,
+): void {
+  properties[name] = mapRepositoryProperties[name](value);
+}
+
 /**
  * A repository property has a name and a value.
  */
-export interface RepositoryProperty {
+interface GitHubRepositoryProperty {
```
mbg: Minor: In general, I think it's a good idea to export types that are (transitively) visible through other, exported types. I'd only leave them as not exported if they are really only used internally in one file.

src/feature-flags/properties.ts
```diff
@@ -7,28 +7,50 @@ import { GitHubVariant, GitHubVersion } from "../util";
...
+
+type AllRepositoryProperties = {
+  [RepositoryPropertyName.DISABLE_OVERLAY]: boolean;
+  [RepositoryPropertyName.EXTRA_QUERIES]: string;
+};
+
+export type RepositoryProperties = Partial<AllRepositoryProperties>;
+
+const mapRepositoryProperties: {
+  [K in RepositoryPropertyName]: (value: string) => AllRepositoryProperties[K];
```
mbg: Minor: Name the function type to make it clearer what it is for? Possibly also a comment on `mapRepositoryProperties` itself?

src/feature-flags/properties.ts
```diff
@@ -7,28 +7,50 @@ import { GitHubVariant, GitHubVersion } from "../util";
...
+} = {
+  [RepositoryPropertyName.DISABLE_OVERLAY]: (value) => value === "true",
```
mbg: Minor: Since this is new, consider making this check case-insensitive?

src/feature-flags/properties.ts
```diff
@@ -7,28 +7,50 @@ import { GitHubVariant, GitHubVersion } from "../util";
...
+};
+
+function setProperty<K extends RepositoryPropertyName>(
+  properties: RepositoryProperties,
+  name: K,
+  value: string,
+): void {
+  properties[name] = mapRepositoryProperties[name](value);
+}
+
```
mbg: Minor/Question: Any particular reason why this is a separate function and isn't inlined below?

src/feature-flags/properties.ts
```diff
@@ -7,28 +7,50 @@ import { GitHubVariant, GitHubVersion } from "../util";
  * Enumerates repository property names that have some meaning to us.
  */
 export enum RepositoryPropertyName {
+  DISABLE_OVERLAY = "github-codeql-disable-overlay",
```
mbg, line 10: Design question: Do you think it would make sense to have control over the base overlay / overlay usage enablement separately?

src/feature-flags/properties.ts
henrymercer: This is consistent with other boolean checks, e.g. in environment variables.  Also the repository properties UI has support for boolean properties, which will be set to either "true" or "false".

src/feature-flags/properties.ts
henrymercer: I couldn't prove that this was type safe when inlining it.  If you can get this to work in a type safe way I'd be happy to change it.

src/feature-flags/properties.ts
henrymercer: Hmm, that kind of distinction seems useful to us when we're rolling stuff out, but can you think of a good reason why customers would want to build overlay-base databases without using them in PR analysis?

src/feature-flags/properties.ts
mbg: Agreed that it is consistent, but is it intuitive? Do we gain anything from being this strict here?

For the repo properties UI, doesn't that require each org to individually configure the property as a boolean? If so, it's easy to imagine someone accidentally configuring it as a string and then using `True` etc.

henrymercer: It's a valid point, but on the flip side if we allow `True` (or `1`, etc) here, it's surprising when that isn't accepted elsewhere.  We can make it clear in our instructions that this should be configured as a boolean property.

src/feature-flags/properties.ts
mbg: I imagine that there could be (unlikely/infrequent) scenarios where it might be desirable to temporarily do one or the other. I mainly wanted to surface the question so that we can think about it, since it might be harder to change down the line.

src/feature-flags/properties.ts
```diff
@@ -24,31 +26,45 @@ type AllRepositoryProperties = {
...
-  [RepositoryPropertyName.DISABLE_OVERLAY]: (value) => value === "true",
-  [RepositoryPropertyName.EXTRA_QUERIES]: (value) => value,
+  [RepositoryPropertyName.DISABLE_OVERLAY]: (name, value, logger) => {
+    if (value !== "true" && value !== "false") {
+      logger.warning(
+        `Repository property '${name}' has unexpected value '${value}'. Expected 'true' or 'false'. Defaulting to false.`,
+      );
+    }
+    return value === "true";
+  },
```
mbg: Minor: given what you wrote in https://github.com/github/codeql-action/pull/3507#discussion_r2849353046, would it make sense to extract this into a function somewhere (`util`?) and make it a little more general to use/allow it to be used in more places where we only accept `true` or `false`?

henrymercer: I don't dislike the idea, but I'd prefer that we tackle that in a separate PR.  For now, I've created `parseBooleanRepositoryProperty` and `parseStringRepositoryProperty` so we have an abstraction for future repository properties.

mbg: That's fine, doesn't have to be in this PR.

src/feature-flags/properties.ts
henrymercer: I can't think of a use case right now.  On the topic of things being easy to change, I considered just an "overlay" property instead (true = enable, false = disable), but since this would always override the default enablement, I think it's clearer just to have a disable property.

mbg: No strong feelings on this. The scenarios I can think of are not likely, and we have other ways of addressing those (at least presently with the FFs).

src/feature-flags/properties.ts
```diff
@@ -7,28 +7,43 @@ import { GitHubVariant, GitHubVersion } from "../util";
...
 export enum RepositoryPropertyName {
+  DISABLE_OVERLAY = "github-codeql-disable-overlay",
   EXTRA_QUERIES = "github-codeql-extra-queries",
 }
 
+/** Parsed types of the known repository properties. */
+type AllRepositoryProperties = {
+  [RepositoryPropertyName.DISABLE_OVERLAY]: boolean;
+  [RepositoryPropertyName.EXTRA_QUERIES]: string;
+};
```
mbg, line 18: I didn't comment on this in my initial review, but I quite like this mapping of properties to expected types and the associated machinery below.

src/config-utils.ts
```diff
@@ -750,16 +753,19 @@ export async function getOverlayDatabaseMode(
...
 ): Promise<{
   overlayDatabaseMode: OverlayDatabaseMode;
   useOverlayDatabaseCaching: boolean;
   skippedDueToCachedStatus: boolean;
+  disabledByRepositoryProperty: boolean;
 }> {
   let overlayDatabaseMode = OverlayDatabaseMode.None;
   let useOverlayDatabaseCaching = false;
   let skippedDueToCachedStatus = false;
+  let disabledByRepositoryProperty = false;
```
mbg: Minor: We have this growing list of causes why overlay analysis is skipped/disabled, but presumably only one of these can be `true` at a given time. (At least based on the current logic, `disabledByRepositoryProperty` seems to imply `skippedDueToCachedStatus` is `false`.) Not blocking for this PR, but perhaps this should be refactored into an `enum` or similar that indicates the reason why overlay analysis was skipped/disabled. Otherwise, we might be mislead into thinking that e.g. `skippedDueToCachedStatus` would be `true` if we'd be skipping overlay analysis based on cache status, even if we already decided to disable it based on the repo property.

src/config-utils.ts
```diff
@@ -1106,6 +1125,31 @@ export async function initConfig(
     );
   }
 
+  if (overlayDisabledByRepositoryProperty) {
```
mbg: (Similarly to my comment above, the way these diagnostics are emitted might suggest that all of these could be generated at the same time.)

src/feature-flags/properties.ts
mbg: Ah, I hadn't realised that a straight-forward inlining wouldn't just work when I wrote that comment. I gave this a quick shot, but also couldn't immediately find a nicer way of doing this, so your solution seems good to me.

src/config-utils.ts
henrymercer: Yeah, great point, let's make this an enum.

mbg:
All good, except for https://github.com/github/codeql-action/pull/3507/commits/03178ce08519198ec64477291c99c11cb5ab5922. I opened https://github.com/github/codeql-action/pull/3512 to address that more permanently.

henrymercer:
We could merge this and then revert https://github.com/github/codeql-action/commit/03178ce08519198ec64477291c99c11cb5ab5922 in https://github.com/github/codeql-action/pull/3512 to get checks on `main` green?

mbg:
@henrymercer that PR is now merged, so we can drop https://github.com/github/codeql-action/commit/03178ce08519198ec64477291c99c11cb5ab5922 from this.

henrymercer:
@mbg This is ready for another look

mbg (approved):

PR #3512, opened 2026-02-25
Use `getDefaultCliVersion` for `start-proxy`

mbg:
Changes the `start-proxy` action to use `getDefaultCliVersion` rather than having `defaults.bundleVersion` hard-coded. Conveniently, this adds a permanent consumer of feature flags to `start-proxy`.

I have put this behind a new FF to gate the new behaviour to make this a less risky change.

### Risk assessment

- **Low risk** 

Workflow types:

Products:

Environments:

### DIFF
--- a/src/start-proxy.ts
+++ b/src/start-proxy.ts
@@ -452,14 +452,14 @@ export async function getDownloadUrl(
     for (const asset of cliRelease.data.assets) {
       if (asset.name === proxyPackage) {
         logger.info(
-          `Found '${proxyPackage}' in release '${defaults.bundleVersion}' at '${asset.url}'`,
+          `Found '${proxyPackage}' in release '${versionInfo.tagName}' at '${asset.url}'`,
         );
         return {
           url: asset.url,
           // The `update-job-proxy` doesn't have a version as such. Since we now bundle it
           // with CodeQL CLI bundle releases, we use the corresponding CLI version to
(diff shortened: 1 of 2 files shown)


src/start-proxy.ts
```diff
@@ -391,46 +397,69 @@ export function getFallbackUrl(proxyPackage: string): string {
...
-    tag: defaults.bundleVersion,
+    tag: version,
   });
 }
 
+/** Uses `features` to determine the default CLI version. */
+async function getCliVersionFromFeatures(
+  features: FeatureEnablement,
+): Promise<CodeQLDefaultVersionInfo> {
+  const gitHubVersion = await getGitHubVersion();
```
sam-robson, line 416: To save an API call, could we thread githubVersion through from https://github.com/github/codeql-action/blob/bce0deb953db43af59053310d09fa3aabd3c04c5/src/start-proxy-action.ts#L49 ?

mbg: We could thread it through, but `getGitHubVersion` should also cache the result from the API on the first call, so that the second call to `getGitHubVersion` here shouldn't result in an additional API call.

sam-robson: 👍🏻 I see. Well I'll leave that as a judgment call to you, happy to approve.

mbg: Let's get this merged to unblock other PRs, thanks 👍🏻

PR #3526, opened 2026-02-28
Convert `sync.py` to TypeScript

mbg:
This PR takes a first step at converting a script from Python to TypeScript. Concretely, it replaces the `sync.py` script that turns our PR check specifications into GitHub Actions workflows with a roughly equivalent TypeScript replacement.

Initially, I aimed for exact textual equivalence in the output between the two scripts, but then I ultimately allowed a few minor changes:

- The formatting of the `steps` from the specification is now preserved more accurately. 
    - This mainly allows more comments from the specification to be preserved and in more accurate locations.
    - There are also some cases where there are now quoted strings in the output where there weren't quotes before. For consistency, all double quoted strings in the `steps` are converted to single quoted strings.
- The `yaml` package allows for unlimited line length, which `ruamel.yaml` didn't quite handle correctly before, so some lines are no longer wrapped as intended.

I initially started by using `js-yaml`, since we already use this elsewhere in the repo, but then switched to `yaml` since it gave me more control over the formatting.

Follow-up in https://github.com/github/codeql-action/pull/3529 for `sync_back.py`.

### Risk assessment

Environments:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,7 +92,7 @@ We typically deprecate a version of CodeQL when the GitHub Enterprise Server (GH
 1. Remove support for the old version of CodeQL.
     - Bump `CODEQL_MINIMUM_VERSION` in `src/codeql.ts` to the new minimum version of CodeQL.
     - Remove any code that is only needed to support the old version of CodeQL. This is often behind a version guard, so look for instances of version numbers between the old minimum version and the new minimum version in the codebase. A good place to start is the list of version numbers in `src/codeql.ts`.
-    - Update the default set of CodeQL test versions in `pr-checks/sync.py`.
+    - Update the default set of CodeQL test versions in `pr-checks/sync.ts`.
         - Remove the old minimum version of CodeQL.
         - Add the latest patch release for any new CodeQL minor version series that have shipped in GHES.
         - Run the script to update the generated PR checks.


pr-checks/sync.ts
```diff
@@ -0,0 +1,528 @@
...
+      Scalar(_key, node) {
+        if (node.type === "QUOTE_DOUBLE") {
+          node.type = "QUOTE_SINGLE";
+        }
+      }
+    });
+
+    // Add the generated steps in front of the ones from the specification.
+    specSteps.items.unshift(...steps);
+
```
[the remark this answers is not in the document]
mbg: This is no longer used and intentionally removed from the new version.

pr-checks/sync.ts
```diff
@@ -0,0 +1,528 @@
...
+      name: "Prepare test",
+      id: "prepare-test",
+      uses: "./.github/actions/prepare-test",
+      with: {
+        version: "${{ matrix.version }}",
+        "use-all-platform-bundle": useAllPlatformBundle,
+        // If the action is being run from a container, then do not setup kotlin.
+        // This is because the kotlin binaries cannot be downloaded from the container.
+        "setup-kotlin": String(
+          !("container" in checkSpecification),
```
esbena: This seems somehow like the dual of the (convoluted IMO) `isTruthy` introduced earlier.
My preference is to be more explicit here with a ternary that produces literal true/false:

```suggestion
        "setup-kotlin": "container" in checkSpecification? "false": "true"
```

pr-checks/sync.ts
```diff
@@ -0,0 +1,528 @@
...
+        specification: checkSpecification,
+        checkName,
+        inputs: workflowInputs,
+      });
+    }
+
+    let extraGroupName = "";
+    for (const inputName of Object.keys(workflowInputs)) {
+      extraGroupName += "-${{inputs." + inputName + "}}";
+    }
```
esbena, line 428: Lets do a very simple sanitization check here, namely that the input name is a valid identifier.
It's not so much for security I worry, it's rather just the ease of diagnosing invalid workflow files.

pr-checks/sync.ts
```diff
@@ -0,0 +1,528 @@
...
+    let combinedInputs: Record<string, WorkflowInput> = {};
+
+    for (const check of collections[collectionName]) {
+      const { checkName, specification, inputs: checkInputs } = check;
+      const checkWith: Record<string, string> = {};
+
+      combinedInputs = { ...combinedInputs, ...checkInputs };
+
+      for (const inputName of Object.keys(checkInputs)) {
+        checkWith[inputName] = "${{ inputs." + inputName + " }}";
```
esbena, line 488: Ditto the earlier comment about sanitization

pr-checks/sync.ts
```diff
@@ -0,0 +1,528 @@
...
+            version,
+          });
+        }
+      }
+    }
+
+    let useAllPlatformBundle = "false"; // Default to false
+    if (checkSpecification.useAllPlatformBundle) {
+      useAllPlatformBundle = checkSpecification.useAllPlatformBundle;
+    }
```
esbena: This should be a one liner. 

```suggestion
    const useAllPlatformBundle = checkSpecification.useAllPlatformBundle? checkSpecification.useAllPlatformBundle: false;
```
or
```suggestion
    const useAllPlatformBundle = !!checkSpecification.useAllPlatformBundle;
```

pr-checks/sync.ts
```diff
@@ -0,0 +1,528 @@
...
+  });
+  const yamlStr = yaml.stringify(workflowDoc, {
+    aliasDuplicateObjects: false,
+    singleQuote: true,
+    lineWidth: 0,
+  });
+  fs.writeFileSync(filePath, stripTrailingWhitespace(header + yamlStr), "utf8");
+}
+
+function isTruthy(value: string | boolean | undefined): boolean {
```
esbena: This function needs a comment, but I'm not sure about the existence at all actually. The truthiness check in JavaScript is widely known to be `!!value`. This implementation changes it very subtly.
Perhaps a rename to something like `isBooleanTrueOrStringTrue`?

esbena:
LGTM, I only have some stylistic comments. The migration is pleasantly simple, and the near zero-diff strategy seems robust enough here.

mbg:
@esbena Thanks for the review! I think I have come up with reasonably good ways to address your feedback. In particular:

- I have eliminated `isTruthy` entirely since that was mainly handling cases where we used strings instead of booleans in the specifications. I have changed them to all use booleans where `isTruthy` was previously used.
- We didn't actually use the custom `inputs` in workflow specifications, so all of them were statically known in the script. I have typed them to make this more explicit.

PR #3528, opened 2026-03-01
Refactor SARIF-related types and functions into a separate module

mbg:
This PR refactors SARIF-related types and functions from `util.ts` and `upload-lib.ts` to a separate `sarif` module. The primary motivation for this is follow-up PRs in which we can then make use of the `sarif` module without introducing circular dependencies and add tests that mock functions from the `sarif` module more easily.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/sarif/index.ts
+++ b/src/sarif/index.ts
@@ -6,8 +6,12 @@ import * as sarif from "sarif";
 
 export type * from "sarif";
 
-// `automationId` is non-standard.
+// Extends `ToolComponent` with the non-standard `automationId` property we use.
 export type RunKey = sarif.ToolComponent & {
+  /**
+   * Describes a SARIF run (either uniquely or not uniquely) based on the criteria used by
+   * Code Scanning to determine analysis categories
+   */
   automationId: string | undefined;
 };
 


src/sarif/index.ts
```diff
@@ -0,0 +1,189 @@
+import * as fs from "fs";
+
+import { Logger } from "../logging";
+
+export interface SarifLocation {
```
henrymercer: An alternative would be taking these types from an npm module.  That would also give us JSDoc.  Perhaps now is a good time to make the switch, or document why not.

mbg: That would make sense. I guess the downsides are an extra dependency and having the full structure might require us to update some tests.

I can look at making this change if we are happy with that.

henrymercer: I think if we use https://www.npmjs.com/package/@types/sarif there should be no impact on the compiled code.

mbg: Yeah that's what I was planning to use.

mbg: I have made this change. For now, I have just imported it in `sarif/index.ts` and then re-exported the relevant types under the old names. If you felt strongly about it, we could replace the types throughout the codebase with those from `@types/sarif`.

henrymercer: > we could replace the types throughout the codebase with those from `@types/sarif`.

It's a moderate preference for me.  If you feel strongly I am happy to approve the PR.

mbg: I have pushed https://github.com/github/codeql-action/pull/3528/commits/b43d146e3737bf31278fa258423a935173411e29 which makes this change, but I am not sure how I feel about it. 

Because the type names from `@types/sarif` do not include `Sarif` and have fairly generic names, the module now essentially has to be imported as qualified so that the naming doesn't become confusing.

src/sarif/index.ts
```diff
@@ -0,0 +1,117 @@
...
+import { Logger } from "../logging";
+
+import * as sarif from "sarif";
+
+export type * from "sarif";
+
+// `automationId` is non-standard.
+export type RunKey = sarif.ToolComponent & {
+  automationId: string | undefined;
+};
```
henrymercer: Could we update the comment to be clearer about the purpose, since this isn't a SARIF object, rather an identifier based on a tool component and an automation ID?

src/sarif/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  return combinedSarif;
+}
+
+/**
+ * Checks whether all the runs in the given SARIF files were produced by CodeQL.
+ * @param sarifObjects The list of SARIF objects to check.
+ */
+export function areAllRunsProducedByCodeQL(sarifObjects: sarif.Log[]): boolean {
```
henrymercer: Nit: `sarifLogs` might be more descriptive than `sarifObjects`

src/util.ts
```diff
@@ -17,6 +17,8 @@ import { EnvVar } from "./environment";
 import { Language } from "./languages";
 import { Logger } from "./logging";
 
+export * from "./sarif";
```
henrymercer: Should this be removed and imports updated?

src/sarif/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  const combinedSarif: sarif.Log = {
+    version: "2.1.0",
+    runs: [],
+  };
+
+  for (const sarifFile of sarifFiles) {
+    logger.debug(`Loading SARIF file: ${sarifFile}`);
+    const sarifObject = readSarifFile(sarifFile);
+    // Check SARIF version
+    if (combinedSarif.version === null) {
```
henrymercer: This cannot be true (I'm surprised this was not flagged by the compiler or a linter)

src/sarif/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+// Takes a list of paths to sarif files and combines them together,
+// returning the contents of the combined sarif file.
+export function combineSarifFiles(
+  sarifFiles: string[],
+  logger: Logger,
+): sarif.Log {
+  logger.info(`Loading SARIF file(s)`);
+  const combinedSarif: sarif.Log = {
+    version: "2.1.0",
```
henrymercer: This seems like a potentially breaking change — now we'll only accept SARIF v2.1.0

src/upload-lib.ts
```diff
@@ -529,9 +447,9 @@ function countResultsInSarif(sarif: string): number {
   return numResults;
 }
 
-export function readSarifFile(sarifFilePath: string): SarifFile {
+export function readSarifFile(sarifFilePath: string): sarif.Log {
```
henrymercer: It might be worth renaming this function since we now have a `readSarifFile` from `sarif` as well.

src/sarif/index.ts
mbg: I am surprised that it wasn't flagged as well, but it can actually be true (and also partially answers your question in https://github.com/github/codeql-action/pull/3528#discussion_r2876944294). We don't perform validation in `readSarifFile` so the `sarif.Log` object here may not have a `version` property. That's no longer explicit in the type though. 

I'll see what the impact of changing the type there to something like `Partial<sarif.Log>` would be.

src/sarif/index.ts
mbg: Not quite. We haven't changed anything about how we parse or validate SARIF files. So although the type from `@types/sarif` won't allow us to construct a `sarif.Log` object without a version "2.1.0" string in TypeScript code, we'll happily parse SARIF files without it. Our validator also didn't previously reject such files.

henrymercer: I suppose since code scanning only supports v2.1.0 it's not necessarily a bad thing if we fail if users try to upload a SARIF file with a different version number.  Though it does seem like a breaking change if the API would previously tolerate different version numbers.

Could you explain why this combining logic wouldn't fail with the "Different SARIF versions encountered" error message if a SARIF file contains a different version number or a missing version number?

src/sarif/index.ts
henrymercer: I don't see how.  We set `combinedSarif.version` to `2.1.0` above, and only overwrite it if it is `null`.

src/sarif/index.ts
henrymercer: e.g. something like "Describes a SARIF run (either uniquely or not uniquely) based on the criteria used by Code Scanning to determine analysis categories"

src/sarif/index.ts
mbg: Right, sorry. I missed that this (and the other comment) were specifically in the context of `combineSarifFiles` and not the `@types/sarif` library's requirement for the version to be `2.1.0`.

I have re-jigged this function slightly now so that the old behaviour is maintained, unless there is no `version` property present. In that case, we insert `2.1.0` in the resulting SARIF.

This was supposed to be a fairly small PR to unblock me on something else, so I am not sure I want to deal with a breaking change here.

PR #3537, opened 2026-03-03
Record the job that published an overlay status

henrymercer:
This makes it easier to find the job that produced the status.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/init-action-post-helper.test.ts
+++ b/src/init-action-post-helper.test.ts
@@ -374,6 +374,7 @@ test("saves overlay status when overlay-base analysis did not complete successfu
         attemptedToBuildOverlayBaseDatabase: true,
         builtOverlayBaseDatabase: false,
         job: {
+          checkRunId: undefined,
           workflowRunId: 12345,
           workflowRunAttempt: 1,
           name: "analyze",
--- a/src/init-action-post-helper.ts
+++ b/src/init-action-post-helper.ts
@@ -283,7 +283,7 @@ async function recordOverlayStatus(
       attemptedToBuildOverlayBaseDatabase: true,
       builtOverlayBaseDatabase: false,
     },
-    Number.isNaN(checkRunId) ? undefined : checkRunId,
+    checkRunId !== undefined && checkRunId >= 0 ? checkRunId : undefined,
   );
 
   const diskUsage = await checkDiskUsage(logger);
--- a/src/overlay/status.ts
+++ b/src/overlay/status.ts
@@ -74,7 +74,7 @@ export function createOverlayStatus(
     workflowRunId: getWorkflowRunID(),
     workflowRunAttempt: getWorkflowRunAttempt(),
     name: getRequiredEnvParam("GITHUB_JOB"),
-    ...(checkRunId !== undefined && { checkRunId }),
+    checkRunId,
   };
   return {
     ...attributes,


src/init-action-post-helper.test.ts
```diff
@@ -316,6 +316,9 @@ test("not uploading failed SARIF when `code-scanning` is not an enabled analysis
 test("saves overlay status when overlay-base analysis did not complete successfully", async (t) => {
   return await util.withTmpDir(async (tmpDir) => {
     process.env["GITHUB_REPOSITORY"] = "github/codeql-action-fake-repository";
+    process.env["GITHUB_RUN_ID"] = "12345";
+    process.env["GITHUB_RUN_ATTEMPT"] = "1";
+    process.env["GITHUB_JOB"] = "analyze";
     process.env["RUNNER_TEMP"] = tmpDir;
```
mbg, line 322: Minor: It might be nice to extend `setupActionsVars` with all environment variables that we'd typically expect in an Actions environment and modify it to accept an argument with a (partial) mapping of specific environment variables we want to set. Not necessarily something for this PR.

henrymercer: Good idea.  I'll take a look at this in a separate PR.

src/overlay/status.ts
```diff
@@ -66,14 +68,17 @@ export interface OverlayStatus {
 /** Creates an `OverlayStatus` populated with the details of the current job. */
 export function createOverlayStatus(
   attributes: Omit<OverlayStatus, "job">,
+  checkRunId?: number,
 ): OverlayStatus {
+  const job: JobInfo = {
+    workflowRunId: getWorkflowRunID(),
+    workflowRunAttempt: getWorkflowRunAttempt(),
+    name: getRequiredEnvParam("GITHUB_JOB"),
+    ...(checkRunId !== undefined && { checkRunId }),
```
mbg: I think we had a similar point on another PR recently: this is mostly equivalent to just `checkRunId`, except that with this the `JobInfo` object doesn't even contain the key? Do you think it's beneficial to explicitly not have the key, rather than it being set to `undefined` here? If so, it would be good to document why. Otherwise, change it to just `checkRunId`.

src/init-action-post-helper.ts
```diff
@@ -274,10 +274,17 @@ async function recordOverlayStatus(
...
+  const checkRunIdInput = actionsUtil.getOptionalInput("check-run-id");
+  const checkRunId =
+    checkRunIdInput !== undefined ? parseInt(checkRunIdInput, 10) : undefined;
+
+  const overlayStatus: OverlayStatus = createOverlayStatus(
+    {
+      attemptedToBuildOverlayBaseDatabase: true,
+      builtOverlayBaseDatabase: false,
+    },
+    Number.isNaN(checkRunId) ? undefined : checkRunId,
```
mbg: Should we also check that this is `>= 0` (or similar)?

src/overlay/status.ts
henrymercer: Since we can't serialize undefined values to JSON, there's no difference in practice.  So we might as well use just `checkRunId`

src/init-action-post-helper.ts
henrymercer: Sure 👍

PR #3538, opened 2026-03-03
Break down overlay disabled reason

henrymercer:
Break down the two most common categories to help us better understand why overlay isn't run.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/config-utils.test.ts
+++ b/src/config-utils.test.ts
@@ -139,7 +139,7 @@ function mockListLanguages(languages: string[]) {
   sinon.stub(api, "getApiClient").value(() => client);
 }
 
-test("load empty config", async (t) => {
+test.serial("load empty config", async (t) => {
   return await withTmpDir(async (tempDir) => {
     const logger = getRunnerLogger(true);
     const languages = "javascript,python";
@@ -180,7 +180,7 @@ test("load empty config", async (t) => {
   });
 });
 
-test("load code quality config", async (t) => {
+test.serial("load code quality config", async (t) => {
   return await withTmpDir(async (tempDir) => {
     const logger = getRunnerLogger(true);
     const languages = "actions";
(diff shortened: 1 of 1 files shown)


src/config-utils.ts
```diff
@@ -653,14 +653,19 @@ const OVERLAY_ANALYSIS_CODE_SCANNING_FEATURES: Record<Language, Feature> = {
   swift: Feature.OverlayAnalysisCodeScanningSwift,
 };
 
-async function isOverlayAnalysisFeatureEnabled(
+/**
+ * Checks whether the overlay analysis feature is enabled for the given
+ * languages and configuration, returning the specific reason it is disabled,
+ * or `undefined` if it is enabled.
+ */
+async function getOverlayFeatureDisabledReason(
```
mbg: I don't think this is a good re-naming for this function. The function name should be focused on what the function does, but the new name suggests this is a getter which just retrieves some existing value. But the function actually performs the checks, and only returns the reason why overlay analysis is disabled if one of those checks is negative. I view it returning the reason why overlay is disabled as secondary.

The old name (`isOverlayAnalysisFeatureEnabled`) is better in this regard, but just `isOverlayAnalysisEnabled` would also be good. (Or indeed `isOverlayAnalysisDisabled`.)

More broadly, I wonder if I'd prefer a `{ enabled: true } | { enabled: false, reason: OverlayDisabledReason }` kind of type for these things, rather than `OverlayDisabledReason | undefined`. I don't particularly like using `undefined` to indicate a successful outcome / the absence of a non-standard condition.

src/overlay/diagnostics.ts
```diff
@@ -12,18 +12,27 @@ import { RepositoryPropertyName } from "../feature-flags/properties";
...
+  InsufficientDiskSpace = "insufficient-disk-space",
+  /** The runner does not have enough memory to perform overlay analysis. */
+  InsufficientMemory = "insufficient-memory",
+  /** Overlay analysis is not enabled for one or more of the configured languages. */
+  LanguageNotEnabled = "language-not-enabled",
   /** The source root is not inside a git repository. */
   NoGitRoot = "no-git-root",
+  /**
+   * Overlay analysis is restricted to code-scanning-only configs but the config uses custom
+   * queries or packs, disables default queries, or specifies query filters.
```
mbg: I know you have already changed this in response to Copilot's comment, but a more significant issue with this comment is the ambiguity over whether it refers to:

- `analysis-kind: code-scanning`, with no customisation; or
- "code-scanning-only configs" is supposed to mean no query customisations.

I.e. it is not clear if "code-scanning-only configs" is about analysis kinds or not.

src/config-utils.ts
```diff
@@ -677,39 +682,35 @@ async function isOverlayAnalysisFeatureEnabled(
...
     }
-    return false;
+    return OverlayDisabledReason.LanguageNotEnabled;
   }
   if (enableForCodeScanningOnly) {
     // A code-scanning configuration runs only the (default) code-scanning suite
     // if the default queries are not disabled, and no packs, queries, or
     // query-filters are specified.
-    return (
+    const isCodeScanningOnly =
```
mbg: See my other comment, but I am concerned about the naming here / relationship with analysis kinds. Since we have a specific `code-scanning` analysis kind, it is confusing that this seems to mean "no customisation".

src/config-utils.ts
```diff
@@ -754,23 +755,31 @@ async function runnerHasSufficientMemory(
...
  */
-async function runnerSupportsOverlayAnalysis(
+async function getResourceDisabledReason(
   codeql: CodeQL,
   diskUsage: DiskUsage | undefined,
   ramInput: string | undefined,
   logger: Logger,
   useV2ResourceChecks: boolean,
-): Promise<boolean> {
+): Promise<OverlayDisabledReason | undefined> {
```
mbg: Same comments as for `getOverlayFeatureDisabledReason`

src/config-utils.ts
```diff
@@ -837,12 +846,12 @@ export async function getOverlayDatabaseMode(
...
     disabledReason = OverlayDisabledReason.DisabledByRepositoryProperty;
   } else if (
-    await isOverlayAnalysisFeatureEnabled(
+    (disabledReason = await getOverlayFeatureDisabledReason(
       features,
       codeql,
       languages,
       codeScanningConfig,
-    )
+    )) === undefined
```
mbg: It would be good to move this assignment out of the condition.

src/config-utils.ts
```diff
@@ -858,18 +867,19 @@ export async function getOverlayDatabaseMode(
...
       performResourceChecks &&
-      !(await runnerSupportsOverlayAnalysis(
+      (resourceDisabledReason = await getResourceDisabledReason(
         codeql,
         diskUsage,
         ramInput,
         logger,
         useV2ResourceChecks,
-      ))
+      )) !== undefined
```
mbg: Same as above

src/config-utils.ts
```diff
@@ -655,17 +655,21 @@ const OVERLAY_ANALYSIS_CODE_SCANNING_FEATURES: Record<Language, Feature> = {
...
  */
-async function getOverlayFeatureDisabledReason(
+async function checkOverlayAnalysisFeatureEnabled(
   features: FeatureEnablement,
   codeql: CodeQL,
   languages: Language[],
   codeScanningConfig: UserConfig,
-): Promise<OverlayDisabledReason | undefined> {
+): Promise<
+  { enabled: true } | { enabled: false; reason: OverlayDisabledReason }
```
mbg: Minor: Since this is used at least a couple of times in this file now, extract it into a type definition `type OverlayEnabledResult = { enabled: true } | { enabled: false; reason: OverlayDisabledReason }`. 

We have similar-ish patterns elsewhere, so we could also move it to `util` as something like `type EnablementResult<T> = { enabled: true } | { enabled: false; reason: T }`, but this doesn't have to be done here.

Perhaps add helper definitions along the lines of `const enabled = { enabled: true }` and `function disabled<T>(reason: T) { return { enabled: false; reason }; }`

src/config-utils.ts
```diff
@@ -831,98 +840,150 @@ export async function getOverlayDatabaseMode(
...
     useOverlayDatabaseCaching: false,
     disabledReason: reason,
   });
 
   if (overlayDatabaseMode === OverlayDatabaseMode.None) {
-    return disabledResult(disabledReason);
+    return {
+      overlayDatabaseMode: OverlayDatabaseMode.None,
+      useOverlayDatabaseCaching: false,
+      disabledReason: undefined,
```
mbg: Minor: `disabledResult` above could accept `reason: OverlayDisabledReason | undefined` if `undefined` should be an acceptable `disabledReason`? 

I see that you had this previously -- was there a reason why you changed it to this? (Also `disabledResult` is duplicated now?)

src/config-utils.ts
```diff
@@ -831,98 +840,150 @@ export async function getOverlayDatabaseMode(
...
+    codeql,
+  ));
+  const useV2ResourceChecks = await features.getValue(
+    Feature.OverlayAnalysisResourceChecksV2,
+  );
+  const checkOverlayStatus = await features.getValue(
+    Feature.OverlayAnalysisStatusCheck,
+  );
+  const diskUsage =
+    performResourceChecks || checkOverlayStatus
```
mbg: Minor: `const needDiskUsage = performResourceChecks || checkOverlayStatus;` so the result can be re-used below.

src/config-utils.ts
```diff
@@ -831,98 +840,150 @@ export async function getOverlayDatabaseMode(
...
-    ));
-    const useV2ResourceChecks = await features.getValue(
-      Feature.OverlayAnalysisResourceChecksV2,
+    logger.warning(
+      `Unable to determine disk usage, therefore setting overlay database mode to ${OverlayDatabaseMode.None}.`,
+    );
+    return disabledResult(OverlayDisabledReason.UnableToDetermineDiskUsage);
+  }
+  const resourceResult =
+    performResourceChecks && diskUsage
```
mbg: Minor: make `diskUsage !== undefined` explicit

src/config-utils.ts
henrymercer: Only to try to avoid accidentally passing an undefined disabledReason when constructing a disabled result.  That's less necessary now I refactored to reduce the amount of nesting, so I'll accept undefined again which will also let me address that duplication.

src/config-utils.ts
henrymercer: I've decided to refactor this into a `Result`, since there's no need to repeat `overlayDatabaseMode: OverlayDatabaseMode.None` and `useOverlayDatabaseCaching: false`.

src/config-utils.ts
```diff
@@ -959,7 +964,7 @@ export async function getOverlayDatabaseMode(
  * appropriate disabled reason.
  */
 async function validateOverlayDatabaseMode(
-  overlayDatabaseMode: OverlayDatabaseMode,
+  overlayDatabaseMode: Exclude<OverlayDatabaseMode, OverlayDatabaseMode.None>,
```
mbg, line 943: Nice!

PR #3540, opened 2026-03-04
Testing: Provide default value for more environment variables in `setupActionsVars`

henrymercer:
As suggested in #3537.  I don't claim that this is comprehensive, but it introduces a pattern we can make use of to hopefully simplify our tests.

### DIFF
--- a/src/testing-utils.ts
+++ b/src/testing-utils.ts
@@ -144,7 +144,7 @@ export function setupTests(test: TestFn<any>) {
  * environment. Tests can override individual variables by passing them in the
  * `overrides` parameter.
  */
-export const DEFAULT_ACTIONS_VARS: Record<string, string> = {
+export const DEFAULT_ACTIONS_VARS = {
   GITHUB_ACTION_REPOSITORY: "github/codeql-action",
   GITHUB_API_URL: "https://api.github.com",
   GITHUB_EVENT_NAME: "push",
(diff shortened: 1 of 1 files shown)


src/testing-utils.ts
```diff
@@ -139,13 +139,40 @@ export function setupTests(test: TestFn<any>) {
...
+  GITHUB_JOB: "test-job",
+  GITHUB_REF: "refs/heads/main",
+  GITHUB_REPOSITORY: "github/codeql-action-testing",
+  GITHUB_RUN_ATTEMPT: "1",
+  GITHUB_RUN_ID: "1",
+  GITHUB_SERVER_URL: "https://github.com",
+  GITHUB_SHA: "0".repeat(40),
+  GITHUB_WORKFLOW: "test-workflow",
+  RUNNER_OS: "Linux",
+};
```
mbg: Minor: Could we remove the `Record<string, string>` type annotation and add `as const satisfies Record<string, string>`? Then we could retrieve the keys from this as values/type of all Actions environment variables.

src/testing-utils.ts
```diff
@@ -139,13 +139,40 @@ export function setupTests(test: TestFn<any>) {
...
+
 // Sets environment variables that make using some libraries designed for
 // use only on actions safe to use outside of actions.
-export function setupActionsVars(tempDir: string, toolsDir: string) {
+export function setupActionsVars(
+  tempDir: string,
+  toolsDir: string,
+  overrides?: Partial<Record<string, string>>,
```
mbg: Following from my first suggestion, we could then type this as something like `Partial<Record<keyof typeof DEFAULT_ACTIONS_VARS, string>>` to limit it to just Actions variables so that this doesn't just turn into a generic `setupEnv` function?

mbg (dismissed):
Thank you for following-up on this suggestion! 💜 

Looks good. I just have one suggestion for how we might want to type this differently.

mbg (approved):

PR #3541, opened 2026-03-04
Add support for validation jobs to `sync.ts`, and refactor

mbg:
The primary goal of this PR is to add support for additional, validation jobs to `sync.ts`. My motivation for this is https://github.com/github/codeql-action/pull/3519#pullrequestreview-3866873456. To add a PR check which does that, the most sensible way of testing it would be to run a second job after the main one which verifies that the artifact was successfully uploaded.

This PR doesn't yet add or modify a PR check which makes use of this new functionality. I will add that as part of https://github.com/github/codeql-action/pull/3519 once this PR is merged. 

I have refactored a few things in the first commits to break up the horrible long `main` function that we had initially ported from the earlier `sync.py` script as-is in https://github.com/github/codeql-action/pull/3526.

The refactoring has introduced some minor changes to the generated workflow files because the order of the various `setup-*` steps is now alphabetical, rather than effectively random. This should not have any impact on the behaviour of the workflows.

**Notes for reviewers**

Best reviewed commit-by-commit. Since these changes do not affect the CodeQL Action itself, a relatively lightweight review which focuses on the following is acceptable:

- The high-level design of this new functionality. I.e. are we happy to support "validation jobs" in this way?
- Are there any issues arising from the slight re-ordering of the setup steps / inputs in the workflows? 
- Any stylistic issues with the changes in this PR?

I have locally verified that the changes here broadly work for what I have in mind for the new workflow for https://github.com/github/codeql-action/pull/3526. Any small adjustments that may be required to support the full workflow could be added in that PR.

### Risk assessment

Environments:

N/A

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -52,18 +52,9 @@ jobs:
       - name: Verify compiled JS up to date
         run: .github/workflows/script/check-js.sh
 
-      - name: Verify PR checks up to date
-        if: always()
-        run: .github/workflows/script/verify-pr-checks.sh
-
       - name: Run unit tests
         if: always()
         run: npm test
-
-      - name: Run pr-checks tests
-        if: always()
-        working-directory: pr-checks
-        run: npm ci && npx tsx --test
 
       - name: Lint
         if: always() && matrix.os != 'windows-latest'
@@ -75,6 +66,40 @@ jobs:
         with:
           sarif_file: eslint.sarif
           category: eslint
+
+  # Verifying the PR checks are up-to-date requires Node 24. The PR checks are not dependent
+  # on the main codebase and therefore do not need to be run as part of the same matrix that
+  # we use for the `unit-tests` job.
+  verify-pr-checks:
+    name: Verify PR checks
+    if: github.triggering_actor != 'dependabot[bot]'
+    permissions:
+      contents: read
+    runs-on: ubuntu-slim
+    timeout-minutes: 10
+
+    steps:
+      - name: Prepare git (Windows)
+        if: runner.os == 'Windows'
+        run: git config --global core.autocrlf false
+
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'npm'
+
+      - name: Verify PR checks up to date
+        if: always()
+        run: .github/workflows/script/verify-pr-checks.sh
+
+      - name: Run pr-checks tests
+        if: always()
+        working-directory: pr-checks
+        run: npm ci && npx tsx --test
 
   check-node-version:
     if: github.triggering_actor != 'dependabot[bot]'


pr-checks/sync.ts
```diff
@@ -134,112 +283,129 @@ function stripTrailingWhitespace(content: string): string {
...
+  return matrix;
+}
+
+/**
+ * Retrieves setup steps and additional input definitions based on specific languages or frameworks
+ * that are requested by the `checkSpecification`.
+ *
+ * @returns An object containing setup steps and additional input specifications.
+ */
+function getSetupSteps(checkSpecification: Specification): {
```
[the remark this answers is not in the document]
mbg: Fun fact: this is not a type error in TypeScript. Because `Specification extends JobSpecification` and all fields added by `Specification` are optional, a minimum required `Specification` is a `JobSpecification` and values of type `JobSpecification` are therefore permitted as arguments where a `Specification` is expected.

pr-checks/sync.ts
```diff
@@ -90,6 +109,136 @@ const defaultTestVersions = [
...
+        type: "string",
+        description: "The version of Java to install",
+        required: false,
+        default: defaultLanguageVersions.java,
+      },
+    },
+    steps: [
+      {
+        name: "Install Java",
+        uses: "actions/setup-java@v5",
```
esbena, line 204: Is this stepping on the toes of the versioning management PR?

pr-checks/sync.ts
```diff
@@ -134,112 +283,130 @@ function stripTrailingWhitespace(content: string): string {
...
-  for (const file of checkFiles) {
-    const checkName = path.basename(file, ".yml");
-    const specDocument = loadYaml(file);
-    const checkSpecification = specDocument.toJS() as Specification;
-
-    console.log(`Processing: ${checkName} — "${checkSpecification.name}"`);
+  for (const version of checkSpecification.versions ?? defaultTestVersions) {
+    if (version === "latest") {
+      throw new Error(
+        'Did not recognise "version: latest". Did you mean "version: linked"?',
```
esbena: This should parameterize the version value to avoid weird error messages (much) later. 

```suggestion
        `Did not recognise "version: ${version}". Did you mean "version: linked"?`,
```

pr-checks/sync.ts
```diff
@@ -134,112 +283,130 @@ function stripTrailingWhitespace(content: string): string {
...
-    for (const version of checkSpecification.versions ?? defaultTestVersions) {
-      if (version === "latest") {
-        throw new Error(
-          'Did not recognise "version: latest". Did you mean "version: linked"?',
-        );
-      }
+    for (const operatingSystem of operatingSystems) {
+      const runnerImagesForOs = runnerImages.filter((image) =>
+        image.startsWith(operatingSystem),
+      );
```
esbena, line 322: The silent fallback to `ubuntu` and the convention of prefix-matching are brittle. But I suppose it works right now. So :+1:

pr-checks/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
```
esbena, line 4: TypeScript supports extending other tsconfig files. Could we leverage that here to ensure consistent TypeScript behavior in the project?

pr-checks/sync.ts
```diff
@@ -134,112 +283,130 @@ function stripTrailingWhitespace(content: string): string {
...
+} {
+  let inputs: WorkflowInputs = {};
+  const steps: any[] = [];
+
+  for (const language of Object.values(KnownLanguage).sort()) {
+    const setupSpec = languageSetups[language];
+
+    if (
+      setupSpec === undefined ||
+      checkSpecification[setupSpec.specProperty] !== true
```
esbena, line 367: How about:
```suggestion
      !checkSpecification[setupSpec.specProperty]
```

pr-checks/sync.ts
```diff
@@ -90,6 +109,136 @@ const defaultTestVersions = [
...
+        type: "string",
+        description: "The version of Python to install",
+        required: false,
+        default: defaultLanguageVersions.python,
+      },
+    },
+    steps: [
+      {
+        name: "Install Python",
+        if: "matrix.version != 'nightly-latest' || !matrix.version",
```
esbena: I think this condition needs an explanation. Both sides of it

pr-checks/sync.ts
```diff
@@ -250,164 +417,186 @@ function main(): void {
...
-    }
+    console.log(`Processing: ${checkName} — "${checkSpecification.name}"`);
+
+    const { checkJob, workflowInputs } = generateJob(
+      specDocument,
+      checkSpecification,
+    );
+    const { validationJobs, workflowInputs: validationJobInputs } =
+      generateValidationJobs(specDocument, checkSpecification, checkName);
+    const combinedInputs = { ...workflowInputs, ...validationJobInputs };
```
esbena: We should have a check here that asserts the input names do not overlap.

pr-checks/sync.ts
mbg: Not especially. The existing `sync_back.ts` finds all `uses: ...` patterns and updates them. The changed one would (depending on details and merge order) need to be updated to use the right version ref here, but it would be straightforward to update.

pr-checks/sync.ts
mbg: Yeah this isn't new in this PR, just moved code.

pr-checks/sync.ts
mbg: We are OK with input names overlapping here, since the assumption is that the input names are well known + have static corresponding specifications. So if an input with the same name appears in both, it should also have the same specification.

pr-checks/sync.ts
mbg: Personal preference, but I generally prefer the explicit check due to JS weirdness about truthiness.

pr-checks/sync.ts
mbg: Like https://github.com/github/codeql-action/pull/3541#discussion_r2916617739, this is just moved code originally inherited from the Python version. I'll change it since it's a quick and easy change, even though it is guaranteed to be `latest` by the `if` condition.

pr-checks/sync.ts
mbg: I have eliminated this condition entirely in https://github.com/github/codeql-action/pull/3541/commits/2e1f08fe70ac5f965f027f606607759df238ca57, by making the `setup-python` step explicit in the (now that https://github.com/github/codeql-action/pull/3549 is merged) only remaining workflow that used `installPython`. Therefore, the `installPython` property now has the same behaviour as other `install*` properties if specified.

pr-checks/sync.ts
mbg: I have refactored the code in https://github.com/github/codeql-action/pull/3541/commits/be7fe2bca68fc0d426462a6c2833e01166d8df3e so that we now compute a set of known input names and then resolve those to their specifications once we have gathered them all. This makes it more explicit that we are not potentially overriding different input specifications.

pr-checks/sync.ts
mbg: So, I'd say that we keep this as-is here and I will deal with any fallout in the other PR.

pr-checks/sync.ts
mbg: I am inclined to leave this as-is here since it's not related to this PR as such.

mbg:
For https://github.com/github/codeql-action/pull/3541/commits/be7fe2bca68fc0d426462a6c2833e01166d8df3e, I bumped the `lib` version in the `tsconfig.json` for `pr-checks` (since apparently the `union` method requires the latest ES version). That should be OK since the scripts in `pr-checks` are only used in development or CI environments.

This required me to [make a subsequent change to the `pr-checks.yml` workflow](https://github.com/github/codeql-action/pull/3541/commits/d1a7580bd31a2da8ab72105a65ba08108e7b29e9) as well, since the `pr-checks` script checks were run as part of the (matrixed) `unit-tests` job there, which runs with both Node 20 and Node 24. That makes sense for the unit tests etc. for the main codebase, but less so for the `pr-checks` scripts where it now [caused a failure](https://github.com/github/codeql-action/actions/runs/22952199044/job/66619454272). I am guessing that this was just an oversight when we added the `node-versions` matrix to run the tests with the node versions we use for both `v3` and `v4`.

pr-checks/sync.ts
esbena: 👍🏻

PR #3542, opened 2026-03-04
Run some unit tests in parallel

henrymercer:
Drop the `--serial` flag we pass to ava in favour of specifying which tests need to be run sequentially.  These are typically ones which read and write environment variables.  This speeds up the unit tests by 4x on my machine from ~24s to ~6s (measured fairly unscientifically by comparing 5 runs of each).


henrymercer:
> I trust that you didn't change any of the test code itself

👍

PR #3529, opened 2026-03-01
Convert `sync_back.py` to TypeScript

mbg:
Follow-up (and based on) https://github.com/github/codeql-action/pull/3526.

This PR converts `sync_back.py` and `test_sync_back.py` to TypeScript equivalents.

### Risk assessment

For internal use only. Please select the risk level of this change:

- **Low risk:** Changes are fully under feature flags, or have been fully tested and validated in pre-production environments and are highly observable, or are documentation or test only.

#### Which use cases does this change impact?

Environments:

- **Testing/None** - This change does not impact any CodeQL workflows in production.

#### How did/will you validate this change?

- **Other** - Please provide details.

#### If something goes wrong after this change is released, what are the mitigation and rollback strategies?

- **Development/testing only** - This change cannot cause any failures in production.

#### How will you know if something goes wrong after this change is released?

- **Other** - Please provide details.

#### Are there any special considerations for merging or releasing this change?

- **No special considerations** - This change can be merged at any time.

### Merge / deployment checklist

- Confirm this change is backwards compatible with existing workflows.
- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.
- Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary.

PR #3547, opened 2026-03-05
Document regular expressions in `sync_back.ts`

mbg:
Follow up to https://github.com/github/codeql-action/pull/3529 to address @esbena's review comment.

### Risk assessment

Environments:

### DIFF
--- a/pr-checks/sync_back.ts
+++ b/pr-checks/sync_back.ts
@@ -28,12 +28,78 @@ const WORKFLOW_DIR = path.join(THIS_DIR, "..", ".github", "workflows");
 const SYNC_TS_PATH = path.join(THIS_DIR, "sync.ts");
 
 /**
+ * Used to find action references (including versions and comments) in a workflow file.
+ *
+ * This pattern captures `action_name` and `version_with_possible_comment` from
+ * `uses: action_name@version_with_possible_comment`. For example, if we have
+ *
+ * ```
+ * uses: ruby/setup-ruby@09a7688d3b55cf0e976497ff046b70949eeaccfd # v1.288.0
+ * ```
+ *
+ * in a workflow file, this regular expression gets us:
+ *
+ * - `ruby/setup-ruby`; and
+ * - `09a7688d3b55cf0e976497ff046b70949eeaccfd # v1.288.0`.
+ */
+const EXTRACT_ACTION_REF_PATTERN: RegExp =
+  /uses:\s+([^/\s]+\/[^@\s]+)@([^@\n]+)/g;
+
+/**
+ * Used to identify characters in `action_name` strings that need to
+ * be escaped before inserting them into TypeScript or YAML strings.
+ */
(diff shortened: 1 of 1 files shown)


pr-checks/sync_back.ts
```diff
@@ -27,13 +27,79 @@ const CHECKS_DIR = path.join(THIS_DIR, "checks");
...
+ * uses: ruby/setup-ruby@09a7688d3b55cf0e976497ff046b70949eeaccfd # v1.288.0
+ * ```
+ *
+ * in a workflow file, this regular expression gets us:
+ *
+ * - `ruby/setup-ruby`; and
+ * - `09a7688d3b55cf0e976497ff046b70949eeaccfd # v1.288.0`.
+ */
+const EXTRACT_ACTION_REF_PATTERN: RegExp =
+  /uses:\s+([^/\s]+\/[^@\s]+)@([^@\n]+)/g;
```
esbena, line 46: Should we maybe anchor it such that we expect it to be at the start of the line? 
```suggestion
  /^\s*uses:\s+([^/\s]+\/[^@\s]+)@([^@\n]+)/g;
```

esbena: Also, it's probably a minor thing, but in terms of YAML semantics, the user could have added quotes around the right-hand side. Meaning that the captured content will contain surrounding quotes.

I'm thinking we could improve robustness against this if we use the current regular expression to find interesting lines, and then reparse them as standalone YAML. We can then take the right-hand side value instead of the right-hand side syntactic definition.

pr-checks/sync_back.ts
```diff
@@ -27,13 +27,79 @@ const CHECKS_DIR = path.join(THIS_DIR, "checks");
...
+  new RegExp(`(uses:\\s*")${actionName}@(?:[^"]+)(")`, "g");
+
+/**
+ * Used to find lines containing action references in a PR check specification.
+ *
+ * Matches `uses: actionName@rest_of_line` in PR check specifications and extracts `uses: actionName`,
+ * allowing `rest_of_line` to be replaced with a new version string.
+ */
+const YAML_PATTERN: SyncBackPattern = (actionName: string) =>
+  new RegExp(`(uses:\\s+${actionName})@(?:[^@\n]+)`, "g");
```
esbena, line 77: As above, we can probably expect this to be the start of the trimmed line. 

```suggestion
  new RegExp(`^\s*(uses:\\s+${actionName})@(?:[^@\n]+)`, "g");
```

mbg:
> Thank you. This matches my earlier request, and it probably would have gone through if it had been done during that pull request. But since I reviewed it in isolation now, I suppose I'm a bit more strict in terms of how we can do this parting.

I ended up questioning why we are using regular expressions here at all and ended up preparing an alternative to this at https://github.com/github/codeql-action/pull/3556 which avoids using regular expressions here entirely.

mbg:
I will close this in favour of https://github.com/github/codeql-action/pull/3556

PR #3561, opened 2026-03-09
Linting: Require unused function parameters to start with `_`

henrymercer:
The default setting for this lint rule doesn't check function parameters, but in our codebase we are often cleaning up code paths for older versions of GHES, therefore it's desirable to check whether something is being passed that we could possibly avoid computing in the first place.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -161,6 +161,7 @@ export default [
       "@typescript-eslint/no-unused-vars": [
         "error",
         {
+          "args": "all",
           "argsIgnorePattern": "^_",
         }
       ],


eslint.config.mjs
```diff
@@ -161,6 +161,7 @@ export default [
       "@typescript-eslint/no-unused-vars": [
         "error",
         {
+          "args": "all",
```
mbg, line 164: Do you think that this is needed as of https://github.com/github/codeql-action/pull/3497 since it seemed to be working as expected before that?

henrymercer: I think that's likely, though I haven't checked.

PR #3562, opened 2026-03-09
Prepare for rolling out skipping computing file coverage information on PRs

henrymercer:
Prepare for rolling out skipping computing file coverage information on PRs to speed up analysis.

- Gate the feature behind a CLI feature that updates the logs to account for file coverage information no longer being expected.
- Add a repository property that can be used to continue computing file coverage information on PRs.
- Add a telemetry diagnostic so we can see whether this is being used.
- Remove the restriction to the `github` org.

Note that the CLI tools feature flag is not yet released, even as a nightly.  Though given it that CLI feature only modifies logs, I don't think we need to wait for a release before merging this PR.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/config-utils.ts
+++ b/src/config-utils.ts
@@ -1556,6 +1556,13 @@ export function isCodeQualityEnabled(config: Config): boolean {
 }
 
 /**
+ * Returns `true` if Code Scanning Risk Assessment analysis is enabled, or `false` if not.
+ */
+export function isRiskAssessmentEnabled(config: Config): boolean {
+  return config.analysisKinds.includes(AnalysisKind.RiskAssessment);
+}
+
+/**
  * Returns the primary analysis kind that the Action is initialised with. If there is only
  * one analysis kind, then that is returned.
  *
--- a/src/init-action-post-helper.test.ts
+++ b/src/init-action-post-helper.test.ts
@@ -1,10 +1,13 @@
+import * as core from "@actions/core";
 import test, { ExecutionContext } from "ava";
 import * as sinon from "sinon";
 
 import * as actionsUtil from "./actions-util";
 import { AnalysisKind } from "./analyses";
+import * as apiClient from "./api-client";
 import * as codeql from "./codeql";
 import * as configUtils from "./config-utils";
+import * as debugArtifacts from "./debug-artifacts";
 import { EnvVar } from "./environment";
 import { Feature } from "./feature-flags";
 import * as initActionPostHelper from "./init-action-post-helper";
@@ -17,6 +20,7 @@ import {
   createTestConfig,
   DEFAULT_ACTIONS_VARS,
   makeVersionInfo,
+  RecordingLogger,
   setupActionsVars,
   setupTests,
 } from "./testing-utils";
@@ -46,7 +50,7 @@ test.serial("init-post action with debug mode off", async (t) => {
     const uploadAllAvailableDebugArtifactsSpy = sinon.spy();
     const printDebugLogsSpy = sinon.spy();
 
-    await initActionPostHelper.run(
+    await initActionPostHelper.uploadFailureInfo(
       uploadAllAvailableDebugArtifactsSpy,
       printDebugLogsSpy,
       codeql.createStubCodeQL({}),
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 ## [UNRELEASED]
 
 - Fixed [a bug](https://github.com/github/codeql-action/issues/3555) which caused the CodeQL Action to fail loading repository properties if a "Multi select" repository property was configured for the repository. [#3557](https://github.com/github/codeql-action/pull/3557)
+- Added an experimental change which skips collecting file coverage information on pull requests to improve analysis performance. File coverage information will still be computed on non-PR analyses.
+
+  Repositories owned by an organization can opt out of this change by creating a custom repository property with the name `github-codeql-file-coverage-on-prs` and the type "True/false", then setting this property to `true` in the repository's settings. For more information, see [Managing custom properties for repositories in your organization](https://docs.github.com/en/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization).
```
mbg: We should add documentation on docs.github.com about these repo properties. I have created an internal issue to track this.

src/init-action.ts
```diff
@@ -378,8 +380,9 @@ async function run(startedAt: Date) {
       repositoryProperties: repositoryPropertiesResult.orElse({}),
       enableFileCoverageInformation: await getFileCoverageInformationEnabled(
         debugMode,
-        repositoryNwo,
+        codeql,
         features,
+        repositoryPropertiesResult.orElse({}),
```
mbg: Minor: we have `repositoryPropertiesResult.orElse({})` at least a couple of times here now. Consider adding `const repositoryProperties = repositoryPropertiesResult.orElse({});` somewhere before the uses, or indeed modifying the code so that we never end up with the "result" here, but only the properties we loaded or `{}`.

src/init-action.ts
```diff
@@ -397,6 +400,24 @@ async function run(startedAt: Date) {
       );
     }
 
+    if (
+      config.enableFileCoverageInformation &&
+      isAnalyzingPullRequest() &&
+      (await features.getValue(Feature.SkipFileCoverageOnPrs, codeql)) &&
+      repositoryPropertiesResult.orElse({})[
+        RepositoryPropertyName.FILE_COVERAGE_ON_PRS
+      ] === true
```
mbg: This now duplicates most of `getFileCoverageInformationEnabled`. While we can't use `getFileCoverageInformationEnabled` as a shortcut here, consider whether you can move the code that emits this diagnostic into `getFileCoverageInformationEnabled` and end up with a cleaner solution in there?

src/init.ts
```diff
@@ -329,6 +333,18 @@ export async function getFileCoverageInformationEnabled(
       showDeprecationWarning: false,
     };
   }
+  // If the user has explicitly opted out via an environment variable, don't
+  // show the deprecation warning.
+  if (
+    (process.env[EnvVar.FILE_COVERAGE_ON_PRS] || "").toLocaleLowerCase() ===
```
mbg, line 339: Minor: I thought we weren't going to accept case-insensitive values for consistency 😅 (No opposition from me here, though.)

src/init.ts
```diff
@@ -358,3 +374,45 @@ export async function getFileCoverageInformationEnabled(
     showDeprecationWarning: false,
   };
 }
+
+/**
+ * Log a warning about the deprecation of file coverage information on PRs, including how to opt
+ * back in via an environment variable or repository property.
+ */
+export function logFileCoverageOnPrsDeprecationWarning(logger: Logger): void {
```
mbg, line 382: This message is getting fairly convoluted now, and is duplicated here and in the changelog. I already suggested we update the public docs, but as a stopgap measure until we have done that, perhaps these instructions could live in a markdown file somewhere in the repo and we link to that from the log message and changelog?

src/init.ts
henrymercer: I'm using the same code as `getOfflineValue` here (and would be open to factoring it out into a function).  I think the argument for repository properties still applies, but don't feel super strongly about it.

src/init.ts
henrymercer: I've added tests so it's clear what the message looks like in each case.  I think it is useful to provide specific instructions for each situation, so there is less information to parse.  How strongly do you feel about duplicating the messaging between the Actions logs / workflow run annotations and the changelog?

mbg: I am mainly concerned about ensuring that the messages (and tests) stay in sync if they are essentially duplicated in three places. Still happy to approve this PR though.

PR #3564, opened 2026-03-10
Fix retries when uploading databases

henrymercer:
The current retry mechanism would pass a read stream that might have already been partially consumed.  Instead, use a custom retry mechanism and pass a fresh read stream on each attempt.

An alternative that wouldn't need the custom retry mechanism would have been to read the database into a buffer, but databases might be multiple GBs so I wanted to keep a streaming approach.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/database-upload.ts
+++ b/src/database-upload.ts
@@ -85,18 +85,6 @@ export async function cleanupAndUploadDatabases(
     await codeql.databaseCleanupCluster(config, cleanupLevel);
   });
 
-  const client = getApiClient();
-
-  const uploadsUrl = new URL(parseGitHubUrl(apiDetails.url));
-  uploadsUrl.hostname = `uploads.${uploadsUrl.hostname}`;
-
-  // Octokit expects the baseUrl to not have a trailing slash,
-  // but it is included by default in a URL.
-  let uploadsBaseUrl = uploadsUrl.toString();
-  if (uploadsBaseUrl.endsWith("/")) {
-    uploadsBaseUrl = uploadsBaseUrl.slice(0, -1);
-  }
-
   const reports: DatabaseUploadResult[] = [];
   for (const language of config.languages) {
     let bundledDbSize: number | undefined = undefined;
@@ -118,30 +106,15 @@ export async function cleanupAndUploadDatabases(
       const maxAttempts = 4; // 1 initial attempt + 3 retries, identical to the default retry behavior of Octokit
       let uploadDurationMs: number | undefined;
       for (let attempt = 1; attempt <= maxAttempts; attempt++) {
-        const bundledDbReadStream = fs.createReadStream(bundledDb);
         try {
-          const attemptStartTime = performance.now();
-          await client.request(
-            `POST /repos/:owner/:repo/code-scanning/codeql/databases/:language?name=:name&commit_oid=:commit_oid`,
-            {
-              baseUrl: uploadsBaseUrl,
-              owner: repositoryNwo.owner,
-              repo: repositoryNwo.repo,
-              language,
-              name: `${language}-database`,
-              commit_oid: commitOid,
-              data: bundledDbReadStream,
-              headers: {
-                authorization: `token ${apiDetails.auth}`,
-                "Content-Type": "application/zip",
-                "Content-Length": bundledDbSize,
-              },
-              request: {
-                retries: 0,
-              },
-            },
+          uploadDurationMs = await uploadBundledDatabase(
+            repositoryNwo,
+            language,
+            commitOid,
+            bundledDb,
+            bundledDbSize,
+            apiDetails,
           );
-          uploadDurationMs = performance.now() - attemptStartTime;
           break;
         } catch (e) {
           const httpError = asHTTPError(e);
@@ -160,8 +133,6 @@ export async function cleanupAndUploadDatabases(
             `Database upload attempt ${attempt} of ${maxAttempts} failed for ${language}: ${util.getErrorMessage(e)}. Retrying in ${backoffMs / 1000}s...`,
           );
           await new Promise((resolve) => setTimeout(resolve, backoffMs));
-        } finally {
-          bundledDbReadStream.close();
         }
       }
       reports.push({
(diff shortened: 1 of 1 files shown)


src/database-upload.ts
```diff
@@ -105,40 +109,63 @@ export async function cleanupAndUploadDatabases(
...
+      // Upload with manual retry logic. We disable Octokit's built-in retries
+      // because the request body is a ReadStream, which can only be consumed
+      // once.
+      const maxAttempts = 4; // 1 initial attempt + 3 retries, identical to the default retry behavior of Octokit
+      let uploadDurationMs: number | undefined;
+      for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+        const bundledDbReadStream = fs.createReadStream(bundledDb);
+        try {
+          const attemptStartTime = performance.now();
+          await client.request(
```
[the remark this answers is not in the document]
henrymercer: The retry mechanism never worked, and I think it's more interesting to know the duration of a successful attempt.

src/database-upload.test.ts
```diff
@@ -214,37 +214,81 @@ test.serial(
...
+      sinon.stub(gitUtils, "isAnalyzingDefaultBranch").resolves(true);
+
+      await mockHttpRequests(500);
+
+      // Stub setTimeout to fire immediately to avoid real delays from retry backoff.
+      const originalSetTimeout = global.setTimeout;
+      sinon
+        .stub(global, "setTimeout")
+        .callsFake((fn: () => void) => originalSetTimeout(fn, 0));
+
```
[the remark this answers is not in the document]
henrymercer: I don't think this is worth addressing.

src/database-upload.ts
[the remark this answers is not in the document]
mbg: I suppose all of the following is interesting:

- Duration of the successful upload
- Overall duration trying to upload
- Number of upload attempts

I'll leave it to you though whether you think that's worth updating the telemetry for, and whether to do it in this PR or not.

As it stands, I agree that only reporting the duration of the successful upload most closely resembles the previous use of `upload_duration_ms` and we can keep it like that here.

src/database-upload.ts
```diff
@@ -105,40 +109,63 @@ export async function cleanupAndUploadDatabases(
...
-        bundledDbReadStream.close();
+          );
+          uploadDurationMs = performance.now() - attemptStartTime;
+          break;
+        } catch (e) {
+          const httpError = asHTTPError(e);
+          const isRetryable =
+            !httpError || !DO_NOT_RETRY_STATUSES.includes(httpError.status);
+          if (!isRetryable || attempt === maxAttempts) {
+            throw e;
```
esbena: Minor debugging improvement.
```suggestion
          if (!isRetryable) {
            throw e;
          } else if (attempt === maxAttempts) {
            logger.error(`Maximum retry attempts exhausted (${attempt}), re-throwing otherwise retryable error...`)
            throw e;
```

src/database-upload.ts
```diff
@@ -105,40 +109,63 @@ export async function cleanupAndUploadDatabases(
...
-              authorization: `token ${apiDetails.auth}`,
-              "Content-Type": "application/zip",
-              "Content-Length": bundledDbSize,
+      // Upload with manual retry logic. We disable Octokit's built-in retries
+      // because the request body is a ReadStream, which can only be consumed
+      // once.
+      const maxAttempts = 4; // 1 initial attempt + 3 retries, identical to the default retry behavior of Octokit
+      let uploadDurationMs: number | undefined;
+      for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+        const bundledDbReadStream = fs.createReadStream(bundledDb);
```
mbg: I see that you marked this PR as high risk. It doesn't seem like it would be too hard to refactor this slightly so that the old code / loop body goes into a new function and then depending on a new FF we either just call that directly or in the loop?

src/database-upload.ts
```diff
@@ -105,40 +109,63 @@ export async function cleanupAndUploadDatabases(
...
+          await new Promise((resolve) => setTimeout(resolve, backoffMs));
+        } finally {
+          bundledDbReadStream.close();
+        }
       }
+      reports.push({
+        language,
+        zipped_upload_size_bytes: bundledDbSize,
+        is_overlay_base: shouldUploadOverlayBase,
+        upload_duration_ms: uploadDurationMs,
```
esbena, line 142: I wonder: Do we separately and in general track how many failures we have for API requests that succeed on retry?

src/database-upload.ts
```diff
@@ -105,40 +109,63 @@ export async function cleanupAndUploadDatabases(
...
-          zipped_upload_size_bytes: bundledDbSize,
-          is_overlay_base: shouldUploadOverlayBase,
-          upload_duration_ms: endTime - startTime,
-        });
-        logger.debug(`Successfully uploaded database for ${language}`);
-      } finally {
-        bundledDbReadStream.close();
+          );
+          uploadDurationMs = performance.now() - attemptStartTime;
+          break;
```
mbg, line 118: Minor: maybe a mild preference for using some boolean to indicate when the upload was successful in the loop condition over using `break`, since that can be brittle if we refactor this in the future. That might not matter / be necessary if you move the upload code into a separate function as suggested in my other comment, though.

src/database-upload.test.ts
```diff
@@ -214,37 +214,81 @@ test.serial(
...
+      );
+
+      t.assert(
+        loggedMessages.find(
+          (v) =>
+            v.type === "warning" &&
+            v.message ===
+              "Failed to upload database for javascript: some error message",
+        ) !== undefined,
+      );
```
mbg: Minor: this is a bit manual. Could we modify `checkExpectedLogMessages` to allow us to match the expected type as well? Also see https://github.com/github/codeql-action/pull/3508 for the direction I may want to take this in.

src/database-upload.test.ts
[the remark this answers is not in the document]
mbg: It doesn't seem too difficult to add some extra assertion(s) for this to at least roughly check whether the retry logic kicks in at all or not? Doesn't have to be too polished if we expect to replace our own retry logic here if/when e.g. https://github.com/octokit/plugin-retry.js/pull/663 is merged.

src/database-upload.ts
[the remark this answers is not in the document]
henrymercer: I agree it's interesting, but I don't think it's interesting enough to add the telemetry.  If we decide otherwise though, I'd prefer that we do this in a separate PR.

src/database-upload.test.ts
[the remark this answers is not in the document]
henrymercer: I don't think that PR would help this situation, since we need to ensure we don't reuse the same stream.

henrymercer:
> This broadly looks good and makes sense. I have a few suggestions, particularly around the "high risk" assessment. I'd suggest we add a FF just in case, which shouldn't be too hard? Otherwise, we probably do want to make sure that we have reasonable test coverage here.

I disagree — we hit this error very rarely, and the problem is remedied by a retry.  I don't think we'd get much signal with a feature flag.  The changes require careful review, but I think a feature flag is overkill.  I'd suggest instead we check the telemetry in a month's time and make sure we're no longer seeing "Response body object should not be disturbed or locked".

src/database-upload.ts
henrymercer: I replied to this as part of your overall review comment

src/database-upload.ts
henrymercer: We don't, but that's a good idea, although for another PR :)

henrymercer:
> I see that you reacted with 👍🏻 to [this comment](https://github.com/github/codeql-action/pull/3564#discussion_r2911653537), but I don't see it addressed yet -- did you mean to?

Thanks, pushed.

PR #3565, opened 2026-03-11
PR checks: Only run Go macOS tests on latest CodeQL versions

henrymercer:
Drop Go macOS tests on older CodeQL versions due to usage in production being predominantly Linux and using up to date CodeQL versions.

### DIFF
--- a/.github/workflows/__bundle-toolcache.yml
+++ b/.github/workflows/__bundle-toolcache.yml
@@ -39,9 +39,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-latest
-            version: linked
           - os: ubuntu-latest
+            version: linked
+          - os: macos-latest
             version: linked
           - os: windows-latest
             version: linked
--- a/.github/workflows/__bundle-zstd.yml
+++ b/.github/workflows/__bundle-zstd.yml
@@ -39,9 +39,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-latest
-            version: linked
           - os: ubuntu-latest
+            version: linked
+          - os: macos-latest
             version: linked
           - os: windows-latest
             version: linked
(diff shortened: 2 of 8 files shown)


pr-checks/sync.ts
```diff
@@ -317,6 +319,13 @@ function generateJobMatrix(
     const operatingSystems = checkSpecification.operatingSystems ?? ["ubuntu"];
 
     for (const operatingSystem of operatingSystems) {
+      // If osCodeQlVersions is set for this OS, only include the specified CodeQL versions.
+      const allowedVersions =
+        checkSpecification.osCodeQlVersions?.[operatingSystem];
+      if (allowedVersions && !allowedVersions.includes(version)) {
+        continue;
+      }
```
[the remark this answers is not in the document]
henrymercer: Might be worth adding a JSON schema for the PR checks to validate this, rather than needing to validate each property individually.  I think this is best left to a future PR.

pr-checks/checks/packaging-config-inputs-js.yml
```diff
@@ -1,6 +1,10 @@
 name: "Packaging: Config and input"
 description: "Checks that specifying packages using a combination of a config file and input to the Action works"
-versions: ["linked", "default", "nightly-latest"] # This feature is not compatible with old CLIs
+# This feature is not compatible with old CLIs
+versions:
+  - linked
+  - default
+  - nightly-latest
```
esbena, line 7: Stray thought: it could be useful to be able to flip this around and specify what we *don't* want to run on.

esbena (approved):
LGTM with one minor thought on maintainability.

pr-checks/checks/packaging-config-inputs-js.yml
henrymercer: Yeah, our approach could use a rethink.  We support a bunch of Action / CLI combinations, but is impractical to run PR checks against every combination so in practice most checks just run against linked, default, and nightly-latest.  Providing CI is sufficiently unflaky though, it could be useful to start running nightly checks against the full set of combinations.

PR #3791, opened 2026-03-30
Fall back to non-overlay analysis when diff-informed analysis is unavailable

sam-robson:
When overlay analysis is enabled for a PR but diff-informed analysis fails (e.g. deleted branch, API error, too many changed files), the action previously continued with overlay-only analysis. This is an untested combination that can produce inaccurate results. Now we fall back to a full non-overlay analysis instead.

### Risk assessment

Workflow types:

Products:

Environments:

- **Other** - Manual code review. The change is a single conditional guard following established patterns in the same file. No unit tests due to the init-action entry point side-effect constraint, but the logic is minimal (3 conditions + mode revert).


redsun82 (dismissed):

src/diff-informed-analysis-utils.ts
```diff
@@ -88,6 +88,10 @@ export function writeDiffRangesJsonFile(
   );
 }
 
+export function hasDiffRangesJsonFile(): boolean {
```
mbg: Minor: Adding a JSDoc comment here would be nice.

src/init-action.ts
```diff
@@ -438,6 +440,30 @@ async function run(startedAt: Date) {
...
+      config.overlayDatabaseMode = OverlayDatabaseMode.None;
+    }
+  } catch (e) {
+    logger.warning(
+      `Failed to determine diff-informed analysis availability, ` +
+        `reverting overlay database mode to ${OverlayDatabaseMode.None}: ${getErrorMessage(e)}`,
+    );
+    config.overlayDatabaseMode = OverlayDatabaseMode.None;
+  }
+
```
mbg: It would be good to move this out of `init-action.ts` and into one of the overlay-specific modules if possible.

There's already a bunch of logic around overlay enablement there, so spreading that out across different locations isn't good.

We also have some frameworks for tests/diagnostics/telemetry there that this could factor into. In any case, this should have test coverage.

mbg (changes requested):

mbg (changes requested):
Thanks for refactoring some of the logic out of `init-action` and into `revertOverlayModeIfDiffInformedUnavailable`, as well as adding tests!

I am still not convinced by the high-level approach here, though. If I see the flow correctly, the key events are:

- In `checkOverlayEnablement` (in `config-utils.ts`), we perform most checks to determine whether overlay analysis should be enabled.
- Afterwards, in `initConfig` (also in `config-utils.ts`), we then call `shouldPerformDiffInformedAnalysis` if overlay analysis is enabled. 
  - `shouldPerformDiffInformedAnalysis` is primarily a check to see if we are analysing a pull request.
  - If diff-informed analysis should be enabled, then we add a query exclusion for `exclude-from-incremental`.
- Back in `init-action.ts`, we call `computeAndPersistDiffRanges` after `initConfig` has returned. This is what produces the file pointed at by `getDiffRangesJsonFilePath`.
- Your change then happens afterwards (outside the normal `try`/`catch` block), since you depend on checking whether the `getDiffRangesJsonFilePath` file exists or not.
  - You also again call `shouldPerformDiffInformedAnalysis` to determine what should have happened. We could probably persist this information in the CodeQL Action state instead of recomputing it.

This seems a bit (unnecessarily) convoluted to me. I am wondering if:

1. We can call `computeAndPersistDiffRanges` as part of the diff-informed analysis check in `initConfig` in `config-utils.ts`? Then the logic isn't spread across two locations and we can potentially also re-use the result of `getDiffInformedAnalysisBranches`. This would require us to move `computeAndPersistDiffRanges` from `init-action.ts` to somewhere more suitable. I don't see a reason for it to have to be in `init-action.ts`, but we should check that calling it from `initConfig` doesn't cause any problems.
2. Then, we could perform the new check that is introduced here immediately afterwards in `initConfig`.
3. We could also investigate whether both the initial diff-informed query enablement check and the new logic here could then be added to `checkOverlayEnablement`

What do you think?

sam-robson:
> 1. We can call `computeAndPersistDiffRanges` as part of the diff-informed analysis check in `initConfig` in `config-utils.ts`? Then the logic isn't spread across two locations and we can potentially also re-use the result of `getDiffInformedAnalysisBranches`. This would require us to move `computeAndPersistDiffRanges` from `init-action.ts` to somewhere more suitable. I don't see a reason for it to have to be in `init-action.ts`, but we should check that calling it from `initConfig` doesn't cause any problems.
> 2. Then, we could perform the new check that is introduced here immediately afterwards in `initConfig`.
> 3. We could also investigate whether both the initial diff-informed query enablement check and the new logic here could then be added to `checkOverlayEnablement`
> 
> What do you think?

Thanks! Pushed a refactor addressing these.

- computeAndPersistDiffRanges moved to diff-informed-analysis-utils.ts. New prepareDiffInformedAnalysis combines the enablement check with computing/persisting ranges, reusing the PullRequestBranches result.
- initConfig calls it once and passes the result to a new applyIncrementalAnalysisSettings helper, the single place deciding both the overlay-mode revert and the exclude-from-incremental exclusion.

src/diff-informed-analysis-utils.ts
```diff
@@ -69,6 +55,61 @@ export async function getDiffInformedAnalysisBranches(
...
+          return false;
+        }
+      },
+    );
+    return { shouldRun: true, isAvailable };
+  } catch (e) {
+    logger.warning(
+      `Failed to determine diff-informed analysis availability: ${getErrorMessage(e)}`,
+    );
+    return { shouldRun: true, isAvailable: false };
```
mbg: This will return `shouldRun: true` if `getDiffInformedAnalysisBranches` throws. I think it should be `false` or, possibly better, `undefined` in that case?

src/diff-informed-analysis-utils.ts
```diff
@@ -69,6 +55,61 @@ export async function getDiffInformedAnalysisBranches(
...
 
+export interface DiffInformedAnalysisPreparation {
+  /**
+   * Whether diff-informed analysis applies to this workflow run.
+   */
+  shouldRun: boolean;
+  /**
+   * Whether the diff ranges were successfully prepared and can be used.
+   */
+  isAvailable: boolean;
```
mbg: Minor: Talking about "available" throughout makes me think that this is dependent on some FF etc. Perhaps it would be clearer to rename this to e.g. `isComputed` or `hasDiffRanges` and then update the wording in `prepareDiffInformedAnalysis` accordingly?

src/config-utils.ts
```diff
@@ -1076,6 +1079,42 @@ function hasQueryCustomisation(userConfig: UserConfig): boolean {
...
+ * actually available.
+ */
+export function applyIncrementalAnalysisSettings(
+  config: Config,
+  diffInformedAnalysis: DiffInformedAnalysisPreparation,
+  logger: Logger,
+): void {
+  if (
+    config.overlayDatabaseMode === OverlayDatabaseMode.Overlay &&
+    diffInformedAnalysis.shouldRun &&
```
mbg: This is the key change in this PR. Conceptually, one would assume that `config.overlayDatabaseMode === OverlayDatabaseMode.Overlay` implies `diffInformedAnalysis.shouldRun`. The PR description explains why that may not be the case, but it would be good to add that as a comment here as well so that it's clear why this is not in fact an implication.

src/config-utils.ts
```diff
@@ -1076,6 +1079,42 @@ function hasQueryCustomisation(userConfig: UserConfig): boolean {
...
+    !diffInformedAnalysis.isAvailable
+  ) {
+    logger.warning(
+      "Diff-informed analysis is not available for this pull request. " +
+        `Reverting overlay database mode to ${OverlayDatabaseMode.None}.`,
+    );
```
mbg: In `checkOverlayEnablement`, these sorts of log messages are `info`-level and I think this one should be as well. That's because `warning`-level messages result in workflow annotations, and it is not likely that a user would be able to do anything about this.

src/config-utils.test.ts
```diff
@@ -2200,3 +2200,80 @@ test.serial(
     });
   },
 );
+
+test.serial(
```
mbg: These `applyIncrementalAnalysisSettings` tests could probably run in parallel with others? I don't see any global state being used/mutated here.

src/config-utils.test.ts
```diff
@@ -2200,3 +2200,80 @@ test.serial(
...
+      logger,
+    );
+
+    t.is(config.overlayDatabaseMode, OverlayDatabaseMode.None);
+    t.deepEqual(config.extraQueryExclusions, []);
+  },
+);
+
+test.serial(
+  "applyIncrementalAnalysisSettings: keeps overlay mode and adds exclusions when diff-informed analysis is disabled",
```
mbg: Minor: "when not analysing a PR" or "when diff-informed analysis shouldn't run"?

src/config-utils.test.ts
```diff
@@ -2200,3 +2200,80 @@ test.serial(
...
+
+    t.is(config.overlayDatabaseMode, OverlayDatabaseMode.Overlay);
+    t.deepEqual(config.extraQueryExclusions, [
+      { exclude: { tags: "exclude-from-incremental" } },
+    ]);
+  },
+);
+
+test.serial(
+  "applyIncrementalAnalysisSettings: reverts to None without exclusions when diff-informed analysis is unavailable",
```
mbg: Minor:
```suggestion
  "applyIncrementalAnalysisSettings: disables overlay analysis when diff-informed analysis is unavailable",
```

src/config-utils.test.ts
```diff
@@ -2200,3 +2200,80 @@ test.serial(
...
+  (t) => {
+    const config = createTestConfig({});
+    config.overlayDatabaseMode =
+      OverlayDatabaseMode.Overlay as OverlayDatabaseMode;
```
mbg: Why the `as OverlayDatabaseMode` here?

src/diff-informed-analysis-utils.ts
```diff
@@ -69,6 +55,61 @@ export async function getDiffInformedAnalysisBranches(
...
+}
+
+/**
+ * Prepares the diff ranges needed for diff-informed analysis for the current
+ * run.
+ *
+ * @returns Whether diff-informed analysis applies to this run, and whether it
+ *   was successfully prepared for use.
+ */
+export async function prepareDiffInformedAnalysis(
```
mbg, line 65: We should add unit tests for `prepareDiffInformedAnalysis`.

mbg (commented):
Thanks for making these changes! This looks good to me now in terms of the overall approach. I added a few, mostly minor comments about polishing a few details.

src/diff-informed-analysis-utils.ts
sam-robson: Makes sense 👍🏻

src/config-utils.ts
sam-robson: Added a comment

src/diff-informed-analysis-utils.ts
sam-robson: Done

src/config-utils.test.ts
sam-robson: To stop TS from narrowing `overlayDatabaseMode` to the `Overlay` literal

src/config-utils.ts
```diff
@@ -1085,8 +1085,11 @@ function hasQueryCustomisation(userConfig: UserConfig): boolean {
  * If overlay mode was selected for a PR but diff-informed analysis should have
  * run and could not be prepared, fall back to a full non-overlay analysis.
  * Query exclusions for incremental-only queries are applied only when the final
- * configuration still uses overlay analysis or diff-informed analysis is
- * actually available.
+ * configuration still uses overlay analysis or the diff ranges are available.
+ *
+ * Note that `overlayDatabaseMode === Overlay` does not imply
+ * `diffInformedAnalysis.shouldRun`: the two are configured independently
+ * so we handle each combination explicitly.
```
mbg: I was hoping for an explanation along the lines of what's in the PR description, i.e. _why_ the former doesn't imply the latter, not just that it doesn't (that much is obvious from the separate checks in the condition below).

src/diff-informed-analysis-utils.ts
```diff
@@ -61,53 +61,54 @@ export interface DiffInformedAnalysisPreparation {
...
-    );
-    return { shouldRun: true, isAvailable };
+    branches = await getDiffInformedAnalysisBranches(codeql, features, logger);
   } catch (e) {
+    // If we cannot determine whether diff-informed analysis applies (for
+    // example, because a feature-flag lookup failed), treat it as not
+    // applicable rather than triggering the overlay fallback.
     logger.warning(
-      `Failed to determine diff-informed analysis availability: ${getErrorMessage(e)}`,
+      `Failed to determine whether diff-informed analysis applies: ${getErrorMessage(e)}`,
```
mbg: Minor:
```suggestion
      `Failed to determine branch information for diff-informed analysis: ${getErrorMessage(e)}`,
```

src/config-utils.test.ts
```diff
@@ -2201,79 +2201,70 @@ test.serial(
...
-      { shouldRun: true, isAvailable: false },
-      logger,
-    );
+test("applyIncrementalAnalysisSettings: disables overlay analysis when diff-informed analysis is unavailable", (t) => {
+  const config = createTestConfig({});
+  // Cast needed so TypeScript doesn't narrow `overlayDatabaseMode` to the
+  // literal `Overlay` type, which would then reject the `None` assertion
+  // below after the function under test mutates the field.
+  config.overlayDatabaseMode =
+    OverlayDatabaseMode.Overlay as OverlayDatabaseMode;
```
mbg: The type cast is not needed if you use the argument to `createTestConfig` to set the option:
```suggestion
  const config = createTestConfig({
    overlayDatabaseMode: OverlayDatabaseMode.Overlay
  });
```

src/diff-informed-analysis-utils.test.ts
```diff
@@ -187,6 +189,148 @@ test.serial(
...
+        codeql,
+        features,
+        logger,
+      );
+
+      t.deepEqual(result, { shouldRun: true, hasDiffRanges: false });
+
+      getPullRequestBranchesStub.restore();
+      getGitHubVersionStub.restore();
+      getApiClientStub.restore();
```
mbg: This is unnecessary, these are automatically restored at the end of tests.

src/diff-informed-analysis-utils.test.ts
```diff
@@ -187,6 +189,148 @@ test.serial(
...
+        codeql,
+        features,
+        logger,
+      );
+
+      t.deepEqual(result, { shouldRun: true, hasDiffRanges: true });
+
+      getPullRequestBranchesStub.restore();
+      getGitHubVersionStub.restore();
+      getApiClientStub.restore();
```
mbg: This is unnecessary, these are automatically restored at the end of tests.

src/diff-informed-analysis-utils.test.ts
```diff
@@ -187,6 +189,148 @@ test.serial(
...
+      const result = await prepareDiffInformedAnalysis(
+        codeql,
+        features,
+        logger,
+      );
+
+      t.deepEqual(result, { shouldRun: false, hasDiffRanges: false });
+
+      getPullRequestBranchesStub.restore();
+      getGitHubVersionStub.restore();
```
mbg: This is unnecessary, these are automatically restored at the end of tests.

src/config-utils.ts
```diff
@@ -1076,6 +1079,48 @@ function hasQueryCustomisation(userConfig: UserConfig): boolean {
...
+  if (
+    config.overlayDatabaseMode === OverlayDatabaseMode.Overlay &&
+    diffInformedAnalysis.shouldRun &&
+    !diffInformedAnalysis.hasDiffRanges
+  ) {
+    logger.info(
+      "Diff-informed analysis is not available for this pull request. " +
+        `Reverting overlay database mode to ${OverlayDatabaseMode.None}.`,
+    );
+    config.overlayDatabaseMode = OverlayDatabaseMode.None;
```
henrymercer, line 1106: It would be great to integrate this into `OverlayDisabledReason` and `addOverlayDisablementDiagnostics`, so we can track all the reasons why overlay is disabled in a single place.

src/config-utils.test.ts
```diff
@@ -2200,3 +2200,68 @@ test.serial(
     });
   },
 );
+
+test("applyIncrementalAnalysisSettings: no-op when mode is not Overlay and diff-informed is unavailable", (t) => {
```
henrymercer: Could you add a happy path test for when the mode is overlay, diff-informed is available and should run?

src/config-utils.ts
```diff
@@ -1076,6 +1079,48 @@ function hasQueryCustomisation(userConfig: UserConfig): boolean {
...
+  logger: Logger,
+): void {
+  if (
+    config.overlayDatabaseMode === OverlayDatabaseMode.Overlay &&
+    diffInformedAnalysis.shouldRun &&
+    !diffInformedAnalysis.hasDiffRanges
+  ) {
+    logger.info(
+      "Diff-informed analysis is not available for this pull request. " +
+        `Reverting overlay database mode to ${OverlayDatabaseMode.None}.`,
```
henrymercer: Minor: To be clearer here, we could say "because the PR diff ranges could not be computed"

henrymercer (commented):
Thanks for all the iteration here!  I have a couple of specific comments, but the most important one is more general: should we be running overlay when diff-informed is disabled (when `shouldRun` is `false`)?  My understanding is overlay has only been validated for diff-informed analyses.  So I think we want to disable overlay if diff-informed can't be run in any circumstance, whether that is because the diff-ranges can't be computed, or because we can't determine the branch information?

sam-robson:
> Thanks for all the iteration here! I have a couple of specific comments, but the most important one is more general: should we be running overlay when diff-informed is disabled (when `shouldRun` is `false`)? My understanding is overlay has only been validated for diff-informed analyses. So I think we want to disable overlay if diff-informed can't be run in any circumstance, whether that is because the diff-ranges can't be computed, or because we can't determine the branch information?

Makes sense 👍🏻

src/config-utils.test.ts
```diff
@@ -2200,3 +2200,79 @@ test.serial(
...
+  await configUtils.applyIncrementalAnalysisSettings(
+    config,
+    false,
+    codeql,
+    logger,
+  );
+
+  t.is(config.overlayDatabaseMode, OverlayDatabaseMode.None);
+  t.is(config.useOverlayDatabaseCaching, false);
+  t.deepEqual(config.extraQueryExclusions, []);
```
henrymercer, line 2205: Minor: Consider also asserting that `addOverlayDisablementDiagnostics` was called with `PrDiffRangesNotComputed`

src/overlay/diagnostics.ts
```diff
@@ -39,6 +39,13 @@ export enum OverlayDisabledReason {
   NotPullRequestOrDefaultBranch = "not-pull-request-or-default-branch",
   /** The top-level overlay analysis feature flag is not enabled. */
   OverallFeatureNotEnabled = "overall-feature-not-enabled",
+  /**
+   * Overlay analysis was selected for a pull request, but the PR diff ranges
+   * needed for diff-informed analysis could not be computed. Overlay analysis
+   * has only been validated in combination with diff-informed analysis, so we
+   * fall back to a non-overlay analysis in this case.
```
henrymercer: Minor: we'll now get this reason whenever diff informed analysis was not enabled.  Consider renaming to `DiffInformedAnalysisNotEnabled` and updating the JSDoc.

henrymercer (commented):
This looks pretty much ready to merge.  Couple of comments, and could we also add a short changelog note about the fact that for performance and accuracy reasons, overlay analysis will now only be enabled when diff-informed analysis is enabled?

sam-robson:
> This looks pretty much ready to merge. Couple of comments, and could we also add a short changelog note about the fact that for performance and accuracy reasons, overlay analysis will now only be enabled when diff-informed analysis is enabled?

Added the changelog note and made the minor adjustments as requested 👍🏻

mbg (dismissed):
Echoing what Henry said: thanks for all the iterations on this! I think the implementation is much better now as a result. I don't see any further issues, so it should be good to merge once the merge conflict is resolved.

henrymercer (approved):
:shipit:

PR #3801, opened 2026-04-02
Mark Swift incompatible OS as configuration error

henrymercer:
Reduce monitor noise by treating the incompatible OS error on Swift as a configuration error rather than an issue with the CodeQL Action.

### DIFF
--- a/src/init-action.ts
+++ b/src/init-action.ts
@@ -391,10 +391,10 @@ async function run(startedAt: Date) {
 
     if (
       config.languages.includes(KnownLanguage.swift) &&
-      process.platform === "linux"
+      process.platform !== "darwin"
     ) {
       throw new ConfigurationError(
-        `Swift analysis on Ubuntu runner images is no longer supported. Please migrate to a macOS runner.`,
+        `Swift analysis is only supported on macOS runner images. Please migrate to a macOS runner.`,
       );
     }
 
--- a/src/init-action.ts
+++ b/src/init-action.ts
@@ -389,6 +389,15 @@ async function run(startedAt: Date) {
       logger,
     });
 
+    if (
+      config.languages.includes(KnownLanguage.swift) &&
+      process.platform === "linux"
+    ) {
+      throw new ConfigurationError(
+        `Swift analysis on Ubuntu runner images is no longer supported. Please migrate to a macOS runner.`,
+      );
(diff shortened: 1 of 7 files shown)


mbg (dismissed):
This looks good to me, but couldn't we catch this earlier on / before we ever invoke the CLI? Or is this specifically to catch the case where there's no `languages` input but we are running on macOS and there's Swift code in the repo?

henrymercer:
> This looks good to me, but couldn't we catch this earlier on / before we ever invoke the CLI? Or is this specifically to catch the case where there's no `languages` input but we are running on macOS and there's Swift code in the repo?

We could, but AFAICT the information about compatible OSes is not machine readable, so this would mean duplicating the knowledge about incompatible OSes between the CLI and the Action.

mbg:
> We could, but AFAICT the information about compatible OSes is not machine readable, so this would mean duplicating the knowledge about incompatible OSes between the CLI and the Action.

Isn't Swift just a special case in that regard or do we have other language/OS pairs that are not compatible? If it's just a special case for Swift, then I wouldn't be overly opposed to adding a check for that in the Action so that we can fail earlier.

henrymercer:
Looks like we already taught the Action about it, but it was a warning rather than a configuration error.  Let's update that.

src/init-action.ts
```diff
@@ -389,6 +389,15 @@ async function run(startedAt: Date) {
...
     });
 
+    if (
+      config.languages.includes(KnownLanguage.swift) &&
+      process.platform === "linux"
+    ) {
+      throw new ConfigurationError(
+        `Swift analysis on Ubuntu runner images is no longer supported. Please migrate to a macOS runner.`,
+      );
+    }
```
mbg, line 399: What happens if someone tries to analyse Swift on Windows? Should that also result in an error here? I imagine that the warning was specifically because we _used to_ support Swift analysis on Linux. Now that we are turning this into an error to fail an analysis early, should we include Windows as well?

mbg (dismissed):
Thanks for finding that! One comment about that change.

src/init-action.ts
henrymercer: I suppose we might as well :)

mbg (approved):

PR #3806, opened 2026-04-09
Store language aliases from linked CLI

henrymercer:
Add a checked-in snapshot of CodeQL language aliases (`src/known-language-aliases.json`) that is updated by the "Update bundle" workflow.  This will have future uses, but for now we just update the `start-proxy` Action to use this instead of a hardcoded map.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/.github/workflows/update-bundle.yml
+++ b/.github/workflows/update-bundle.yml
@@ -57,6 +57,20 @@ jobs:
       - name: Update bundle
         uses: ./.github/actions/update-bundle
 
+      - name: Set up CodeQL CLI from new bundle
+        id: setup-codeql
+        uses: ./setup-codeql
+        with:
+          tools: https://github.com/github/codeql-action/releases/download/${{ github.event.release.tag_name }}/codeql-bundle-linux64.tar.gz
+
+      - name: Update language aliases
+        env:
+          CODEQL_PATH: ${{ steps.setup-codeql.outputs.codeql-path }}
+        run: |
+          "$CODEQL_PATH" resolve languages --format=betterjson --extractor-include-aliases \
+            | jq -S '.aliases // {}' \
+            > src/known-language-aliases.json
+
       - name: Bump Action minor version if new CodeQL minor version series
         id: bump-action-version
         run: |
--- a/src/start-proxy.ts
+++ b/src/start-proxy.ts
@@ -18,6 +18,7 @@ import {
   FeatureEnablement,
 } from "./feature-flags";
 import * as json from "./json";
+import * as knownLanguageAliases from "./known-language-aliases.json";
 import { KnownLanguage } from "./languages";
 import { Logger } from "./logging";
 import {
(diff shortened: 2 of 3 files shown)


src/start-proxy.ts
```diff
@@ -188,43 +189,30 @@ export const UPDATEJOB_PROXY_VERSION = "v2.0.20250624110901";
...
 
   // Check language aliases
-  if (language in LANGUAGE_ALIASES) {
-    return LANGUAGE_ALIASES[language];
+  if (Object.hasOwn(knownLanguageAliases, language)) {
+    language =
+      knownLanguageAliases[language as keyof typeof knownLanguageAliases];
+    if (Object.hasOwn(KnownLanguage, language)) {
+      return language as KnownLanguage;
+    }
```
mbg, line 215: I think that this should be a given for the languages we have in `known-language-aliases.json` and shouldn't be checked here. (Instead we should check this at the time `known-language-aliases.json` is generated.) Adding a `known-language-aliases.json.d.ts` file with:

```typescript
import { KnownLanguage } from "./languages";

declare const knownLanguageAliases: Record<string, KnownLanguage>;

export default knownLanguageAliases;
```

can accomplish this. You may need to change the import here slightly and modify the `github/filenames-match-regex` eslint rule to accept `.json.d`. The code here then becomes:

```typescript
  // Check language aliases
  if (Object.hasOwn(knownLanguageAliases, language)) {
    return knownLanguageAliases[language];
```

.github/workflows/update-bundle.yml
```diff
@@ -57,6 +57,20 @@ jobs:
...
+        with:
+          tools: https://github.com/github/codeql-action/releases/download/${{ github.event.release.tag_name }}/codeql-bundle-linux64.tar.gz
+
+      - name: Update language aliases
+        env:
+          CODEQL_PATH: ${{ steps.setup-codeql.outputs.codeql-path }}
+        run: |
+          "$CODEQL_PATH" resolve languages --format=betterjson --extractor-include-aliases \
+            | jq -S '.aliases // {}' \
+            > src/known-language-aliases.json
```
mbg, line 72: This is simpler than the TS you had before, but does make it harder to perform the check I am proposing in my other comment. Though, we could just have a simple script which performs that check and runs after this.

mbg (approved):
This looks fine, and thanks for simplifying! Fortunately, I hadn't reviewed those new scripts yet before you removed them.

One suggestion about typing `known-language-aliases.json` and performing the type checks statically, rather than dynamically.

src/start-proxy.ts
henrymercer: We filter the aliases down to the known languages to avoid a behaviour change in the start proxy Action, but I'm not sure if it's desirable in general to restrict to known languages.  If we did want to, I like your solution.  But I don't see the benefit of adding the restriction.

henrymercer:
I'll merge this as is, and make a followup PR implementing the improvements we've discussed offline.

PR #3808, opened 2026-04-10
Improve esbuild metadata usage

mbg:
Follow-up to #3787 which makes the following changes:

- The `bundle-metadata.ts` script is no longer run as part of the `build` script since it wasn't particularly useful in its existing form.
- The bundle metadata is now uploaded as workflow artifacts in the `pr-checks.yml` workflow.
- The `bundle-metadata.ts` script is improved so that it can download the bundle metadata from workflow artifacts for a given branch and compare the current, local metadata to that.

Since we don't yet have these artifacts on `main`, this PR does not automatically invoke `bundle-metadata.ts` as part of any workflow. This is intentional. The script can be run manually if desired. E.g. to compare local bundle metadata against that generated for this branch: `./bundle-metadata.ts --token "$(gh auth token)" --branch "mbg/bundle-metadata-improvements"`

### Risk assessment

Environments:


pr-checks/config.ts
```diff
@@ -11,3 +11,9 @@ export const PR_CHECK_EXCLUDED_FILE = path.join(PR_CHECKS_DIR, "excluded.yml");
 
 /** The path to the esbuild metadata file. */
 export const BUNDLE_METADATA_FILE = path.join(PR_CHECKS_DIR, "..", "meta.json");
```
henrymercer: I realise that this wasn't introduced here, but perhaps we could give this a more descriptive name, e.g. `build-metadata.json`?

henrymercer (commented):
I like the idea of highlighting changes that will affect how long it takes for clients to download the CodeQL Action.  However my understanding is that clients will download the entire contents of the repository, especially now that the Actions packaging experiment has been discontinued.  By measuring just the size of the compiled artifacts I think we risk missing other changes that would slow down clients.

I would encourage us to try to measure and optimise for the client Action download time as best as possible, perhaps by having a PR check that finds out the change in the size of the entire contents of the repo?

mbg:
> I like the idea of highlighting changes that will affect how long it takes for clients to download the CodeQL Action. 

That's not necessarily the primary/sole motivation. Analysing the bundle metadata can also help us understand whether/when we end up bundling things that don't have to be bundled and that might lead to unnecessary (or unnecessarily large) bundle changes.

> I would encourage us to try to measure and optimise for the client Action download time as best as possible, perhaps by having a PR check that finds out the change in the size of the entire contents of the repo?

We could do that separately from this, or think again about whether we can find a way to tag an optimised version of the repo that only contains the bundles.

pr-checks/config.ts
mbg: Makes sense.

henrymercer (commented):
> Analysing the bundle metadata can also help us understand whether/when we end up bundling things that don't have to be bundled and that might lead to unnecessary (or unnecessarily large) bundle changes.

Is there a reason that we'd care about that other than it taking longer to download the Action?  If not, then I think it makes more sense from a complexity point of view to just measure the size of the repo.

mbg:
> Is there a reason that we'd care about that other than it taking longer to download the Action?

Diff sizes / potential merge conflicts is one.

mbg:
Closing this for now, since we merged https://github.com/github/codeql-action/pull/3884.

PR #3809, opened 2026-04-10
Overlay: Determine which versions of CodeQL are compatible with cached base DBs

henrymercer:
Add an unused `getCodeQlVersionsForOverlayBaseDatabases(rawLanguages: string[], logger: Logger)` function that looks up overlay base databases for the provided languages and returns a list of the CodeQL versions that were used to generate those overlay base databases.

The intention is to use this in a future PR to determine what CodeQL version to set up.

Returning a list rather than just the latest means we can filter the list down based on the enabled default version feature flags.  This gives us the ability to skip a CodeQL version that we might initially roll out but then decide to rollback.

### Risk assessment

None yet.

### DIFF
--- a/src/overlay/caching.test.ts
+++ b/src/overlay/caching.test.ts
@@ -7,7 +7,7 @@ import * as sinon from "sinon";
 
 import * as actionsUtil from "../actions-util";
 import * as apiClient from "../api-client";
-import { ResolveDatabaseOutput } from "../codeql";
+import { type ResolveDatabaseOutput } from "../codeql";
 import * as gitUtils from "../git-utils";
 import { BuiltInLanguage } from "../languages";
 import { getRunnerLogger } from "../logging";
--- a/src/overlay/caching.test.ts
+++ b/src/overlay/caching.test.ts
@@ -23,6 +23,7 @@ import {
   downloadOverlayBaseDatabaseFromCache,
   getCacheRestoreKeyPrefix,
   getCacheSaveKey,
+  getCodeQlVersionsForOverlayBaseDatabases,
 } from "./caching";
 import { OverlayDatabaseMode } from "./overlay-database-mode";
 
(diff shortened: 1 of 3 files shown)


src/overlay/caching.ts
```diff
@@ -404,25 +406,115 @@ export async function getCacheRestoreKeyPrefix(
...
+ * Resolves a language name to its canonical form using the static
+ * `known-language-aliases.json` mapping. If the language is not an alias, it is
+ * returned unchanged.
+ */
+function resolveLanguageAlias(language: string): string {
+  const normalized = language.trim().toLowerCase();
+  if (Object.hasOwn(knownLanguageAliases, normalized)) {
+    return knownLanguageAliases[
+      normalized as keyof typeof knownLanguageAliases
+    ];
```
mbg: Note that this is a breaking change in the sense that any previously uploaded caches which include language aliases in the cache key will no longer be restored once this ships.

src/overlay/caching.ts
```diff
@@ -404,25 +406,115 @@ export async function getCacheRestoreKeyPrefix(
...
+
+  logger.info(
+    `Found ${caches.length} overlay-base database cache ` +
+      `${caches.length === 1 ? "entry" : "entries"}.`,
+  );
+
+  // Parse CodeQL versions from cache keys.
+  // After the prefix, the remaining key format starts with
+  // `${codeQlVersion}-`.
+  const versionRegex = /^([\d.]+)-/;
```
mbg, line 493: What happens here if a cache was uploaded for a pre-release CLI?

Also should the `.` be escaped?

src/overlay/caching.ts
```diff
@@ -404,25 +406,115 @@ export async function getCacheRestoreKeyPrefix(
...
+        "cache keys.",
+    );
+    return [];
+  }
+
+  const compatibleVersions = [...versions].sort(semver.rcompare);
+
+  logger.info(
+    `Compatible CodeQL versions in overlay-base database caches: ${compatibleVersions.join(", ")}`,
+  );
```
mbg, line 517: Minor: This could be quite long. Also, isn't this just the list of CodeQL versions, _not_ the compatible ones?

src/overlay/caching.ts
```diff
@@ -404,25 +406,115 @@ export async function getCacheRestoreKeyPrefix(
...
+ * Searches the GitHub Actions cache for overlay-base databases matching the
+ * languages in the given config, and returns all compatible CodeQL versions
+ * found across matching cache entries.
+ *
+ * @param config The configuration object containing the languages to match
+ * @param logger The logger instance
+ * @returns Unique compatible CodeQL versions found in cached overlay-base
+ * databases, sorted from latest to earliest.
+ */
+export async function getCompatibleCodeQlVersionsForOverlayBaseDatabases(
```
mbg: I am confused about the notion of "compatible" here, since this function just seems to extract all the CodeQL CLI versions from the cache keys. I don't see anything that would make a given CLI version compatible or not compatible.

src/overlay/caching.ts
```diff
@@ -404,25 +406,115 @@ export async function getCacheRestoreKeyPrefix(
...
+  logger: Logger,
+): Promise<string[]> {
+  const cacheKeyPrefix = await getCacheKeyPrefixBase(config);
+
+  logger.debug(
+    `Searching for overlay-base databases in Actions cache with ` +
+      `prefix ${cacheKeyPrefix}`,
+  );
+
+  const caches = await listActionsCaches(cacheKeyPrefix);
```
mbg: Note that the approach of listing available caches, performing some checks, and then fetching a particular cache later opens us up to race conditions where a cache may be available when we list them, but no longer when we try to download it.

That's probably not going to happen often and is generally not very likely, but we should consider it.

mbg (changes requested):
Thanks for working on this! This is much simpler than the previous attempt at addressing this problem.

I have picked "Request changes" here because there are a few comments that I had which we should think about before we merge this. 

Also, it might be good to have a separate PR for the first commit. That would make it easier to verify that it's just a refactoring, which I think we can merge straight away. It would then reduce the noise here.

src/overlay/caching.ts
henrymercer: The cache keys only contain parsed languages.  I've updated the methods so the existing code paths do not resolve any aliases.

src/overlay/caching.ts
henrymercer: I've documented and added a test that we ignore nightly CLIs.  There is a caveat I've documented around the prerelease semver format, which I believe we have never used.

henrymercer: The `.` doesn't need to be escaped since it's in a character class.

src/overlay/caching.ts
henrymercer: Yes, updated.  The restriction on the age of the cache entries to 6 days should avoid this being too long a list, and I'm not too worried about a long log line in an edge case.

src/overlay/caching.ts
henrymercer: We want to match the CLI version used in the base database as closely as possible.  Since we're only dealing with stable releases, our goal is to just to download the same CodeQL version as the base DB.  (Though we provide the full list so if the feature flags explicitly disable a particular CodeQL version due to a bug, we can ignore that version.)

src/overlay/caching.ts
henrymercer: Good point, it indeed may be unlikely, but I've added a limit on the age of the cache entry which should prevent this from happening.

oscarsj (dismissed):

src/overlay/caching.ts
```diff
@@ -435,6 +445,39 @@ async function getCacheKeyPrefixBase(
...
 }
 
+/**
+ * Lists overlay-base database cache entries with the given key prefix, ignoring entries that are
+ * old enough that they may be evicted by the Actions cache before we attempt to download them.
+ */
+async function listRecentOverlayBaseDatabaseCaches(
+  cacheKeyPrefix: string,
+  logger: Logger,
+): Promise<ActionsCacheItem[]> {
```
mbg: Sorry to be very critical here, but I don't think this is architecturally a good mitigation for the problem, and adds extra complexity for (IMHO) little gain.

Cache eviction is based on several criteria, and age is only one of them. This mitigation guards against the case where:

- We are analysing a low-traffic repo 
- Around 7 days after the cache was last accessed

That's a valid problem and will happen once in a while, but isn't very common.

I am probably more concerned about high-traffic repos where workflows are uploading caches all the time, and may cause a cache we have identified to be evicted at essentially any time for space reasons. Checking the age of the cache we want doesn't guard against that. (Unless we also check all other caches to see how likely it is that ours would be evicted first, but I don't think that's worth the effort.)

I think realistically, a better solution might just be to stick this logic in a retry loop: find a suitable cache, try to download it, and if it fails attempt the process another time or two.

src/overlay/caching.test.ts
```diff
@@ -7,7 +7,7 @@ import * as sinon from "sinon";
 
 import * as actionsUtil from "../actions-util";
 import * as apiClient from "../api-client";
-import { ResolveDatabaseOutput } from "../codeql";
+import { type ResolveDatabaseOutput } from "../codeql";
```
mbg: Minor (`type` in the `{ }` only disambiguates, while in front doesn't require the module at runtime iirc? Not overly relevant here since it's a test file, but a general point):
```suggestion
import type { ResolveDatabaseOutput } from "../codeql";
```

src/overlay/caching.ts
mbg: Sorry, I think you misunderstood my comment. I obviously understand what the goal of this work is, but I didn't see how the implementation accomplished it.

Reviewing it again now, I see that the key detail is that the call to `listActionsCaches(cacheKeyPrefix)` is given the `cacheKeyPrefix` which already includes the target CodeQL CLI version. I seem to have missed that previously, and so didn't see how the caches were filtered to the right CLI version. It looked to me as if we just retrieved a list of all overlay caches.

It might be good to document this (critical) detail in the code and link to  https://docs.github.com/en/rest/actions/cache?apiVersion=2026-03-10#list-github-actions-caches-for-a-repository which documents that the `key` parameter can be a prefix.

src/overlay/caching.ts
mbg: This seems to have completely disappeared from the code now (for both stores and restores). Is that intentional?

mbg (changes requested):
(Requesting changes to prevent accidental merges.)

I have re-reviewed, thanks for making changes and responding to the previous comments. There is one larger design point I want to flag, and otherwise just a couple of smaller points that follow-up on previous comments.

src/overlay/caching.ts
henrymercer: We discussed this online, and I've added some documentation to the function to acknowledge the possibility that the overlay-base database might get evicted before we can use it, and why we don't think a retry loop is worth it.

src/overlay/caching.ts
henrymercer: I've added links to API docs for each of the functions in `api-client.ts`.

src/overlay/caching.ts
henrymercer: This now happens in the line `const languages = rawLanguages.map(parseBuiltInLanguage);`.

mbg (approved):
These docs changes are all great, thank you!

PR #3811, opened 2026-04-10
Store all built-in languages

henrymercer:
While we want the CodeQL Action to work with third-party language support, having a list of all built-in languages can help us create better type-level checks to ensure that we don't miss things that we want to customize for each of our built-in languages.

For example, in the `start-proxy` Action, we want to make sure each built-in language is associated with a registries config.  If we add a new language, we want to be reminded to update that config.

This means that when we add a new language, we will need to update the `BuiltInLanguage` enum (and potentially other places in the code that exhaustively enumerate `BuiltInLanguage` for safety reasons).  This is a rare event and probably worth it from a safety and correctness perspective.

Commit-by-commit review recommended as we rename `KnownLanguage` to `BuiltInLanguage` to reflect the new motivation behind this type. 

### Risk assessment

Workflow types:

Products:

Environments:


.github/workflows/script/update-builtin-languages.ts
```diff
@@ -0,0 +1,88 @@
...
+const codeqlPath = process.argv[2] || "codeql";
+
+// Step 1: Resolve all language extractor directories.
+const resolveJson: Record<string, string[]> = JSON.parse(
+  execFileSync(
+    codeqlPath,
+    ["resolve", "languages", "--format=json"],
+    { encoding: "utf8" },
+  ),
+);
```
redsun82: typically new languages start out as experimental, and as such require `CODEQL_ENABLE_EXPERIMENTAL_FEATURES` variable to be resolved. To be future ready, we probably need to set that here

src/languages/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  cpp = "cpp",
+  csharp = "csharp",
+  go = "go",
+  java = "java",
+  javascript = "javascript",
+  python = "python",
+  ruby = "ruby",
+  rust = "rust",
+  swift = "swift",
+}
```
redsun82, line 18: maybe it's not worth it, but I wonder if there would be an easy way to have `builtin.json` as the source of truth without having to repeat the list here. I know it should be easy if we use a union type instead of an `enum`. Or we could have a const record built from `builtin.json`. Or we could have codegen do this for us. Anyway, I'll leave to you to decide whether it's worth it, it's true that adding new language support is a rare enough event (for now 😉)

.github/workflows/script/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "lib": ["esnext"],
+    "rootDir": "../../..",
+    "sourceMap": false,
+    "noEmit": true,
+  },
+  "include": ["./*.ts"],
+  "exclude": ["node_modules"]
+}
```
mbg: Minor: We should probably agree on where the TypeScript scripts used in CI live going forward. When I converted some of the scripts from `.github/workflows` to TypeScript I put them in `pr-checks/` since I had already set up the infrastructure for TypeScript scripts there. In general, it would be good to avoid duplicating this secondary project setup across two locations.

.github/workflows/script/update-builtin-languages.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { execFileSync } from "node:child_process";
+import * as fs from "node:fs";
+import * as path from "node:path";
+
+import * as yaml from "yaml";
+
+const codeqlPath = process.argv[2] || "codeql";
+
+// Step 1: Resolve all language extractor directories.
```
mbg: It would be good to encapsulate the different steps in functions so that we can add tests if desired.

.github/workflows/script/update-builtin-languages.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  const extractorYml = yaml.parse(fs.readFileSync(extractorYmlPath, "utf8"));
+  const defaultQueries: unknown[] | undefined = extractorYml.default_queries;
+
+  if (Array.isArray(defaultQueries) && defaultQueries.length > 0) {
+    console.log(`  ✅ ${language}: included (default_queries: ${JSON.stringify(defaultQueries)})`);
+    languages.push(language);
+  } else {
+    console.log(`  ❌ ${language}: excluded (no default queries)`);
```
mbg: Tell me Copilot generated this code without telling me that Copilot generated this code.

.github/workflows/script/update-builtin-languages.ts
```diff
@@ -0,0 +1,88 @@
...
+// Step 4: Write builtin.json.
+const outputPath = path.join(
+  __dirname,
+  "..",
+  "..",
+  "..",
+  "src",
+  "languages",
+  "builtin.json",
+);
```
mbg: In the `pr-checks` setup, we'd want this to be a constant in `config.ts`

src/languages/index.ts
```diff
@@ -27,3 +29,26 @@ export enum JavaEnvVars {
...
 }
+
+const knownLanguageSet = new Set<string>(knownLanguagesData.languages);
+
+export function isKnownLanguage(language: string): language is KnownLanguage {
+  return knownLanguageSet.has(language);
+}
+
+/**
+ * Parses a built-in language name or alias.
```
mbg: Minor: Make it clearer in the wording that this function parses an arbitrary string into a built-in language name, resolving aliases in the process.

The `parseLanguage` function that's now removed had more extensive documentation:

```
/**
 * Parse the start-proxy language input into its canonical CodeQL language name.
 *
 * This uses the language aliases shipped with the Action and will not be able to resolve aliases
 * added by versions of the CodeQL CLI newer than the one mentioned in `defaults.json`. However,
 * this is sufficient for the start-proxy Action since we are already specifying proxy
 * configurations on a per-language basis.
 */
```

src/languages/index.ts
mbg: I believe Henry's intention with this is to explicitly have a duplicate of languages that we treat in special ways in the Action vs the complete list of languages (although the comment disagrees?).

henrymercer: > maybe it's not worth it, but I wonder if there would be an easy way to have `builtin.json` as the source of truth without having to repeat the list here. I know it should be easy if we use a union type instead of an `enum`. Or we could have a const record built from `builtin.json`. Or we could have codegen do this for us.

I explored different approaches to this, but I couldn't find a simple way of using the JSON directly without losing exhaustivity checking on `BuiltInLanguage`.  I think using a codegen library would allow us to keep exhaustivity checking, but it seems too heavyweight given the frequency at which we add new languages.

> I believe Henry's intention with this is to explicitly have a duplicate of languages that we treat in special ways in the Action vs the complete list of languages (although the comment disagrees?).

This PR reflects a change of approach to store the full list of supported languages in the linked CodeQL bundle in service of better type-level checks (like the start-proxy Action checks we discussed offline).

.github/workflows/script/tsconfig.json
henrymercer: I'm happy with moving this to `pr-checks`.

.github/workflows/script/update-builtin-languages.ts
henrymercer: Hmm, I'll fix the `default_queries` / `default queries` discrepancy.  I'm keeping the emoji though :)

.github/workflows/script/update-builtin-languages.ts
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env npx tsx
```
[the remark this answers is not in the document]
henrymercer: Leaving this as is for consistency with other scripts

redsun82 (dismissed):
LGTM!

src/languages/index.ts
mbg: I am not sure that we'd need a codegen library if we wanted to do this. It should be relatively simple (famous last words) to emit a suitable `.d.ts` file from `pr-checks/update-builtin-languages.ts` in addition to the JSON. 

I agree that this doesn't have to be part of this PR though and I am happy to have a look at that once this is merged.

mbg (dismissed):

src/languages/index.ts
henrymercer: I had a look at that approach but couldn't work out how to do it without compromising on exhaustivity checking.  By all means have a look though as I agree it's more elegant than keeping this up to date via a unit test.

mbg (approved):

PR #3850, opened 2026-04-25
Private registries: Add support for Cloudsmith and GCP OIDC configurations

mbg:
This PR primarily adds validation for Cloudsmith and GCP OIDC configurations, so that they can be correctly propagated to the authentication proxy for private package registries when given to the `start-proxy` action as input in GitHub-managed workflows.

The first few commits add a lightweight JSON schema implementation to simplify the specification and validation of different credential objects. The existing private registry configuration types are then migrated to make use of this.

We take advantage of this to simplify some of the validation and testing logic so that we have to update fewer places when adding new configuration types.

Then we add the two new configuration types.

Finally, we add validation for the `replaces-base` setting, which has been around for a while, but wasn't accepted until now.

I'd suggest reviewing this commit-by-commit.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/json/index.ts
+++ b/src/json/index.ts
@@ -36,3 +36,77 @@ export function isStringOrUndefined(
 ): value is string | undefined {
   return value === undefined || isString(value);
 }
+
+/**
+ * Represents a field of type `T` in a schema.
+ * Carries a validation function and flag indicating whether the field is required or not.
+ */
+export type Validator<T> = {
+  validate: (val: unknown) => val is T;
+  required: boolean;
+};
+
+/** Extracts `T` from `Validator<T>`. */
+export type UnwrapValidator<V> =
+  V extends Validator<infer A>
+    ? V["required"] extends true
+      ? A
+      : A | undefined
+    : never;
+
+/** A validator for string fields in schemas. */
+export const string = {
+  validate: isString,
+  required: true,
+} as const satisfies Validator<string>;
+
+/** Transforms a validator to be optional. */
+export function optional<T>(validator: Validator<T>) {
+  return {
+    validate: (val: unknown) => {
+      return val === undefined || val === null || validator.validate(val);
+    },
+    required: false,
+  } as const satisfies Validator<T | undefined | null>;
+}
+
+/** Represents an arbitrary object schema. */
+export type Schema = Record<string, Validator<any>>;
+
+/** Constructs an object type corresponding to a schema. */
+export type FromSchema<S extends Schema> = {
+  [K in keyof S]: UnwrapValidator<S[K]>;
+};
+
+/**
+ * Validates `obj` against `schema`.
+ *
+ * @param schema The schema to validate against.
+ * @param obj The object to validate.
+ * @returns Asserts that `obj` is of the `schema`'s type if validation is successful.
+ */
+export function validateSchema<S extends Schema>(
+  schema: S,
+  obj: UnvalidatedObject<any>,
+): obj is FromSchema<S> {
+  for (const [key, validator] of Object.entries(schema)) {
+    const hasKey = key in obj;
+
+    // If the property is required, but absent, fail.
+    if (validator.required && !hasKey) {
+      return false;
+    }
+
+    // If the property is required, but undefined or null, fail.
+    if (validator.required && (obj[key] === undefined || obj[key] === null)) {
+      return false;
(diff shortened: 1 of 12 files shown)


YiChunLinlifeguidebook (approved):

src/start-proxy/types.ts
```diff
@@ -9,144 +9,165 @@ import { isDefined } from "../util";
...
-  password?: string;
-} & Username;
+export type UsernamePassword = json.FromSchema<typeof usernamePasswordSchema>;
 
 /** Decides whether `config` is based on a username and password. */
 export function isUsernamePassword(
   config: AuthConfig,
 ): config is UsernamePassword {
-  return hasUsername(config) && "password" in config;
+  return json.validateSchema(usernamePasswordSchema, config);
```
henrymercer: Should this still include `hasUsername(config) && "password" in config` like `isToken`?  Otherwise how are we validating that these fields are present given they are both optional in the schema?

src/start-proxy/validation.ts
```diff
@@ -0,0 +1,81 @@
...
+import { isDefined } from "../util";
+
+import type { AuthConfig, UsernamePassword } from "./types";
+import * as types from "./types";
+
+/** Constructs a new object from `obj` with only keys that exist in `schema`. */
+export function cloneCredential<
+  T extends json.FromSchema<S>,
+  S extends json.Schema,
+>(schema: S, obj: T): T {
```
henrymercer: Should the return type be `json.FromSchema<S>` instead?  If `T` extends `json.FromSchema<S>` with some new properties, those won't be included in the output, so I think we're returning a `json.FromSchema<S>` rather than a `T`.

src/json/index.ts
```diff
@@ -36,3 +36,77 @@ export function isStringOrUndefined(
...
+/** Represents an arbitrary object schema. */
+export type Schema = Record<string, Validator<any>>;
+
+/** Constructs an object type corresponding to a schema. */
+export type FromSchema<S extends Schema> = {
+  [K in keyof S]: UnwrapValidator<S[K]>;
+};
+
+/**
+ * Validates `obj` against `schema`.
```
henrymercer: Perhaps mention that extra keys are fine.

src/json/index.ts
```diff
@@ -36,3 +36,77 @@ export function isStringOrUndefined(
...
+    required: false,
+  } as const satisfies Validator<T | undefined | null>;
+}
+
+/** Represents an arbitrary object schema. */
+export type Schema = Record<string, Validator<any>>;
+
+/** Constructs an object type corresponding to a schema. */
+export type FromSchema<S extends Schema> = {
+  [K in keyof S]: UnwrapValidator<S[K]>;
```
henrymercer: Should this only include required keys to be consistent with the validation and avoid the need to specify `username: null` in the tests?

henrymercer (commented):
What happens when we merge this?  Does the authentication proxy already support these new registries or is there more work to do?

src/start-proxy/validation.ts
mbg: I have changed it to remove `T` entirely, and just use `json.FromSchema<S>` since that's enough for how it is used.

mbg:
> What happens when we merge this? Does the authentication proxy already support these new registries or is there more work to do?

@henrymercer The authentication proxy already supports it ([Cloudsmith](https://github.com/dependabot/proxy/pull/47) and [GCP](https://github.com/dependabot/proxy/pull/108)). Also, it parses the JSON configuration as a generic map and would ignore keys it doesn't understand, so even if we ended up in a scenario where we have newer configurations than the version of the proxy that we have supports, it shouldn't cause problems. In general, the Dependabot team have been working on the implementation in the proxy and backend.

src/start-proxy.test.ts
henrymercer: Should we also have tests for Cloudsmith and GCP?

henrymercer (dismissed):
Just one thing I missed before.  Feel free to merge if you don't think it's important.

src/start-proxy.test.ts
mbg: I replaced these with some code that constructs tests based on `oidcSchemas`, so that all are covered automatically.

henrymercer (approved):

PR #3852, opened 2026-04-28
Add random suffix when writing diagnostics to avoid filename collisions

henrymercer:
Fixed a bug where two diagnostics produced within the same millisecond could overwrite each other on disk, causing one of them to be lost.

We could also add the source ID to the filename, but that doesn't protect against multiple diagnostics from the same source, and would be more likely to hit file path length limits.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -73,6 +73,13 @@ let unwrittenDiagnostics: UnwrittenDiagnostic[] = [];
 let unwrittenDefaultLanguageDiagnostics: DiagnosticMessage[] = [];
 
 /**
+ * Counter used to generate a unique suffix for each diagnostic filename, so that
+ * two diagnostics produced within the same millisecond do not overwrite each
+ * other on disk.
+ */
+let diagnosticCounter = 0;
+
+/**
  * Constructs a new diagnostic message with the specified id and name, as well as optional additional data.
  *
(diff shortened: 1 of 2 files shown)


navntoft (dismissed):
Good call with the suffix instead of source id.

src/diagnostics.ts
```diff
@@ -167,10 +167,21 @@ function writeDiagnostic(
     // Create the directory if it doesn't exist yet.
     mkdirSync(diagnosticsPath, { recursive: true });
 
+    // Include a random suffix to avoid filename collisions between diagnostics
+    // produced within the same millisecond. This doesn't need to be
+    // cryptographically secure, so `Math.random` is fine.
+    const uniqueSuffix = Math.floor(Math.random() * 0x100000000)
+      .toString(16)
+      .padStart(8, "0");
```
mbg: I wonder if, given the relatively low volume of diagnostics we expect the Action to produce, it would be a better approach to just have a global counter that's incremented with each diagnostic that's written, and append that? That way, we'd avoid the very unlikely, but possible case where the random numbers clash (as Copilot notes).

henrymercer: Agreed, my first approach was to avoid global state but everything is running on the same thread so there isn't a risk of race conditions.

mbg (approved):

PR #3853, opened 2026-04-28
Improve connection tests

mbg:
A few minor improvements to the private registry connection checks:

- Use `GET` instead of `HEAD` requests, for better compatibility. Some registries don't like `HEAD` requests. We still call `res.destroy()` immediately after inspecting the status code, so never read the response body. So using `GET` over `HEAD` should have a minimal impact on performance.
- For NuGet feeds, we now test the service index instead of the base URL. The service index should always be present and is more reliable than making a request to the base URL.
- Some logging improvements.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/start-proxy/reachability.ts
+++ b/src/start-proxy/reachability.ts
@@ -41,7 +41,7 @@ class NetworkReachabilityBackend implements ReachabilityBackend {
         url,
         {
           agent: this.agent,
-          method: "HEAD",
+          method: "GET",
           ca: this.proxy.cert,
           timeout: 5 * 1000, // 5 seconds
         },
--- a/src/start-proxy/reachability.ts
+++ b/src/start-proxy/reachability.ts
@@ -7,6 +7,35 @@ import { getErrorMessage } from "../util";
 
 import { getAddressString, ProxyInfo, Registry } from "./types";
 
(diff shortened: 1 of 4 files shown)


src/start-proxy/reachability.ts
```diff
@@ -2,11 +2,41 @@ import * as https from "https";
...
+export interface ConnectionTestConfig {
+  /** An optional path to append to the end of the base url. */
+  path?: string;
+}
+
+/** A partial mapping of registry types to extra connection test configurations. */
+export const connectionTestConfig: Partial<
+  Record<string, ConnectionTestConfig>
+> = {
+  nuget_feed: { path: "v3/index.json" },
```
henrymercer, line 21: I presume v3 is old enough that we can rely on this existing / the proxy only supports v3 anyway?

henrymercer (approved):

src/start-proxy/reachability.ts
mbg: AFAIK nothing else exists. I.e. I am not aware that `v2/index.json` was ever a thing. See also https://learn.microsoft.com/en-us/nuget/api/service-index#versioning

PR #3874, opened 2026-05-06
Tests: Run slow `scanArtifactsForTokens` test in CI only by default

henrymercer:
The "scanArtifactsForTokens finds token in debug artifacts" test takes 4.7 seconds to complete on my machine.  Run it in CI only by default.

[Here](https://github.com/github/codeql-action/actions/runs/25451545833/job/74669781947#step:7:133)'s proof that it actually runs.

### DIFF
--- a/src/artifact-scanner.test.ts
+++ b/src/artifact-scanner.test.ts
@@ -141,7 +141,12 @@ test("scanArtifactsForTokens handles files without tokens", async (t) => {
   }
 });
 
-if (os.platform() !== "win32") {
+// This test is slow (extracts and scans a zip artifact), so by default we only run it in CI. Set
+// RUN_SLOW_TESTS=1 to run it locally.
+if (
+  os.platform() !== "win32" &&
+  (process.env.CI === "true" || process.env.RUN_SLOW_TESTS === "1")
+) {
   test("scanArtifactsForTokens finds token in debug artifacts", async (t) => {
     t.timeout(15000); // 15 seconds
     const messages: LoggedMessage[] = [];
--- a/src/artifact-scanner.ts
+++ b/src/artifact-scanner.ts
@@ -156,6 +156,10 @@ async function scanArchiveFile(
     );
   }
 
+  if (process.platform === "win32") {
+    throw new Error("Scanning archives is not supported on Windows.");
+  }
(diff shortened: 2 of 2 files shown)


src/artifact-scanner.test.ts
```diff
@@ -141,7 +141,12 @@ test("scanArtifactsForTokens handles files without tokens", async (t) => {
   }
 });
 
-if (os.platform() !== "win32") {
+// This test is slow (extracts and scans a zip artifact), so by default we only run it in CI. Set
+// RUN_SLOW_TESTS=1 to run it locally.
+if (
+  os.platform() !== "win32" &&
```
mbg, line 147: If we are only running this test in CI, why not also run it on Windows? Given the nature of the function being tested, it seems like we could easily miss a Windows-specific problem.

If there is a good reason why we aren't running it on Windows, it would be good if the comment noted it and explained why.

mbg (commented):
Just a question about why this also changes the test to not run on Windows, given that it would only run in CI.

Aside from that, it makes sense to me to only run this in CI.

src/artifact-scanner.test.ts
henrymercer: The artifact scanner isn't compatible with Windows — it's only used in our debug artifact tests, which always run on Linux.  It's not compatible because it directly shells out to extract archives.

mbg: Ah, I had a look at the implementation but didn't see an obvious guard against running on Windows. I saw the `isInTestMode()` check, but I didn't verify whether we always run on Linux there.

In that case, it might be good to make this more explicit in the main implementation in `scanArtifactsForTokens` as well and return with a suitable message there if it is somehow running on Windows?

henrymercer: Agreed, done!

src/artifact-scanner.ts
```diff
@@ -156,6 +156,10 @@ async function scanArchiveFile(
     );
   }
 
+  if (process.platform === "win32") {
```
mbg, line 159: Of course it is `win32` regardless of architecture.

mbg (approved):

PR #3880, opened 2026-05-06
Overlay: Use overlay-aware CLI version when analyzing PRs

henrymercer:
When analyzing PRs, prefer CLI versions that have cached overlay-base databases to speed up analysis time.  However to ensure we can effectively rollback new versions, do not use a CodeQL version whose feature flag is disabled, even if this means running without overlay analysis.

This will be shipped via two feature flags:

- `overlay_analysis_match_codeql_version_dry_run` logs a diagnostic when the overlay-aware version differs from the latest enabled version
- `overlay_analysis_match_codeql_version` uses the overlay-aware version when analysing PRs

### Risk assessment

Workflow types:

Products:

Environments:


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Added an experimental change which, when running a Code Scanning analysis for a PR with [improved incremental analysis](https://github.com/github/roadmap/issues/1158) enabled, prefers CodeQL CLI versions that have a cached overlay-base database for the configured languages. This speeds up analysis when the latest version does not yet have a cached overlay-base database. We expect to roll this change out to everyone in May. [#3880](https://github.com/github/codeql-action/pull/3880)
```
mbg:
```suggestion
- Added an experimental change which, when running a Code Scanning analysis for a PR with [improved incremental analysis](https://github.com/github/roadmap/issues/1158) enabled, prefers CodeQL CLI versions that have a cached overlay-base database for the configured languages. This speeds up analysis for a repository when there is not yet a cached overlay-base database for the latest CLI version. We expect to roll this change out to everyone in May. [#3880](https://github.com/github/codeql-action/pull/3880)
```

src/feature-flags.ts
```diff
@@ -637,26 +669,27 @@ class GitHubFeatureFlags {
...
       );
       const result: CodeQLDefaultVersionInfo = {
-        cliVersion: defaults.cliVersion,
-        tagName: defaults.bundleVersion,
+        enabledVersions: [
+          {
+            cliVersion: defaults.cliVersion,
+            tagName: defaults.bundleVersion,
+          },
+        ],
```
mbg: Minor: Could we define this as a constant to avoid some duplication?

src/start-proxy.ts
```diff
@@ -440,7 +440,7 @@ export async function getDownloadUrl(
     // Retrieve information about the CLI version we should use. This will be either the linked
     // version, or the one enabled by FFs.
     const versionInfo = useFeaturesToDetermineCLI
-      ? await getCliVersionFromFeatures(features)
+      ? (await getCliVersionFromFeatures(features)).enabledVersions[0]
```
mbg, line 443: This is an interesting change. It might mean that the bundle release used by `start-proxy` and `init` could diverge. That's certainly not a problem right now, since we don't depend on them lining up. (I.e. we have thus far not had any reason to require the proxy to be of at least a particular version for changes made in the CLI/extractors.) 

It could also mean that a base database is extracted using the version of the proxy that shipped with the corresponding CLI, and we then end up using a newer proxy for the overlay analysis on a PR. I don't see an immediate problem with it, but it is worth noting.

We might want to think about whether this is something we want to document somehow so that it doesn't become a problem if we ever want to make that assumption.

src/setup-codeql-action.ts
```diff
@@ -145,6 +145,7 @@ async function run(startedAt: Date): Promise<void> {
       getTemporaryDirectory(),
       gitHubVersion.type,
       codeQLDefaultVersionInfo,
+      undefined, // rawLanguages: currently, setup-codeql is not language aware
```
mbg: A potential issue down the line if we ever want to expand on `setup-codeql` and allow it to run before `init`. It might be worth documenting the implications of this on overlay analysis here.

src/init-action.ts
```diff
@@ -301,12 +301,16 @@ async function run(startedAt: Date) {
     const codeQLDefaultVersionInfo =
       await features.getEnabledDefaultCliVersions(gitHubVersion.type);
     toolsFeatureFlagsValid = codeQLDefaultVersionInfo.toolsFeatureFlagsValid;
+    const rawLanguages = configUtils.getRawLanguagesNoAutodetect(
```
mbg, line 304: Could we add a comment here to say what this is for? E.g. "Split the comma-separated input string (if any) into an array of its components to use for identifying overlay base databases".

src/setup-codeql.test.ts
```diff
@@ -601,3 +610,155 @@ test.serial(
...
+      throw new Error("not implemented");
+    },
+    getValue: async (feature) => {
+      if (feature === Feature.OverlayAnalysisMatchCodeqlVersion) {
+        return matchFlagEnabled;
+      }
+      return false;
+    },
+  };
+}
```
mbg: Question: Why add this instead of using the existing `createFeatures`?

src/setup-codeql.ts
```diff
@@ -264,12 +270,84 @@ async function findOverridingToolsInCache(
...
+
+  let cachedVersions: string[] | undefined;
+  try {
+    cachedVersions = await getCodeQlVersionsForOverlayBaseDatabases(
+      rawLanguages,
+      logger,
+    );
+  } catch (e) {
+    logger.warning(
+      `While setting up CodeQL, was unable to list overlay-base databases in the Actions cache. Details: ${e}`,
```
mbg: Minor: The wording here isn't great. How about "While setting up CodeQL, the CodeQL Action was unable to [..]"?

src/setup-codeql.ts
```diff
@@ -264,12 +270,84 @@ async function findOverridingToolsInCache(
...
+  logger: Logger,
+): Promise<CodeQLVersionInfo[]> {
+  if (rawLanguages === undefined || rawLanguages.length === 0) {
+    return [];
+  }
+  if (!(await features.getValue(Feature.OverlayAnalysisMatchCodeqlVersion))) {
+    return [];
+  }
+
+  let cachedVersions: string[] | undefined;
```
mbg, line 303: I wish we had a `SemVer` type instead of just `string`.

src/setup-codeql.ts
```diff
@@ -454,9 +539,14 @@ export async function getCodeQLSource(
...
-    cliVersion = defaultCliVersion.enabledVersions[0].cliVersion;
-    tagName = defaultCliVersion.enabledVersions[0].tagName;
+    const version = await resolveDefaultCliVersion(
+      defaultCliVersion,
+      rawLanguages,
+      features,
+      logger,
+    );
+    cliVersion = version.cliVersion;
+    tagName = version.tagName;
```
mbg, line 603: It would be nice to improve the inline documentation a bit where you have made changes here in order to explain the intended behaviour / flow of this function so we don't have to trace out all possible paths through it every time 😅 

(Or majorly refactor it.)

src/setup-codeql.test.ts
```diff
@@ -601,3 +610,155 @@ test.serial(
...
+  },
+);
+
+test.serial(
+  "getEnabledVersionsWithOverlayBaseDatabases includes the highest version when it is cached",
+  async (t) => {
+    sinon.stub(api, "getAutomationID").resolves("test/");
+    sinon.stub(api, "listActionsCaches").resolves([
+      {
+        key: "codeql-overlay-base-database-1-aaaaaaaaaaaaaaaa-javascript-2.20.2-abc-1-1",
```
mbg: Minor: Could we recycle one of the functions for computing the cache key here so that this doesn't break immediately if we make any changes to the cache key format?

src/setup-codeql.test.ts
```diff
@@ -601,3 +610,155 @@ test.serial(
...
+      {
+        key: "codeql-overlay-base-database-1-aaaaaaaaaaaaaaaa-javascript-2.21.0-abc-1-1",
+      },
+      // Flag-enabled versions present in the cache.
+      {
+        key: "codeql-overlay-base-database-1-aaaaaaaaaaaaaaaa-javascript-2.20.1-def-2-1",
+      },
+      {
+        key: "codeql-overlay-base-database-1-aaaaaaaaaaaaaaaa-javascript-2.20.0-ghi-3-1",
+      },
```
mbg, line 727: Swap the order of these to exercise the "sorted desc" part of the test?

src/setup-codeql.test.ts
```diff
@@ -601,3 +610,155 @@ test.serial(
...
+);
+
+test.serial(
+  "getEnabledVersionsWithOverlayBaseDatabases does not list caches when rawLanguages is empty",
+  async (t) => {
+    const listStub = sinon.stub(api, "listActionsCaches").resolves([]);
+
+    const result = await setupCodeql.getEnabledVersionsWithOverlayBaseDatabases(
+      overlayMatchEnabledVersions,
+      undefined,
```
mbg: Test name says "empty", but here we pass in `undefined` directly. What if we get `[]`?

src/setup-codeql.test.ts
```diff
@@ -601,3 +610,155 @@ test.serial(
...
+      ["javascript"],
+      makeOverlayMatchFeatures(true),
+      getRunnerLogger(true),
+    );
+    t.deepEqual(result, []);
+  },
+);
+
+test.serial(
+  "getEnabledVersionsWithOverlayBaseDatabases includes the highest version when it is cached",
```
mbg: Could we make this test name a bit clearer? The test body seems to just set up a single cache and the test then checks that it is included in the results. I don't see how this is specific to "the highest"?

src/init-action.ts
```diff
@@ -304,13 +304,17 @@ async function run(startedAt: Date) {
     const rawLanguages = configUtils.getRawLanguagesNoAutodetect(
       getOptionalInput("languages"),
     );
+    const useOverlayAwareDefaultCliVersion = !!analysisKinds?.includes(
+      AnalysisKind.CodeScanning,
+    );
```
mbg: Just to be clear, is the intention to:

- Use this for only `code-scanning` analyses?
- Or for any analysis involving `code-scanning`?

mbg (commented):
Thanks for breaking this up into (reasonably) sane commits and a separate PR from the previous one! Overall, this looks like a pretty good first stab at this change and I don't see any major issues here.

I have left a bunch of detailed comments. There are a couple in particular about possible issues down the road that we might want to consider documenting or guarding against.

I am wondering if the second FF is a bit overkill and adds an extra layer of complexity on top of an already fairly complex change. I don't feel strongly about this, but the thought crossed my mind whether a reasonable thing to do here to reduce complexity / risk would be to consider breaking this PR up into three separate changes:

- The first commit that modifies some of the existing logic to return all FF-enabled CLI versions, but doesn't make use of this information. We can ship that change without a new FF.
- Then ship the new logic that's guarded by the second FF. I.e. everything except actually using the CLI version that we determined using the new process.
- Then ship the first FF + the logic that uses the CLI version determined using the new process.

src/init-action.ts
henrymercer: I've added JSDoc to that function to avoid needing to repeat ourselves.

src/start-proxy.ts
henrymercer: I was originally thinking we'd do this in a separate PR, but let's just add the languages and analysis-kinds inputs to avoid confusion here.

src/setup-codeql-action.ts
henrymercer: See above.

src/setup-codeql.ts
henrymercer: The `semver` package does have a `SemVer` class — perhaps worth looking at, but I'd suggest we do so in a separate PR.

src/setup-codeql.ts
henrymercer: I agree that this is a pain but (as you mention above) this PR has got large enough and I'd like to avoid further scope creep.  Would you be happy if I look at this as followup?

src/init-action.ts
henrymercer: I've implemented the latter for now, but we don't have a motivating use case yet for multiple analysis kinds, so that might be something we change in the future.  I'd say we'd be most likely to want to use the overlay-aware version, but I'm happy to change it to be more conservative if you prefer.

henrymercer:
> I don't feel strongly about this, but the thought crossed my mind whether a reasonable thing to do here to reduce complexity / risk would be to consider breaking this PR up into three separate changes:
>   * The first commit that modifies some of the existing logic to return all FF-enabled CLI versions, but doesn't make use of this information. We can ship that change without a new FF.

I have a mild preference against splitting this PR up further now, but I agree it would have been better to introduce just the "dry run" changes in a separate PR.  Happy to factor this into a separate PR if you have a strong preference.

I would prefer to keep the "dry run" changes behind a feature flag though — we're making an extra API request to list the caches, and even though it's not much code, it's safer to roll it out with a flag.

src/init-action.ts
mbg: (We discussed my reason for asking elsewhere.) Either is fine practically, since there should only ever be one analysis kind, but it might be safer to add a more conservative `analysisKinds.length` check as well.

src/setup-codeql.ts
mbg: That's fine.

setup-codeql/action.yml
```diff
@@ -19,6 +19,25 @@ inputs:
       If not specified, the Action will check in several places until it finds
       the CodeQL tools.
     required: false
+  languages:
+    description: >-
+      A comma-separated list of CodeQL languages that will be analyzed in subsequent
+      `github/codeql-action/init` and `github/codeql-action/analyze` invocations. If specified, the
+      Action may use this list to select a CodeQL CLI version that is best suited to analyzing those
+      languages, for example by preferring a version that has a cached overlay-base database for the
```
mbg, line 27: Minor: I'd drop the "If specified, [..], for example by preferring a version that has a cached overlay-base database for the specified languages." part from this description. My thinking here is that:

- `setup-codeql` is mainly about installing the CLI.
- So the main thing anyone using this should care about is the `tools` input.
- If `tools` is provided, I'd expect us to honour it.
- If it is unspecified, then `languages` may affect the CLI version we choose, but so do other things.
- Therefore, I am not sure it's worth pointing out here explicitly.

src/setup-codeql-action.ts
```diff
@@ -139,14 +141,18 @@ async function run(startedAt: Date): Promise<void> {
...
     const initCodeQLResult = await initCodeQL(
       getOptionalInput("tools"),
       apiDetails,
       getTemporaryDirectory(),
       gitHubVersion.type,
       codeQLDefaultVersionInfo,
-      undefined, // rawLanguages: currently, setup-codeql is not language aware
-      false, // useOverlayAwareDefaultCliVersion: setup-codeql is not language aware
+      rawLanguages,
+      analysisKinds.includes(AnalysisKind.CodeScanning),
```
mbg: Minor: If we make this check more conservative `init`, then we should make the same change here.

src/config-utils.ts
```diff
@@ -407,6 +407,7 @@ export async function getLanguages(
   return languages;
 }
 
+/** Parses the `languages` input into a list of languages without checking if they are supported by CodeQL. */
```
mbg: Minor: Really all this does is split a comma-separated string into an array, removes excess space characters, converts the strings to lower-case, and removes empty elements. The current description might suggest that something more specific to languages is happening here. How about:

```
Splits a comma-separated string into an array. Excess spaces are removed and all characters are converted to lower-case.
```

setup-codeql/action.yml
```diff
@@ -19,6 +19,25 @@ inputs:
...
+  analysis-kinds:
+    description: >-
+      [Internal] A comma-separated list of analysis kinds that subsequent
+      `github/codeql-action/init` invocations will enable. If specified, the Action may use this
+      list to select a CodeQL CLI version that is best suited to those analysis kinds. This input is
+      not remembered and must also be passed to `github/codeql-action/init`.
+
+      Available options are the same as for the `analysis-kinds` input on the `init` Action.
+    default: 'code-scanning'
+    required: true
```
mbg, line 40: `required: false`?

src/start-proxy.ts
mbg: That solution is OK, but a concern here is that it takes the `setup-codeql` action a step away from an action that just handles installing the CLI in the sense that it now needs to be aware of what is going to be analysed (if anything). I think the changes are OK though because:

- The `languages` and `analysis-kinds` inputs are optional (ish - `analysis-kinds` has a default).
- We are still just installing the CLI, but for overlay-enabled Code Scanning analyses the new logic is unavoidable and that requires the information about the (expected) analysis.

mbg (dismissed):
> I have a mild preference against splitting this PR up further now, but I agree it would have been better to introduce just the "dry run" changes in a separate PR. Happy to factor this into a separate PR if you have a strong preference.

It's fine to keep it as-is. I don't feel strongly enough about this to push for splitting it up.

Otherwise the changes look good, thanks for addressing my feedback! I have left a few minor comments.

setup-codeql/action.yml
henrymercer: I think this is mostly relevant to us as we develop default setup, indeed it seems unlikely anyone would want to use `setup-codeql` in an advanced setup.  I'm leaning on the side of keeping it to explain why it's there when `setup-codeql` is otherwise language agnostic.

setup-codeql/action.yml
henrymercer: Since this input has a default, we're disallowing specifying `analysis-kinds: null`.  I don't see why we'd want to allow that, hence disallowing it.

src/start-proxy.ts
henrymercer: Indeed, since we're now basing our choice of CLI on the languages being analysed when doing a code scanning analysis, I think this is a necessity.

src/config-utils.ts
henrymercer: That's a little implementation specific.  Perhaps the main opposition is with "Parses"?  I'll update that.

mbg (approved):

PR #3899, opened 2026-05-14
Reduce duplication across JS bundles by creating one bundle with smaller entry point wrappers

mbg:
Currently, we have a lot of duplication across the JS bundles that esbuild produces for each entry point. This PR fundamentally changes this as follows:

- All CodeQL Action code is now bundled into `entry-points.js`. This avoids significant duplication across the individual bundles, since most of the Action's code is shared among the different actions.
- We have new entry points for the individual actions, which import `entry-points` and call the relevant `run` function.
- In the esbuild configuration, we mark `entry-points` as external, so that it is not bundled up with the entry point wrappers.

This reduces the size of the compressed repo. I tested the impact of this locally with a fresh, shallow checkout of the repo:

```
$ git clone git@github.com:github/codeql-action --depth=1 --branch mbg/esbuild/split
$ cd codeql-action
$ tar -czvf *
```

The resulting `.tar.gz` file is around ~2.9M. This is approximately a 70% reduction in archive size.

### Risk assessment

Workflow types:

Products:

Environments:


src/analyze-action-env.test.ts
```diff
@@ -69,7 +69,8 @@ test("analyze action with RAM & threads from environment variables", async (t) =
     // level but does not wait for it to finish. To ensure that calls to
     // runFinalize and runQueries are correctly captured by spies, we explicitly
     // wait for the action promise to complete before starting verification.
-    await analyzeAction.runPromise;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    await analyzeAction.runWrapper();
```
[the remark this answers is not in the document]
henrymercer: Agreed

src/analyze-action-input.test.ts
```diff
@@ -67,7 +67,8 @@ test("analyze action with RAM & threads from action inputs", async (t) => {
     // level but does not wait for it to finish. To ensure that calls to
     // runFinalize and runQueries are correctly captured by spies, we explicitly
     // wait for the action promise to complete before starting verification.
-    await analyzeAction.runPromise;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    await analyzeAction.runWrapper();
```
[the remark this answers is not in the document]
henrymercer: Agreed

build.mjs
```diff
@@ -62,18 +62,118 @@ const onEndPlugin = {
...
+        const isPost = match[2] !== undefined;
+
+        actions.push({
+          path: actionFile,
+          name: actionName,
+          isPost,
+          pascalCaseName: `${toPascal(actionName)}${isPost ? "Post" : ""}Action`,
+        });
+      }
+    });
```
henrymercer, line 106: This works, but it might be simpler to get this information from the `action.yml` files.

mbg: I have played around with this, but I don't actually think it's simpler, because the `action.yml` files point at the `-entry.js` files. So finding the `action.yml` files, extracting the paths to the `-entry.js` files, and then reverse-engineering what the paths of the `-action.ts` / `-action-post.ts` files is ends up being more convoluted.

src/analyze-action-env.test.ts
henrymercer, line 25: I think this reasoning is no longer true.

src/analyze-action-input.test.ts
henrymercer, line 25: Likewise here

build.mjs
```diff
@@ -62,18 +62,123 @@ const onEndPlugin = {
...
 };
 
+/** The name of the virtual `entry-points` module. */
+const SHARED_ENTRYPOINT = "entry-points";
+
+/**
+ * This plugin finds all source files that contain action entry points.
+ * It then generates the virtual `entry-points` module which imports all identifies files,
+ * and re-exports their `runWrapper` functions with suitable aliases.
+ * A tiny stub file is emitted for each Action entrypoint. Each stub imports the shared bundle
```
henrymercer, line 72: Nit: "action" vs "Action" inconsistency in documentation.  I mildly prefer "Action", but don't mind as long as it's consistent.

build.mjs
```diff
@@ -62,18 +62,123 @@ const onEndPlugin = {
...
+        "utf-8",
+      );
+
+      const actionsSorted = actions.sort((a, b) =>
+        a.name.localeCompare(b.name),
+      );
+      const imports = actionsSorted
+        .map(
+          (action) =>
+            `import * as ${action.pascalCaseName} from "./src/${basename(action.path)}"`,
```
henrymercer, line 130: Minor: would it be better to use `;` rather than relying on automated semicolon insertion?

henrymercer (approved):
I'd like to see a couple of things addressed, but it can be done as followup.

mbg:
@henrymercer I agree with all points. I will merge this and then prepare a follow-up PR to address the comments.

PR #3901, opened 2026-05-14
Minify test debug artifacts

henrymercer:
Quick win: minify the test debug artifacts to include only the TRAP file containing the secret.  Preserve the zipped structure so the test is realistic, but remove the other files.

This reduces the size of the test debug artifacts from 1.14 MB to 22 KB on disk.  The tarred checkout size decreases from 10.02 MB to 9.17 MB (-8.4%).

This does make it harder to ensure the scanner can scan artifacts in a reasonable amount of time, but the scanner isn't being changed much, and we will still do a full test in the PR checks.  So the repo size saving seems worth the cost.


src/artifact-scanner.test.ts
```diff
@@ -141,14 +141,9 @@ test("scanArtifactsForTokens handles files without tokens", async (t) => {
...
 });
 
-// This test is slow (extracts and scans a zip artifact), so by default we only run it in CI. Set
-// RUN_SLOW_TESTS=1 to run it locally.
-if (
-  os.platform() !== "win32" &&
-  (process.env.CI === "true" || process.env.RUN_SLOW_TESTS === "1")
-) {
+// `scanArchiveFile` does not support Windows, so we skip this test there.
+if (os.platform() !== "win32") {
```
mbg, line 145: Since we are not running the test on Windows, could we compress the debug artifacts as `.tar.gz` or `.tar.zstd` instead?

src/artifact-scanner.test.ts
```diff
@@ -141,14 +141,9 @@ test("scanArtifactsForTokens handles files without tokens", async (t) => {
   }
 });
 
-// This test is slow (extracts and scans a zip artifact), so by default we only run it in CI. Set
-// RUN_SLOW_TESTS=1 to run it locally.
-if (
-  os.platform() !== "win32" &&
-  (process.env.CI === "true" || process.env.RUN_SLOW_TESTS === "1")
-) {
```
mbg, line 149: Why are we reverting this?

mbg (approved):
LGTM, with one further suggestion and one question.

src/artifact-scanner.test.ts
henrymercer: Now that the size of the debug artifacts has reduced, the test doesn't have a significant impact on the time it takes to run `npm test`.

src/artifact-scanner.test.ts
henrymercer: We'd only save ~0.2 KB.  I don't think it's worth it.

mbg: Thanks for investigating :)

PR #3903, opened 2026-05-14
PR checks: Run slowest macOS checks on larger runners

henrymercer:
Run our slowest macOS checks on larger runners to improve the overall turnaround time of PR checks.

See https://github.com/github/codeql-action/actions/metrics/performance?tab=jobs — the 15 slowest jobs all run on macOS default size runners.

This speeds up the PR checks from ~19m to ~11m.

We'll need to update the set of required checks before merging this.

### DIFF
--- a/pr-checks/sync.ts
+++ b/pr-checks/sync.ts
@@ -43,7 +43,7 @@ interface Specification extends JobSpecification {
   inputs?: Record<string, WorkflowInput>;
   /** CodeQL bundle versions to test against. Defaults to `DEFAULT_TEST_VERSIONS`. */
   versions?: string[];
-  /** Operating system prefixes used to select runner images (e.g. `["ubuntu", "macos"]`). */
+  /** Operating system prefixes, either as strings or with explicit runner image labels. */
   operatingSystems?: OperatingSystem[];
   /** Per-OS version overrides. If specified for an OS, only those versions are tested on that OS. */
   osCodeQlVersions?: Record<string, string[]>;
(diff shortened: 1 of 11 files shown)


pr-checks/sync.ts
```diff
@@ -28,6 +28,13 @@ interface WorkflowInput {
 /** A partial mapping from known input names to input definitions. */
 type WorkflowInputs = Partial<Record<KnownInputName, WorkflowInput>>;
 
+type OperatingSystem =
+  | string
+  | {
+      os: string;
+      "runner-image"?: string;
+    };
```
mbg, line 47: Minor: JSDocs? Perhaps rather than having the extended description below, move some of that here and then simplify the doc for `operatingSystems`.

mbg: I also wonder if having both `os` and `runner-image` is a bit redundant since `runner-image` typically starts with the `os` string, so you could infer `os` from `runner-image` when that's specified.

mbg (dismissed):
LGTM, with a couple minor comments

pr-checks/sync.ts
henrymercer: I've added JSDoc and a more specific type for `os`.  You're right that we could probably infer `runner-image` but there's no guarantee about the structure of the labels in general so I'll leave this as is.

pr-checks/sync.ts
```diff
@@ -28,10 +28,21 @@ interface WorkflowInput {
 /** A partial mapping from known input names to input definitions. */
 type WorkflowInputs = Partial<Record<KnownInputName, WorkflowInput>>;
 
+/** An operating system identifier. */
+type OperatingSystemIdentifier = "ubuntu" | "macos" | "windows";
```
mbg, line 32: Nice catch!

mbg (approved):

PR #3910, opened 2026-05-18
Action size: Add a PR check that comments on significant repo size changes

henrymercer:
The compressed checkout of the repo is downloaded at the start of every CodeQL job.  Significant jumps or drops are worth surfacing since they directly affect job startup time.

To that end, this PR adds a "Check repo size" PR check that streams `git archive --format=tar.gz` for both the PR base and `HEAD`, compares the compressed sizes, and posts a sticky comment when the difference is at least 10% in either direction.


.github/workflows/check-repo-size.yml
```diff
@@ -0,0 +1,53 @@
+name: Check repo size
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
mbg: Could we add this on to an existing workflow (pr-checks?) rather than adding a totally new one?

pr-checks/check-repo-size.ts
```diff
@@ -0,0 +1,327 @@
...
+
+/**
+ * Fraction of the base archive size at which a delta is considered
+ * significant enough to warrant a new sticky comment. We always update an
+ * existing comment regardless, so the comment stays in sync as the diff
+ * evolves.
+ */
+export const SIGNIFICANT_DELTA_FRACTION = 0.1;
+
+export type Octokit = ReturnType<typeof getOctokit>;
```
mbg: `api-client.ts` exports `ApiClient` (the return type of `getOctokit`). Why this roundabout way of getting hold of that type?

mbg (commented):
Thanks for looking at this! Two high-level design questions:

1. Could we add this on to an existing workflow (pr-checks?) rather than adding a totally new one?
2. Rather than comparing to the PR base ref, would it make sense to compare to the latest CodeQL Action release instead?

henrymercer:
> Rather than comparing to the PR base ref, would it make sense to compare to the latest CodeQL Action release instead?

With that approach, if we merge a PR that significantly increases / decreases the repo size, this would create noise on all subsequent PRs until the next release.  Since we compare against the base, we'll get the comparison against the latest release when we run the release process, so we have another opportunity to see significant changes at that point.

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,130 @@ jobs:
...
+
+      - name: Post repo size comment
+        env:
+          COMMENT_MARKER: "<!-- repo-size-diff-bot -->"
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          significant=$(jq -r '.significant' repo-size-comment/metadata.json)
+          body=$(cat repo-size-comment/body.md)
```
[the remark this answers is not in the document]
henrymercer: This is an acceptable risk.

pr-checks/check-repo-size.ts
```diff
@@ -0,0 +1,221 @@
...
+  console.log(`Wrote body.md and metadata.json to ${args.outputDir}.`);
+  return 0;
+}
+
+if (require.main === module) {
+  void (async () => {
+    try {
+      process.exit(await main());
+    } catch (err) {
+      console.error(err instanceof Error ? err.message : String(err));
```
mbg: Minor: `getErrorMessage` could be imported and used here.

pr-checks/check-repo-size.ts
```diff
@@ -0,0 +1,221 @@
...
+      process.exit(1);
+    }
```
mbg: Consider moving this into a named function to avoid the awkward `void (async () => {` here.

pr-checks/check-repo-size.ts
```diff
@@ -0,0 +1,221 @@
...
+    outputDir,
+  };
+}
+
+async function main(): Promise<number> {
+  const args = readArgs();
+
+  // The script lives at `<repoRoot>/pr-checks/check-repo-size.ts`, so the repo root is the parent
+  // directory.
+  const repoRoot = path.resolve(__dirname, "..");
```
mbg: Consider adding a `REPO_ROOT` constant to `config.ts` since there are other definitions which do the `path.resolve(__dirname, "..")` dance there.

pr-checks/check-repo-size.ts
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Format a byte count into a human-readable string with binary units. If `signed` is true, a
+ * leading `+` is prepended for non-negative values so gains and losses are visually distinct.
+ */
+export function formatBytes(bytes: number, signed = false): string {
+  const sign = bytes < 0 ? "-" : signed ? "+" : "";
+  const abs = Math.abs(bytes);
+  if (abs < 1024) return `${sign}${abs} B`;
+  if (abs < 1024 * 1024) return `${sign}${(abs / 1024).toFixed(2)} KiB`;
+  return `${sign}${(abs / 1024 / 1024).toFixed(2)} MiB`;
```
mbg: Minor: This seems a bit over-engineered and might make it harder to spot differences between units (although always including the bytes in the output mitigates that somewhat). I think it might be better if we stick to one unit, especially since we are unlikely to end up with repo sizes that can be measured in bytes anytime soon (unless something is really wrong).

.github/workflows/pr-checks.yml
```diff
@@ -67,22 +71,21 @@ jobs:
...
 
-  # Verifying the PR checks are up-to-date requires Node 24. The PR checks are not dependent
-  # on the main codebase and therefore do not need to be run as part of the same matrix that
-  # we use for the `unit-tests` job.
-  verify-pr-checks:
-    name: Verify PR checks
+  # These checks do not need to be run as part of the same matrix that we use for the `unit-tests`
+  # job.
+  pr-checks:
+    name: PR Checks
```
mbg: I deliberately tried to avoid naming this "PR Checks" when I added this job, since the overall workflow is already called "PR Checks" and it's a bit confusing to have a workflow named that which does a bunch of things, and then a job that only does part of that.

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,129 @@ jobs:
         run: npm ci
 
       - name: Verify PR checks up to date
-        if: always()
         run: .github/workflows/script/verify-pr-checks.sh
 
       - name: Run pr-checks tests
-        if: always()
```
mbg, line 103: I have a slight preference for trying to run this if `verify-pr-checks.sh` fails, since they perform different checks and it would be useful to know if both are failing for a PR rather than seeing the `verify-pr-checks.sh` failure, fixing that, only to then get a failure here. Perhaps instead of just running it `always()` we can check that at least `npm ci` was successful?

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,129 @@ jobs:
...
-
-    steps:
-      - uses: actions/checkout@v6
-      - id: head-version
-        name: Verify all Actions use the same Node version
+      - name: Verify all Actions use the same Node version
+        id: head-version
         run: |
-          NODE_VERSION=$(find . -name "action.yml" -exec yq -e '.runs.using' {} \; | grep node | sort | uniq)
+          NODE_VERSION=$(find . -path "*/node_modules" -prune -o -name "action.yml" -exec yq -o=json '.runs.using' {} \; | jq -rs '[.[] | select(. != null and startswith("node"))] | unique | .[]')
```
mbg, line 114: I am not keen on the complexity here. I would prefer to have this kind of check implemented elsewhere and have drafted https://github.com/github/codeql-action/pull/3917 which adds an esbuild plugin for the same task. We could alternative have a script in `pr-checks` instead.

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,129 @@ jobs:
...
-      BASE_REF: ${{ github.base_ref }}
-
-    permissions:
-      contents: read
-
-    steps:
-      - uses: actions/checkout@v6
-      - id: head-version
-        name: Verify all Actions use the same Node version
+      - name: Verify all Actions use the same Node version
```
mbg, line 111: This should probably also run even if the previous two steps have failed.

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,129 @@ jobs:
...
-        name: 'Backport: Check out base ref'
+      - name: Fetch base commit
+        # Forks and Dependabot PRs don't have permission to write comments, so skip the repo size
+        # check in those cases.
+        if: >-
+          github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.full_name == github.repository &&
+          github.event.pull_request.user.login != 'dependabot[bot]'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
```
mbg, line 131: I assume that, if a PR is opened against `main`, `base.sha` is always going to be the `HEAD` of `main`. Just a sanity check that it won't be the PR branch base (since that might be stale).

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,129 @@ jobs:
...
+        run: git fetch --no-tags --depth=1 origin "$BASE_SHA"
+
+      - name: Check repo size
+        # Forks and Dependabot PRs don't have permission to write comments, so skip the repo size
+        # check in those cases.
+        if: >-
+          github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.full_name == github.repository &&
+          github.event.pull_request.user.login != 'dependabot[bot]'
```
mbg: Minor: To avoid repetition, could we check the status of the "Fetch base commit" step and only run these if that was successful (rather than skipped)?

.github/workflows/pr-checks.yml
```diff
@@ -96,54 +99,129 @@ jobs:
...
-          BASE_VERSION=$(find . -name "action.yml" -exec yq -e '.runs.using' {} \; | grep node | sort | uniq)
+          BASE_VERSION=$(find . -path "*/node_modules" -prune -o -name "action.yml" -exec yq -o=json '.runs.using' {} \; | jq -rs '[.[] | select(. != null and startswith("node"))] | unique | .[]')
           echo "HEAD_VERSION: ${HEAD_VERSION}"
           echo "BASE_VERSION: ${BASE_VERSION}"
           if [[ "$BASE_VERSION" != "$HEAD_VERSION" ]]; then
             echo "::error::Cannot change the Node version of an Action in a backport PR."
             exit 1
           fi
+
+  post-repo-size-comment:
```
mbg, line 180: We have discussed elsewhere whether this is a useful addition or not.

mbg (commented):
Thanks for investing the time into putting this together! This makes sense both given the work we have done recently on decreasing the repo size and as a follow-up to https://github.com/github/codeql-action/pull/3808

I don't think there's anything blocking here, but I have had a few comments that we'll want to review.

.github/workflows/pr-checks.yml
henrymercer: I was thinking the PR branch base would be preferable, so that we compare just the commits in the PR, rather than comparing against size increases or decreases on `main` that aren't part of the PR branch.

mbg: Perhaps we should do both. I see your point about comparing the size difference to the PR branch's starting point, but I could also see a case where e.g. we reduce the size of the repo on `main`, the PR branch is based on a commit before that improvement, the PR branch contains a change that when merged to `main` would undo the improvement, but that's not visible in a comparison against the branch base?

henrymercer: This seems like an edge case — I think I'd rather catch this at release time than introduce more complexity here.  Or if you feel strongly about this, we can look at this as followup.

.github/workflows/pr-checks.yml
henrymercer: I'll keep it unless you feel strongly.

mbg (dismissed):

mbg (approved):

PR #3912, opened 2026-05-18
Action size: Reduce duplication between `upload-lib` and `entry-points`

henrymercer:
Reduce the duplication between `upload-lib` and `entry-points` by exporting `lib/upload-lib.js` from `entry-points` and having `lib/upload-lib.js` simply re-export those exports instead of duplicating implementations.

### Risk assessment

Only impacts internal workflows that were directly importing `upload-lib`.

Internal only.


henrymercer:
Passed internal testing, so marking as ready for review.

build.mjs
```diff
@@ -65,12 +65,22 @@ const onEndPlugin = {
 /** The name of the virtual `entry-points` module. */
 const SHARED_ENTRYPOINT = "entry-points";
 
+/** The property name under which `upload-lib`'s namespace is exposed on `entry-points`. */
```
mbg: Minor: `in` or `from`:
```suggestion
/** The property name under which `upload-lib`'s namespace is exposed in `entry-points`. */
```

build.mjs
```diff
@@ -136,15 +146,19 @@ const entryPointsPlugin = {
         )
         .join("\n\n");
 
+      // Also re-export the `upload-lib` namespace so that external consumers can reach it
+      // via the `lib/upload-lib.js` stub without us having to bundle a second copy.
+      const uploadLibReExport = `export * as ${UPLOAD_LIB_EXPORT} from "${UPLOAD_LIB_SRC}";`;
+
       return {
-        contents: `"use strict";\n${imports}\n\n${wrappers}\n`,
+        contents: `"use strict";\n${imports}\n\n${wrappers}\n\n${uploadLibReExport}\n`,
```
mbg: Super minor since the generated file never sees the light of day, but usually we'd have `export`s following `import`s:
```suggestion
        contents: `"use strict";\n${imports}\n\n${uploadLibReExport}\n\n${wrappers}\n`,
```

build.mjs
```diff
@@ -163,16 +177,22 @@ const entryPointsPlugin = {
...
       }
+
+      // Write a small stub for `upload-lib` that re-exports it from the shared bundle.
+      // External callers (e.g. internal testing environments) `require("./lib/upload-lib")`
+      // and expect the same shape as before, so we expose the namespace as `module.exports`.
+      await writeFile(
+        join(OUT_DIR, "upload-lib.js"),
+        `// Automatically generated stub re-exporting '${UPLOAD_LIB_SRC}.ts' from '${SHARED_ENTRYPOINT}.js'.\n\n` +
+          `"use strict";\n\n` +
+          `module.exports = require("./${SHARED_ENTRYPOINT}").${UPLOAD_LIB_EXPORT};\n`,
```
mbg: Consider whether this could also be a template `.js.tpl` file instead of inlined (or at least a `.join`ed array).

mbg (dismissed):

build.mjs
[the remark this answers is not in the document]
henrymercer: True, but in practice this doesn't matter since we are only exporting this for internal purposes and this doesn't affect our internal usage.

build.mjs
henrymercer: Ah I wasn't aware of that convention.  I'll apply.

mbg (approved):

PR #3930, opened 2026-05-22
Do not use `core.exportVariable` in unit tests

mbg:
Although [we already clear environment variables that are set during unit tests](https://github.com/github/codeql-action/blob/0fb8a6672b5f51d3dc5bf6c25f7c19591ddd842c/src/testing-utils.ts#L146), that does not affect the behaviour of `core.exportVariable` which additionally sets environment variables for subsequent steps in a workflow. If the unit tests are run in CI, then `core.exportVariable` sets environment variables for subsequent steps in the workflow job which can interfere with them. 

This PR improves the situation by introducing a wrapper around `core.exportVariable` which does not call `core.exportVariable` when `NODE_ENV` is `test` (which is set automatically by `ava`).

### Risk assessment

Environments:


src/environment.ts
```diff
@@ -154,3 +156,29 @@ export enum EnvVar {
...
+export function exportVariable(name: string, val: any): void {
+  if (process.env["NODE_ENV"] === "test") {
+    // Setting the environment variable for the current process is OK since we reset
+    // those at the end of each test. This allows tests to pass that rely on that
+    // part of the `core.exportVariable` behaviour.
+    process.env[name] = val;
+  } else {
+    // Call `core.exportVariable` whenever we are not in a test environment.
+    core.exportVariable(name, val);
+  }
```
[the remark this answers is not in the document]
mbg: Like I wrote in the PR description, `ava` sets this automatically.

src/environment.ts
```diff
@@ -154,3 +156,29 @@ export enum EnvVar {
...
+  return process.env[EnvVar.TEST_MODE] === "true";
+}
+
+/**
+ * Wrapper around `core.exportVariable` which does not call `core.exportVariable`
+ * when running unit tests. This is important, because otherwise `core.exportVariable`
+ * sets environment variables for other steps in a workflow when we run unit tests in CI.
+ */
+export function exportVariable(name: string, val: any): void {
+  if (process.env["NODE_ENV"] === "test") {
```
oscarsj, line 175: Shouldn't this just use the `isInTestMode()` function declared earlier? why the difference?

mbg: `isInTestMode` is based on the `CODEQL_ACTION_TEST_MODE` environment variable, which we primarily set for e2e tests. Historically, it was only about whether or not we should upload SARIF results, but has been used for some other purposes in that context.

The `CODEQL_ACTION_TEST_MODE` environment variable is not set for the unit tests. I did initially have a version of `exportVariable` that was based on `isInTestMode` before I noticed that. However, rather than trying to decide if it would be OK to set `CODEQL_ACTION_TEST_MODE` for the unit tests or not (since it changes the behaviour), I figured that testing whether we are running the unit tests or not is a safer approach that involves less potential for change.

PR #3932, opened 2026-05-26
Fix ghs_ token regex: rebuild lib/ bundle and add missing hyphen

hagould:
Updates the ghs_ token regex to support the new token format which allows dots and underscores ([A-Za-z0-9._-]) and variable length (no longer fixed at 36 chars).

See the changelog: https://github.blog/changelog/2026-05-15-github-app-installation-tokens-per-request-override-header/

Supersedes #3931 (from fork, had the correct source but stale bundle).
Fixes the compiled `lib/entry-points.js` bundle which was missing the `-` (hyphen) character in the `ghs_` token regex character class.

The source `.ts` file correctly uses `[A-Za-z0-9._-]` but the committed bundle had `[A-Za-z0-9._]` — missing the hyphen that base64url encoding (RFC 4648 §5) uses in JWT signatures.

This PR rebuilds the bundle from the source to fix the mismatch.

### DIFF
--- a/src/artifact-scanner.test.ts
+++ b/src/artifact-scanner.test.ts
@@ -23,6 +23,9 @@ test("makeTestToken", (t) => {
   t.is(makeTestToken(255).length, 255);
 });
 
+const NEW_FORMAT_GHS_TOKEN =
+  "ghs_abc123.def456.ghi789_abc123.def456.ghi789";
+
 test("isAuthToken", (t) => {
   // Undefined for strings that aren't tokens
   t.is(isAuthToken("some string"), undefined);
@@ -32,9 +35,10 @@ test("isAuthToken", (t) => {
   // Token types for strings that are tokens.
   t.is(isAuthToken(`ghp_${makeTestToken()}`), TokenType.PersonalAccessClassic);
   t.is(isAuthToken(`ghp_${makeTestToken()}`), TokenType.PersonalAccessClassic);
+  t.is(isAuthToken(NEW_FORMAT_GHS_TOKEN), TokenType.ServerToServer);
   t.is(
     isAuthToken(`ghs_${makeTestToken(255)}`),
-    TokenType.AppInstallationAccess,
+    TokenType.ServerToServer,
   );
   t.is(
     isAuthToken(`github_pat_${makeTestToken(22)}_${makeTestToken(59)}`),
@@ -51,6 +55,15 @@ test("isAuthToken", (t) => {
       GITHUB_PAT_CLASSIC_PATTERN,
     ]),
     undefined,
+  );
+  t.is(
+    isAuthToken(NEW_FORMAT_GHS_TOKEN, [
+      {
+        type: TokenType.AppInstallationAccess,
+        pattern: /ghs_[A-Za-z0-9._]{36,}/g,
+      },
+    ]),
+    TokenType.AppInstallationAccess,
   );
 });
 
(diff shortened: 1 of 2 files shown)


src/artifact-scanner.test.ts
```diff
@@ -52,6 +52,15 @@ test("isAuthToken", (t) => {
...
   );
+  t.is(
+    isAuthToken(NEW_FORMAT_GHS_TOKEN, [
+      {
+        type: TokenType.ServerToServer,
+        pattern: /ghs_[A-Za-z0-9._-]{36,}/g,
+      },
+    ]),
+    TokenType.ServerToServer,
+  );
```
mbg: This addition misses the point of what the preceding two tests are testing. We are not interested in custom pattern configurations, just custom subsets of the set of available configurations. The preceding two assertions test that, so this addition is not required:

```suggestion
```

src/artifact-scanner.test.ts
```diff
@@ -76,16 +85,12 @@ const testTokens = [
   },
   {
     type: TokenType.ServerToServer,
-    value: `ghs_${makeTestToken()}`,
+    value: NEW_FORMAT_GHS_TOKEN,
   },
```
mbg, line 88: This results in the new format being tested by the tests below, but the old format is no longer being tested. Please restore a test case for the old format.

src/artifact-scanner.ts
```diff
@@ -55,15 +55,15 @@ const GITHUB_TOKEN_PATTERNS: TokenPattern[] = [
   },
   {
     type: TokenType.ServerToServer,
-    pattern: /\bghs_[a-zA-Z0-9]{36}\b/g,
+    pattern: /ghs_[A-Za-z0-9._-]{36,}/g,
```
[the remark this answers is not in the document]
mbg: Copilot is correct. For our use case(s), we would not want this to be unconstrained.

src/artifact-scanner.ts
```diff
@@ -109,16 +104,27 @@ function scanFileForTokens(
...
         }
-        logger.debug(`Found ${matches.length} ${type}(s) in ${relativePath}`);
+
+        seenMatches.add(index);
+        findings.push({ tokenType: type, filePath: relativePath });
+        matchCount++;
+      }
+
+      if (matchCount > 0) {
+        logger.debug(`Found ${matchCount} ${type}(s) in ${relativePath}`);
```
mbg: Could you explain what the goal of these changes is in relation to the objectives of the PR?

mbg (changes requested):
Thank you for taking the initiative to update our code to be aware of the new token format. That's a welcome contribution.

Please note how the `artifact-scanner` module is used, which is twofold:

- For use in **tests** as a **best-effort** check that we don't accidentally include tokens in artifacts. 
- For use in production code to decide whether a given credential is a kind of GitHub token or not.

I have left a few detailed comments.

src/artifact-scanner.test.ts
hagould: Done, removed in latest commit

src/artifact-scanner.test.ts
hagould: Thank you -  restored.

src/artifact-scanner.ts
hagould: > Copilot is correct. For our use case(s), we would not want this to be unconstrained.

 Updated to `\bghs_[A-Za-z0-9._-]{36,516}(?![A-Za-z0-9._-])`. The upper bound of 516 comes from babeld's MAX_PASSWORD_SIZE of 520 bytes, which is the hard ceiling.
 
 > Could you explain what the goal of these changes is in relation to the objectives of the PR?
 
They were added to handle overlapping patterns between ServerToServer and AppInstallationAccess, but since those are now collapsed into a single pattern, they're no longer needed.

hagould:
@mbg Anything else need adjustment?

PR #3937, opened 2026-05-28
Change waitForProcessing to use exponential backoff

robertbrignull:
The goal is to change `waitForProcessing` to backoff a bit more aggressively and reduce the maximum number of requests that it makes. During periods where analysis processing is being delayed, we're seeing a large number of extra GET requests that I believe are coming from here. See linked internal issue for more info.

In this PR I'm suggesting changing it to exponential backoff that'll make a maximum of 5 requests, instead of the current behaviour which can make up to 24 requests. I also add a wait before the first poll because processing will always take at least a couple of seconds so it's unlikely that an immediate poll will get anything. Let me know if you think this is reasonable. We can also tweak values to keep the maximum timeout to 2 minutes if that would be better.

### Risk assessment

- **Low risk:** Hopefully can be safely tested before merge (either by tests or on a test repository)

Workflow types:

Products:

Environments:

### DIFF
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -858,7 +858,9 @@ export async function waitForProcessing(
 
     // Do an initial wait because processing will always take a minimum of 2-3 seconds
     let statusCheckBackoff = STATUS_CHECK_INITIAL_BACKOFF_MILLISECONDS;
-    await util.delay(statusCheckBackoff, { allowProcessExit: false });
+    if (process.env["NODE_ENV"] !== "test") {
+      await util.delay(statusCheckBackoff, { allowProcessExit: false });
+    }
 
     for (
       let statusCheckCount = 1;
--- a/src/upload-lib.ts
+++ b/src/upload-lib.ts
@@ -829,8 +829,10 @@ function dumpSarifFile(
   fs.writeFileSync(outputFile, sarifPayload);
 }
 
-const STATUS_CHECK_FREQUENCY_MILLISECONDS = 5 * 1000;
-const STATUS_CHECK_TIMEOUT_MILLISECONDS = 2 * 60 * 1000;
+// Should lead to status checks after 5s, 15s, 35s, 75s, and 155s.
+const STATUS_CHECK_INITIAL_BACKOFF_MILLISECONDS = 5 * 1000;
+const STATUS_CHECK_BACKOFF_MULTIPLIER = 2;
+const STATUS_CHECK_MAX_TRIES = 5;
 
 type ProcessingStatus = "pending" | "complete" | "failed";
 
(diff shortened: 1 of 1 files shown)


src/upload-lib.ts
```diff
@@ -912,9 +919,8 @@ export async function waitForProcessing(
         util.assertNever(status);
       }
 
-      await util.delay(STATUS_CHECK_FREQUENCY_MILLISECONDS, {
-        allowProcessExit: false,
-      });
+      statusCheckBackoff *= STATUS_CHECK_BACKOFF_MULTIPLIER;
+      await util.delay(statusCheckBackoff, { allowProcessExit: false });
```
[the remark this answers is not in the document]
robertbrignull: I don't think this comment is quite right about the worst case calculations, but it's still a good idea and it would be cleaner to put the timeout check at the end of the loop, instead of doing an extra loop iteration just to break out early.

src/upload-lib.ts
```diff
@@ -854,20 +856,15 @@ export async function waitForProcessing(
...
-        // It's possible the analysis will eventually finish processing, but it's not worth spending more
-        // Actions time waiting.
-        logger.warning(
-          "Timed out waiting for analysis to finish processing. Continuing.",
-        );
-        break;
-      }
+    // Do an initial wait because processing will always take a minimum of 2-3 seconds
+    let statusCheckBackoff = STATUS_CHECK_INITIAL_BACKOFF_MILLISECONDS;
+    await util.delay(statusCheckBackoff, { allowProcessExit: false });
```
mbg: Minor: I think this also adds a `STATUS_CHECK_INITIAL_BACKOFF_MILLISECONDS` long delay to tests where we can get a "response" from a stubbed `client.request` faster. If that is the case, we could either make this initial wait conditional on e.g. the `NODE_ENV` (like in https://github.com/github/codeql-action/pull/3930) or stub `util.delay` in relevant tests to not wait.

mbg (dismissed):
Thanks for preparing this! The change makes to me as-is and there's nothing blocking here. 

We may want to review how well chosen the backoff pattern is, but that's something we can adjust later if needed. 

Just one comment about the (potential) delay this imposes on the unit tests.

Also agree with Copilot that it would be nice to have a test that checks that this works as expected.

src/upload-lib.ts
robertbrignull: Good point. I think skipping this initial wait during tests makes sense.

robertbrignull:
> We may want to review how well chosen the backoff pattern is, but that's something we can adjust later if needed.

I don't have any good data on what a good backoff pattern would be. This was just the idea that came to me first. But yes, it's easy to adjust it later on.

src/upload-lib.ts
robertbrignull: Looking more closely, the code from https://github.com/github/codeql-action/pull/3930 has now changed and there aren't any other references to `NODE_ENV` is the codebase. I've changed this PR to use `isInTestMode()`. Do you think that's the right way to do things?

henrymercer (commented):
Echoing Michael's review, this looks good.  One comment: I think it would be better to use `NODE_ENV` as you originally implemented vs `isInTestMode` — the latter is mainly intended for disabling uploading SARIFs to code scanning, and I don't think it's true when we run the unit tests locally or in CI.  We can add `NODE_ENV` to `queries/default-setup-environment-variables.ql` to suppress the warning, sorry about the noise from that check.

robertbrignull:
> I think it would be better to use `NODE_ENV` as you originally implemented vs `isInTestMode` — the latter is mainly intended for disabling uploading SARIFs to code scanning, and I don't think it's true when we run the unit tests locally or in CI.

Thanks for explaining. I'll just remove my last commit.

I looked earlier for the test mode env var and saw lots of cases of `CODEQL_ACTION_TEST_MODE: true` in workflow files, but you're right it not in the `pr-checks` workflow file and wouldn't be set locally either.

henrymercer (approved):

robertbrignull:
I also tested this with an advanced setup workflow in a test repo and everything worked as expected.

PR #3943, opened 2026-06-02
Cache CLI version information across Actions steps

henrymercer:
Persist CLI version information across Actions steps to avoid calling `codeql version` in each step.  This saves 3 to 6 calls (and associated JVM startups) in a typical analysis job.

We use the path to the CodeQL CLI as a cache key.  This is robust to setting up a different CLI with `init` or `setup-codeql` in the same job, but if a user hotswaps the CLI with a different mechanism and keeps the path the same, we are going to have the wrong version information.  That seems like a rare enough case to be acceptable, but if we care about this, we could introduce more data into the cache key (at the cost of some performance).

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/codeql.ts
+++ b/src/codeql.ts
@@ -523,7 +523,7 @@ async function getCodeQLForCmd(
       return cmd;
     },
     async getVersion() {
-      let result = util.getCachedCodeQlVersion();
+      let result = util.getCachedCodeQlVersion(cmd);
       if (result === undefined) {
         const output = await runCli(cmd, ["version", "--format=json"], {
           noStreamStdout: true,
@@ -535,12 +535,13 @@ async function getCodeQLForCmd(
             `Invalid JSON output from \`version --format=json\`: ${output}`,
           );
         }
-        util.cacheCodeQlVersion(result);
+        util.cacheCodeQlVersion(cmd, result);
       }
       return result;
     },
     async printVersion() {
-      await runCli(cmd, ["version", "--format=json"]);
+      // Reuse the cached version information rather than invoking the CLI again.
+      core.info(JSON.stringify(await this.getVersion(), null, 2));
     },
     async supportsFeature(feature: ToolsFeature) {
       return isSupportedToolsFeature(await this.getVersion(), feature);
--- a/src/util.ts
+++ b/src/util.ts
@@ -619,15 +619,51 @@ export function asHTTPError(arg: any): HTTPError | undefined {
 
 let cachedCodeQlVersion: undefined | VersionInfo = undefined;
 
-export function cacheCodeQlVersion(version: VersionInfo): void {
+/** The persisted version together with the CLI path it was obtained from. */
+interface PersistedVersionInfo {
+  cmd: string;
+  version: VersionInfo;
+}
+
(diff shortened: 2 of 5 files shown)


src/util.ts
```diff
@@ -619,15 +619,51 @@ export function asHTTPError(arg: any): HTTPError | undefined {
...
+    return undefined;
+  }
+  if (
+    typeof persisted?.version?.version !== "string" ||
+    (cmd !== undefined && persisted.cmd !== cmd)
+  ) {
+    return undefined;
+  }
+  return persisted.version;
 }
```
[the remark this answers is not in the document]
henrymercer: This isn't relevant at the moment.

src/util.ts
```diff
@@ -619,15 +619,51 @@ export function asHTTPError(arg: any): HTTPError | undefined {
...
+    return undefined;
+  }
+  let persisted: PersistedVersionInfo;
+  try {
+    persisted = JSON.parse(serialized) as PersistedVersionInfo;
+  } catch {
+    return undefined;
+  }
+  if (
+    typeof persisted?.version?.version !== "string" ||
```
mario-campos: Nit: I think this would be a bit more readable as a type predicate:


```suggestion
    isPersistedVersionInfo(persisted) ||
```

Where `isPersistedVersionInfo` is defined as:

```TypeScript
function isPersistedVersionInfo(x: unknown): x is PersistedVersionInfo {
  return typeof (x as any)?.version?.version === "string" &&
         typeof (x as any)?.cmd === "string";
}
```

src/codeql.ts
```diff
@@ -535,12 +535,13 @@ async function getCodeQLForCmd(
             `Invalid JSON output from \`version --format=json\`: ${output}`,
           );
         }
-        util.cacheCodeQlVersion(result);
+        util.cacheCodeQlVersion(cmd, result);
```
mario-campos, line 538: There's a hypothetical "risk" here that a subsequent step/process may be inefficiently and repeatedly parsing the `CODEQL_VERSION_INFO` environment variable if the process calls `getVersion()` more than once.

The reason being that when `getCachedCodeQlVersion()` parses `CODEQL_VERSION_INFO` it does not store that result in the local variable `cachedCodeQlVersion`. That variable is only assigned when `result === undefined`—i.e.`cachedCodeQlVersion` is `undefined` _and_ `CODEQL_VERSION_INFO` is unset/invalid—which is the first time that `getVersion()` is ever called.

Of course, this is all hypothetical because I don't think `getVersion()` is called more than once in the same step/process. And, it's probably a small penalty because it probably wouldn't take much time to parse the output of `codeql version`. But, the possibility would exist.

One option could be to always call `cacheCodeQlVersion(...)` just before returning from `getVersion()`:

```TypeScript
if (result === undefined) {
   ...
}
util.cacheCodeQlVersion(cmd, result);
return result;
```

mario-campos (dismissed):
Some minor nits, but overall LGTM!

mario-campos (approved):

De-ASI-INTERFACE (commented):
Thank you !

PR #3950, opened 2026-06-04
Cache CLI extractor paths across Actions steps

mario-campos:
Similar to https://github.com/github/codeql-action/pull/3943, this PR caches the output of `codeql resolve languages`, which contains the paths to the various extractors so that repeated calls to `resolveLanguages()` are idempotent. Additionally, re-implement `resolveExtractor()` as a wrapper over `resolveLanguages()` (to re-use the cached output) rather than shell out to `codeql resolve extractor`.

In one experiment, I counted **seven** instances of shelling out to `codeql resolve extractor`. When you dig into the code, you can see why: `resolveExtractor()` is not called often or from many places; But one caller is `isTracedLanguage()`, which is wrapped by `isScannedLanguage()`. And these functions are often used in a loop/map over all/some languages. This can explain why we see consecutive executions of `codeql resolve extractor`.

In support of the above goals, this PR also adds some additional functions to the `json` module, to enable validation of the `codeql version` output.

### Risk assessment

Workflow types:

Products:

Environments:

- **Other** - Manual/local testing


src/codeql.ts
```diff
@@ -969,32 +974,11 @@ async function getCodeQLForCmd(
...
-            },
-          },
-        },
-      ).exec();
-      return JSON.parse(extractorPath) as string;
+      // A previous implementation executed `codeql resolve extractor`.
+      // This can be a bit slow due to the JVM startup cost. Instead, get
+      // the extractor path from resolveLanguages(), which caches its output.
+      const extractors = await this.resolveLanguages();
+      return extractors[language][0];
```
henrymercer: Does this correctly handle language aliases?  We do currently normalise languages to their original names, but it would be good to at least document this change of behaviour if not resolve aliases.

Also, we should make sure that `extractors[language]` is defined and if not, error nicely.

henrymercer (commented):
Caching these invocations makes a lot of sense!  I have a high level comment and a couple of lower level comments.

The main point is that now that we're caching multiple invocations, it might be a good opportunity to generalise the design.  For instance, you could imagine something like:

```ts
const versionCache = createPersistedCliCache({ envVar: EnvVar.CODEQL_VERSION_INFO, validate: isVersionInfo });
const resolveLanguagesCache = createPersistedCliCache({ envVar: EnvVar.CODEQL_RESOLVE_LANGUAGES, validate: isResolveLanguagesOutput });
```

where `createPersistedCliCache` handles memoising in the Action and persisting between Actions steps with an environment variable.

Some smaller things:

- Ideally the cache entry would also depend on `getExtraOptionsFromEnv(["resolve","languages"])`
- We should remove the cache in `testing-utils.ts` like we do for the CodeQL version cache

src/environment.ts
```diff
@@ -23,6 +23,12 @@ export enum EnvVar {
    */
   CODEQL_VERSION_INFO = "CODEQL_ACTION_CLI_VERSION_INFO",
 
+  /**
+   * `ResolveLanguagesOutput` for the CodeQL CLI, so later Actions steps can reuse it instead of
```
mbg: Minor: `ResolveLanguagesOutput` is a type, but the environment variable presumably stores some representation of a _value_ of that type. For example:

```suggestion
   * Used to store a stringified JSON representation of a `ResolveLanguagesOutput` value for the CodeQL CLI, so later Actions steps can reuse it instead of
```

src/util.ts
```diff
@@ -701,6 +701,75 @@ export function getCachedCodeQlVersion(cmd?: string): undefined | VersionInfo {
...
+  value: unknown,
+): value is PersistedResolveLanguagesOutput {
+  return (
+    typeof value === "object" &&
+    value !== null &&
+    typeof (value as Record<string, unknown>).cmd === "string" &&
+    typeof (value as Record<string, unknown>).output === "object" &&
+    (value as Record<string, unknown>).output !== null
+  );
+}
```
mbg: I know this mirrors what was done in https://github.com/github/codeql-action/pull/3943, but it would probably make sense to use the helper functions from the `./json` module.

mbg (commented):
I agree with @henrymercer's comments regarding a more generalised design for this. I am wondering about the use of environment variables here vs using a file on disk. I don't know if you have already considered this, but we store e.g. the Action configuration on disk as a file. Perhaps that would make sense for these cached CLI results as well.

A general point: could we also make sure to add doc comments for new top-level definitions before merging?

src/util.ts
mario-campos: Done

src/codeql.ts
mario-campos: According to Copilot, at the point of this function call, `language` has already been normalized.

mario-campos:
I've taken your comments into consideration and overhauled the design to be more comprehensive and unified. The design now backs to a temporary file instead of the environment. I also identified a few opportunities to refactor some duplicated code into helper functions.

I kept the use of `cmd` as a key in the cache, but I question whether it's really necessary. I think it's safe to assume that, in most cases, there will only be one instance of `codeql` in use per job. And, even in the event that there's more than one instance, how likely is it that `init` would use a different version than `autobuild` or `analyze`? If it's not necessary, I would opt to delete it to simplify the code a bit.

src/codeql.ts
```diff
@@ -731,26 +791,23 @@ async function getCodeQLForCmd(
...
-      } catch (e) {
-        throw new Error(
-          `Unexpected output from codeql resolve languages with --format=betterjson: ${e}`,
-        );
-      }
+      return getCachedOrRun(
+        CommandCacheKey.ResolveLanguages,
+        cmd,
+        () =>
+          runCliJson<ResolveLanguagesOutput>(cmd, [
```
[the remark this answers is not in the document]
mario-campos: Good catch!

It looks like none of the callers of `resolveLanguages` actually use or rely on these filtered extractors (`html`, `csv`, `xml`). So, while we could just always include the flag `--filter-to-languages-with-queries` in the command, not all versions of the CLI would support it. Alas, we would have to raise the `CODEQL_MINIMUM_VERSION` to v2.23.0 (2025-09-04)—which I'm open to doing, but perhaps unnecessary.

Instead, I opted to include the flag if it's supported. Otherwise, don't. This new implementation is spiritually similar to the current implementation, so I don't expect any surprises from it.

henrymercer (commented):
Thanks for implementing the generic cache mechanism, that's a nice simplification!  To make this PR easier to review and get merged, what do you think about splitting this PR into two: the first would implement the generic cache mechanism just for `codeql version`, and the second would use it to cache `codeql resolve languages`?

src/json/index.ts
```diff
@@ -65,6 +85,19 @@ export function optional<T>(validator: Validator<T>) {
   } as const satisfies Validator<T | undefined | null>;
 }
 
+/**
+ * Transforms a validator to be optional, accepting `undefined` for an absent
+ * value but, unlike `optional`, rejecting `null`.
+ */
+export function undefinable<T>(validator: Validator<T>) {
```
mbg: Minor: I think this would be better named `optional` and the existing `optional` would be better named as `optionalOrNull` (or something like that). I probably didn't think about this scenario when I named the existing `optional`.

src/cache.ts
mbg, line 1: We have several different types of caches that are handled by the CodeQL Action, so the name `cache.ts` is a bit too generic for this since it suggests to me that `cache.ts` contains implementations that are shared between all the different types of caches / caching implementations. 

How about `cli-cache.ts` or `cli-config-cache.ts`? Or, alternatively, we could add a new `cli` folder and then move `cache.ts` into it (or call it `config-cache.ts`).

src/json/index.test.ts
```diff
@@ -44,3 +44,26 @@ test("validateSchema - optional properties are optional", async (t) => {
   t.true(json.validateSchema(optionalSchema, { optionalKey: "" }));
   t.true(json.validateSchema(optionalSchema, { optionalKey: "foo" }));
 });
+
+const undefinableSchema = {
+  optionalKey: json.undefinable(json.string),
+};
+
+test("validateSchema - undefinable properties are optional but reject null", async (t) => {
+  // Optional fields may be absent or explicitly undefined
```
mbg: Minor: This comment still refers to "optional" (which is fine if you follow the renaming suggestion in my other comment).

src/config-utils.ts
```diff
@@ -277,9 +275,9 @@ async function getSupportedLanguageMap(
   const supportedLanguages: Record<string, string> = {};
   // Populate canonical language names
   for (const extractor of Object.keys(resolveResult.extractors)) {
-    // If the CLI supports resolving languages with default queries, use these
-    // as the set of supported languages. Otherwise, require the language to be
-    // a built-in language.
+    // TODO: Delete this `if` condition once CODEQL_MINIMUM_VERSION
+    //       is at least v2.23.0 — the first version to support the
+    //       BuiltinExtractorsSpecifyDefaultQueries feature.
```
mbg: TODO?

src/config-utils.ts
```diff
@@ -266,9 +266,7 @@ async function getSupportedLanguageMap(
   const resolveSupportedLanguagesUsingCli = await codeql.supportsFeature(
     ToolsFeature.BuiltinExtractorsSpecifyDefaultQueries,
   );
-  const resolveResult = await codeql.resolveLanguages({
-    filterToLanguagesWithQueries: resolveSupportedLanguagesUsingCli,
-  });
+  const resolveResult = await codeql.resolveLanguages();
```
mbg: This change seems unrelated to the PR?

src/codeql.ts
```diff
@@ -120,9 +126,7 @@ export interface CodeQL {
   /**
    * Run 'codeql resolve languages' with '--format=betterjson'.
    */
-  resolveLanguages(options?: {
-    filterToLanguagesWithQueries: boolean;
-  }): Promise<ResolveLanguagesOutput>;
+  resolveLanguages(): Promise<ResolveLanguagesOutput>;
```
mbg: Unrelated to the PR?

src/codeql.ts
```diff
@@ -230,6 +234,20 @@ export interface VersionInfo {
...
 
+export function isVersionInfo(x: unknown): x is VersionInfo {
+  return (
+    json.isObject(x) &&
+    json.validateSchema(
+      {
+        version: json.string,
+        features: json.undefinable(json.object),
+        overlayVersion: json.undefinable(json.number),
+      },
```
mbg: Minor: This would be better as a named constant than inline here.

src/codeql.ts
```diff
@@ -230,6 +234,20 @@ export interface VersionInfo {
   overlayVersion?: number;
 }
 
+export function isVersionInfo(x: unknown): x is VersionInfo {
```
mbg: JSDoc?

src/codeql.ts
```diff
@@ -246,6 +264,27 @@ export interface ResolveLanguagesOutput {
   };
 }
 
+export function isResolveLanguagesOutput(
```
mbg: JSDoc?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
+import * as fs from "fs";
+import * as path from "path";
+
+import { getTemporaryDirectory } from "./actions-util";
+import * as json from "./json";
+
+/** The name of the temporary file backing the cache (tier 2). */
```
mbg: Minor: "tier 2" is a bit abstract. Something like "[..] the on-disk cache that persists outputs between workflow steps" would be better.

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+const COMMAND_CACHE_FILENAME = "codeql-action-command-cache.json";
+
+/**
+ * The keys under which the output of cached `codeql` commands is stored. Each
+ * key is shared by the producer (the corresponding method in `codeql.ts`) and
+ * any consumers (e.g. `status-report.ts`, which peeks the cached version
+ * without invoking the CLI).
+ */
```
mbg: Minor: The keys just seem to correspond to CLI commands; this explanation about them being shared by "producers" and "consumers" is maybe unnecessary.

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/** A single cached command output together with the CLI path it came from. */
+interface CacheEntry {
+  /**
+   * The path to the CodeQL CLI that produced `output`. Persisted so that a
+   * different step using a different CodeQL bundle doesn't pick up a stale
+   * value.
+   */
+  cmd: string;
+  output: unknown;
```
mbg: Could we type this by having a type parameter on `CacheEntry` and a mapping from `CommandCacheKey` to the respective output types?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * Tier 1: the in-process memo. Consulted first on every lookup and populated
+ * whenever a value is read from the file (tier 2) or computed via the CLI
+ * (tier 3).
+ */
+const inMemoryCache = new Map<string, CacheEntry>();
```
mbg: Why `string` and not `CommandCacheKey`?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+  } catch {
+    return {};
+  }
+  try {
+    const parsed = json.parseString(contents);
+    if (json.isObject(parsed)) {
+      return parsed;
+    }
+  } catch {
+    // Fall through and treat a malformed file as empty.
```
mbg: This case might be worth logging, since we wouldn't expect someone to deliberately create a malformatted file.

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+ * Reads and parses the temporary cache file. Best-effort: a missing, malformed,
+ * or otherwise unreadable file is treated as an empty cache.
+ */
+function readCommandCacheFile(): Record<string, CacheEntry> {
+  let contents: string;
+  try {
+    contents = fs.readFileSync(getCommandCacheFilePath(), "utf8");
+  } catch {
+    return {};
+  }
```
mbg: You could stat the file first to check if it exists, rather than relying on the exception. Then you could treat exceptions (e.g. permissions errors etc.) as exceptional and as actual errors.

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Persists the cache to the temporary file. Best-effort: a failure to write
+ * just means a later step re-runs the CLI.
+ */
+function writeCommandCacheFile(data: Record<string, CacheEntry>): void {
+  try {
+    fs.writeFileSync(getCommandCacheFilePath(), JSON.stringify(data));
+  } catch {
+    // Best-effort; ignore write failures.
```
mbg: Log?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Stores the output of a command under `key`, writing it to both the in-memory
+ * memo (tier 1) and the temporary file (tier 2).
+ *
+ * Should only be called once per key within a single process; doing otherwise
+ * indicates a logic error, since a value that has already been cached should be
+ * served from the memo rather than recomputed.
+ */
+export function cacheCommandOutput(
+  key: string,
```
mbg: `CommandCacheKey`?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+  cmd: string,
+  output: unknown,
+): void {
+  if (inMemoryCache.has(key)) {
+    throw new Error(
+      `cacheCommandOutput() should be called only once per key, but was called more than once for '${key}'.`,
+    );
```
mbg: Should this be a hard error like this? We are already performing the check for whether the key is present in the cache, so we could just return and not do anything. We could log that `cacheCommandOutput()` was called more than once for the same key, but overall this doesn't seem beneficial as an error unless we want to ensure that producers perform the check and never call this function. But since we are performing the check here anyway, there's no real benefit in that.

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+  const entry: CacheEntry = { cmd, output };
+  inMemoryCache.set(key, entry);
+
+  const data = readCommandCacheFile();
+  data[key] = entry;
+  writeCommandCacheFile(data);
```
mbg: This seems a bit wasteful and introduces a good amount of overhead. Why not just ensure that `readCommandCacheFile` is called once (perhaps when `getCodeQLForCmd` is called?) and then call `writeCommandCacheFile` once when the Action is about to exit (perhaps only when it is about to exit successfully?).

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+  validate?: (output: unknown) => output is T,
+): T | undefined {
+  // Tier 1: the in-memory variable.
+  const memoized = inMemoryCache.get(key);
+  if (memoized !== undefined) {
+    return memoized.output as T;
+  }
+
+  // Tier 2: the temporary file persisted by an earlier step, if any.
+  const entry = readCommandCacheFile()[key] as unknown;
```
mbg: Similar comment as above. I think we should use this to (attempt to) restore the in-memory cache once on "startup" and not call it repeatedly.

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+export function getCachedCommandOutput<T>(
+  key: string,
+  cmd?: string,
+  validate?: (output: unknown) => output is T,
+): T | undefined {
+  // Tier 1: the in-memory variable.
+  const memoized = inMemoryCache.get(key);
+  if (memoized !== undefined) {
+    return memoized.output as T;
+  }
```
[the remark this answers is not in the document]
mbg: This seems correct? At least with respect to the `cmd` aspect?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  // Tier 2: the temporary file persisted by an earlier step, if any.
+  const entry = readCommandCacheFile()[key] as unknown;
+  if (
+    !json.isObject<CacheEntry>(entry) ||
+    !json.isString(entry.cmd) ||
+    (cmd !== undefined && entry.cmd !== cmd) ||
+    (validate !== undefined && !validate(entry.output))
+  ) {
+    return undefined;
```
mbg: Should this be an error or logged?

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  // Memoize so subsequent lookups in this process hit tier 1.
+  inMemoryCache.set(key, { cmd: entry.cmd, output: entry.output });
```
mbg: With the current approach where the on-disk file is read/written to repeatedly, shouldn't this be `cacheCommandOutput`? (Not that I suggest we stick with this approach.)

src/cache.ts
```diff
@@ -0,0 +1,147 @@
...
+  return entry.output as T;
+}
+
+/**
+ * Clears the in-process memo (tier 1). Only for use in tests, which exercise
+ * multiple "steps" within a single process.
+ */
+export function resetCachedCommandOutputs(): void {
+  inMemoryCache.clear();
+}
```
mbg: In some other files, we export these test-only functions in a nested object to make it clearer that they are for test-use only. 

That said, my general preference for global state in the CodeQL Action is to try and make it explicit in the way I have done recently in e.g. https://github.com/github/codeql-action/pull/3963

mbg (changes requested):
I haven't finished reviewing everything (particularly the new tests as well as `codeql.ts`), but I already have a bunch of comments that I think should be addressed. So to not end up with too much noise in one go, I am submitting my review without reviewing those last bits.

The main thing to look at is that the on-disk cache is repeatedly written to and read from, which obviously has unnecessary overheads. It would be better if the on-disk cache was only read once (if it exists) to initialise the in-memory cache. It should then only be written to when the Action is about to exit.

Secondly, it would be nice (but is not essential) if we could avoid having the in-memory cache as global state and instead propagate it between the relevant use sites. Since we already have a `CodeQL` object, we could probably make the command cache part of it. That would make the data flow more explicit and simplify the tests, which then don't have to clear the global cache and can additionally run in parallel.

src/config-utils.ts
mario-campos: It's actually a result of this [finding](https://github.com/github/codeql-action/pull/3950#discussion_r3437146339).

src/codeql.ts
mario-campos: Similarly, this is a result of this [finding](https://github.com/github/codeql-action/pull/3950#discussion_r3437146339).

src/config-utils.ts
mario-campos: `TODO` may not be the right word. Really, this comment is meant to serve as a reminder (to the future maintainers) that this part of the code is only necessary for now; so as not to forget to "clean up" these workarounds/fallbacks.

src/cache.ts
mario-campos: Fair enough. If it's all the same, I think I'll go with either `cli-cache.ts` or `cli/cache.ts`.

src/json/index.ts
mario-campos: Done

src/cache.ts
mario-campos: Yeah, good point.

src/cache.ts
mario-campos: Oh, that's a goof. Good catch.

src/cache.ts
mario-campos: Fixed.

src/codeql.ts
mario-campos: Why is that? It would only be used once, here.

src/cache.ts
mario-campos: Fixed

src/cache.ts
mario-campos: Great catch. I wrestled with this one for a minute, but I think it's okay to delete this guard. As you said, we only set the cache when we know something isn't in the cache already. In that case, this check does nothing. And since this only has one call site—a helper function—this behaviour isn't likely to change.

src/cache.ts
mario-campos: Good call. I had this thought too, early on, but must've forgot.

src/cache.ts
mario-campos: Makes sense. I think it's safe to use `cacheCommandOutput` now that it's only backing to the object in memory.

src/cache.ts
mario-campos: I think logging a warning seems more appropriate. That way, the CLI can still be invoked.

src/cache.ts
mbg: One concern I have with both of those (and why I suggested the even more verbose ones) is that we also cache the CLI bundle itself (in the Actions toolcache) so `cli-cache.ts` or `cli/cache.ts` could still be ambiguous in that regard. I'd have a mild preference for something like `cli/config-cache.ts` or `cli/output-cache.ts` as a result. Perhaps if you wanted something shorter, then just `cli/output.ts` would be enough? It doesn't necessarily have to mention the caching aspect explicitly.

src/config-utils.ts
mbg: If I understand the relationship between that finding and this change correctly, then there are two options:

1. We make this change in a separate PR, **if** specifying `filterToLanguagesWithQueries: resolveSupportedLanguagesUsingCli` is no longer needed because we don't support CLI versions that are old enough to require this anymore.
2. We properly support the checks that Copilot mentioned in that finding. (Those can potentially be simplified by including a hash of all relevant factors that affect the cached entry, similarly to what we do with e.g. the Actions caches).

src/codeql.ts
mbg: Like I said, it's a minor point. The two reasons I can think of why it's better as a named (top-level) constant are:

1. We can refer to it in tests, if needed.
2. It makes it more visible if we need to update the schema.

Both are minor reasons, and this is obviously not blocking.

src/cache.ts
mario-campos: Renamed to `cli/output-cache.ts`.

src/cache.ts
mario-campos: Done

src/config-utils.ts
mario-campos: I think (1) is the preferred choice, so I've reverted this hunk.

src/cache.ts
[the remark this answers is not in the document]
mario-campos: Fixed

src/cache.ts
mario-campos: I like the idea of not using global state as a cache, but I wonder if that might be a big shift from the current implementation. And, in the spirit of keeping this PR "small" and tightly-scoped, I think that should be a separate PR.

I'd like to keep this PR focused on generalizing the caching implementation into memory and file backings.

ellisonantoniafelicia-star (approved):

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  /**
+   * The overlay version helps deal with backward incompatible changes for
+   * overlay analysis. When a precompiled query pack reports the same overlay
+   * version as the CodeQL CLI, we can use the CodeQL CLI to perform overlay
+   * analysis with that pack. Otherwise, if the overlay versions are different,
+   * or if either the pack or the CLI does not report an overlay version,
+   * we need to revert to non-overlay analysis.
+   */
+  overlayVersion?: number;
+}
```
mbg, line 32: This type doesn't seem like it really belongs here; i.e. it has nothing to do with the output caching specifically. I assume that you've moved it to avoid a circular dependency. There are two ways of addressing this:

1. (Preferably) Moving this into a location that avoids the cycle. For example, a new `cli/types.ts` file.
2. Importing just `type { VersionInfo } from "../codeql.ts` which (afaik) is also fine.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+        version: json.string,
+        features: {
+          validate: isBooleanRecord,
+          check: (obj) => ({
+            unknownKeys: [],
+            invalidKeys: [],
+            valid: isBooleanRecord(obj),
+          }),
+          required: false,
+        },
```
mbg, line 53: This could probably be turned into a reusable `record` validator in `json/index.ts`. It would also be good to populate `invalidKeys` correctly.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  return output;
+}
+
+/**
+ * Clears the in-process memo (tier 1). Only for use in tests, which exercise
+ * multiple "steps" within a single process.
+ */
+export function resetCachedCommandOutputs(): void {
+  inMemoryCache.clear();
+}
```
mbg, line 238: See my earlier comment about exporting this via an object named to indicate that this is for testing only.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  output: CommandCacheKeyOutputMap[K],
+): void {
+  const entry: CommandCacheEntry<K> = { cmd, output };
+  inMemoryCache.set(key, entry);
+}
+
+/**
+ * Returns the cached output for `key`, or `undefined` if it isn't cached.
+ *
+ * Resolves tier 1 (in-memory memo) first, then tier 2 (temporary file). A value
```
mbg, line 190: Minor: The phrase "in-memory memo" is used a few times. This seems odd to me for two reasons:

1. Memoisation isn't usually abbreviated like that (unless it is a function name or similar where the length becomes a problem) 
2. Memoisation implies "in-memory", so the "in-memory" part is redundant.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+ * values are memoized into tier 1 before being returned.
+ *
+ * A return value of `undefined` signals the caller to fall back to tier 3 (the
+ * CLI).
+ */
+export function getCachedCommandOutput<K extends CommandCacheKey>(
+  key: K,
+  cmd?: string,
+): CommandCacheKeyOutputMap[K] | undefined {
+  // Tier 1: the in-memory variable.
```
mbg, line 202:
```suggestion
  // Try to retrieve the result from memory first.
```

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  const memoized = inMemoryCache.get(key);
+  if (memoized !== undefined) {
+    if (cmd === undefined || memoized.cmd === cmd) {
+      return memoized.output as CommandCacheKeyOutputMap[K];
+    }
+    // If the memoized entry doesn't match the requested CLI,
```
mbg, line 208: What is "doesn't match the requested CLI" supposed to mean?

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+    // treat it as a miss and fall back to tier 2, the file.
+    inMemoryCache.delete(key);
```
mbg, line 210: I don't understand the relationship between this comment and deleting the entry from `inMemoryCache`. A "miss" would usually not involve removing something.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  // Tier 2: the temporary file persisted by an earlier step, if any.
```
mbg, line 213:
```suggestion
  // Try to retrieve the result from disk next.
```

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+    return {};
+  }
+  try {
+    const contents = fs.readFileSync(getCommandCacheFilePath(), "utf8");
+    const parsed = json.parseString(contents);
+    if (json.isObject(parsed)) {
+      return parsed;
+    }
+  } catch (e) {
+    logger.warning(`Failed to read or parse command cache file: ${e}`);
```
mbg, line 155: Since this is not something a user can act on in e.g. Default Setup, and we just fall back to returning `{}`, consider making this an `.info`-level message.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  ) {
+    logger.warning("Received invalid data from the command-cache file.");
+    return undefined;
+  }
+
+  // Memoize so subsequent lookups in this process hit tier 1.
+  const cachedEntry = entry as StoredCommandCacheEntry;
+  const output = cachedEntry.output as CommandCacheKeyOutputMap[K];
+  cacheCommandOutput(key, cachedEntry.cmd, output);
+  return output;
```
mbg, line 229: Why is this on a per-entry basis and not once-per-action?

src/init-action.ts
```diff
@@ -769,6 +774,9 @@ async function run(actionState: ActionState<["Logger", "Env", "Actions"]>) {
 
     core.setOutput("codeql-path", config.codeQLCmd);
     core.setOutput("codeql-version", (await codeql.getVersion()).version);
+
+    // Persist the command cache to disk at the end of a successful init.
+    writeCommandCacheFile();
```
mbg, line 779: Why is this only done for the `init` action? Consider hooking this into `runInActions` in `action-common.ts`.

src/init-action.ts
```diff
@@ -326,6 +327,10 @@ async function run(actionState: ActionState<["Logger", "Env", "Actions"]>) {
     toolsSource = initCodeQLResult.toolsSource;
     zstdAvailability = initCodeQLResult.zstdAvailability;
 
+    // Populate the in-memory command cache with CLI version.
+    // This result will be persisted to disk at the end of the init action.
+    await codeql.getVersion();
```
mbg, line 332: Does this have any benefit over just waiting for the first, existing `getVersion` call?

src/codeql.ts
mbg, line 1: Should `resolveLanguages` use `getCachedOrRun` as well?

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+  aliases?: {
+    [alias: string]: string;
+  };
+  extractors: {
+    [language: string]: Array<{
+      extractor_root: string;
+      extractor_options?: any;
+    }>;
+  };
+}
```
mbg, line 71: Same comment as for `VersionInfo`.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,238 @@
...
+/** A single cached command output together with the CLI path it came from. */
+export interface CommandCacheEntry<K extends CommandCacheKey> {
+  /**
+   * The path to the CodeQL CLI that produced `output`. Persisted so that a
+   * different step using a different CodeQL bundle doesn't pick up a stale
+   * value.
+   */
+  cmd: string;
+  output: CommandCacheKeyOutputMap[K];
+}
```
mbg, line 123: These two types are identical except for the type of `output`. Could you simplify this down by making the type of `output` polymorphic in a type variable? 

If the `K extends CommandCacheKey` constraint is hard to lift, then perhaps just the following might work:

```suggestion
/** A single cached command output together with the CLI path it came from. */
interface BaseCommandCacheEntry<O = unknown> {
  /**
   * The path to the CodeQL CLI that produced `output`. Persisted so that a
   * different step using a different CodeQL bundle doesn't pick up a stale
   * value.
   */
  cmd: string;
  output: O;
}

export type CommandCacheEntry<K extends CommandCacheKey> = BaseCommandCacheEntry<CommandCacheKeyOutputMap[K]>;
```

mbg (changes requested):
Thanks for continuing to work on this, and addressing earlier feedback.

Unfortunately, I still don't think this is ready. There are a few things going on which suggest that the changes aren't finished yet, such as that the caching is not used for `resolveLanguages` and that only the `init` action stores the on-disk cache.

Further, there continue to be two high-level design points that I am concerned about:

1. The cache is only restored from disk on a per-command basis in `getCachedCommandOutput`. Restoring all cache entries from the on-disk cache would make sense. This could reasonably be done in either `runInActions` or the first time that `readCommandCacheFile` is called and successfully reads the on-disk cache. 
2. Cache entries are keyed by the command and CLI path, but not the arguments. This probably doesn't matter for `version` and maybe not even for `resolve languages`, but would make sense to avoid potential surprises now or later. To make this change, entries in the cache shouldn't primarily be indexed by `CommandCacheKey` values, but by keys derived (i.e. hashes computed) from the CLI path, command, arguments, and possibly even other factors such as environment variables. 

A concern is also self-hosted runners, where the on-disk cache may persist between workflows. We should make sure that this either doesn't happen (by explicitly deleting the file in e.g. `init-post`) or storing e.g. the analysis UUID in the file and deleting it if it doesn't match.

I'd also like us to be careful with the new validators you've added. These shouldn't accidentally end up in the path of direct CLI output, because that would make this a high-risk change.

To get on with things and have something that can be merged sooner-than-later, perhaps it would be good to break this down into at least two parts (each as separate PR):

1. An implementation of only "Tier 1" and "Tier 3". I.e. no on-disk cache at all, but addressing my concern about the cache keys.
2. Adding "Tier 2" back in, and addressing my concerns about reading/writing the on-disk cache and potential issues on self-hosted runners.

src/cli/output-cache.ts
mario-campos: It means `memoized.cmd !== cmd`, in that case it deletes the in-memory cache entries and falls back to the next "tier" (file).

src/cli/output-cache.ts
mario-campos: This was a [suggestion from Copilot](https://github.com/github/codeql-action/pull/3950#discussion_r3437146269). As I understand it, it's validating that the cached output we have in memory has come from the same CLI that we expect. Otherwise, if it has come from a different CLI, then we delete what we have in memory, retrieve the cached output from the file, and then store _that_ into the memory cache (towards the end of the function).

src/init-action.ts
mario-campos: My thinking here is that many (most?) of the CLI commands that codeql-action invokes are not worth caching—either because we won't need the output again or because the value of the command is in the side-effect, not the output (e.g. ,`codeql database create`).

Therefore, I think it would make sense to invoke all of the cache-able CLI commands in `init`, either as-needed or pre-emptively. Once that's done, the cache would presumably be set for the life of the workflow run. There would be no need to write the cache to file in `autobuild` or `analyze` or `upload-sarif`.

I could be wrong here, of course, but I cannot think of a CLI command that we would invoke for the first time in those steps that we'd want to cache.

src/init-action.ts
mario-campos: Right now, no.

I was setting this up for a subsequent PR in which the cache is pre-loaded concurrently at the start of `init`, to avoid some of the "waiting" penalty of executing these commands.

src/codeql.ts
mario-campos: Yes, but that's for a follow-up PR, per [@henrymercer's comment](https://github.com/github/codeql-action/pull/3950#pullrequestreview-4533904350).

src/cli/output-cache.ts
mario-campos: Yeah... that's Copilot 😞

src/codeql.ts
mbg: But this PR still includes other `resolveLanguages`-related changes (like moving the output type and adding the validator). Either leave them all for a follow-up PR as per Henry's suggestion, or do it all in this PR.

src/init-action.ts
mbg: > Therefore, I think it would make sense to invoke all of the cache-able CLI commands in `init`, either as-needed or pre-emptively. Once that's done, the cache would presumably be set for the life of the workflow run. There would be no need to write the cache to file in `autobuild` or `analyze` or `upload-sarif`.

I agree with your assessment that probably not all actions run CLI commands that are worth caching. If you don't have this already, it would be good to put together an overview of which CLI commands are used by which action. That can then guide us.

That said, I'd also like to avoid an unintuitive implementation where restoring/storing the on-disk cache is handled in a fairly ad-hoc manner that may not be transparent when making changes down the line.

src/cli/output-cache.ts
mbg: > [..] it's validating that the cached output we have in memory has come from the same CLI that we expect. Otherwise, if it has come from a different CLI, then we delete what we have in memory, retrieve the cached output from the file, and then store _that_ into the memory cache (towards the end of the function).

Copilot's comment (as I indicated in my reply) was correct in that the earlier version of the implementation here ignored `cmd`, even though you explicitly chose to include it in the cached entries. So it makes sense to check that the `cmd` value of the cached entry matches the one we have. Note that this doesn't mean that Copilot's suggested change makes sense.

What my comment here is about is the call to `inMemoryCache.delete(key);`. While that "fixes" the immediate problem of having a cache entry that doesn't match `cmd`, it doesn't use this as an opportunity to reflect on the wider design:

- Why are cache entries only indexed by the command name, even though they depend on other factors? See the high-level point about this in my main review comment.
- The path to the CodeQL CLI should not normally change during an analysis. So why do we have cache entries that have CLI paths that are different from the CLI path we have during the lookup? Perhaps this should be an error or handled differently. I'd like to better understand whether there are genuine use cases (e.g. our own integration tests or the overlay analysis CLI version matching) which make this an expected scenario, or whether this is exceptional and indicative of a problem elsewhere.

src/cli/output-cache.ts
mbg: Sorry if this wasn't clear, but my question wasn't necessarily asking for an explanation, but a suggestion to update this comment. The current wording suggests that you are somehow comparing the entire entry to the "requested CLI". You are actually comparing _the CLI path stored in the entry_ to _the current path of the CLI_ (or similar).

mario-campos:
This PR is too big and vast. Per the recommendations, I've been broken it up into several smaller PRs:

- https://github.com/github/codeql-action/pull/4096
- https://github.com/github/codeql-action/pull/4081
- https://github.com/github/codeql-action/pull/3981
- https://github.com/github/codeql-action/pull/3980
- and a few more that have yet to be created...

PR #3959, opened 2026-06-13
fix: reduce redundant git-call-failed log messages in upload-sarif

### LINKED ISSUE #3383: upload-sarif: `git call failed.` more times than necessary for a working directory without a git repository
https://github.com/github/codeql-action/blob/f67ec12472e1b361f5e1e2085f5e6f85a57e3cd6/src/git-utils.ts#L87

https://github.com/check-spelling-sandbox/skipper/actions/runs/20649925837/job/59292941788

```js
Run github/codeql-action/upload-sarif@7434149006143a4d75b82a2f411ef15b03ccc2d7
git call failed. Continuing with commit SHA from user input or environment. Error: The checkout path provided to the action does not appear to be a git repository.
Uploading code scanning results
  Processing sarif files: ["cifuzz-sarif/results.sarif"]
  Validating cifuzz-sarif/results.sarif
  Adding fingerprints to SARIF file. See https://docs.github.com/en/enterprise-cloud@latest/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#providing-data-to-track-code-scanning-alerts-across-runs for more information.
  (node:12002) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
  (Use `node --trace-deprecation ...` to show where the warning was created)
  git call failed. Continuing with commit SHA from user input or environment. Error: The checkout path provided to the action does not appear to be a git repository.
  git call failed. Continuing with commit SHA from user input or environment. Error: The checkout path provided to the action does not appear to be a git repository.
  git call failed. Will calculate the base branch SHA on the server. Error: The checkout path provided to the action does not appear to be a git repository.
  Uploading results
  Successfully uploaded results
Waiting for processing to finish
git call failed. Continuing with commit SHA from user input or environment. Error: The checkout path provided to the action does not appear to be a git repository.
```

Ideally, the action should be able to "quickly" determine that there isn't a repository checked out and not waste additional time running git commands.

Note that this flow (using google/clusterfuzzlite/actions/run_fuzzers and then ideally uploading sarif output) is a flow that people expect to be able to do (yes, Google is missing documentation for how to do it, but...).

mvanhorn:
When `upload-sarif` runs in a directory that is not a git repository, `runGitCommand` was calling `core.info` on every git call failure. Since callers such as `getCommitOid` and `determineBaseBranchHeadCommitOid` silently swallow the exception and fall back to environment variables, the same "not a git repository" diagnostic was emitted multiple times per run, adding noise without adding value.

This PR changes the log level in `runGitCommand` from `core.info` to `core.debug` so the full diagnostic is preserved for maintainers with debug logging enabled, but does not appear in normal output. A single `core.info` fallback notice is added in `getCommitOid` so callers still see exactly one actionable message when git is unavailable.

Fixes #3383.

### Risk assessment

- **Low risk:** Changes are test-covered, only affect log output levels, and have no impact on action behavior or correctness.

- **Unit tests** - I am depending on unit test coverage (i.e. tests in `.test.ts` files). Existing `determineBaseBranchHeadCommitOid` tests have been updated to assert `core.debug` is called instead of `core.info`. All 30 tests in `src/git-utils.test.ts` pass.

- **Other** - Log level changes are observable; if users report missing diagnostics, debug logging can be enabled.

### DIFF
--- a/src/git-utils.test.ts
+++ b/src/git-utils.test.ts
@@ -252,6 +252,7 @@ test.serial(
   "determineBaseBranchHeadCommitOid not git repository",
   async (t) => {
     const infoStub = sinon.stub(core, "info");
+    const debugStub = sinon.stub(core, "debug");
 
     process.env["GITHUB_EVENT_NAME"] = "pull_request";
     process.env["GITHUB_SHA"] = "100912429fab4cb230e66ffb11e738ac5194e73a";
@@ -260,17 +261,19 @@ test.serial(
       await gitUtils.determineBaseBranchHeadCommitOid(tmpDir);
     });
 
-    t.deepEqual(1, infoStub.callCount);
-    t.deepEqual(
-      infoStub.firstCall.args[0],
-      "git call failed. Will calculate the base branch SHA on the server. Error: " +
-        "The checkout path provided to the action does not appear to be a git repository.",
+    t.deepEqual(0, infoStub.callCount);
+    t.assert(
+      debugStub.calledWithMatch(
+        "git call failed. Will calculate the base branch SHA on the server. Error: " +
+          "The checkout path provided to the action does not appear to be a git repository.",
+      ),
     );
   },
 );
 
 test.serial("determineBaseBranchHeadCommitOid other error", async (t) => {
   const infoStub = sinon.stub(core, "info");
+  const debugStub = sinon.stub(core, "debug");
 
   process.env["GITHUB_EVENT_NAME"] = "pull_request";
   process.env["GITHUB_SHA"] = "100912429fab4cb230e66ffb11e738ac5194e73a";
@@ -278,14 +281,14 @@ test.serial("determineBaseBranchHeadCommitOid other error", async (t) => {
     path.join(__dirname, "../../i-dont-exist"),
   );
   t.deepEqual(result, undefined);
-  t.deepEqual(1, infoStub.callCount);
+  t.deepEqual(0, infoStub.callCount);
   t.assert(
-    infoStub.firstCall.args[0].startsWith(
+    debugStub.calledWithMatch(
       "git call failed. Will calculate the base branch SHA on the server. Error: ",
     ),
   );
   t.assert(
-    !infoStub.firstCall.args[0].endsWith(
+    !debugStub.firstCall.args[0].endsWith(
       "The checkout path provided to the action does not appear to be a git repository.",
     ),
   );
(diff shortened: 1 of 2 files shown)


mbg (commented):
Hi @mvanhorn 👋🏻 

Thanks for opening this PR! I think this is probably one of those innocent-seeming changes that's unfortunately a bit of a rabbit hole.

Typically, we expect that the CodeQL Action is run with reference to some Git repository. In the general case, we therefore want to know when that is not the case and `git` commands fail unexpectedly. The change here mostly just masks these unexpected failures, since `runGitCommand` is used by several callers and the change shifts the responsibility of communicating the failure from `runGitCommand` to the caller. See e.g. the comment in `getGitRoot` which notes that `// Errors are already logged by runGitCommand()` which would no longer be true with this change.

I agree with the [comment in the issue](https://github.com/github/codeql-action/issues/3383#issue-3775565380) that:

> Ideally, the action should be able to "quickly" determine that there isn't a repository checked out and not waste additional time running git commands.

This seems feasible and we could then emit one warning about that and skip subsequent `git` calls. That's a more involved change than this one though.

mvanhorn:
@mbg thanks, that is a fair read and I think you are right.

`runGitCommand` has seven callers in `src/git-utils.ts`: `getGitVersionOrThrow`, `getCommitOid`, `determineBaseBranchHeadCommitOid`, `getGitRoot`, `getFileOidsUnderPath`, `listFiles` and `getGeneratedFiles`. My diff demotes the shared line to `debug` for all seven and only adds a replacement `info` back in `getCommitOid`, so the other six lose their user-visible failure message with nothing taking over. And the comment you pointed at is exactly the tripwire: `// Errors are already logged by runGitCommand()` in `getGitRoot` stops being true the moment that line moves to `debug`. That is masking rather than deduplicating, and I should have caught it.

On the direction you sketched, detecting once that there is no repository checked out, warning once, and skipping the later git calls: that is what I actually wanted, and it treats the cause instead of the symptom. The `not a git repository` condition is already special-cased inside `runGitCommand`, so the signal is there to hang it on.

Two things I would rather have your call on before writing it:

1. Scope. Do you want it in this PR, or should I close this one and open a fresh PR against #3383 so the history reads cleanly?
2. Placement. Should the check live inside `runGitCommand`, so all seven callers inherit it, or one level up in the callers? Inside is less code, but it means `runGitCommand` grows a piece of memoized state, and I would rather not introduce that unless you want it there.

Either is fine by me. And if you would rather keep #3383 for someone on the team, say so and I will close this one without any fuss.

PR #3963, opened 2026-06-16
Add repository property for configuration files

mbg:
This PR is the first in a series of changes which will add the ability to point the CodeQL Action at a configuration file using a repository property.

This PR adds the new repository property and uses its value, if provided and no explicit `config-file` input is provided for the CodeQL Action in the workflow. In other words, the `config-file` input has priority over the value of the repository property.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/config/file.test.ts
+++ b/src/config/file.test.ts
@@ -1,0 +1,26 @@
+import test from "ava";
+import sinon from "sinon";
+
+import { getTestActionsEnv, setupTests } from "../testing-utils";
+
+import { getConfigFileInput } from "./file";
+
+setupTests(test);
+
+test("getConfigFileInput returns undefined by default", async (t) => {
+  const actionsEnv = getTestActionsEnv();
+  const result = getConfigFileInput(actionsEnv);
+  t.is(result, undefined);
+});
+
+test("getConfigFileInput returns input value", async (t) => {
+  const actionsEnv = getTestActionsEnv();
+  const testInput = "/some/path";
+  sinon
+    .stub(actionsEnv, "getOptionalInput")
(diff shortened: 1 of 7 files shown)


src/config/file.test.ts
```diff
@@ -0,0 +1,69 @@
...
+test("getConfigFileInput ignores empty repository property value", async (t) => {
+  const logger = new RecordingLogger();
+  const actionsEnv = getTestActionsEnv();
+
+  // Since the repository property value is an empty/whitespace string, we should ignore it.
+  const result = getConfigFileInput(logger, actionsEnv, {
+    [RepositoryPropertyName.CONFIG_FILE]: "   ",
+  });
+  t.is(result, undefined);
+});
```
mario-campos, line 69: This test doesn't entirely make sense to me. If we expect the repository property to have whitespace, then I think `getConfigFileInput()` should have stronger input validation, and possibly have the whitespace  stripped. That being said, I think we can reasonably expect to not have to deal with whitespace in these properties/inputs—obviating the need for this test.

mario-campos (approved):
LGTM with one minor comment/question.

src/config/file.test.ts
mbg: After initially being hesistant, I added this in response to https://github.com/github/codeql-action/pull/3963#discussion_r3424247085

I also don't really expect that we'd get an empty string from the repo properties API, but I figured it wouldn't hurt to rule out that case in case it is possible (or becomes possible). 

There is downstream validation from the result of `getConfigFileInput()` and we will also add the shorthand stuff in the next PR, so I think this is OK as-is for now.

PR #3971, opened 2026-06-23
Pin Swift macOS PR checks to macOS 15 for Xcode 16

henrymercer:
The `macos-latest-xlarge` label has started resolving to **macOS 26**, which ships only Xcode 26 (Swift 6.2) and no longer includes Xcode 16. Our Swift-analysing PR checks explicitly select Xcode 16 (`xcode-select -s /Applications/Xcode_16.app`) because CodeQL CLI versions before 2.24.0 only support Swift up to 6.1. That step now fails on macOS 26, which is why the *Multi-language repository* check is red.

We still want to exercise the full matrix of supported CodeQL versions (back to 2.19.4) against Swift, so we can't simply move to Xcode 26. Instead, this pins the affected checks to macOS 15, where Xcode 16 is still available:

- `multi-language-autodetect`: `macos-latest-xlarge` -> `macos-15-xlarge`

This restores the runner that `macos-latest-xlarge` pointed at before the migration, so it is a known-good configuration. The generated `__*.yml` workflows were regenerated via `npm run update-pr-checks`. Once the minimum supported CodeQL version reaches 2.24.0, these checks can drop the Xcode 16 pin and move back to `macos-latest`.

The nightly-only Swift checks (`swift-autobuild`, `export-file-baseline-information`) are intentionally left on macOS 26, since nightly CodeQL supports Swift 6.2 with the default Xcode.

See https://github.com/actions/runner-images/issues/14167.

### Risk assessment

Environments:

- **Other** - The affected PR checks will fail in CI.

- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change. N/A - CI/test-only change with no user-facing impact.
- Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary. N/A

### DIFF
--- a/pr-checks/checks/multi-language-autodetect.yml
+++ b/pr-checks/checks/multi-language-autodetect.yml
@@ -2,16 +2,8 @@ name: "Multi-language repository"
 description: "An end-to-end integration test of a multi-language repository using automatic language detection"
 operatingSystems:
   - ubuntu
-  # Newer CodeQL CLI versions support Swift 6.2, so analyse Swift on the latest macOS runner
-  # (currently macOS 26) with its default Xcode. This exercises the common combination of a
-  # recent CLI on a recent runner.
   - os: macos
     runner-image: macos-latest-xlarge
-    codeql-versions:
-      - stable-v2.24.3
-      - default
-      - linked
-      - nightly-latest
   # Older CodeQL CLI versions only support Swift up to 6.1, which requires Xcode 16. That is
   # not available on macOS 26, so run these versions on macOS 15 where we select Xcode 16
   # below. See https://github.com/actions/runner-images/issues/14167.
@@ -22,7 +14,6 @@ operatingSystems:
       - stable-v2.20.7
       - stable-v2.21.4
       - stable-v2.22.4
-      - stable-v2.23.9
 env:
   CODEQL_ACTION_RESOLVE_SUPPORTED_LANGUAGES_USING_CLI: true
 installGo: true
--- a/pr-checks/checks/multi-language-autodetect.yml
+++ b/pr-checks/checks/multi-language-autodetect.yml
@@ -2,11 +2,27 @@ name: "Multi-language repository"
 description: "An end-to-end integration test of a multi-language repository using automatic language detection"
 operatingSystems:
   - ubuntu
-  # Pin to macOS 15 rather than `macos-latest`: the older CodeQL CLI versions in the
-  # matrix only support Swift up to 6.1 (Xcode 16), which is not available on macOS 26
-  # (ships Xcode 26 / Swift 6.2). See https://github.com/actions/runner-images/issues/14167.
+  # Newer CodeQL CLI versions support Swift 6.2, so analyse Swift on the latest macOS runner
+  # (currently macOS 26) with its default Xcode. This exercises the common combination of a
+  # recent CLI on a recent runner.
+  - os: macos
+    runner-image: macos-latest-xlarge
+    codeql-versions:
+      - stable-v2.24.3
+      - default
+      - linked
+      - nightly-latest
+  # Older CodeQL CLI versions only support Swift up to 6.1, which requires Xcode 16. That is
+  # not available on macOS 26, so run these versions on macOS 15 where we select Xcode 16
+  # below. See https://github.com/actions/runner-images/issues/14167.
   - os: macos
     runner-image: macos-15-xlarge
+    codeql-versions:
+      - stable-v2.19.4
+      - stable-v2.20.7
+      - stable-v2.21.4
+      - stable-v2.22.4
+      - stable-v2.23.9
 env:
   CODEQL_ACTION_RESOLVE_SUPPORTED_LANGUAGES_USING_CLI: true
 installGo: true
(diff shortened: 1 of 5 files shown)


pr-checks/checks/swift-custom-build.yml
```diff
@@ -5,7 +5,11 @@ versions:
   - default
   - nightly-latest
 operatingSystems:
-  - macos
+  # Pin to macOS 15 rather than `macos-latest`: the `linked`/`default` CodeQL CLI versions
+  # are analysed with Xcode 16, which is not available on macOS 26 (ships Xcode 26 / Swift
+  # 6.2). See https://github.com/actions/runner-images/issues/14167.
+  - os: macos
+    runner-image: macos-15
```
mbg: As mentioned, I don't think this is necessary since this workflow can now succeed without the Xcode 16 step.

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -2,8 +2,11 @@ name: "Multi-language repository"
 description: "An end-to-end integration test of a multi-language repository using automatic language detection"
 operatingSystems:
   - ubuntu
+  # Pin to macOS 15 rather than `macos-latest`: the older CodeQL CLI versions in the
+  # matrix only support Swift up to 6.1 (Xcode 16), which is not available on macOS 26
+  # (ships Xcode 26 / Swift 6.2). See https://github.com/actions/runner-images/issues/14167.
   - os: macos
-    runner-image: macos-latest-xlarge
+    runner-image: macos-15-xlarge
```
mbg, line 11: Ah, this is what you had in mind. My concern here would be that this means we run all of the `macos` tests on `macos-15` as a result, even though we only need it for a couple of old CLI versions. Consequently, we are then no longer testing newer versions on the `-latest` runners, which is by far the more common combination (latest CLI + latest runner).

henrymercer: Addressed with a small extension to the workflow generator.

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -2,8 +2,27 @@ name: "Multi-language repository"
...
+  # Newer CodeQL CLI versions support Swift 6.2, so analyse Swift on the latest macOS runner
+  # (currently macOS 26) with its default Xcode. This exercises the common combination of a
+  # recent CLI on a recent runner.
   - os: macos
     runner-image: macos-latest-xlarge
+    codeql-versions:
+      - stable-v2.24.3
+      - default
+      - linked
+      - nightly-latest
```
mbg: Minor: It might be annoying to keep this list updated. It would be nice if we could just specify which `codeql-versions` we want to use `macos-15-xlarge` for and default to `macos-latest-xlarge` for all others.

pr-checks/checks/multi-language-autodetect.yml
```diff
@@ -2,8 +2,27 @@ name: "Multi-language repository"
...
+  # not available on macOS 26, so run these versions on macOS 15 where we select Xcode 16
+  # below. See https://github.com/actions/runner-images/issues/14167.
+  - os: macos
+    runner-image: macos-15-xlarge
+    codeql-versions:
+      - stable-v2.19.4
+      - stable-v2.20.7
+      - stable-v2.21.4
+      - stable-v2.22.4
+      - stable-v2.23.9
```
mbg: In [my testing](https://github.com/github/codeql-action/actions/runs/28032302145/job/82975993686?pr=3970), `v2.19.4`, `v2.20.7`, and `v2.22.4` failed. `v2.21.4` got lucky with an older runner image, and should have failed as well. However, `v2.23.9` got the new runner image and succeeded - should it be in this list?

mbg (dismissed):
Thanks for addressing those comments! This looks OK to merge, but with one maintenance concern and one question about the included CodeQL versions.

pr-checks/checks/multi-language-autodetect.yml
henrymercer: I think it wasn't officially supported at that point, but I don't think it matters hugely either way.

pr-checks/checks/multi-language-autodetect.yml
henrymercer: Sure, it makes the workflow generator a bit more complicated but perhaps that's worth it.

pr-checks/sync.ts
```diff
@@ -358,6 +359,28 @@ function generateJobMatrix(
 ): Array<Record<string, any>> {
   let matrix: Array<Record<string, any>> = [];
 
+  const operatingSystems = checkSpecification.operatingSystems ?? ["ubuntu"];
+
+  // For each OS, collect the CodeQL versions explicitly claimed by entries that specify
+  // `codeql-versions`. A sibling entry for the same OS that omits `codeql-versions` runs all
+  // versions not in this set.
+  const claimedVersionsByOs = new Map<string, Set<string>>();
```
mbg, line 367: Minor: This mapping is indexed by the `.os` key, which is shared among all siblings. The "claimedVersionsByOs" naming here is therefore a bit confusing because it seems to say something about which OS has claimed which versions. Changing this to e.g. `handledVersionsByOS` or similar might be slightly clearer.

mbg (approved):
Looks good, thanks! Nothing blocking here, just one comment that resulted from my experience reviewing the most recent change. Feel free to merge as-is.

pr-checks/sync.ts
henrymercer: I'll leave this as is so we can get the checks passing on main 👍

PR #3973, opened 2026-06-23
Support shorthand forms for `getRemoteConfig`

mbg:
This is a follow-up to #3963 which modifies `getRemoteConfig` to accept shorthand addresses for remote files. Concretely, this adds a new format (gated behind a FF) where the `owner`, `path`, and `ref` components are optional and default to the owner of the current repo (as given by `GITHUB_REPOSITORY`), `.github/codeql-action.yaml`, and `main` respectively.

Examples:

- `owner/repo` resolves to `owner/repo@main:.github/codeql-action.yaml`
- `owner/repo@dev` resolves to `owner/repo@dev:.github/codeql-action.yaml`
- `owner/repo:.github/codeql/config.yml` resolves to `owner/repo@main:.github/codeql/config.yml`

**Notes for reviewers**

- We should discuss whether there's a better default we want for the `path` component
- The default ref of `main` may not always be the default branch for the target repo, but that's OK since it can be manually set to a different ref if needed.
- ~~I originally implemented the shorthand forms from my proposal, but then noticed that `getRemoteConfig` already had a similar (but not equivalent) shorthand format. I then updated that instead so that the components other than the repo are optional. However, this isn't as nice, because `/` is used as a separator and can also appear in the `path` component. That means e.g. it is not possible with this format to specify a repo + path since it would be interpreted as owner + repo by the regex instead.~~ 

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/action-common.ts
+++ b/src/action-common.ts
@@ -2,13 +2,36 @@ import { Env } from "./environment";
 import { FeatureEnablement } from "./feature-flags";
 import { Logger } from "./logging";
 
-export interface ActionState {
-  /** The logger that is in use. */
-  logger: Logger;
-
-  /** Information about environment variables. */
-  env: Env;
-
-  /** Information about enabled feature flags. */
-  features: FeatureEnablement;
+/** Describes different state features that an Action may have. */
+export interface FeatureState {
+  Logger: {
+    /** The logger that is in use. */
+    logger: Logger;
+  };
+  Env: {
+    /** Information about environment variables. */
+    env: Env;
+  };
+  FeatureFlags: {
+    /** Information about enabled feature flags. */
+    features: FeatureEnablement;
+  };
 }
+
+/** Identifies a type of state an Action may have. */
+export type StateFeature = keyof FeatureState;
+
+/** Constructs the union of all state types identifies by `Fs`. */
+export type FieldsOf<Fs extends readonly StateFeature[]> = Fs extends [
+  infer Head extends StateFeature,
+]
+  ? FeatureState[Head]
+  : Fs extends [
+        infer Head extends StateFeature,
+        ...infer Tail extends StateFeature[],
+      ]
+    ? FeatureState[Head] & FieldsOf<Tail>
+    : never;
+
+/** Describes the state of an Action that has access to the state corresponding to `Fs`. */
+export type ActionState<Fs extends readonly StateFeature[]> = FieldsOf<Fs>;
--- a/src/action-common.ts
+++ b/src/action-common.ts
@@ -1,0 +1,14 @@
+import { Env } from "./environment";
+import { FeatureEnablement } from "./feature-flags";
+import { Logger } from "./logging";
+
+export interface ActionState {
+  /** The logger that is in use. */
+  logger: Logger;
+
+  /** Information about environment variables. */
+  env: Env;
+
+  /** Information about enabled feature flags. */
+  features: FeatureEnablement;
+}
--- a/src/action-common.ts
+++ b/src/action-common.ts
@@ -1,6 +1,23 @@
+import * as core from "@actions/core";
+
+import { ActionsEnv, getActionsEnv } from "./actions-util";
 import { Env } from "./environment";
 import { FeatureEnablement } from "./feature-flags";
-import { Logger } from "./logging";
+import { getActionsLogger, Logger } from "./logging";
+import {
+  ActionName,
+  getDisplayActionName,
+  sendUnhandledErrorStatusReport,
+} from "./status-report";
+import { getEnv, getErrorMessage } from "./util";
+
+/** Common state that is always available in `ActionState`. */
+export interface BaseState {
+  /** The name of the Action. */
+  name: ActionName;
+  /** When the Action was started. */
+  startedAt: Date;
+}
 
 /** Describes different state features that an Action may have. */
 export interface FeatureState {
@@ -12,6 +29,10 @@ export interface FeatureState {
     /** Information about environment variables. */
     env: Env;
   };
+  Actions: {
+    /** Access to Actions-related functionality. */
+    actions: ActionsEnv;
+  };
   FeatureFlags: {
     /** Information about enabled feature flags. */
     features: FeatureEnablement;
@@ -21,17 +42,54 @@ export interface FeatureState {
 /** Identifies a type of state an Action may have. */
 export type StateFeature = keyof FeatureState;
 
-/** Constructs the union of all state types identifies by `Fs`. */
-export type FieldsOf<Fs extends readonly StateFeature[]> = Fs extends [
-  infer Head extends StateFeature,
-]
-  ? FeatureState[Head]
+/** Constructs the intersection of all state types identifies by `Fs`. */
+export type FieldsOf<Fs extends readonly StateFeature[]> = Fs extends []
+  ? BaseState
   : Fs extends [
         infer Head extends StateFeature,
-        ...infer Tail extends StateFeature[],
+        ...infer Tail extends readonly StateFeature[],
       ]
     ? FeatureState[Head] & FieldsOf<Tail>
     : never;
 
 /** Describes the state of an Action that has access to the state corresponding to `Fs`. */
 export type ActionState<Fs extends readonly StateFeature[]> = FieldsOf<Fs>;
+
+/** The type of an Action's main entry point. This is a function that is provided
+ * with a basic `ActionState` object with features that are always available.
+ * Each Action can then augment the `state` further if additional features are required.
+ */
+export type ActionMain = (
+  state: ActionState<["Logger", "Env", "Actions"]>,
+) => Promise<void>;
+
+/** A specification for a CodeQL Action step. */
+export interface Action {
+  /** The name of the Action. */
+  name: ActionName;
+  /** The entry point for the Action. */
+  run: ActionMain;
+}
+
+/** A generic entry point that sets up the basic environment for the `action` and runs it. */
+export async function runInActions(action: Action) {
+  const startedAt = new Date();
+  const logger = getActionsLogger();
+  const env = getEnv();
+  const actionsEnv = getActionsEnv();
+
+  try {
+    await action.run({
+      name: action.name,
+      startedAt,
+      logger,
+      env,
+      actions: actionsEnv,
+    });
+  } catch (error) {
+    core.setFailed(
+      `${getDisplayActionName(action.name)} action failed: ${getErrorMessage(error)}`,
+    );
+    await sendUnhandledErrorStatusReport(action.name, startedAt, error, logger);
+  }
+}
--- a/src/config/file.ts
+++ b/src/config/file.ts
@@ -6,7 +6,7 @@ import {
   RepositoryPropertyName,
 } from "../feature-flags/properties";
 import { Logger } from "../logging";
-import { ConfigurationError } from "../util";
+import { ConfigurationError, getEnv } from "../util";
 
 import { parseUserConfig, UserConfig } from "./db-config";
 import { parseRemoteFileAddress } from "./remote-file";
@@ -55,7 +55,7 @@ export async function getRemoteConfig(
   apiDetails: api.GitHubApiCombinedDetails,
   validateConfig: boolean,
 ): Promise<UserConfig> {
-  const address = parseRemoteFileAddress(configFile);
+  const address = parseRemoteFileAddress(getEnv(), configFile);
 
   const response = await api
     .getApiClientWithExternalAuth(apiDetails)
--- a/src/config/file.ts
+++ b/src/config/file.ts
@@ -1,5 +1,4 @@
 import { ActionState } from "../action-common";
-import { ActionsEnv } from "../actions-util";
 import * as api from "../api-client";
 import * as errorMessages from "../error-messages";
 import { Feature } from "../feature-flags";
@@ -7,7 +6,6 @@ import {
   RepositoryProperties,
   RepositoryPropertyName,
 } from "../feature-flags/properties";
-import { Logger } from "../logging";
 import { ConfigurationError } from "../util";
 
 import { parseUserConfig, UserConfig } from "./db-config";
@@ -16,11 +14,14 @@ import { parseRemoteFileAddress } from "./remote-file";
 /**
  * Gets the value that is configured for the configuration file, if any.
  */
-export function getConfigFileInput(
-  logger: Logger,
-  actions: ActionsEnv,
+export async function getConfigFileInput(
+  {
+    logger,
+    actions,
+    features,
+  }: ActionState<["Logger", "Actions", "FeatureFlags"]>,
   repositoryProperties: Partial<RepositoryProperties>,
-): string | undefined {
+): Promise<string | undefined> {
   const input = actions.getOptionalInput("config-file");
 
   if (input !== undefined) {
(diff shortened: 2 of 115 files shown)


src/config/remote-file.ts
```diff
@@ -0,0 +1,78 @@
...
+  const currentRepoNwo = env.getRequired(ActionsEnvVars.GITHUB_REPOSITORY);
+  const nwoParts = currentRepoNwo.split("/");
+
+  if (nwoParts.length !== 2 || nwoParts[0].trim().length === 0) {
+    // This shouldn't happen, so we should throw if `GITHUB_REPOSITORY` doesn't match
+    // our expectations.
+    throw new Error(
+      `Expected ${ActionsEnvVars.GITHUB_REPOSITORY} to contain a name with owner, but got '${currentRepoNwo}'.`,
+    );
+  }
```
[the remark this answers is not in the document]
mbg: I think this is OK, since we only care about the owner and this is unlikely to ever be the case or cause problems.

src/config/remote-file.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export function parseRemoteFileAddress(
+  env: Env,
+  configFile: string,
+): RemoteFileAddress {
+  // retrieve the various parts of the config location, and ensure they're present
+  const format = new RegExp(
+    "((?<owner>[^/]+)/)?(?<repo>[^/@]+)(/(?<path>[^@]+))?(@(?<ref>.*))?",
+  );
+  const pieces = format.exec(configFile);
```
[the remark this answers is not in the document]
mbg: The regex was already not anchored previously, so this is not new in this PR, but I agree and have changed this.

src/config/remote-file.ts
```diff
@@ -0,0 +1,84 @@
...
+  const currentRepoNwo = env.getRequired(ActionsEnvVars.GITHUB_REPOSITORY);
+  const nwoParts = currentRepoNwo.split("/");
+
+  if (nwoParts.length !== 2 || nwoParts[0].trim().length === 0) {
+    // This shouldn't happen, so we should throw if `GITHUB_REPOSITORY` doesn't match
+    // our expectations.
+    throw new Error(
+      `Expected ${ActionsEnvVars.GITHUB_REPOSITORY} to contain a name with owner, but got '${currentRepoNwo}'.`,
+    );
+  }
```
mario-campos, line 37: Nit: if it shouldn't happen, should we check for it?

src/config/remote-file.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    ref: "ref",
+  } satisfies RemoteFileAddress);
+
+  t.deepEqual(parseRemoteFileAddress(env, "repo"), {
+    owner,
+    repo: "repo",
+    path: DEFAULT_CONFIG_FILE_NAME,
+    ref: DEFAULT_CONFIG_FILE_REF,
+  } satisfies RemoteFileAddress);
+});
```
mario-campos, line 158: What about test cases for `repo/path` and `repo/path@ref`?

src/config/remote-file.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+test("parseRemoteFileAddress throws for invalid `GITHUB_REPOSITORY`", async (t) => {
+  const env = getTestEnv();
+  const getRequired = sinon.stub(env, "getRequired");
+  getRequired.withArgs(ActionsEnvVars.GITHUB_REPOSITORY).returns(`not-valid`);
+
+  t.throws(() => parseRemoteFileAddress(env, "repo@ref"), {
+    instanceOf: Error,
+  });
+});
```
mario-campos, line 173: Similarly to my comment above, I think this is overkill, but also know that it's a nitpick.

src/config/remote-file.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  const env = getTestEnv();
+
+  t.deepEqual(parseRemoteFileAddress(env, "owner/repo/path"), {
+    owner: "owner",
+    repo: "repo",
+    path: "path",
+    ref: DEFAULT_CONFIG_FILE_REF,
+  } satisfies RemoteFileAddress);
+
+  t.deepEqual(parseRemoteFileAddress(env, "owner/repo/path@"), {
```
mario-campos: I think this should be invalid, for the same reason that `/repo@ref` is invalid.

Edit: Also, I just noticed the expected format below does not seem to indicate that this should be a correct format:

```
Expected format [<owner>/]<repository>[/<file-path>][@<ref>]
```

src/config/file.ts
```diff
@@ -32,3 +38,51 @@ export function getConfigFileInput(
...
+  } else {
+    throw new ConfigurationError(
+      errorMessages.getConfigFileFormatInvalidMessage(configFile),
+    );
+  }
+
+  return parseUserConfig(
+    logger,
+    configFile,
+    Buffer.from(fileContents, "base64").toString("binary"),
```
mario-campos, line 100:
```suggestion
    Buffer.from(fileContents, "base64").toString("utf8"),
```

UTF-8?

mario-campos (dismissed):
Looks solid! I left a few comments, but minor stuff.

src/config/file.ts
mbg: This isn't new code in this PR. I just moved it from another file. I am not sure why it uses `binary` rather than `utf8` since I didn't write the code, but AFAIK we haven't run into an issue with this yet, so I am hesitant to change it here without understanding why it is like this to begin with. 

I'd say we leave it like this here, but we can revisit it in a follow-up and either change it to `utf8` if we think that it's unlikely to break something or add a FF-gated change that we can roll out.

src/config/remote-file.test.ts
mbg: I wouldn't have added it if I didn't know that CCR wouldn't complain about a missing test case otherwise.

src/config/remote-file.test.ts
mbg: As per the PR description, that doesn't work with the existing format because `repo/path` is indistinguishable from `owner/repo`.

src/config/remote-file.ts
mbg: Yes, for the following reasons:

1. There's no obvious default value we can return if our expectations aren't met.
2. We'd end up with a less clear error if we don't check our assumptions at all.
3. Although it shouldn't happen in Actions workflows based on the current Actions behaviour, it doesn't mean that the behaviour won't change in the future.
4. Although this shouldn't happen in Actions workflows, it could happen in tests if we don't set them up right. So there it's more useful to have the meaningful error in that case.

src/config/remote-file.test.ts
mbg: I have no strong feelings on this, but agree that we should be consistent. I can change this to also be invalid.

src/config/remote-file.test.ts
```diff
@@ -171,10 +252,43 @@ test("parseRemoteFileAddress rejects invalid values", async (t) => {
...
+  });
+  t.throws(() => parseRemoteFileAddress(env, ":path"), {
+    instanceOf: ConfigurationError,
+  });
+  t.throws(() => parseRemoteFileAddress(env, "@ref"), {
+    instanceOf: ConfigurationError,
+  });
+  t.throws(() => parseRemoteFileAddress(env, "@ref:path"), {
+    instanceOf: ConfigurationError,
+  });
```
mario-campos: Aren't all components optional? If so, then this should be a valid, supported format.

src/config/remote-file.ts
mario-campos, line 131: If you don't mind losing the clear error message, this could be implemented in the regular-expression pattern with a negative lookahead: `(:(?!/)(?<path>.+))`

src/action-common.ts
```diff
@@ -0,0 +1,14 @@
+import { Env } from "./environment";
+import { FeatureEnablement } from "./feature-flags";
+import { Logger } from "./logging";
+
+export interface ActionState {
```
mario-campos: A JSDoc explaining the purpose of `ActionState` would be helpful here.

src/config/remote-file.test.ts
```diff
@@ -12,230 +14,220 @@ import {
...
-    parseRemoteFileAddress(env, "owner/repo@ref/feature:path/to/codeql.yml"),
+  // New format.
+  const newFormatInputs: ParseRemoteFileAddressTest[] = [
+    { input: "owner/repo@ref:path", expected },
+    { input: "owner  /repo@ref:path", expected },
+    { input: "owner/   repo@ref:path", expected },
+    { input: "owner/repo   @ref:path", expected },
+    { input: "owner/repo@   ref:path", expected },
+    { input: "owner/repo@ref   :path", expected },
+    { input: "owner/repo@ref:   path", expected },
```
mario-campos, line 64: Whitespace is significant in a file path. How do you know that `   path` is a typo, and that the user—for whatever reason—didn't actually intend `   path`?

But, taking a step back, I don't think we should necessarily support (or expect) whitespace in the address. I think it's perfectly acceptable to treat a typo/mistake as an error, if it is, in fact, a mistake. That would allow us to treat `owner   `, `   repo`, `ref    ` literally and simplify the logic a bit.

src/config/remote-file.ts
mario-campos, line 62: `pieces.length < 5` is a useless condition, because it will always be false. The regular expression pattern will either match with 4 groups + the whole expression, or it will not match at all.

In fact, `pieces.length` will also never be greater than 5.

src/config/remote-file.ts
```diff
@@ -89,6 +91,16 @@ export function parseRemoteFileAddress(
...
+  );
+  if (!allowNewFormat) {
+    throw new ConfigurationError(
+      errorMessages.getConfigFileRepoOldFormatInvalidMessage(configFile),
+    );
+  }
+
   // retrieve the various parts of the config location, and ensure they're present
   const format = new RegExp(
     "^((?<owner>[^:@/]+)/)?(?<repo>[^:@/]+)(@(?<ref>[^:]+))?(:(?<path>.+))?$",
```
mario-campos, line 106: - Since `/` will always follow the owner, I think it would be simpler and less error-prone to make the character class `[^/]`.
- Likewise, I think the character class for the `repo` group could be `[^:@]`, since `/` is used in the old format, not this new one.

src/config/remote-file.ts
mario-campos, line 113: Regarding `!repo`, I am not sure how the JS regular-expression engine works with regards to `undefined`, but since the quantifier is `+`, I expect `repo` to always be a one character string, at least. So, it seems unnecessary 🤷 

Similarly, `repo.length` should always be at least 1 (because of the `+` quantifier). So, this seems like a useless condition.

mario-campos, line 47: 1. Since all components are required, the `ref` group should have the `+` quantifier. The `*` quantifier can consume zero characters, which can mean that the `ref` group is the empty string, which would need to be handled in a special case.
2. This pattern should be anchored to match the whole string.

mario-campos (changes requested):
I left a few comments, most of them minor/simple.

If you're wondering about the _Outdated_ label on the comments... I commented on the individual commits, because I reviewed them chronologically.

src/config/remote-file.ts
mbg: What you wrote mirrors what Copilot complained about originally when I updated the regex to make parts of it optional. However, now that I have added a new regex for the new format, `OLD_REMOTE_ADDRESS_FORMAT` is just the regex from the existing code before any changes in this PR. In other words, making these adjustments (while perhaps reasonable) would alter the original behaviour. The old behaviour is not FF-gated, and so we'd be making a breaking change to existing code. That seems unnecessary to me.

src/config/remote-file.test.ts
mbg: All components except the `repo` are optional. These tests are a combination of "no repo" and the sort of thing you previously complained about in https://github.com/github/codeql-action/pull/3973#discussion_r3468738979

src/config/remote-file.ts
mbg: Personally, I don't like using regular expressions to perform complex parsing tasks. This approach is more readable/maintainable. I'd consider adding the regular error message on to this one if you think that's beneficial, but I am not sure it's necessary since the regex already matched (meaning the overall format was correct) and this error is more specific than the overall format.

src/action-common.ts
mbg: Done in the commit I pushed just after you posted your review, I think.

src/config/remote-file.test.ts
mbg: > Whitespace is significant in a file path.

The path is used in `getRemoteConfig` as argument to the `path` parameter for the `rest.repos.getContent` call. I am not sure that leading or trailing whitespace would be meaningful there, or how happy `git`/`GitHub` is with repos containing whitespace at the start / end of file paths. If that is supported, then it would of course make sense to keep it.

src/config/remote-file.ts
mbg: This is a check kept from the previous implementation and not new in this PR.

src/config/remote-file.ts
mbg: Changing the regex as suggested leads to potentially incorrect results, because the groups end up being too eager in what they accept. For example, `foo:some/path/to/a/file.yml` should use `foo` as the `repo` and `some/path/to/a/file.yml` as the `path`. What would happen in practice (with your suggestion) is that `foo:some` ends up being the `owner` (`owner` reads until it finds a `/`) and `path/to/a/file.yml` ends up as the `repo` (`repo` reads until it finds a `:` or `@`, neither of which exist). That satisfies the regex, but is unlikely what was intended.

src/config/remote-file.ts
mbg: There are two separate things going on here:

1. The `!repo` check is there to make TypeScript happy, since it does't know that `!pieces?.groups` implies that `repo !== undefined`.
2. The `repo.length` check follows from the call to `.trim()` in the initial assignment of `repo`. Since, in theory, the regex could accept a whitespace-only `repo` value, `.trim()` may then reduce it to `0` length, which we check for here to reject it. That could be captured in the regex itself, but would then require us to specify which character ranges are acceptable (and maintain it).

src/config/remote-file.test.ts
mario-campos: >I am not sure that leading or trailing whitespace would be meaningful there, or how happy git/GitHub is with repos containing whitespace at the start / end of file paths.

I would expect `git`/GitHub to do its own input validation. And error appropriately if whitespace is invalid.

src/config/remote-file.ts
mario-campos: That's fair. It's not always the best trade-off.

src/config/remote-file.ts
```diff
@@ -0,0 +1,139 @@
...
+  return nwoParts[0].trim();
+}
+
+/**
+ * The old remote address format that's always been supported for the `config-file` input.
+ * All the components are required. Unchanged from the previous implementation.
+ */
+const OLD_REMOTE_ADDRESS_FORMAT = new RegExp(
+  "(?<owner>[^/]+)/(?<repo>[^/]+)/(?<path>[^@]+)@(?<ref>.*)",
+);
```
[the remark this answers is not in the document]
mbg: Didn't I already dismiss this comment before?

src/config/remote-file.ts
mario-campos: You're right! My bad.

mario-campos (approved):

PR #3979, opened 2026-06-26
Measure overlay-base database size at the clear cleanup level

henrymercer:
This lets us compare the storage cost of overlay-base and trimmed databases for the same commit.

This only happens when the feature flag for uploading overlay DBs to the GitHub API is enabled (currently staff only) and when running on the default branch.  There is some performance penalty to taking this measurement, but since it's on the default branch, this seems justifiable for the time period that we are rolling out uploading overlay DBs to the GitHub API.  We could also just compare the zipped DB size with and without the feature flag, but this change has the benefit of giving us the numbers on the same commit, which improves data quality.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/database-upload.ts
+++ b/src/database-upload.ts
@@ -189,8 +189,8 @@ export async function cleanupAndUploadDatabases(
 
 /**
  * Cleans up the databases at the `clear` cleanup level and records the resulting zipped size for
- * each language in `clear_cleanup_zipped_size_bytes`, along with the time spent taking the
- * measurement in `clear_cleanup_measurement_duration_ms`.
+ * each language in `clear_cleanup_zipped_size_bytes`. If the cleanup succeeds, also records the
+ * time spent taking the measurement in `clear_cleanup_measurement_duration_ms`.
  *
  * This mutates the entries of `reports` in place. It must run only after all overlay-base uploads
  * have completed, since the `clear` cleanup discards overlay data that the uploaded database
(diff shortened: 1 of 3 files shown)


src/database-upload.ts
```diff
@@ -156,9 +169,87 @@ export async function cleanupAndUploadDatabases(
...
 }
 
+/**
+ * Cleans up the databases at the `clear` cleanup level and records the resulting zipped size for
+ * each language in `clear_cleanup_zipped_size_bytes`. If the cleanup succeeds, also records the
+ * time spent taking the measurement in `clear_cleanup_measurement_duration_ms`.
+ *
+ * This mutates the entries of `reports` in place. It must run only after all overlay-base uploads
+ * have completed, since the `clear` cleanup discards overlay data that the uploaded database
+ * depends on.
```
mbg, line 197: Minor: Generally, I have a mild preference for not mutating state and returning new values instead since it simplifies data flow and testing.

src/database-upload.ts
```diff
@@ -156,9 +169,87 @@ export async function cleanupAndUploadDatabases(
...
+
+  for (const language of config.languages) {
+    const report = reports.find((r) => r.language === language);
+    if (report === undefined) {
+      continue;
+    }
+    try {
+      const bundledDb = await bundleDb(config, language, codeql, language, {
+        includeDiagnostics: false,
+      });
```
mbg, line 232: Question: The comment for `bundleDb` suggests that this will only do something if no bundle exists already. The implementation of `bundleDb` looks as if it deletes existing bundles and then re-creates them. If it is the latter, should we update the comment on `bundleDb` (and turn it into a JSDoc)? If it is actually the former, what happens if `bundleDb` is called multiple times, but with different `includeDiagnostics` values?

src/database-upload.ts
```diff
@@ -156,9 +169,87 @@ export async function cleanupAndUploadDatabases(
...
+        `Failed to measure the '${CleanupLevel.Clear}' cleanup database size ` +
+          `for ${language}: ${util.getErrorMessage(e)}`,
+      );
+    }
+  }
+
+  const durationMs = performance.now() - startTime;
+  for (const report of reports) {
+    report.clear_cleanup_measurement_duration_ms = durationMs;
+  }
```
mbg, line 252: Should this come after the `try`/`catch` for `databaseCleanupCluster`? Why or why not? If not, maybe add a comment here explaining why you chose to include the `bundleDb` loop in this measurement.

mbg (dismissed):
The changes look reasonable. I had a couple of questions and one minor comment, but nothing blocking.

src/database-upload.ts
henrymercer: Going to leave this as is — I plan to delete this code within a quarter.

src/database-upload.ts
henrymercer: Done

src/database-upload.ts
henrymercer: Done

src/database-upload.ts
mbg: And we know that those plans always work out 😄

mbg (approved):

PR #3980, opened 2026-06-30
Support `undefined`, `number` in JSON module

mario-campos:
### Summary

This pull request introduces several additions to the JSON schema-validation utilities:

* Renamed the `optional` validator to `optionalOrNull` to clarify that schema fields may be optional and accept `null` as a valid value, in addition to `undefined`.
* Redefined the `optional` validator so that it only accepts `undefined` (not `null`), making its behavior more precise and distinct from `optionalOrNull`.
* Added new built-in validators for `number`.
* Added an `isNumber` type guard function for type checking.

This PR is a prerequisite for #3950.

### Risk assessment

Workflow types:

Products:

Environments:

- [x] Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.

### DIFF
--- a/src/json/index.test.ts
+++ b/src/json/index.test.ts
@@ -10,8 +10,8 @@ const testSchema = {
   requiredKey: json.string,
 };
 
-const optionalSchema = {
-  optionalKey: json.optional(json.string),
+const optionalOrNullSchema = {
+  optionalKey: json.optionalOrNull(json.string),
 };
 
 test("validateSchema - required properties are required", async (t) => {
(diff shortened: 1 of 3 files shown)


src/json/index.ts
```diff
@@ -55,8 +60,23 @@ export const string = {
...
+export const number = {
+  validate: isNumber,
+  required: true,
+} as const satisfies Validator<number>;
+
+/** A validator for object fields in schemas. */
+export const object = {
+  validate: isObject,
+  required: true,
+} as const satisfies Validator<UnvalidatedObject<unknown>>;
```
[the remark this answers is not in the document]
mario-campos: Fixed

src/json/index.ts
```diff
@@ -30,6 +30,10 @@ export function isString(value: unknown): value is string {
   return typeof value === "string";
 }
 
+export function isNumber(value: unknown): value is number {
+  return typeof value === "number" && Number.isFinite(value);
```
mbg: Since this module is intended for validating the results of parsing JSON, I am not sure that the `Number.isFinite` check is required here (the reasons Copilot gave for adding it are also the reasons why I don't think it would be needed).

src/json/index.ts
```diff
@@ -55,8 +59,23 @@ export const string = {
...
+export const number = {
+  validate: isNumber,
+  required: true,
+} as const satisfies Validator<number>;
+
+/** A validator for object fields in schemas. */
+export const object = {
+  validate: isObject<any>,
+  required: true,
+} as const satisfies Validator<UnvalidatedObject<any>>;
```
mbg: For top-level objects, we'd call `validateSchema` on them along with the relevant schema. Adding this `object` validator to schemas seems a bit simplistic. Is the intention here to say "I want to accept an arbitrary object"? If so, then it might be better off as being named `any` or similar, rather than `object`. If not, then perhaps this should be parameterised over a schema:
```suggestion
export const object = <S extends Schema>(schema: S) => ...;
```

src/json/index.test.ts
```diff
@@ -30,11 +30,34 @@ test("validateSchema - required properties are required", async (t) => {
 
 test("validateSchema - optional properties are optional", async (t) => {
```
mbg:
```suggestion
test("validateSchema - optionalOrNullSchema properties are optional or null", async (t) => {
```

src/json/index.test.ts
```diff
@@ -30,11 +30,34 @@ test("validateSchema - required properties are required", async (t) => {
...
+  t.false(json.validateSchema(optionalOrNullSchema, { optionalKey: {} }));
+  t.true(json.validateSchema(optionalOrNullSchema, { optionalKey: "" }));
+  t.true(json.validateSchema(optionalOrNullSchema, { optionalKey: "foo" }));
+});
+
+const optionalSchema = {
+  optionalKey: json.optional(json.string),
+};
+
+test("validateSchema - optional properties may be absent or undefined, but reject null", async (t) => {
```
mbg:
```suggestion
test("validateSchema - optional properties are optional", async (t) => {
```

src/json/index.ts
mario-campos: Here's the use-case:

```TypeScript
json.validateSchema(
  {
    version: json.string,
    features: json.optional(json.object),
    overlayVersion: json.optional(json.number),
  },
  x,
);
```

where `features` here refers to `VersionInfo.features`.

mbg: OK, so there is a schema for `features` (a mapping of strings to booleans) that could be expressed if we wanted to. I think it's better for us to leave the conversation about the schema validation for the CLI output to the other PR, but the relevant flow chart for my comment was:

- If you just want to validate arbitrary objects, rename it to `any` or similar instead of `object`.
- If not, change `object` to validate a specific object schema.

src/json/index.ts
mario-campos: Deleted commit.

src/json/index.ts
mario-campos: On second thought, I think I'd rather perform a deeper validation of the `features` object, obviating the need for `object`. So, I'll just delete it.

mbg (approved):

PR #3981, opened 2026-07-01
Create wrapper `runCliJson`

mario-campos:
### Summary

A few of the CLI commands emit JSON. This PR creates a `runCli` wrapper function called `runCliJson` that expects JSON output from the CLI command. This change centralizes and reduces the number of JSON-parsing call sites in `codeql.ts`, making the code more readable and D.R.Y.er.

This PR is a prerequisite for #3950.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/codeql.ts
+++ b/src/codeql.ts
@@ -1140,7 +1140,8 @@ async function runCliJson<T>(
   try {
     return JSON.parse(output) as T;
   } catch {
-    throw Error(`Invalid JSON output from \`${args.join(" ")}\`: ${output}`);
+    const command = [cmd, ...args].join(" ");
+    throw Error(`Invalid JSON output from \`${command}\`: ${output}`);
   }
 }
 
--- a/src/codeql.ts
+++ b/src/codeql.ts
@@ -1131,6 +1131,15 @@ async function runCli(
   }
 }
 
+/**
+ * Wraps the command executor {@link runCli} and tries to parse the output as JSON.
+ * @param cmd The command to run.
+ * @param args The arguments to pass to the command.
+ * @param opts The options for running the command.
+ * @param opts.stdin Optional string to pass to the command's standard input.
+ * @param opts.noStreamStdout Optional boolean to indicate whether to stream the command's standard output.
+ * @returns The parsed JSON output from the command.
+ */
 async function runCliJson<T>(
   cmd: string,
   args: string[] = [],
--- a/src/codeql.ts
+++ b/src/codeql.ts
@@ -514,16 +514,13 @@ async function getCodeQLForCmd(
     async getVersion() {
       let result = util.getCachedCodeQlVersion(cmd);
       if (result === undefined) {
-        const output = await runCli(cmd, ["version", "--format=json"], {
-          noStreamStdout: true,
-        });
-        try {
-          result = JSON.parse(output) as VersionInfo;
-        } catch {
-          throw Error(
-            `Invalid JSON output from \`version --format=json\`: ${output}`,
-          );
-        }
+        result = await runCliJson<VersionInfo>(
+          cmd,
+          ["version", "--format=json"],
+          {
+            noStreamStdout: true,
+          },
+        );
         util.cacheCodeQlVersion(cmd, result);
       }
       return result;
@@ -731,26 +728,20 @@ async function getCodeQLForCmd(
         filterToLanguagesWithQueries: boolean;
       } = { filterToLanguagesWithQueries: false },
     ) {
-      const codeqlArgs = [
+      return runCliJson<ResolveLanguagesOutput>(cmd, [
         "resolve",
         "languages",
         "--format=betterjson",
         "--extractor-options-verbosity=4",
         "--extractor-include-aliases",
+        // TODO: Unconditionally include `--filter-to-languages-with-queries`
+        //       once CODEQL_MINIMUM_VERSION is at least v2.23.0
+        //       — the first version to support this flag.
         ...(filterToLanguagesWithQueries
           ? ["--filter-to-languages-with-queries"]
           : []),
         ...getExtraOptionsFromEnv(["resolve", "languages"]),
-      ];
-      const output = await runCli(cmd, codeqlArgs);
-
-      try {
-        return JSON.parse(output) as ResolveLanguagesOutput;
-      } catch (e) {
-        throw new Error(
-          `Unexpected output from codeql resolve languages with --format=betterjson: ${e}`,
-        );
-      }
+      ]);
     },
     async resolveBuildEnvironment(
       workingDir: string | undefined,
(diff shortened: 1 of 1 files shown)


src/codeql.ts
```diff
@@ -512,21 +512,9 @@ async function getCodeQLForCmd(
...
-          throw Error(
-            `Invalid JSON output from \`version --format=json\`: ${output}`,
-          );
-        }
-        util.cacheCodeQlVersion(cmd, result);
-      }
-      return result;
+      return runCliJson<VersionInfo>(cmd, ["version", "--format=json"], {
+        noStreamStdout: true,
+      });
```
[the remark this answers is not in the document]
mario-campos: Fixed

src/codeql.ts
```diff
@@ -1160,6 +1122,19 @@ async function runCli(
...
+  cmd: string,
+  args: string[] = [],
+  opts: { stdin?: string; noStreamStdout?: boolean } = {},
+): Promise<T> {
+  const output = await runCli(cmd, args, opts);
+  try {
+    return JSON.parse(output) as T;
+  } catch {
+    throw Error(`Invalid JSON output from \`${args.join(" ")}\`: ${output}`);
+  }
```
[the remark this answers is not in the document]
mario-campos: Fixed

src/codeql.ts
```diff
@@ -1160,6 +1131,20 @@ async function runCli(
   }
 }
 
+async function runCliJson<T>(
```
mbg: JSDoc?

src/codeql.ts
```diff
@@ -1160,6 +1131,20 @@ async function runCli(
...
+  cmd: string,
+  args: string[] = [],
+  opts: { stdin?: string; noStreamStdout?: boolean } = {},
+): Promise<T> {
+  const output = await runCli(cmd, args, opts);
+  try {
+    return JSON.parse(output) as T;
+  } catch {
+    const command = [cmd, ...args].join(" ");
+    throw Error(`Invalid JSON output from \`${command}\`: ${output}`);
```
mbg: Minor: No strong feelings, but I don't love that we are constructing the command here, since it could go out-of-sync with the actual command that's run by `runCli` in the future. 

Also, I assume that the `command` string wouldn't include the CodeQL binary name/path? If so, I note that the `Error` message doesn't mention the CLI at all. It might be good to include `"codeql"` before `cmd` or mention the CLI.

As a simple improvement, how about:
```suggestion
    const command = [cmd, ...args].join(" ");
    throw Error(`Invalid JSON output from CLI '${cmd}' command with arguments '${args.join(" ")}': ${output}`);
```
_Ideally_ we'd keep the new error messages as close as possible to the old ones, to avoid changing them unnecessarily.

src/codeql.ts
```diff
@@ -1160,6 +1131,20 @@ async function runCli(
   }
 }
 
+async function runCliJson<T>(
+  cmd: string,
+  args: string[] = [],
+  opts: { stdin?: string; noStreamStdout?: boolean } = {},
+): Promise<T> {
```
mbg: In the spirit of your use of the `json` module, should this be:
```suggestion
): Promise<unknown> {
```
or (with an `isObject` check, since we presumably always expect objects from the CLI):
```suggestion
): Promise<UnvalidatedObject<T>> {
```
?

src/codeql.ts
```diff
@@ -514,16 +514,13 @@ async function getCodeQLForCmd(
...
-        try {
-          result = JSON.parse(output) as VersionInfo;
-        } catch {
-          throw Error(
-            `Invalid JSON output from \`version --format=json\`: ${output}`,
-          );
-        }
+        result = await runCliJson<VersionInfo>(
+          cmd,
+          ["version", "--format=json"],
```
mbg: Could the `--format=json` or `--format=betterjson` argument be added automatically by `runCliJson`?

src/codeql.ts
```diff
@@ -731,26 +728,20 @@ async function getCodeQLForCmd(
...
-      const codeqlArgs = [
+      return runCliJson<ResolveLanguagesOutput>(cmd, [
         "resolve",
         "languages",
         "--format=betterjson",
         "--extractor-options-verbosity=4",
         "--extractor-include-aliases",
+        // TODO: Unconditionally include `--filter-to-languages-with-queries`
+        //       once CODEQL_MINIMUM_VERSION is at least v2.23.0
+        //       — the first version to support this flag.
```
mbg, line 739: Elaborating on my comment on the other PR, what does this comment have to do with the changes here? Just a drive-by improvement?

mario-campos: >what does this comment have to do with the changes here? Just a drive-by improvement?

Nothing, really—a "drive-by improvement" is a nice way to word it. But, it seemed more relevant here than in #3950. I'm fine to delete it, but I think it adds some value.

src/codeql.ts
mario-campos: That would be nice! Unfortunately, there doesn't seem to be consistency between all of the calls of `runCliJson`:

- `getVersion`, `resolveDatabase` use `--format=json`
- `resolveLanguages` uses `--format=betterjson`
- `resolveBuildEnvironment` uses nothing??
- `resolveQueriesStartingPacks` uses `--format=startingpacks`

src/codeql.ts
mario-campos: >Ideally we'd keep the new error messages as close as possible to the old ones, to avoid changing them unnecessarily.

I think this makes sense. I'll push a commit to make it more consistent.

src/codeql.ts
mario-campos: Done

src/codeql.ts
```diff
@@ -1139,9 +1139,8 @@ async function runCliJson<T>(
   const output = await runCli(cmd, args, opts);
   try {
     return JSON.parse(output) as T;
-  } catch {
-    const command = [cmd, ...args].join(" ");
-    throw Error(`Invalid JSON output from \`${command}\`: ${output}`);
+  } catch (e) {
+    throw Error(`Unexpected output from codeql ${args.join(" ")}: ${e}`);
```
mbg: We might want to wrap `e` in `getErrorMessage`.

src/codeql.ts
```diff
@@ -1131,6 +1131,15 @@ async function runCli(
   }
 }
 
+/**
+ * Wraps the command executor {@link runCli} and parses the JSON output.
```
mbg: Change `parses the JSON output.` to `tries to parse the output as JSON.`

mbg (commented):
Thanks for addressing the feedback! I am slightly in two minds about https://github.com/github/codeql-action/pull/3981/commits/8ff2a75b6fbbd7dd028e307c72177663211e86e7. On one hand, the type of `runCliJson` is more accurate now, but on the other hand, we still just assert the types (and that now makes the rest of the code uglier). If the intention is not to perform any validation on the JSON as part of this PR (which is OK, since that's the status quo) then the previous version where `runCliJson` returned `Promise<T>` was probably better.

mario-campos:
>If the intention is not to perform any validation on the JSON as part of this PR (which is OK, since that's the status quo) then the previous version where `runCliJson` returned `Promise<T>` was probably better.

Yes, attempting to validate the output—while justified and correct—is beyond the scope of this simple PR. I'll delete the commit.

mbg (approved):

PR #3983, opened 2026-07-01
Add FF for configuration file repository property

mbg:
Follow-up to #3963 which adds a FF to control whether that functionality is enabled or not.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/config/file.ts
+++ b/src/config/file.ts
@@ -21,19 +21,21 @@ export function getConfigFileInput(
     return input;
   }
 
-  // Don't take the repository property into consideration if the FF is not enabled.
-  if (!useRepositoryProperty) {
-    return undefined;
-  }
-
   const propertyValue =
     repositoryProperties[RepositoryPropertyName.CONFIG_FILE];
 
   if (propertyValue !== undefined && propertyValue.trim().length > 0) {
-    logger.info(
-      `Using configuration file input from repository property: ${propertyValue}`,
-    );
-    return propertyValue;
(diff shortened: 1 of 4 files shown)


src/config/file.ts
```diff
@@ -12,6 +12,7 @@ export function getConfigFileInput(
   logger: Logger,
   actions: ActionsEnv,
   repositoryProperties: Partial<RepositoryProperties>,
+  useRepositoryProperty: boolean,
```
henrymercer, line 15: Minor: It would be clearer to use an object like `{ useRepositoryProperty: boolean }` here so it's clearer when the function is called what the last argument means.

src/config/file.ts
```diff
@@ -20,6 +21,11 @@ export function getConfigFileInput(
     return input;
   }
 
+  // Don't take the repository property into consideration if the FF is not enabled.
+  if (!useRepositoryProperty) {
+    return undefined;
```
henrymercer: Minor: we could add a log here saying we're ignoring the repo property because the feature isn't enabled.

henrymercer (dismissed):

src/config/file.ts
mbg: Agreed with the point, but I expect to change this in line with the changes in https://github.com/github/codeql-action/pull/3973 (so that the `features` are available via the indexed `ActionState`).

henrymercer (approved):

PR #3990, opened 2026-07-01
Add `ActionState` type and make some initial use of it

mbg:
This PR extracts some changes from #3973 that add the new `ActionState` type. The `ActionState` type is indexed (at the type-level) over the types of state a corresponding value provides (essentially providing a primitive state effect system). Concretely, this allows:

- `ActionState` values can be propagated easily throughout the implementation.
- `ActionState` values can easily be mocked in tests.
- By moving from using global state to `ActionState`, we will be able to run more tests in parallel in the future.
- The type-level array of features allows functions to declare which state they require access to.
- TypeScript checks that against usage in the function body. I.e. if a function tries to access state that it hasn't declared, then that is a type error.
- Broader `ActionState` values can be passed to functions that require only a subset of features, but cannot use the features they don't declare.
- TypeScript doesn't check that the declared usage is minimal, i.e. that functions only declare the features they need, but that is OK for the moment.

This PR adds the new type and surrounding infrastructure. It also updates some parts of the codebase to make use of it, but the goal of this PR is by no means to refactor the entire codebase to make use of `ActionState` values. The main example uses in this PR are:

- A common `runInActions` wrapper to de-duplicate common code for the different Actions' entry points.
- Simplifying interface of `getConfigFileInput` following #3983 and making use of the new `TestEnv` framework for the corresponding tests.

**Notes for reviewers**

- For functions which don't pass on `ActionState` to other functions, I have deconstructed the objects with pattern-matching. This will go away more as we update more of the codebase to accept relevant `ActionState` values as inputs.
- I have thought about the semantics of the `logger` in `TestEnv` for a while, but am not sure I have reached a satisfactory conclusion there. It probably makes sense for each `TestEnv` object to have its own logger, so that they are isolated from each other if multiple calls are made / assertions are run, but this makes it more awkward to inspect the results (since we have to acquire a reference to the logger of the `TestEnv` we call `passes` on). This could probably be improved, but may be outside the scope of this PR.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/action-common.ts
+++ b/src/action-common.ts
@@ -1,3 +1,4 @@
+import { ActionsEnv } from "./actions-util";
 import { Env } from "./environment";
 import { FeatureEnablement } from "./feature-flags";
 import { Logger } from "./logging";
@@ -11,6 +12,10 @@ export interface FeatureState {
   Env: {
     /** Information about environment variables. */
     env: Env;
+  };
+  Actions: {
+    /** Access to Actions-related functionality. */
+    actions: ActionsEnv;
   };
   FeatureFlags: {
     /** Information about enabled feature flags. */
--- a/src/action-common.ts
+++ b/src/action-common.ts
@@ -4,7 +4,11 @@ import { ActionsEnv, getActionsEnv } from "./actions-util";
 import { Env } from "./environment";
 import { FeatureEnablement } from "./feature-flags";
 import { getActionsLogger, Logger } from "./logging";
-import { ActionName, sendUnhandledErrorStatusReport } from "./status-report";
+import {
+  ActionName,
+  getDisplayActionName,
+  sendUnhandledErrorStatusReport,
+} from "./status-report";
 import { getEnv, getErrorMessage } from "./util";
 
 /** Common state that is always available in `ActionState`. */
@@ -83,7 +87,9 @@ export async function runInActions(action: Action) {
       actions: actionsEnv,
     });
   } catch (error) {
-    core.setFailed(`${action.name} action failed: ${getErrorMessage(error)}`);
+    core.setFailed(
+      `${getDisplayActionName(action.name)} action failed: ${getErrorMessage(error)}`,
+    );
     await sendUnhandledErrorStatusReport(action.name, startedAt, error, logger);
   }
 }
--- a/src/action-common.ts
+++ b/src/action-common.ts
@@ -1,0 +1,37 @@
+import { Env } from "./environment";
+import { FeatureEnablement } from "./feature-flags";
+import { Logger } from "./logging";
+
+/** Describes different state features that an Action may have. */
+export interface FeatureState {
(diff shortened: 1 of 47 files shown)


src/action-common.ts
```diff
@@ -0,0 +1,89 @@
...
+  try {
+    await action.run({
+      name: action.name,
+      startedAt,
+      logger,
+      env,
+      actions: actionsEnv,
+    });
+  } catch (error) {
+    core.setFailed(`${action.name} action failed: ${getErrorMessage(error)}`);
```
henrymercer: This is probably undesirable for the `analyze` Action, since the "finish" telemetry name will now leak out into the logs.  Consider adding a `displayName` property to `Action`.

src/testing-utils.ts
```diff
@@ -181,6 +195,108 @@ export function getTestActionsEnv(): ActionsEnv {
...
+    this.logger = new RecordingLogger();
+    this.state =
+      cloneFrom !== undefined
+        ? { ...cloneFrom.state, logger: this.logger }
+        : {
+            name: ActionName.Init,
+            startedAt: new Date(),
+            logger: this.logger,
+            env: getTestEnv(),
+            actions: getActionsEnv(),
```
henrymercer: Is the idea that this is temporary, to avoid needing to rewrite existing tests to use new `getOptionalInput` stubs?

henrymercer (dismissed):

src/action-common.ts
mbg: Thanks, that was well spotted! I thought I had checked this when porting the entry points to `runInActions`, but clearly missed this. 

Since in most cases the `ActionName` value is the correct display name, I have added a `getDisplayActionName` function which maps `ActionName` values to the right display string. That might be better than an extra property in `Action` that has to be set for each action. We should later be able to use `getDisplayActionName` for the post Actions as well.

src/testing-utils.ts
mbg: This part wouldn't be temporary. The highlighted part is for initialising the `TestEnv` state when an instance is first created. The idea would be that new tests adopt the pattern of wrapping functions in `callee` and then provide the context with the `withX` functions of the `TestEnv`.

henrymercer: I was thinking about `actions: getActionsEnv()` in particular — presumably unit tests shouldn't depend on the Actions environment any more than they should depend on `process.env`?

mbg: Ah, I see what you're asking now. I thought you were worried about having all possible `ActionState` state in `TestEnv`, but now I spotted the oversight that this uses `getActionsEnv` instead of `getTestActionsEnv`. I have replaced this with `getTestActionsEnv` now.

henrymercer (dismissed):

henrymercer (approved):

PR #3994, opened 2026-07-06
Allow merging Default Setup `config` with config file

mbg:
This PR introduces support for merging the `config` inputs that are provided by Default Setup with custom configuration files. That can be the case if, for example, a configuration file is configured by the corresponding repository property. With the change in this PR, a supported scenario is that e.g.:

- A CodeQL workflow runs in Default Setup on a repository where the `github-codeql-config-file` is set.
- `getConfigFileInput` returns the value from the repository property.
- `getOptionalInput("config")` returns the configuration provided by Default Setup.
- In `determineUserConfig`, the two configurations are merged if the corresponding FF is enabled.

### Notes for reviewers / deliberate implementation aspects

- `determineUserConfig` is refactored out of `initConfig`. This PR does not make any changes to the refactored code that aren't required for the objective of this PR.
- `inputs.configFile` is mutated to the resulting configuration file path as in `initConfig` before. No change in behaviour. The new tests for `determineUserConfig` check that this behaves as expected in relevant cases.
- No new validation of the `inputs` properties is performed. Just like `initConfig` before, `determineUserConfig` assumes that the necessary validation before this step has taken place. Further validation may occur in e.g. `loadUserConfig`.
- The new tests do not cover `Feature.ValidateDbConfig`. This is deliberate because there are already tests exercising `loadUserConfig`'s behaviour if that FF is enabled. In any case, `Feature.ValidateDbConfig` needs more work to work correctly that is not in the scope of this PR.
- The tests make use of the framework merged in https://github.com/github/codeql-action/pull/3993 to allow the tests to run in parallel.
- The tests do not cover cases with remote files, because whether a file is or is not remote does not matter to `determineUserConfig`. That's for the existing `loadUserConfig` function and its tests.

### Risk assessment

Workflow types:

Products:

Environments:


src/actions-util.ts
```diff
@@ -399,9 +419,10 @@ const persistedInputsKey = "persisted_inputs";
...
-export const persistInputs = function () {
-  const inputEnvironmentVariables = Object.entries(process.env).filter(
-    ([name]) => name.startsWith("INPUT_"),
+export const persistInputs = function (env?: Env) {
+  const entries = env?.entries() || Object.entries(process.env);
+  const inputEnvironmentVariables = entries.filter(([name]) =>
+    name.startsWith("INPUT_"),
   );
   core.saveState(persistedInputsKey, JSON.stringify(inputEnvironmentVariables));
 };
```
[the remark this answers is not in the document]
mbg: This matches the previous behaviour.

src/config/db-config.test.ts
```diff
@@ -488,3 +489,67 @@ test("parseUserConfig - throws no ConfigurationError if validation should fail,
...
+    logger,
+    { "threat-models": ["a", "b"] },
+    { "threat-models": ["local", "remote"] },
+  );
+
+  const threatModels = result["threat-models"];
+
+  if (t.truthy(threatModels)) {
+    t.deepEqual(threatModels, ["a", "b", "local", "remote"]);
+  }
```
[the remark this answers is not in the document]
mbg: This is nonsense. `truthy` has a return value and can be used as a type guard.

src/config/db-config.ts
```diff
@@ -46,6 +54,65 @@ export interface UserConfig {
...
+ * Merges supported properties from two configuration files. This is intended only for
+ * use with merging the `config` input provided by Default Setup with a potentially
+ * richer configuration file provided by a user.
+ *
+ * @param logger The logger to use.
+ * @param fromConfigInput The configuration from Default Setup.
+ * @param fromConfigFile The user-supplied configuration.
+ * @returns The combination of both configuration files.
+ */
+export function mergeUserConfigs(
```
henrymercer: Can we indicate in the function name that this is about merging a user config with a default setup config?

src/config/db-config.ts
```diff
@@ -46,6 +54,65 @@ export interface UserConfig {
...
+  delete result["default-setup"];
+
+  if (fromConfigInput["default-setup"]) {
+    result["default-setup"] = fromConfigInput["default-setup"];
+  }
+  if (threatModels.size > 0) {
+    result["threat-models"] = Array.from(threatModels);
+  }
+
+  return result;
```
henrymercer, line 179: It is probably worth logging a warning that is sent to telemetry if we find any other properties in `fromConfigInput`.  If we format this as a telemetry diagnostic, we can set the `internal-error` tag on the reporting descriptor to send exceptions to Sentry.

src/config-utils.test.ts
```diff
@@ -2272,3 +2278,317 @@ test("applyIncrementalAnalysisSettings: adds exclusions for diff-informed-only r
     { exclude: { tags: "exclude-from-incremental" } },
   ]);
 });
+
+test("determineUserConfig - empty config when neither input is specified", async (t) => {
+  await withTmpDir(async (tmpDir) => {
+    const logger = new RecordingLogger();
+    const env = util.getEnv(DEFAULT_ACTIONS_VARS);
```
henrymercer: Should this act on a copy of `DEFAULT_ACTIONS_VARS`, to avoid potential calls to `Env.set` mutating `DEFAULT_ACTIONS_VARS`?  I know that `determineUserConfig` is read only, but this could be confusing if behaviour changes.

src/config-utils.test.ts
```diff
@@ -2272,3 +2278,317 @@ test("applyIncrementalAnalysisSettings: adds exclusions for diff-informed-only r
...
+      tmpDir,
+      createTestInitConfigInputs({
+        configInput: defaultSetupConfigInput,
+        configFile: configFilePath,
+        workspacePath: tmpDir,
+      }),
+    );
+
+    // The loaded configuration should match the result of merging
+    // `defaultSetupConfigInput` and `simpleConfigFileContents`.
```
henrymercer, line 2429: Let's also check that `inputs.configFile` was mutated and that the contents on disk match the returned config

src/testing-utils.ts
```diff
@@ -323,11 +323,19 @@ export type ActionVarOverrides = Partial<
  * excluding some that are expected to be set to paths. See `setupActionsVars`.
  *
  * @param overrides Overrides for the defaults.
+ * @param env The environment to set the variables for.
  */
-export function setupBaseActionsVars(overrides?: ActionVarOverrides) {
+export function setupBaseActionsVars(
+  overrides?: ActionVarOverrides,
+  env?: Env,
```
mario-campos: Rather than make this optional, it would be better to have it default to `process.env` via `getEnv`:


```suggestion
  env = getEnv(),
```

src/testing-utils.ts
```diff
@@ -323,11 +323,19 @@ export type ActionVarOverrides = Partial<
...
+) {
   const vars = { ...DEFAULT_ACTIONS_VARS, ...overrides };
   for (const [key, value] of Object.entries(vars)) {
-    process.env[key] = value;
+    if (env) {
+      env.set(key, value);
+    } else {
+      process.env[key] = value;
+    }
```
mario-campos: That change above would allow this branching to be avoided, simplifying this to a simple straightline `set`:

```suggestion
    env.set(key, value);
```

Another option could be to do away with the loop entirely and use [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign):

```TypeScript
const vars = { ...DEFAULT_ACTIONS_VARS, ...overrides };
Object.assign(env, vars);
```

src/actions-util.ts
```diff
@@ -7,12 +7,13 @@ import * as github from "@actions/github";
...
 
 import type { Config } from "./config-utils";
+import { Env, EnvVar } from "./environment";
 import { Logger } from "./logging";
 import {
   doesDirectoryExist,
   getCodeQLDatabasePath,
-  getRequiredEnvParam,
   ConfigurationError,
+  getEnv,
```
mario-campos, line 16: Perhaps `getEnv` should live in `environment.ts`.

src/actions-util.ts
```diff
@@ -83,17 +84,22 @@ export const getOptionalInput = function (name: string): string | undefined {
...
+ * Gets the temporary directory used by the CodeQL Action. This will either be the temporary
+ * directory that has been set in `CODEQL_ACTION_TEMP` by e.g. a previous step, or the
+ * value of `RUNNER_TEMP` otherwise.
+ */
+export function getTemporaryDirectory(env: Env = getEnv()): string {
+  const value = env.getOptional(EnvVar.TEMP);
   return value !== undefined && value !== ""
     ? value
-    : getRequiredEnvParam(ActionsEnvVars.RUNNER_TEMP);
+    : env.getRequired(ActionsEnvVars.RUNNER_TEMP);
```
mario-campos: If `EnvVar.TEMP` has been set by a previous step, then I think it's safe to assume that it wouldn't be an empty string. That assumption allows for this simplification:


```suggestion
return env.getOptional(EnvVar.TEMP) ?? env.getRequired(ActionsEnvVars.RUNNER_TEMP);
```

src/actions-util.ts
```diff
@@ -290,18 +296,18 @@ export const getFileType = async (filePath: string): Promise<string> => {
   }
 };
 
-export function isSelfHostedRunner() {
-  return process.env.RUNNER_ENVIRONMENT === "self-hosted";
+export function isSelfHostedRunner(env: Env = getEnv()) {
+  return env.getOptional("RUNNER_ENVIRONMENT") === "self-hosted";
```
mario-campos: I think this should be `getRequired` because we can expect this environment variable to exist in any GitHub Actions job, which is why the previous implementation just assumed it existed in `process.env`.

```suggestion
  return env.getRequired("RUNNER_ENVIRONMENT") === "self-hosted";
```

Also, `RUNNER_ENVIRONMENT` should be in `ActionsEnvVars`, right? So, this should actually be:

```
  return env.getRequired(ActionsEnvVars.RUNNER_ENVIRONMENT) === "self-hosted";
```

src/actions-util.ts
```diff
@@ -443,8 +452,8 @@ export function getPullRequestBranches(): PullRequestBranches | undefined {
 
   // PR analysis under Default Setup does not have the pull_request context,
   // but it should set CODE_SCANNING_REF and CODE_SCANNING_BASE_BRANCH.
-  const codeScanningRef = process.env.CODE_SCANNING_REF;
-  const codeScanningBaseBranch = process.env.CODE_SCANNING_BASE_BRANCH;
+  const codeScanningRef = env.getOptional("CODE_SCANNING_REF");
+  const codeScanningBaseBranch = env.getOptional("CODE_SCANNING_BASE_BRANCH");
```
mario-campos: Should these environment variables be in an `enum`?

src/config-utils.ts
```diff
@@ -1132,36 +1150,98 @@ export async function applyIncrementalAnalysisSettings(
...
-    if (inputs.configFile) {
-      logger.warning(
-        `Both a config file and config input were provided. Ignoring config file.`,
+    const computedConfigPath = userConfigFromActionPath(tempDir);
+
+    // Get a function which enables us to determine whether the FF that allows us to
+    // merge supported configuration file properties is enabled. We only execute
+    // this lazily if the other checks pass.
+    const allowMergeConfigs = () =>
+      features.getValue(Feature.AllowMergeConfigFiles);
```
mario-campos: I understand what this is doing, but I don't know why. Perhaps a little more elaboration on why this is necessary. Why should this be evaluated lazily?

src/config-utils.ts
mario-campos, line 1: I had/have a hard time understanding the difference between `configInput` and `configFileInput`. I think it would be clearer to refer to the config-file as a repository property. Or, make the distinction clearer somehow.

mario-campos (changes requested):
A couple of minor suggestions, small improvements for clarity.

src/config-utils.ts
mbg: I have updated this a bit already since the version you commented on, but the main point is just that we don't query it until we need it. That's probably overly conservative since we have normally already loaded all the FFs and the lookup is cheap, but in case we haven't, it would save us on making an API call.

src/actions-util.ts
mbg: > I think this should be `getRequired` because we can expect this environment variable to exist in any GitHub Actions job, which is why the previous implementation just assumed it existed in `process.env`.

I generally agree that we would expect this variable to exist in an Actions environment. That said, with the previous implementation, if `RUNNER_ENVIRONMENT` wasn't set, the left-hand side of the `===` would have evaluated to `undefined` and therefore the overall expression would have evaluated to `false`. The call to `getOptional` works similarly and returns `undefined` if the environment variable isn't set. If we changed it to `getRequired`, then it would throw if `RUNNER_ENVIRONMENT` is not set.

I'd rather avoid this change to cause unnecessary exceptions in environments where `RUNNER_ENVIRONMENT` is not set (e.g. tests).

> Also, `RUNNER_ENVIRONMENT` should be in `ActionsEnvVars`, right? 

Yes. That `enum` is relatively new though, so we are still working on migrating all relevant environment variable names to it. I'll add it.

src/testing-utils.ts
mbg: > Another option could be to do away with the loop entirely and use [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign):
> 
> ```ts
> const vars = { ...DEFAULT_ACTIONS_VARS, ...overrides };
> Object.assign(env, vars);
> ```

`env` is not a `Record`-style object, so that doesn't work.

src/actions-util.ts
mbg: While `CODEQL_ACTION_TEMP` is expected to be set by a previous CodeQL Action step if it set, we can't verify that it is. It could have been set by something else (e.g. us in a test workflow, or more generally in any Advanced Setup workflow). 

That said, we already perform the emptiness check in `getOptionalEnvVarFrom` (called by `getOptional`) and so `value` would never be empty here anyway (it would be `undefined` instead) so the simplification is still applicable.

mbg:
@mario-campos Since most of your feedback was about the changes for environment variable handling, I have refactored those changes out into https://github.com/github/codeql-action/pull/4004 and addressed your comments with commits there.

src/config-utils.test.ts
mbg: I have pushed some improvements to the `TestEnv` implementation in https://github.com/github/codeql-action/pull/4005, which will help address this here once those changes are merged.

src/json/index.test.ts
mario-campos, line 1: I don't see any tests for the newly-added `CheckSchemaOptions` parameter `opts` to `Validator.check`. Is that deliberate?

src/json/index.test.ts
```diff
@@ -121,6 +121,22 @@ test("validateSchema - checkSchema reports unknown keys", async (t) => {
...
+      objectKey: {
+        arrayKey: ["foo"],
+      },
+    },
+  });
+
+  t.false(result.valid);
+  t.deepEqual(
+    result.invalidKeys.sort(),
+    [".rootKey.objectKey.arrayKey[0]"].sort(),
```
mario-campos, line 140: Why sort an array of 1?

```suggestion
    [".rootKey.objectKey.arrayKey[0]"],
```

src/config/db-config.ts
```diff
@@ -112,6 +112,22 @@ export function mergeDefaultSetupAndUserConfigs(
...
+      undefined,
+      makeTelemetryDiagnostic(
+        "codeql-action/invalid-default-setup-config-keys",
+        "Invalid Default Setup configuration keys",
+        {
+          invalidKeys: schemaCheckResult.invalidKeys,
+        },
+      ),
+    );
+  }
```
mario-campos, line 130: I'm not sure what `makeTelemetryDiagnostic` does and I'm just presuming based on the name, so take this with a grain of salt:

If `makeTelemetryDiagnostic` uploads some data to a backend paid service, like Datadog, it may not be wise to send untrusted user input to it—or at least _unbounded_ untrusted user input.

src/config/db-config.test.ts
```diff
@@ -572,13 +572,56 @@ test("mergeDefaultSetupAndUserConfigs - ignores, but warns about, unknown keys f
...
+    },
+    configFile,
+  );
+
+  t.deepEqual(result, {
+    ...configFile,
+    "default-setup": { org: { "model-packs": [123] } },
+  });
+
+  const expectedInvalidKeys = [".default-setup.org.model-packs[0]"].join(", ");
```
mario-campos, line 623: I think this constant expression can be simplified:

```suggestion
  const expectedInvalidKeys = ".default-setup.org.model-packs[0]";
```

mario-campos (dismissed):
A couple of minor suggestions that can be ignored, and some food for thought. But otherwise, looks good 👍

src/config/db-config.ts
mbg: I replied to you elsewhere about this, but needless to say that this is not an issue.

src/json/index.test.ts
mbg: It's a minor optimisation, and although the effect is not explicitly tested, both code paths are exercised in respective tests for: `validateSchema` explicitly sets `failFast: true` while `checkSchema` uses the default `failFast: false`. I am not sure that it's worth adding explicit tests for this behaviour.

src/json/index.test.ts
mbg: Mainly since I copied it from the other test that needs it and it's not causing any harm if we extended the test to return more keys here.

src/config/db-config.test.ts
mbg: It's here since I copied it from the other test that needs it and it's not causing any harm if we extended the test to return more keys here.

src/config-utils.test.ts
mbg: That PR was merged and I have updated this branch with the merged changes, which has allowed me to update the tests to make use of the new `withDefaultActionsEnv()`. See https://github.com/github/codeql-action/pull/3994/commits/2773684c2965b383257f03aa009863baaab9ade4 and https://github.com/github/codeql-action/pull/3994/commits/f181d70282000de7652b8f7fe4f558d5d9781f44

src/config/db-config.ts
mbg: I have added diagnostics for unrecognised and invalid keys in https://github.com/github/codeql-action/pull/3994/commits/7c961fa35788ac1b3c29a1bf8f5ad097b2259697 and https://github.com/github/codeql-action/pull/3994/commits/f630f8fd043520cb9d4e0ccda16c4ed4b635683d

henrymercer (approved):
I'm happy merging this as is, but consider making use of the `internal-error` tag on the telemetry diagnostics to report exceptions to Sentry.

PR #4014, opened 2026-07-15
Add support for an explicit remote path prefix

mbg:
This is an alternative to #4012 which addresses the same issue. The first few commits here are the same as there and increase test coverage and refactor a couple of functions.

The approach to resolving the issue here is different, however: instead of trying to find the file locally as in #4012 and always treat ambiguous paths as remote paths, we instead add support for a prefix that allows user to explicitly indicate that a configuration file path is supposed to refer to a remote file. 

### Risk assessment

Workflow types:

Products:

Environments:


src/config/file.ts
```diff
@@ -11,6 +11,19 @@ import { ConfigurationError } from "../util";
...
+ */
+export const LOCAL_PATH_PREFIX = "./";
+
+/**
+ * The prefix that can be specified to indicate that a path should be treated as a remote file address.
+ * The new remote file address format must start with either an owner or repository name. Both
+ * are restricted to ASCII characters, '.', and '-'. The prefix chosen here does not interfere with
+ * those and is _unlikely_ (but not impossible) to appear in a local file path.
+ */
+export const REMOTE_PATH_PREFIX = "::";
```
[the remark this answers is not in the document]
mbg: Copilot is wrong because the feature is not rolled out and this is a bug fix. It previously would have failed for ambiguous paths.

henrymercer: What about something like `remote:`?  It's five additional characters but can't be confused with two `:` separator characters.

henrymercer (dismissed):
Bike shed about the prefix, but otherwise LGTM!

src/config/file.ts
mbg: I have no strong feelings on `::` and am happy to change it to something else, but I don't think that `remote:` would be a good choice because, if you have a repo in your organisation named `remote`, e.g. `remote:codeql.yml` would be a valid remote file address that instructs us to fetch `codeql.yml` from `org/remote`, and a `remote:` prefix would clash with that and cause us to truncate it to just `codeql.yml`. That's why I think the prefix should start with a character that's not allowed in repo or org names.

henrymercer: Good point, how about `remote=org/repo@ref:path` or `remote(org/repo@ref:path)`?

mbg: I have changed the prefix to `remote=` in https://github.com/github/codeql-action/pull/4014/commits/3492b7e9ab96e28b1d8b971345d30e929c6f8fee

henrymercer (approved):

PR #4043, opened 2026-07-24
Convert last remaining Python CI scripts to TypeScript

mbg:
This PR converts the last remaining Python scripts that are used in our CI to TypeScript equivalents.

### Risk assessment

Environments:

- **Other** - CI only.


pr-checks/rollback-changelog.ts
```diff
@@ -0,0 +1,93 @@
...
+        },
+      },
+      strict: true,
+    });
+
+    for (const [key, val] of Object.entries(values)) {
+      if (val === undefined || val.trim() === "") {
+        throw new Error(`Argument '--${key}' is required.`);
+      }
+    }
```
navntoft: Nit: Note that, unlike in `rollback_changelog.py`, Node's `parseArgs` with `strict` doesn't check for missing args.

All args are given in the workflow though, and as this scripts just writes to stdout, we will likely catch `undefined` in that case if we happen to run it locally. Your call wrt. whether it's worth the complexity:

```suggestion
    const options = {
      "target-version": { type: "string", short: "t" },
      "rollback-version": { type: "string", short: "r" },
      "new-version": { type: "string", short: "n" },
    } as const;

    const { values } = parseArgs({ options, strict: true });

    for (const key of Object.keys(options)) {
      const val = values[key as keyof typeof values];
      if (val === undefined || val.trim() === "") {
        throw new Error(`Argument '--${key}' is required.`);
      }
    }
```

pr-checks/changelog.ts
```diff
@@ -66,70 +155,58 @@ export function processChangelogForBackports(
...
-      output += `\n${line}\n\n`;
-    } else {
+      // Keep the line.
+      filteredEntries.push(line);
+
+      // Set `foundContent` to `true` if the line is not empty.
       if (line.trim() !== "") {
         foundContent = true;
-        output += `${line}\n`;
       }
```
navntoft, line 199: Good improvement! This also seems to fix a minor bug where blank lines between paragraphs get stripped when backporting ([example](https://github.com/github/codeql-action/commit/b264e152591cf9e60453d86f7644cc602e2d2770)).

Nit: Maybe we should add a regression test for good measure? Your call wrt. whether it is worth the complexity.

```ts
const testChangelogMultiParagraph = `# CodeQL Action Changelog

## 4.12.3 - 14 Aug 2026

- Bullet 1

- Bullet 2
`;

const testChangelogMultiParagraphResult = `# CodeQL Action Changelog

## 3.12.3 - 14 Aug 2026

- Bullet 1

- Bullet 2
`;

// Regression test for a minor bug that caused blank lines between paragraphs to get stripped when backporting.
await it("preserves blank lines between paragraphs of a multi-paragraph entry", async () => {
  const result = processChangelogForBackports(
    "4",
    "3",
    testChangelogMultiParagraph,
  );

  assert.deepEqual(
    result.split("\n"),
    testChangelogMultiParagraphResult.split("\n"),
  );
});
```

navntoft (dismissed):
(reviewed to the best of my ability as someone with limited knowledge about this codebase)

Looks well thought out and consistent with the rest of the codebase, as far as I can tell. Only two nits; your call on both.

pr-checks/rollback-changelog.ts
mbg: I must have been asleep while working on the validation for the arguments here 😅 Your suggestion is what I had intended to do here, but not only did I make the mistake that CCR had already pointed out, but also based the check on the wrong object 🙄

pr-checks/changelog.ts
mbg: Yeah, this caused me some confusion when I ported the code, until I realised that there was a bug in the Python variant. Adding a test makes sense, but we can do that separately later since there's probably more scope-creep there to validate that we are doing the right thing / the input changelog is in the right format.

navntoft (approved):

PR #4049, opened 2026-07-28
Generate job UUID in `start-proxy` action

mbg:
The `start-proxy` action precedes the `init` action in Default Setup, but a UUID for the analysis is only generated in the `init` action. That means that telemetry from the `start-proxy` step is not easy to associate with subsequent steps in the workflow.

This PR refactors the existing code that's duplicated across `init` and `setup-codeql` into a function, allows it to retrieve the existing job UUID from the environment, adds tests, and then uses the new function in `start-proxy`. This ensures that telemetry from `start-proxy` includes a unique identifier and the same one is used for the rest of the workflow, while workflows that don't use `start-proxy` also still generate one in `init`.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/setup-codeql-action.ts
+++ b/src/setup-codeql-action.ts
@@ -95,7 +95,7 @@ async function sendCompletedStatusReport(
 
 /** The main behaviour of this action. */
 async function run(
-  actionState: ActionState<["Base", "Logger", "Actions"]>,
+  actionState: ActionState<["Base", "Logger", "Env", "Actions"]>,
 ): Promise<void> {
   // To capture errors appropriately, keep as much code within the try-catch as
   // possible, and only use safe functions outside.
@@ -141,7 +141,7 @@ async function run(
     const actionStateWithFeatures = { ...actionState, features };
 
     // Create a unique identifier for this run.
-    getJobUUID(logger);
+    getJobUUID(actionState);
 
     const statusReportBase = await createStatusReportBase(
       ActionName.SetupCodeQL,
--- a/src/setup-codeql-action.ts
+++ b/src/setup-codeql-action.ts
@@ -1,5 +1,4 @@
 import * as core from "@actions/core";
-import { v4 as uuidV4 } from "uuid";
 
 import { Action, ActionState, runInActions } from "./action-common";
 import {
@@ -26,6 +25,7 @@ import {
   InitToolsDownloadFields,
   createStatusReportBase,
   getActionsStatus,
+  getJobUUID,
   sendStatusReport,
 } from "./status-report";
 import { ToolsDownloadStatusReport } from "./tools-download";
@@ -140,9 +140,8 @@ async function run(
 
     const actionStateWithFeatures = { ...actionState, features };
 
-    const jobRunUuid = uuidV4();
-    logger.info(`Job run UUID is ${jobRunUuid}.`);
-    core.exportVariable(EnvVar.JOB_RUN_UUID, jobRunUuid);
+    // Create a unique identifier for this run.
+    getJobUUID(logger);
 
     const statusReportBase = await createStatusReportBase(
       ActionName.SetupCodeQL,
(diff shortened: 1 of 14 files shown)


src/start-proxy-action.ts
```diff
@@ -35,6 +41,15 @@ async function run(startedAt: Date) {
   let language: BuiltInLanguage | undefined;
 
   try {
+    const action: ActionState<["Logger", "Env", "Actions"]> = {
+      logger,
+      env: new Env(process.env),
```
mario-campos: Nit: use `getEnv()` instead.

mario-campos (dismissed):
One nit but otherwise makes sense and LGTM.

src/status-report.ts
```diff
@@ -59,6 +61,30 @@ export function getDisplayActionName(actionName: ActionName): string {
...
 }
 
+/**
+ * Either creates a UUIDv4 for the analysis or retrieves an existing one from the
+ * environment and returns it.
+ * If a new UUID is generated, it is also exported as an environment variable.
+ */
+export function getJobUUID(
+  action: ActionState<["Logger", "ReadOnlyEnv", "Actions"]>,
+) {
```
henrymercer, line 71: This means that if `JOB_RUN_UUID` is already set by another step for some reason, we'll use that rather than overwriting it.  That is probably fine, but perhaps we should add the `CODEQL_ACTION_` prefix to the environment variable to reduce the chances of this happening.

src/setup-codeql-action.ts
```diff
@@ -140,9 +140,8 @@ async function run(
 
     const actionStateWithFeatures = { ...actionState, features };
 
-    const jobRunUuid = uuidV4();
-    logger.info(`Job run UUID is ${jobRunUuid}.`);
-    core.exportVariable(EnvVar.JOB_RUN_UUID, jobRunUuid);
+    // Create a unique identifier for this run.
+    getJobUUID(actionState);
```
henrymercer: Perhaps it would make sense to do this in `createStatusReportBase` so we don't have to remember to call this in every Action that might run as the first step.

src/testing-utils.ts
```diff
@@ -187,7 +187,11 @@ export function getTestEnv(testEnv: NodeJS.ProcessEnv = {}): Env {
  */
 export function getTestActionsEnv(): ActionsEnv {
   return {
+    getRequiredInput: (name) => {
+      throw new Error(`Input required and not supplied: ${name}`);
+    },
     getOptionalInput: () => undefined,
+    exportVariable: () => {},
```
henrymercer: This seems like it could be surprising — it'll silently discard all env var writes, while `getTestEnv` preserves writes.

src/status-report.ts
mbg: Makes sense.

src/setup-codeql-action.ts
mbg: I was actually thinking about putting it in `runInActions` since that's intended to be the common entry point for all actions, but that would mean some more work to make that suitable for all actions.

src/testing-utils.ts
mbg: Yeah, I agree. I cherry-picked this from a different branch where I was working on adding `exportVariable` to `ActionsEnv` and had stalled on what I wanted the test behaviour of `exportVariable` to be there. I'll have a go at improving this.

mbg: I have pushed a few commits which make the test `ActionsEnv` backed by an `Env` and then added a way of asserting that the `Env` contains particular entries after the main assertion has run.

src/setup-codeql-action.ts
mbg: I have pushed some changes which call it in `runInActions`, make `runInActions` work for `start-proxy`, and add tests.

henrymercer (approved):

PR #4050, opened 2026-07-29
Include `registry_types` in more status reports

mbg:
Currently, the `registry_types` field in status reports is only populated by the `start-proxy` action. However, other steps such as `init` and `analyze` have access to the registry types via the `CODEQL_PROXY_URLS` environment variable. It is useful for us to be able to see at a glance when `init` or `analyze` steps benefit from configured registries. 

This PR updates the `createStatusReportBase` function to try and parse the contents of `CODEQL_PROXY_URLS` when it is set to include the unique registry types in the status report when possible.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/json/index.ts
+++ b/src/json/index.ts
@@ -35,6 +35,11 @@ export function isNumber(value: unknown): value is number {
   return typeof value === "number";
 }
 
+/** Asserts that `value` is a boolean. */
+export function isBoolean(value: unknown): value is boolean {
+  return typeof value === "boolean";
+}
+
 /** Asserts that `value` is either a string or undefined. */
 export function isStringOrUndefined(
   value: unknown,
@@ -78,6 +83,9 @@ export const string = makeValidator(isString);
 
 /** A validator for number fields in schemas. */
 export const number = makeValidator(isNumber);
+
+/** A validator for boolean fields in schemas. */
+export const boolean = makeValidator(isBoolean);
 
 /** A validator for arrays. */
 export function array<T>(validator: Validator<T>) {
--- a/src/json/index.ts
+++ b/src/json/index.ts
@@ -62,14 +62,11 @@ function defaultCheck(
   return (arg) => ({ unknownKeys: [], invalidKeys: [], valid: validate(arg) });
 }
 
-function makeValidator<T>(
-  validate: (arg: unknown) => arg is T,
-  required: boolean = true,
-) {
+function makeValidator<T>(validate: (arg: unknown) => arg is T) {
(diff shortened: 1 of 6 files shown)


sam-robson:
Is it intentional that we omit host registries in the the status reports? I believe `CODEQL_PROXY_URLS` drops these, but the current status report sent by `start-proxy` includes them.

mbg:
@sam-robson Yes, we intentionally do not include hostnames in the telemetry. The hostnames are filtered out for the existing `start-proxy` telemetry by `proxyConfig.all_credentials.map((c) => c.type)`.

sam-robson:
> @sam-robson Yes, we intentionally do not include hostnames in the telemetry. The hostnames are filtered out for the existing `start-proxy` telemetry by `proxyConfig.all_credentials.map((c) => c.type)`.

I meant the registry types for host-only credentials, not the hostname values themselves. `proxyConfig.all_credentials.map((c) => c.type)` includes the type of every credential, including those identified only by host. By contrast, `proxy_urls` first filters with `credential.url !== undefined`, so later status reports omit the types of host-only credentials.

src/json/index.ts
```diff
@@ -62,14 +67,11 @@ function defaultCheck(
...
-function makeValidator<T>(
-  validate: (arg: unknown) => arg is T,
-  required: boolean = true,
-) {
+function makeValidator<T>(validate: (arg: unknown) => arg is T) {
   return {
     validate,
     check: defaultCheck(validate),
-    required,
+    required: true,
```
oscarsj, line 74: why this change?

oscarsj (approved):
just a heads up on a change that I don't understand,but otherwise lgtm

mbg:
> I meant the registry types for host-only credentials, not the hostname values themselves. 

@sam-robson Ah, thanks for clarifying and sorry I misunderstood your comment. AFAIK, the `host` field isn't used anymore by the backend and every credential _should_ be one with a `url` key anyway, so this shouldn't be a meaningful distinction in practice. We could probably look at removing the `host` field, but will need to check whether there isn't something that depends on it somewhere.

src/json/index.ts
mbg: See the relevant commit message. The existing code inferred an unhelpful type for the results of this function, since we require the value of `required` in the type. I could have made a different change here to still have the `required` parameter, but that would have complicated the function more than necessary considering all consumers create validators for required properties.

sam-robson (approved):

PR #4060, opened 2026-07-30
Fix `releases/v3` branch

mbg:
On every recent CodeQL Action release, we have run into a merge conflict on the PR that merges back changes to the `releases/v3` branch. That is because part of the mergeback PR process is to revert the most recent commit titled "Rebuild". Each time, the conflict has been resolved by rebuilding the JS and committing the resolution as "Rebuild", causing it to be discarded again the next time.

This PR mirrors the mergeback process, with the exception of a separate commit to resolve the merge conflict, by:

- Reverting the "Update version" and "Rebuild" commits added during the last actual mergeback process.
- Resolving the merge conflict.
- Re-applying the "Update version" commit
- Rebuilding

During the next actual mergeback process, reverting the last two commits should then not revert the merge conflict resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,62 +2,62 @@
 
 See the [releases page](https://github.com/github/codeql-action/releases) for the relevant changes to the CodeQL CLI and language packs.
 
-## 4.37.4 - 29 Jul 2026
+## 3.37.4 - 29 Jul 2026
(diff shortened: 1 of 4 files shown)


oscarsj (approved):
👍🏾  thanks, this was a confusing step in the process

Issue #3367, opened 2025-12-16
'init' action doesn't seem to retry on socket exceptions and network hiccups


hossein-nexl:
Using `github/codeql-action/init@v4` in our CodeQL Advanced workflow like this:

```yaml
- name: Initialize CodeQL
  uses: github/codeql-action/init@v4
  with:
    languages: ruby
    build-mode: none
    config-file: ./.github/codeql.yml
```

```yaml
# ./.github/codeql.yml

paths-ignore:
  - path/to/file1.rb
  - path/to/file2.rb
```

I randomly received the following error today (December 16), which by the look of it was using tag [v4.31.8](https://github.com/github/codeql-action/releases/tag/v4.31.8) released on December 12:

```
/usr/bin/tar --version
tar (GNU tar) 1.35
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
Found gnu tar version 1.35.
Did not find CodeQL tools version 2.23.8 in the toolcache.
Using CodeQL CLI version 2.23.8 sourced from https://github.com/github/codeql-action/releases/download/codeql-bundle-v2.23.8/codeql-bundle-linux64.tar.zst .
Downloading CodeQL tools from https://github.com/github/codeql-action/releases/download/codeql-bundle-v2.23.8/codeql-bundle-linux64.tar.zst . This may take a while.
Streaming the extraction of the CodeQL bundle.
node:events:486
      throw er; // Unhandled 'error' event
      ^

Error: socket hang up
    at TLSSocket.socketOnEnd (node:_http_client:598:25)
    at TLSSocket.emit (node:events:520:35)
    at endReadableNT (node:internal/streams/readable:1701:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on Writable instance at:
    at eventHandlers.<computed> (/home/runner/work/_actions/github/codeql-action/v4/lib/init-action.js:81114:28)
    at ClientRequest.emit (node:events:508:28)
    at emitErrorEvent (node:_http_client:107:11)
    at TLSSocket.socketOnEnd (node:_http_client:598:5)
    at TLSSocket.emit (node:events:520:35)
    at endReadableNT (node:internal/streams/readable:1701:12)
[...]

PR #4061, opened 2026-07-30
Handle network errors when streaming the CodeQL bundle download

henrymercer:
When we stream the download and extraction of the CodeQL bundle, a network error such as `ECONNRESET` terminates the `init` Action rather than falling back to downloading the bundle before extracting it. This is because we never attached an `error` listener to the request returned by `https.get`, so `follow-redirects` emits the error as an unhandled `error` event, which is fatal. See [#3367](https://github.com/github/codeql-action/issues/3367).

Attaching the listener is enough to turn the crash into the fallback path we already have. This PR also switches the pipe from the response into `tar` over to `stream.pipeline`, since `tar.pipe(stdin).on("error", ...)` only listens for errors on `tar`'s standard input: an error on the response itself was unhandled, and would have left `tar` waiting indefinitely on a pipe that was never going to be closed. Finally, we abort the request if it stalls for five minutes so that a hung connection can't take the whole job with it.

I've deliberately not added a retry loop around the streaming attempt: `toolcache.downloadTool`, which we fall back to, already retries three times with exponential backoff and already avoids retrying non-transient status codes. The bug here was never that we don't retry, it's that we crashed before we got the chance.

### Risk assessment

- **Low risk:** This only affects what happens when the streaming download fails, which currently means the job dies. Both new failure paths are covered by unit tests that were confirmed to fail without the fix.

Workflow types:

Products:

Environments:

- **Unit tests** - New tests simulate a mid-connection `ECONNRESET` and assert that we fall back rather than crashing.
- **End-to-end tests** - The existing bundle toolcache PR checks continue to exercise the streaming download on Linux.

- **Telemetry** - Streaming failures now surface as the existing fallback warning and show up in the download duration status field, rather than as a hard job failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- Fixed a bug where a network error while streaming the download of the CodeQL bundle could terminate the `init` Action instead of falling back to downloading the bundle before extracting it. [#3367](https://github.com/github/codeql-action/issues/3367)
 
 ## 4.37.4 - 29 Jul 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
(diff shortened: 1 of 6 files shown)


mbg (approved):
Nice, thanks for improving this!

PR #4062, opened 2026-07-31
ci: update Enterprise release checkout credential

sam-robson:
Updates the supported-version workflow to use the federated read-only credential when checking out Enterprise release metadata.

No workflow behavior changes beyond the credential reference.


mbg (commented):
@sam-robson I've had a look and have some questions I'll ask elsewhere.

navntoft (approved):

PR #4072, opened 2026-08-04
Add support for Linux Arm64 runners

redsun82:
The CodeQL CLI will ship native `linux-arm64` artifacts, but the Action does not yet recognize that platform: it rejects the CLI as an unsupported platform/architecture combination and, even if it got past that, would resolve the wrong bundle asset. This change teaches the Action about `linux/arm64` so it accepts the CLI and downloads the correct bundle.

### Approach

Two small semantic edits, plus faithful test/mock coverage:

- `src/cli-errors.ts` - add `["linux", "arm64"]` to `SUPPORTED_PLATFORMS` so `isUnsupportedPlatform()` no longer rejects it.
- `src/setup-codeql.ts` - make the linux branch of `getCodeQLBundleName()` architecture-aware, resolving `codeql-bundle-linux-arm64.tar.zst` on arm64 and `codeql-bundle-linux64` elsewhere. This flows automatically to the download and nightlies URL builders.

Arm64 is shipped as a per-platform bundle only and is deliberately excluded from the combined all-platforms gzip (which is near the release-asset size limit), so it is important that resolution lands on the exact per-platform asset name rather than the combined-gzip fallback.

The bundle-download mock in `src/testing-utils.ts` was updated to mirror the same arch logic so it stays faithful on arm64 hosts (x64 behavior is unchanged). Tests were updated accordingly: the now-supported `linux/arm64` case was removed from the unsupported-platform list in `src/cli-errors.test.ts`, and `src/setup-codeql.test.ts` gained an explicit `arch` per case plus a new `linux/arm64` case.

x64, macOS, and Windows behavior is unchanged; the only new behavior is on Linux arm64 runners, which do not work at all today.

### Risk assessment

- **Low risk:** Changes are documentation or test only, or have been fully tested and validated. This change only adds behavior for Linux arm64 (a platform that fails today) and leaves all existing platforms untouched; it is covered by unit tests.

Workflow types:

Products:

Environments:

Only relevant when running on a Linux arm64 runner; other architectures are unaffected.

- **Unit tests** - I am depending on unit test coverage (i.e. tests in `.test.ts` files) for the new arch branch and platform-support change.
- **Test repository** - This change will be tested on a Linux arm64 test repository before merging.

- **Rollback** - Change can only be disabled by rolling back the release or releasing a new version with a fix. There is no x64/macOS/Windows impact, so the blast radius is limited to arm64 runners.

- **Other** - Failures would surface as CodeQL bundle download or platform errors in the Action logs on Linux arm64 runners.

- **Special considerations** - This relies on the CLI publishing the per-platform `codeql-bundle-linux-arm64.tar.zst` asset. It should only be merged once that asset is being produced.


redsun82:
@mbg or @henrymercer do you mind giving a very early quick look whether this seems sensible? We can postpone merging until we decide to go with it in the releases. Tomorrow I plan on running some tests by hand on this branch with the nightly release, so I don't mind if the first review is shallow.

src/setup-codeql.test.ts
```diff
@@ -120,24 +120,35 @@ test.serial(
...
+    expectedBundleName: "codeql-bundle-linux-arm64.tar.zst",
+    expectedCompressionMethod: "zstd",
+  },
   {
     platform: "darwin",
+    arch: "arm64",
     tarSupportsZstd: true,
     expectedBundleName: "codeql-bundle-osx64.tar.zst",
     expectedCompressionMethod: "zstd",
   },
```
henrymercer, line 141: Can you add a test for darwin x64?

henrymercer (commented):
Looks good to start testing!  Can we add a PR check based on `ubuntu-24.04-arm`?

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The CodeQL Action now supports running on Linux Arm64 runners, downloading the native `linux-arm64` CodeQL bundle. [#4072](https://github.com/github/codeql-action/pull/4072)
```
mbg: It's the CLI that supports it. The Action changes are just in support of it: 
```suggestion
- The CodeQL Action now supports CodeQL releases that are compatible with Linux Arm64 and download the native `linux-arm64` CodeQL bundle when available. [#4072](https://github.com/github/codeql-action/pull/4072)
```

mbg (commented):
Thanks for preparing these changes! They generally look reasonable. A few points that we'd probably want to tackle on top of them:

- Add at least one end-to-end test that runs on arm64 Linux. (This might also require updating `sync.ts`) Possibly run other e2e tests on arm64 as well.
- It would be nice if we could provide a useful error message if someone attempts to use an older CLI that doesn't support arm64 (e.g. by checking the release assets and provide an appropriate error message if on that platform)

src/setup-codeql.test.ts
redsun82: makes sense, will do!

pr-checks/checks/linux-arm64.yml
```diff
@@ -0,0 +1,31 @@
...
+  - uses: ./../action/init
+    with:
+      languages: javascript,python,go
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
+  - name: Build Go code
+    run: go build main.go
+  - uses: ./../action/analyze
+    with:
+      skip-queries: true
+      upload-database: false
```
henrymercer, line 22: Perhaps we should run the queries, to check the database can be finalized and that it works to some extent.  I'm not too concerned about the additional run time, as long as this doesn't become the longest check.

src/cli-errors.ts
```diff
@@ -8,6 +8,7 @@ import { ConfigurationError } from "./util";
 
 const SUPPORTED_PLATFORMS = [
   ["linux", "x64"],
+  ["linux", "arm64"],
```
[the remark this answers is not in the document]
henrymercer: This is a good idea, we should stub the OS and architecture in "wrapCliConfigurationError - supported platform" rather than just testing on the OS / architecture on which we happen to be running.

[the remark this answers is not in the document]
redsun82: good point, done in 025009006. I turned the supported-platform test into a serial loop over all the supported platform/arch pairs (linux/arm64 included), stubbing `process.platform` and `process.arch` for each and asserting `wrapCliConfigurationError` hands back the original `CliError`. so it doesn't depend on whatever host the tests happen to run on any more.

pr-checks/checks/linux-arm64.yml
redsun82: makes sense, I dropped `skip-queries` so the check now actually runs the default queries and finalizes each database (regenerated `__linux-arm64.yml` too). gives us a bit of end-to-end signal that analysis works on arm64, and it shouldn't come anywhere near being the longest check :slightly_smiling_face:

pr-checks/checks/linux-arm64.yml
```diff
@@ -0,0 +1,30 @@
...
+# The native linux-arm64 CodeQL bundle is only available in recent CLI releases, so we restrict this
+# check to `nightly-latest`, which is guaranteed to ship it. Older stable versions do not have an
+# arm64 asset, and `prepare-test` would resolve an x64 bundle URL for them on this runner.
+versions:
+  - nightly-latest
+installGo: true
+steps:
+  - uses: ./../action/init
+    with:
+      languages: javascript,python,go
```
mbg, line 15: Minor and doesn't have to be addressed in this PR, but it might be good to analyse all languages that we support on arm64 (ignore if that's already what this list is)

pr-checks/checks/linux-arm64.yml
```diff
@@ -0,0 +1,30 @@
...
+      tools: ${{ steps.prepare-test.outputs.tools-url }}
+  - name: Build Go code
+    run: go build main.go
+  - uses: ./../action/analyze
+    with:
+      upload-database: false
+  - name: Assert databases exist
+    run: |
+      cd "$RUNNER_TEMP/codeql_databases"
+      for lang in javascript python go; do
```
mbg, line 25: Minor: It would be nice if the list of languages wasn't duplicated between the `init` input and here.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-No user facing changes.
+- The CodeQL Action now supports CodeQL releases that are compatible with Linux Arm64 and download the native `linux-arm64` CodeQL bundle when available. [#4072](https://github.com/github/codeql-action/pull/4072)
```
mbg, line 7: Minor: "download" -> "downloads" to make the tense consistent

mbg (approved):
Looks good, thanks for adding the tests! Just a few very minor things that can mostly be done in follow-ups or don't need to be addressed before merging.

redsun82:
@mbg thanks, I'll merge as is for now but spin a follow-up right away 👍

PR #4079, opened 2026-08-06
Support CodeQL versions, ranges, release offsets, and nightly transitions

felickz:
## What this changes

The `tools` input for `github/codeql-action/init` and `github/codeql-action/setup-codeql` can now select CodeQL using a version, a SemVer range, a release offset, or an automatic nightly transition.

The Action resolves the requested version and downloads the correct bundle for the runner's operating system and architecture.

## Select an exact version

```yaml
- uses: github/codeql-action/init@v4
  with:
    languages: javascript-typescript
    tools: 2.20.1
```

An optional `v` prefix is also accepted:

```yaml
tools: v2.20.1
```

This replaces the need to provide a full platform-specific bundle URL.

## Select a SemVer range

The input supports npm-style SemVer ranges and selects the newest stable CodeQL release satisfying the range:

```yaml
tools: 2.20.x
tools: 2.x
tools: ~2.20.0
tools: ^2.20.0
tools: ">=2.20.0 <2.21.0"
```

If no release satisfies the range, the error reports the available stable CodeQL version range.

## Stay behind the latest release

Use `latest-N` to select a stable release by its position behind the newest stable release:

```yaml
tools: latest-0  # Newest stable release
tools: latest-1  # One stable release behind
tools: latest-2  # Two stable releases behind
```

If the offset is outside the available release history, the error reports how many stable releases are available.

## Include prereleases

Use `latest-prerelease` to select the newest semantic CodeQL bundle while allowing GitHub releases marked as prereleases:

```yaml
tools: latest-prerelease
```

Stable releases remain eligible, so the newest eligible version wins. Draft and date-based bundle releases are ignored. Existing `latest-N` and SemVer range inputs remain stable-only.

## Use nightly until a release is available

Use `nightly-until-<version>` to test a feature expected in a future CodeQL release:

```yaml
tools: nightly-until-2.26.3
```

The Action uses the latest nightly until a published CodeQL release at or above the threshold exists. GitHub prereleases are eligible. It then switches to the newest eligible published release.

## Use nightly until the normal default is ready

Use `nightly-until-default-<version>` when you want to return to the Action's normal default selection only after the CodeQL CLI that would actually be selected contains the required feature:

```yaml
tools: nightly-until-default-2.26.3
```

The Action first resolves its normal default source and continues using `nightly` while that selected CLI is below the threshold. Once the CLI that would actually be used is at or above the threshold, normal default selection resumes.

This form does not list CodeQL releases or switch merely because a release was published.

## Existing inputs remain supported

Existing `tools` values—including URLs, local paths, `latest`, `nightly`, `linked`, and `toolcache`—continue to work as before.

### Risk assessment

**Risk level: Low risk**

The new behavior is opt-in through previously unsupported `tools` input values. Existing inputs and workflows remain unchanged. The implementation has unit coverage and was exercised in a dedicated cross-platform test workflow.

Workflow types:

- **Advanced setup** - Impacts users who explicitly configure the `tools` input.
- **Managed** - May impact managed workflows when `github-codeql-tools` supplies one of the new values.

Products:

- **Code Scanning** - Impacts analyses using the new `tools` values.
- **Code Quality** - Impacts analyses using the new `tools` values.
- **Other first-party** - May impact first-party analyses using the new `tools` values.

Environments:

- **Dotcom** - Supports GitHub.com and GitHub Enterprise Cloud with Data Residency.
- **GHES** - Exact/default downloads retain existing instance-aware behavior. Dynamic release discovery uses the canonical public GitHub.com release inventory without forwarding enterprise credentials.

- **Test repository** - Tested through a dedicated matrix workflow using Python analysis across Ubuntu, Windows, and macOS.
- **Unit tests** - Added focused coverage for exact versions, SemVer ranges, release offsets, prereleases, nightly transitions, malformed inputs, release filtering, overlay-aware defaults, and GHES toolcache overrides.
- **End-to-end tests** - PR checks completed successfully except for unavailable macOS xlarge runners, which received no runner and executed no steps.
- **Other** - The matrix exercised 37 input forms across three operating systems. Expected unsupported inputs produced clear errors, while supported inputs resolved the expected versions and platform-specific assets. (see [run](https://github.com/forks-felickz/codeql-action/actions/runs/31056695944?pr=3))

- **Rollback** - The change can be disabled by rolling back the release or releasing a new version with a fix.
- **Other** - The behavior is opt-in. Users can immediately return to existing supported inputs such as an exact bundle URL, `linked`, `latest`, `nightly`, `toolcache`, or an omitted `tools` value.

- **Telemetry** - Existing Action telemetry records the tools source and selected version.
- **Other** - Version resolution, download, compatibility, and malformed-input failures produce explicit Action errors describing the requested input and available release range.

- [x] Add a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.


CHANGELOG.md
```diff
@@ -4,7 +4,10 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
...
+- The `tools` input also now accepts a CodeQL CLI version range, for example `2.24.x`, `2.x`, `~2.24.0`, or `^2.24.0`, as well as `latest-<N>` (for example `latest-1`), which resolves to the CodeQL Bundle release `N` stable releases before the most recent one.
+- The `tools` input also now accepts `latest-prerelease`, which resolves to the most recent CodeQL Bundle release, whether it is a stable release or a GitHub prerelease, as well as `nightly-until-<version>` (for example `nightly-until-2.24.0`), which resolves to the most recent CodeQL Bundle release satisfying that version threshold, considering GitHub prereleases too, or falls back to the latest nightly version of the CodeQL tools if no such release exists.
+- The `tools` input also now accepts `nightly-until-default-<version>` (for example `nightly-until-default-2.24.0`), which uses the default CodeQL CLI version for this environment if it satisfies that version threshold, or falls back to the latest nightly version of the CodeQL tools otherwise.
```
mbg, line 10: This is probably an excessive amount of release notes for this group of changes, especially given that we would not expect most people to use any of this. Consider condensing it to just something like 
```suggestion
- The `tools` input for the `init` and `setup-codeql` Actions now accepts additional input formats. See the respective Action specifications for more details. [#4079](https://github.com/github/codeql-action/pull/4079)
```

init/action.yml
```diff
@@ -10,6 +10,24 @@ inputs:
...
+      - A special value `latest-prerelease` (matching is case insensitive) which uses the most
+        recent CodeQL Bundle release, whether it is a stable release or a GitHub prerelease, or
+      - `nightly-until-<version>`, for example `nightly-until-2.24.0` or
+        `NIGHTLY-UNTIL-2.24.0` (matching is case insensitive), which uses the most recent CodeQL
+        Bundle release satisfying that version threshold, considering GitHub prereleases too, or
+        falls back to the latest nightly version of the CodeQL tools if no such release exists, or
+      - `nightly-until-default-<version>`, for example `nightly-until-default-2.24.0` (matching
+        is case insensitive), which uses the default CodeQL CLI version if it satisfies that
+        version threshold, or falls back to the latest nightly version of the CodeQL tools
+        otherwise, or
```
mbg, line 30: Currently, we (deliberately) don't document all existing inputs here because they generally cause more harm than good when used incorrectly. We should think carefully about which formats are worth documenting here once we have agreed on all the formats.

setup-codeql/action.yml
```diff
@@ -10,6 +10,24 @@ inputs:
...
+      - A special value `latest-prerelease` (matching is case insensitive) which uses the most
+        recent CodeQL Bundle release, whether it is a stable release or a GitHub prerelease, or
+      - `nightly-until-<version>`, for example `nightly-until-2.24.0` or
+        `NIGHTLY-UNTIL-2.24.0` (matching is case insensitive), which uses the most recent CodeQL
+        Bundle release satisfying that version threshold, considering GitHub prereleases too, or
+        falls back to the latest nightly version of the CodeQL tools if no such release exists, or
+      - `nightly-until-default-<version>`, for example `nightly-until-default-2.24.0` (matching
+        is case insensitive), which uses the default CodeQL CLI version if it satisfies that
+        version threshold, or falls back to the latest nightly version of the CodeQL tools
+        otherwise, or
```
mbg, line 30: Same point as for `init`.

src/api-client.ts
```diff
@@ -181,6 +182,38 @@ export function getApiClientWithExternalAuth(
...
+ * history -- regardless of which GitHub instance this Action itself is running on.
+ *
+ * This must be used instead of `getApiClient`/`getApiClientWithExternalAuth` whenever the current
+ * GitHub instance is not GitHub.com (for example GitHub Enterprise Server or GHEC with data
+ * residency), since in that case the current instance's authentication token cannot authenticate
+ * to GitHub.com. Sending it there anyway would both fail and needlessly expose that token to a
+ * different host. Reading public data, such as the release history of a public repository, does
+ * not require authentication.
+ */
+export function getUnauthenticatedApiClientForDotcom(
```
mbg, line 197: I don't love having this, both because it is messy to use and duplicates a lot of `getApiClientWithExternalAuth`. I'd probably prefer us to look into whether we can better identify/track which instance a credential is for and then "do the right thing" automatically (i.e. use the token if it is for the instance the request/the Octokit we are initialising is for, or not otherwise).

mbg (commented):
Thank you for taking initiative on this work! Like I wrote elsewhere, this had been on my radar as well after adding the new repository property (and bumping quite hard into the 75-character constraint while testing it 😄). Though also read why I didn't work on this there.

I have added a few detailed comments attached to relevant places in the changes, but there are some more significant high-level points to look at as well:

1. A common concern with all the new formats is that `tools` can be pointed at either just a CLI release or (more commonly) a CLI bundle release. The new formats all implicitly refer to the latter, which makes sense since it's more common, but means that there is no equivalent mechanism for just the CLI.
1. This PR adds a bunch of different new input formats all in one go. We should think carefully about which of these we actually want, particularly since this is a relatively niche feature that is only useful for troubleshooting / testing / temporarily resolving problems. My concerns focus around:
    - Ongoing maintenance of these in an already complex context for determining the CLI version to install. This is made worse by the extra `getUnauthenticatedApiClientForDotcom` logic.
    - Potential overlap between formats, which can be alleviated with better test coverage.
    - Agreeing on all the new formats. This would be easier if each was in a separate PR.
1. There is no new test coverage for the new input formats and the corresponding changes in `setup-codeql.ts`.
1. The logic in `setup-codeql.ts` is already a nightmare as-is and has long been due a refactor. We shouldn't tackle this at the same time as introducing new formats, but adding more does make it harder to do later.

### Thoughts on the new formats and semantics

- The "bare version" and `semver` formats make sense and are probably uncontroversial in terms of notation with the caveat of the CLI / CLI bundle distinction mentioned above.
- I am less confident in the `latest-<N>` format. This is because it's not immediately obvious if `-<N>` means "whatever version before the `latest`" (e.g. if `2.27.2` is the `latest`, it would get you `2.27.1`) or "the previous minor release" (e.g. if `2.27.2` is `latest`, it would get you the latest patch-level release of `2.26.x`), ...
  - I am also unsure of why someone would want this, because whatever the behaviour, it would result in sudden changes to the CLI version. So customers would lose out of new features when they are released while simultaneously making it harder to identify issues when they are finally rolled over to a newer release.  
- I would avoid `latest-prerelease` since we do not really use `prerelease` releases anymore (certainly for bundles) and internally assume that these aren't used by anyone other than ourselves.
- `nightly-until-default-<version>` is a bit horrible. We would only want customers to use `nightly` releases if they resolve an otherwise blocking issue for them, so we generally don't want customers to use this themselves or have to figure out when to stop using nightlies. Implementation-wise, we might want to check against the `linked` version rather than querying the releases via the API.

I don't want to make this more complicated than it already is, but  I had some thoughts on possible, more general formats we might want to consider for the input specifications. E.g. `cli:^2.27.0` or `bundle:nightly` to indicate whether the specification relates to a CLI release or a CLI bundle release. Instead of `nightly-until-default-2.28.0`, we could have fallbacks like `bundle:^2.28.0,nightly` meaning "use at least `2.28.0`, but if unavailable use `nightly`".

----

In conclusion, I think it would be good to narrow the scope of this down to:

- Introducing short-hands for inputs that are already supported now. E.g. the "bare" version format that is essentially just expanded to a URL.
- Adding tests for that.

And then going from there.

PR #4080, opened 2026-08-06
Determine the overlay minimum disk space requirement from feature flags

henrymercer:
Overlay analysis skips runners that don't have enough free disk space. Until now that threshold was a single hard-coded value with one flag-gated alternative: 20 GB normally, or 14 GB when `overlay_analysis_resource_checks_v2` was enabled. That gave us exactly one step to roll out, and reaching a lower limit meant another flag and another release.

This replaces that with a set of flags that each name a threshold, so we can tune the requirement without shipping code:

- `overlay_analysis_min_disk_8_gb`
- `overlay_analysis_min_disk_9_gb`
- `overlay_analysis_min_disk_10_gb`
- `overlay_analysis_min_disk_11_gb`
- `overlay_analysis_min_disk_12_gb`
- `overlay_analysis_min_disk_13_gb`

When several are enabled the lowest threshold wins, so we can roll a lower limit out to a subset of repositories without having to disable the flag above it first. When none are enabled the default of 14 GB applies.

`overlay_analysis_resource_checks_v2` is removed and its 14 GB value becomes the unconditional baseline, so the 20 GB limit is gone. Note that this part is not itself behind a flag — anyone previously on the 20 GB path moves to 14 GB when this ships. On GHES that only affects users who have explicitly opted into overlay analysis via `CODEQL_ACTION_OVERLAY_ANALYSIS`, since `overlay_analysis` defaults to off there. Anyone setting `CODEQL_ACTION_OVERLAY_ANALYSIS_RESOURCE_CHECKS_V2` will find it no longer has any effect.

Thresholds stay in decimal MB (`N * 1000 MB`), matching the existing bytes-per-MB convention in the disk check, so the effective byte values are unchanged from the previous 14 GB path. The check also now logs the available and required space at debug level when it passes, so it's possible to tell from a run log which flag took effect.

### Risk assessment

Workflow types:

Products:

Environments:

- **Feature flags** - Lowering the threshold below the 14 GB default is fully controlled by the new feature flags and can be reverted by disabling them.
- **Rollback** - The change of the default threshold from 20 GB to 14 GB is not behind a feature flag, and would need a rollback or a follow-up release to revert. Overlay analysis as a whole can still be disabled via the `overlay_analysis` feature flag, the `CODEQL_OVERLAY_DATABASE_MODE` environment variable, or the disable-overlay repository property.

- **Telemetry** - I rely on existing telemetry.
    - **Dashboards** - I will watch relevant dashboards for issues after the release.

- **Special considerations** - The six new feature flags need to exist before this is released, otherwise the threshold stays at the 14 GB default.

- [ ] Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change.
- [ ] Confirm the [readme](https://github.com/github/codeql-action/blob/main/README.md) and docs have been updated if necessary.


src/feature-flags.ts
```diff
@@ -121,12 +121,19 @@ export enum Feature {
...
+   * enabled.
    */
-  OverlayAnalysisResourceChecksV2 = "overlay_analysis_resource_checks_v2",
+  OverlayAnalysisMinDisk8Gb = "overlay_analysis_min_disk_8_gb",
+  OverlayAnalysisMinDisk9Gb = "overlay_analysis_min_disk_9_gb",
+  OverlayAnalysisMinDisk10Gb = "overlay_analysis_min_disk_10_gb",
+  OverlayAnalysisMinDisk11Gb = "overlay_analysis_min_disk_11_gb",
+  OverlayAnalysisMinDisk12Gb = "overlay_analysis_min_disk_12_gb",
+  OverlayAnalysisMinDisk13Gb = "overlay_analysis_min_disk_13_gb",
+  OverlayAnalysisPython = "overlay_analysis_python",
```
mbg: Minor, but maybe worth doing since you need to rebase these changes anyway / resolve the merge conflicts: it might be good to have a docs comment for each FF individually, particularly to explain the intended behaviour of each FF more clearly. E.g. something along the lines of "if all FFs are off, the minimum is determined by the `OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_MB` constant in `config-utils.ts` - otherwise, the FFs decrease the minimum to the value indicated by their name. The lowest value out of all enabled FFs is picked."

src/feature-flags.ts
```diff
@@ -121,12 +121,19 @@ export enum Feature {
...
-  OverlayAnalysisPython = "overlay_analysis_python",
   /**
-   * Controls whether lower disk space requirements are used for overlay hardware checks.
-   * Has no effect if `OverlayAnalysisSkipResourceChecks` is enabled.
+   * Feature flags that lower the amount of available disk space that the overlay hardware check
+   * requires. The lowest threshold that is enabled takes effect; if none are enabled, the default
+   * threshold applies. These flags have no effect if `OverlayAnalysisSkipResourceChecks` is
+   * enabled.
    */
-  OverlayAnalysisResourceChecksV2 = "overlay_analysis_resource_checks_v2",
```
mbg: I cherry-picked the change to remove `OverlayAnalysisResourceChecksV2` in https://github.com/github/codeql-action/pull/4083 which was merged and has shipped. This PR will need to be updated based on that.

src/config-utils.ts
```diff
@@ -101,19 +101,28 @@ export { type Config } from "./config/action-config";
...
+ * Feature flags that lower the minimum available disk space required to perform
+ * overlay analysis, paired with the threshold (in MB) that each one enables.
+ *
+ * If several of these are enabled, the lowest threshold takes effect.
  */
-const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_MB = 14000;
-const OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_BYTES =
-  OVERLAY_MINIMUM_AVAILABLE_DISK_SPACE_V2_MB * 1_000_000;
+const OVERLAY_MINIMUM_DISK_SPACE_FEATURES: ReadonlyArray<
+  [FeatureWithoutCLI, number]
```
mbg: Minor: No strong feelings, but why a tuple instead of an object?

mbg (dismissed):
As discussed elsewhere, let's move ahead with this for now. The changes look fine I think, but the branch will need to be updated to account for the changes I cherry-picked into https://github.com/github/codeql-action/pull/4083 that has already been merged/released. Let me know if you'd like me to take over updating this branch, since I am responsible for the conflicts.

src/feature-flags.ts
henrymercer: Good idea, done.

src/config-utils.ts
henrymercer: No particular reason — changed.

src/config-utils.ts
```diff
@@ -103,26 +103,21 @@ export { type Config } from "./config/action-config";
...
-  [Feature.OverlayAnalysisMinDisk13Gb, 13000],
-];
+const OVERLAY_MINIMUM_DISK_SPACE_MB_BY_FEATURE = {
+  [Feature.OverlayAnalysisMinDisk8Gb]: 8000,
+  [Feature.OverlayAnalysisMinDisk9Gb]: 9000,
+  [Feature.OverlayAnalysisMinDisk10Gb]: 10000,
+  [Feature.OverlayAnalysisMinDisk11Gb]: 11000,
+  [Feature.OverlayAnalysisMinDisk12Gb]: 12000,
+  [Feature.OverlayAnalysisMinDisk13Gb]: 13000,
+} satisfies Partial<Record<FeatureWithoutCLI, number>>;
```
mbg, line 120: Minor: Consider whether it would make sense to define a type for the relevant subset of `FeatureWithoutCLI` so that we can drop the `Partial` here and ensure that all the `OverlayAnalysisMinDisk` features have a size mapping.

mbg (approved):
LGTM as-is, with just one minor suggestion.

src/config-utils.ts
henrymercer: I'll leave this one as is.

PR #4081, opened 2026-08-07
Persist CodeQL version output to file rather than environment

mario-campos:
Currently, the output of `codeql version` is cached and persisted between steps as an environment variable. This is fine, for now, but, as we begin to cache more of the CLI, we should persist that output to a file instead because:

1. There is a limit to the size of a process' environment. On GitHub Actions, the runners already have a populated environment. And, the output of these CLI commands may be verbose.
2. There is data to suggest that a larger environment has performance implications (more environment variables -> more memory for the kernel to copy during `execve(2)`).

To that end, and as a precaution, this PR replaces `$CODEQL_ACTION_CLI_VERSION_INFO` with `$CODEQL_ACTION_TEMP/version.json`, a file-backed cache for `codeql version`.

**Changes**

* Moved the CodeQL CLI version caching logic from `util.ts` (which used environment variables) into a new module `src/cli/output-cache.ts`, which now uses a temporary on-disk file (`codeql-action-command-cache.json`) for cross-step persistence. This includes new functions for caching, retrieving, and resetting the cached version, as well as improved validation of persisted data.
* Moved the `VersionInfo` type definition from `codeql.ts` to a new shared file `src/cli/types.ts`, and updated all references to use the new location. This centralizes type definitions to avoid cyclic dependencies.
* Removed the now-obsolete `CODEQL_VERSION_INFO` environment variable from `EnvVar` in `environment.ts`.

---

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/cli/output-cache.ts
+++ b/src/cli/output-cache.ts
@@ -9,21 +9,21 @@ import type { VersionInfo } from "./types";
 /**
  * The keys of the command cache. Each key corresponds to a command whose output we cache.
  */
-enum CommandCacheKey {
+export enum CommandCacheKey {
   Version = "version",
 }
 
 /**
  * The mapping of CLI commands to the types of the output of each command that we cache.
  */
-type CommandCacheKeyOutputMap = {
+export type CommandCacheKeyOutputMap = {
   [CommandCacheKey.Version]: VersionInfo;
 };
 
 /**
  * The type of the command cache that is persisted to disk.
  */
-interface CommandCacheRecord<K extends CommandCacheKey> {
+export interface CommandCacheRecord<K extends CommandCacheKey> {
   cmd: string;
   entries: Map<K, CommandCacheKeyOutputMap[K]>;
 }
--- a/src/cli/output-cache.ts
+++ b/src/cli/output-cache.ts
@@ -23,7 +23,7 @@ export type CommandCacheKeyOutputMap = {
 /**
  * The type of the command cache that is persisted to disk.
  */
-export interface CommandCacheRecord<K extends CommandCacheKey> {
+export interface OutputCache<K extends CommandCacheKey> {
   cmd: string;
   entries: Map<K, CommandCacheKeyOutputMap[K]>;
 }
@@ -109,7 +109,7 @@ export function getCachedCodeQlVersion(
     return undefined;
   }
   if (
-    !isCommandCacheRecord(persisted) ||
+    !isOutputCache(persisted) ||
     (cmd !== undefined && persisted.cmd !== cmd)
   ) {
     return undefined;
@@ -139,15 +139,11 @@ function isVersionInfo(x: unknown): x is VersionInfo {
 }
 
 /**
- * Determines whether a value is a `CommandCacheRecord` object.
+ * Determines whether a value is a `OutputCache` object.
  * @param x The value to test
  */
-function isCommandCacheRecord(
-  x: unknown,
-): x is CommandCacheRecord<CommandCacheKey.Version> {
-  const candidate = x as Partial<
-    CommandCacheRecord<CommandCacheKey.Version>
-  > | null;
+function isOutputCache(x: unknown): x is OutputCache<CommandCacheKey.Version> {
+  const candidate = x as Partial<OutputCache<CommandCacheKey.Version>> | null;
   return (
     typeof candidate === "object" &&
     candidate !== null &&
--- a/src/cli/output-cache.ts
+++ b/src/cli/output-cache.ts
@@ -59,14 +59,14 @@ function getCommandCacheFilePath(env: Env): string {
 
 /**
  * Caches the CodeQL CLI version both in-memory and on disk.
+ * @param env The environment variables to use.
  * @param cmd The path to the CodeQL CLI.
  * @param version The version information to cache.
- * @param env The environment variables to use.
  */
 export function cacheCodeQlVersion(
+  env: Env,
   cmd: string,
   version: VersionInfo,
-  env: Env,
 ): void {
   if (cachedCodeQlVersion !== undefined) {
     throw new Error("cacheCodeQlVersion() should be called only once");
--- a/src/cli/output-cache.ts
+++ b/src/cli/output-cache.ts
@@ -1,0 +1,17 @@
+import path from "path";
+
+import { getTemporaryDirectory } from "../actions-util";
+
+/**
+ * The name of the temporary file that backs the on-disk cache of
+ * CLI responses between workflow steps.
+ */
+const COMMAND_CACHE_FILENAME = "codeql-action-command-cache.json";
+
+/**
+ * Returns the path to the temporary file that backs the
+ * on-disk cache of CLI responses between workflow steps.
+ */
+function getCommandCacheFilePath(): string {
+  return path.join(getTemporaryDirectory(), COMMAND_CACHE_FILENAME);
+}
--- a/src/cli/output-cache.ts
+++ b/src/cli/output-cache.ts
@@ -25,7 +25,9 @@ export type CommandCacheKeyOutputMap = {
  */
 export interface OutputCache<K extends CommandCacheKey> {
   cmd: string;
-  entries: Map<K, CommandCacheKeyOutputMap[K]>;
+  entries: {
+    [P in K]: CommandCacheKeyOutputMap[K];
+  };
 }
 
 /**
@@ -70,13 +72,17 @@ export function cacheCodeQlVersion(
     throw new Error("cacheCodeQlVersion() should be called only once");
   }
   cachedCodeQlVersion = version;
+  const outputCache = {
+    cmd,
+    entries: { [CommandCacheKey.Version]: version },
+  } satisfies OutputCache<CommandCacheKey.Version>;
   // Persist the version so that subsequent Actions steps, which run in separate
   // processes, can reuse it rather than invoking `codeql version` again. We
   // record the CLI path so that a different step using a different CodeQL bundle
   // doesn't pick up a stale version.
   fs.writeFileSync(
     getCommandCacheFilePath(env),
-    JSON.stringify({ cmd, entries: { [CommandCacheKey.Version]: version } }),
+    JSON.stringify(outputCache),
     "utf8",
   );
 }
--- a/src/cli/output-cache.ts
+++ b/src/cli/output-cache.ts
@@ -9,25 +9,14 @@ import type { VersionInfo } from "./types";
 /**
  * The keys of the command cache. Each key corresponds to a command whose output we cache.
  */
-export enum CommandCacheKey {
-  Version = "version",
-}
-
-/**
- * The mapping of CLI commands to the types of the output of each command that we cache.
- */
-export type CommandCacheKeyOutputMap = {
-  [CommandCacheKey.Version]: VersionInfo;
-};
+export type CommandCacheKey = string;
 
 /**
  * The type of the command cache that is persisted to disk.
  */
-export interface OutputCache<K extends CommandCacheKey> {
+export interface OutputCache {
   cmd: string;
-  entries: {
-    [P in K]: CommandCacheKeyOutputMap[K];
-  };
+  entries: Record<CommandCacheKey, unknown>;
 }
 
 /**
@@ -74,8 +63,8 @@ export function cacheCodeQlVersion(
   cachedCodeQlVersion = version;
   const outputCache = {
     cmd,
-    entries: { [CommandCacheKey.Version]: version },
-  } satisfies OutputCache<CommandCacheKey.Version>;
+    entries: { version },
+  } satisfies OutputCache;
   // Persist the version so that subsequent Actions steps, which run in separate
   // processes, can reuse it rather than invoking `codeql version` again. We
   // record the CLI path so that a different step using a different CodeQL bundle
@@ -122,7 +111,7 @@ export function getCachedCodeQlVersion(
   }
   // Memoize the parsed value so that subsequent calls in this process don't
   // re-parse the environment variable.
-  cachedCodeQlVersion = persisted.entries[CommandCacheKey.Version];
+  cachedCodeQlVersion = persisted.entries.version as VersionInfo;
   return cachedCodeQlVersion;
 }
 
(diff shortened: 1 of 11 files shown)


src/util.ts
```diff
@@ -638,7 +639,25 @@ function isPersistedVersionInfo(x: unknown): x is PersistedVersionInfo {
   );
 }
 
-export function cacheCodeQlVersion(cmd: string, version: VersionInfo): void {
+/**
+ * Returns the file path to the `codeql version` output cache.
+ * @param env The environment variables to use—only necessary for testing.
```
mbg: Here and elsewhere: `env` is not just used for testing. Remove everything after "use".

src/util.ts
```diff
@@ -638,7 +639,25 @@ function isPersistedVersionInfo(x: unknown): x is PersistedVersionInfo {
...
+ */
+function getPathToCodeQLVersionCacheFile(env: Env): string {
+  return path.join(getTemporaryDirectory(env), "version.json");
+}
```
mbg: Are you planning to have a separate file for each command? If so, why?

src/util.ts
```diff
@@ -638,7 +639,25 @@ function isPersistedVersionInfo(x: unknown): x is PersistedVersionInfo {
...
+/**
+ * Caches the CodeQL CLI version both in-memory and on disk.
+ * @param cmd The path to the CodeQL CLI.
+ * @param version The version information to cache.
+ * @param env The environment variables to use—only necessary for testing.
+ */
+export function cacheCodeQlVersion(
+  cmd: string,
+  version: VersionInfo,
+  env: Env = getEnv(),
```
mbg: Avoid the default `getEnv()` which could lead to unexpected results.

src/util.ts
```diff
@@ -647,21 +666,33 @@ export function cacheCodeQlVersion(cmd: string, version: VersionInfo): void {
...
   }
   // Fall back to the value persisted by an earlier Actions step, if any. This is
   // best-effort: any malformed or mismatched value is ignored so that the caller
   // invokes `codeql version` instead.
-  const serialized = process.env[EnvVar.CODEQL_VERSION_INFO];
-  if (!serialized) {
+  let serialized: string;
+  try {
+    serialized = fs.readFileSync(getPathToCodeQLVersionCacheFile(env), "utf8");
+  } catch {
```
mbg: The error here is swallowed, which might make it hard to troubleshoot issues. Consider logging it at debug-level.

src/util.ts
```diff
@@ -647,21 +666,33 @@ export function cacheCodeQlVersion(cmd: string, version: VersionInfo): void {
...
-export function getCachedCodeQlVersion(cmd?: string): undefined | VersionInfo {
+/**
+ * Returns the cached CodeQL CLI version, if any. If not cached,
+ * attempts to read and parse it from disk.
+ * @param cmd The path to the CodeQL CLI.
+ * @param env The environment variables to use—only necessary for testing.
+ */
+export function getCachedCodeQlVersion(
+  cmd?: string,
+  env: Env = getEnv(),
```
mbg: Avoid the default `getEnv()` which could lead to unexpected results.

src/util.ts
```diff
@@ -647,21 +666,33 @@ export function cacheCodeQlVersion(cmd: string, version: VersionInfo): void {
...
+    getPathToCodeQLVersionCacheFile(env),
     JSON.stringify({ cmd, version }),
+    "utf8",
   );
 }
 
-export function getCachedCodeQlVersion(cmd?: string): undefined | VersionInfo {
+/**
+ * Returns the cached CodeQL CLI version, if any. If not cached,
+ * attempts to read and parse it from disk.
```
mbg: This description is a bit nonsensical: it seems to imply that the file on disk is not a cache. 

Additionally, I'd view the different caching layers as more of an implementation detail that doesn't have to be exposed in the caller-facing documentation.

```suggestion
 * Returns the cached CodeQL CLI version, if any.
```

src/util.test.ts
```diff
@@ -535,55 +535,83 @@ test("Failure.orElse returns the default value for a failure result", (t) => {
 
 test.serial(
```
mbg: This test, and others, probably don't need to be `serial` anymore since `process.env` is no longer mutated.

src/util.test.ts
```diff
@@ -535,55 +535,83 @@ test("Failure.orElse returns the default value for a failure result", (t) => {
...
+        JSON.stringify({ version: { version: "2.20.0" } }),
+        JSON.stringify({
+          cmd: "/path/to/codeql",
+          version: { version: "2.20.0", overlayVersion: "1" },
+        }),
+        JSON.stringify({
+          cmd: "/path/to/codeql",
+          version: { version: "2.20.0", features: "nope" },
+        }),
+      ]) {
```
mbg: Minor: I know this isn't a change here, but consider lifting the array out of the loop expression to make it more readable.

```suggestion
      const testValues = [
        JSON.stringify({ cmd: "/path/to/codeql" }),
        JSON.stringify({ cmd: "/path/to/codeql", version: {} }),
        JSON.stringify({ cmd: "/path/to/codeql", version: { version: 2 } }),
        JSON.stringify({ version: { version: "2.20.0" } }),
        JSON.stringify({
          cmd: "/path/to/codeql",
          version: { version: "2.20.0", overlayVersion: "1" },
        }),
        JSON.stringify({
          cmd: "/path/to/codeql",
          version: { version: "2.20.0", features: "nope" },
        }),
      ];
      
      for (const value of testValues) {
```

You could simplify this further and remove all the individual calls to `JSON.stringify`:

```suggestion
      const testValues = [
        { cmd: "/path/to/codeql" },
        { cmd: "/path/to/codeql", version: {} },
        { cmd: "/path/to/codeql", version: { version: 2 } },
        { version: { version: "2.20.0" } },
        {
          cmd: "/path/to/codeql",
          version: { version: "2.20.0", overlayVersion: "1" },
        },
        {
          cmd: "/path/to/codeql",
          version: { version: "2.20.0", features: "nope" },
        },
      ].map(JSON.stringify);
      
      for (const value of testValues) {
```

mbg (commented):
Thanks for picking this back up! I have added a few inline (mostly style) comments.

Some high-level points:

- Also added as an in-line comment, but do you plan to have a separate file for each command? If so, why? On the same topic, if not, then should the file be more structured to allow for relevant metadata and other command output down the line?
- Have you had any more thoughts about either checking that the cached file is intended for the current analysis (e.g. by checking the generated `jobUuid`) or deleting cache files on "startup"?
- Add a unit test for the case where the file cannot be read (e.g. because it doesn't exist).
- The linter CI is failing. Remember to run `npm lint` locally before committing / pushing changes.

----

Much more minor, but I spotted this in the PR description:

> This is fine, for now, but, as we begin to cache more of the CLI, we should persist that output to a file instead.

The reasoning here ("as we begin to cache more of the CLI") doesn't really justify the conclusion ("we should persist that output to a file instead). Could you edit that to actually highlight _why_ files are preferable to environment variables?

src/util.ts
mario-campos: No, with this PR, I'm simply trying to do a direct environment-variable-to-file translation. In subsequent PRs, I plan to generalize the file to include additional commands. The only reason that the file's name is associated with the command is to keep this PR focused on this one change, without scope creep of future intentions.

src/util.test.ts
mario-campos: `.map(JSON.stringify);` doesn't quite work but I've got the essence of it.

src/util.test.ts
mario-campos: I had the same thought, but Copilot assured me it's still necessary:

>Yes, keep `.serial()` for this one.
>
>Even though this test itself only reads malformed cache entries, it exercises `getCachedCodeQlVersion`, which depends on the module-global `cachedCodeQlVersion` state in `util.ts`. Other nearby tests in the same file can populate that cache, and parallel execution can interleave despite beforeEach resets, since they all share one process/module instance.
>
>So `.serial()` is the safe choice unless you fully refactor these cache-state tests to avoid shared global state.

mbg: The point about `cachedCodeQlVersion` is fair. That said, I wouldn't be opposed to refactoring that away so that the state is threaded to these functions rather than global as part of this PR.

src/util.ts
mario-campos: Could the logger be instantiated locally within the module or within the function itself? I tried threading it through the call stack, but that would require a lot of modifications (changes to 16+ files).

src/util.test.ts
mario-campos: Agreed, but perhaps best saved for another PR.

src/util.ts
mario-campos: Same problem here: threading `env` through the call stack would require touching a lot of files/functions. Is this still preferable?

src/util.test.ts
```diff
@@ -618,3 +618,10 @@ test.serial(
     });
   },
 );
+
+test.serial("getCachedCodeQlVersion ignores non-existent file", async (t) => {
+  await util.withTmpDir(async (tmpDir: string) => {
+    const env = getTestEnv({ [EnvVar.TEMP]: tmpDir });
```
mbg: Minor, but maybe wrap at least this in `t.notThrows` as well.

src/cli/output-cache.ts
```diff
@@ -6,10 +6,24 @@ import { Env, getEnv } from "../environment";
...
+ * The keys of the command cache. Each key corresponds to a command whose output we cache.
+ */
+enum CommandCacheKey {
+  Version = "version",
+}
+
+/**
+ * The mapping of CLI commands to the types of the output of each command that we cache.
+ */
+type CommandCacheKeyOutputMap = {
```
mbg: Minor: `export`

src/cli/output-cache.ts
```diff
@@ -6,10 +6,24 @@ import { Env, getEnv } from "../environment";
 
 import type { VersionInfo } from "./types";
 
+/**
+ * The keys of the command cache. Each key corresponds to a command whose output we cache.
+ */
+enum CommandCacheKey {
+  Version = "version",
+}
```
mbg: Let's simplify this to just `export type CommandCacheKey = string;` for now. While stronger typing is better, the contents of the "cache key" was a big discussion point on the previous PR.

src/cli/output-cache.ts
```diff
@@ -60,7 +74,7 @@ export function cacheCodeQlVersion(
   // doesn't pick up a stale version.
   fs.writeFileSync(
     getCommandCacheFilePath(env),
-    JSON.stringify({ cmd, version }),
+    JSON.stringify({ cmd, entries: { [CommandCacheKey.Version]: version } }),
```
mbg: Here and in the test, nothing guarantees that the object is a valid `CommandCacheRecord` / `OutputCache`. Either assert this with a `satisfies` clause or store it in a constant with an explicit type annotation before stringify-ing.

src/cli/output-cache.ts
```diff
@@ -126,12 +140,17 @@ function isVersionInfo(x: unknown): x is VersionInfo {
...
+  const candidate = x as Partial<
+    PersistedVersionInfo<CommandCacheKey.Version>
+  > | null;
   return (
     typeof candidate === "object" &&
     candidate !== null &&
     typeof candidate.cmd === "string" &&
-    isVersionInfo(candidate.version)
+    candidate.entries !== undefined &&
+    isVersionInfo(candidate.entries[CommandCacheKey.Version])
```
mbg: Note that `candidate.entries[CommandCacheKey.Version]` could be `undefined`. Check that it doesn't cause any unwanted exceptions here or in `isVersionInfo`.

src/cli/output-cache.ts
```diff
@@ -0,0 +1,158 @@
...
+  [CommandCacheKey.Version]: VersionInfo;
+};
+
+/**
+ * The type of the command cache that is persisted to disk.
+ */
+interface CommandCacheRecord<K extends CommandCacheKey> {
+  cmd: string;
+  entries: Map<K, CommandCacheKeyOutputMap[K]>;
+}
```
mbg: `export` this and drop at least the `Record` part of the name. Maybe consider renaming it to `OutputCache`.

src/util.ts
mbg: It's fine to pass the result of `getEnv()` into the function as an explicit argument. I am just worried about the default here.

src/util.ts
mbg: I looked into what it would take to thread a `Logger` to `getCodeQLForCmd` and it wasn't too bad: https://github.com/github/codeql-action/pull/4086

It's fine for `getCodeQLForCmd` to be given the `Logger` and the `CodeQL` functions to capture it, since for production use the `Logger` is expected to be the same throughout in any case.

src/util.test.ts
mario-campos: Good suggestion 👍

src/cli/output-cache.ts
mario-campos: As I understand it, it would only throw if `candidate.entries` were `null` or `undefined`, which is guarded against. `candidate.entries[CommandCacheKey.Version]` may be `undefined`, but `isVersionInfo` handles that case. Nevertheless, I've added a few more test cases to one of the unit tests to assert this behavior.

src/cli/output-cache.ts
```diff
@@ -64,7 +64,7 @@ function getCommandCacheFilePath(env: Env): string {
 export function cacheCodeQlVersion(
   cmd: string,
   version: VersionInfo,
-  env: Env = getEnv(),
+  env: Env,
```
mbg: Minor: You moved `env` to be the first parameter of `getCachedCodeQlVersion` but not here. It would be good to have that consistency (and having it at the start of the parameter list if preferable if we want to upgrade it to `ActionState` down the line).

src/cli/output-cache.ts
mario-campos: Done

src/util.ts
mbg: This still needs addressing now that https://github.com/github/codeql-action/pull/4086 is merged.

src/util.test.ts
mbg: OK, that's fine.

mbg (dismissed):
LGTM, aside from addressing the minor point about the logging the error and resolving the merge conflict.

src/util.ts
mario-campos: I merged in `main` and included some logging. Thanks

mbg (approved):

PR #4096, opened 2026-08-14
Use `json` module for JSON validation in `output-cache`

mario-campos:
- Use the included `json` module to improve (make more readable) the validation of the output of `codeql version`.
- Refactor `getCachedCodeQlVersion` and `cacheCodeQlVersion` to accept the cache-file-path as an argument, rather than construct it; this makes it easier to test.
- Change the `VersionInfo` type to be derived from its JSON schema.

### Risk assessment

Workflow types:

Products:

Environments:

### DIFF
--- a/src/cli/output-cache.test.ts
+++ b/src/cli/output-cache.test.ts
@@ -7,7 +7,10 @@ import { getRunnerLogger } from "../logging";
 import { getTestEnv, setupTests } from "../testing-utils";
 import * as util from "../util";
 
-import * as outputCache from "./output-cache";
+import {
+  getCachedCodeQlVersion,
+  getCommandCacheFilePath,
+} from "./output-cache";
 
 setupTests(test);
 
@@ -18,7 +21,7 @@ test.serial(
   async (t) => {
     await util.withTmpDir(async (tmpDir: string) => {
       const env = getTestEnv({ [EnvVar.TEMP]: tmpDir });
-      const cacheFile = outputCache.getCommandCacheFilePath(env);
+      const cacheFile = getCommandCacheFilePath(env);
       fs.writeFileSync(
         cacheFile,
         JSON.stringify({
@@ -27,12 +30,9 @@ test.serial(
         }),
         "utf8",
       );
-      t.deepEqual(
-        outputCache.getCachedCodeQlVersion(logger, env, "/path/to/codeql"),
-        {
-          version: "2.20.0",
-        },
-      );
+      t.deepEqual(getCachedCodeQlVersion(logger, env, "/path/to/codeql"), {
+        version: "2.20.0",
+      });
     });
   },
 );
@@ -42,7 +42,7 @@ test.serial(
   async (t) => {
     await util.withTmpDir(async (tmpDir: string) => {
       const env = getTestEnv({ [EnvVar.TEMP]: tmpDir });
-      const cacheFile = outputCache.getCommandCacheFilePath(env);
+      const cacheFile = getCommandCacheFilePath(env);
       fs.writeFileSync(
         cacheFile,
         JSON.stringify({
@@ -51,10 +51,7 @@ test.serial(
         }),
         "utf8",
       );
-      t.is(
-        outputCache.getCachedCodeQlVersion(logger, env, "/path/to/codeql"),
-        undefined,
-      );
+      t.is(getCachedCodeQlVersion(logger, env, "/path/to/codeql"), undefined);
     });
   },
 );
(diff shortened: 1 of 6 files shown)


src/cli/output-cache.ts
```diff
@@ -127,16 +128,16 @@ export function getCachedCodeQlVersion(
...
-    (candidate.overlayVersion === undefined ||
-      typeof candidate.overlayVersion === "number")
+    json.isObject(x) &&
+    json.validateSchema(
+      {
+        version: json.string,
+        features: json.optional(json.object({})),
+        overlayVersion: json.optional(json.number),
+      } as const satisfies json.Schema,
+      x,
```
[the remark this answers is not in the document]
mbg: Good catch, @mario-campos can you address this? You already have `COMMAND_CACHE_FILENAME` as a constant, export it and use it in the tests instead of duplicating the filename there.

src/cli/output-cache.ts
```diff
@@ -127,16 +128,16 @@ export function getCachedCodeQlVersion(
...
-        candidate.features !== null)) &&
-    (candidate.overlayVersion === undefined ||
-      typeof candidate.overlayVersion === "number")
+    json.isObject(x) &&
+    json.validateSchema(
+      {
+        version: json.string,
+        features: json.optional(json.object({})),
+        overlayVersion: json.optional(json.number),
+      } as const satisfies json.Schema,
```
mbg: Minor: Lift the schemas out of the function definitions to make them named, top-level definitions in this file. 

You might almost be able to derive `VersionInfo` from the schema then, which would be preferable over having a disjoint schema and type. 

If the goal is not to diverge from the existing validation by making it stricter, then you could have e.g. something like:

```typescript
export const versionInfoBaseSchema = {
  version: json.string,
  features: json.optional(json.object({})),
  overlayVersion: json.optional(json.number),
} as const satisfies json.Schema;

export type VersionInfoBase = json.FromSchema<typeof versionInfoBaseSchema>;
export type VersionInfo = VersionInfoBase & {
  // `features` remains optional, but the more specific type takes precedence
  // over the `any` type derived by `FromSchema`.
};
```

mbg (commented):
Looks OK, but a good catch from Copilot there that we missed on the previous PR. I also added a comment about avoiding the disconnect between the schema and `VersionInfo` type. When using the `json` module for schemas, it is desirable to derive the corresponding type with `FromSchema` to avoid them going out of sync.

src/cli/output-cache.ts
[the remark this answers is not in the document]
mario-campos: Yes, can't believe I missed that 🤦

src/cli/output-cache.ts
mario-campos: Makes sense. Done.

src/cli/types.ts
```diff
@@ -9,5 +14,21 @@ export interface VersionInfo {
...
    * we need to revert to non-overlay analysis.
    */
-  overlayVersion?: number;
-}
+  overlayVersion: json.optional(json.number),
+} as const satisfies json.Schema;
+
+/**
+ * The base type that describes the expected output of the `codeql version` command.
+ * This type is partially derived from {@link versionInfoBaseSchema}.
```
mbg:
```suggestion
 * This type is derived from {@link versionInfoBaseSchema}.
```
But really, you can just remove this sentence.

src/cli/output-cache.test.ts
```diff
@@ -18,7 +17,8 @@ test.serial(
   "getCachedCodeQlVersion reuses a version persisted by an earlier step",
   async (t) => {
     await util.withTmpDir(async (tmpDir: string) => {
-      const cacheFile = path.join(tmpDir, "codeql-action-command-cache.json");
+      const env = getTestEnv({ [EnvVar.TEMP]: tmpDir });
+      const cacheFile = outputCache.getCommandCacheFilePath(env);
```
mbg: A slightly better approach to address this issue than to set the right environment variable to make the test work is to make the path of the cache file a parameter of the `getCachedCodeQlVersion` and `cacheCodeQlVersion` functions.

Since you already have `getCommandCacheFilePath`, it is easy to provide the right argument in production code. In the test code, it can then be completely arbitrary.

mbg:
@mario-campos Also looks like the linter check is failing again. If you haven't already enabled `editor.formatOnSave`, that might be a useful thing to do to help prevent this.

mario-campos:
>Also looks like the linter check is failing again.

That was a downstream (in [`tools-feature.ts`](https://github.com/github/codeql-action/blob/d97b3428e8eebbb1810cf454d6397886d136b4ba/src/tools-features.ts#L20-L25)) effect of the changes made to `VersionInfo`. It resulted in `VersionInfo.features` having type `any`, despite the override. Copilot suggested using `Omit<>` to fix it, so I did.

That being said, I wonder if it would be better to define a JSON validator for `VersionInfo.features` such that we can derive `VersionInfo` directly from its schema?

src/cli/output-cache.test.ts
mario-campos: Yeah, that makes more sense. Good suggestion! Thanks.

mbg (approved):

PR #4105, opened 2026-08-24
Improve CodeQL tools download time telemetry

henrymercer:
We are considering shipping trimmed CodeQL bundles containing the CLI and a single language, as an alternative to today's combined per-platform bundle. This PR establishes a baseline for how long the current bundle takes, so the effect can be measured.

`downloadAndExtract` has two paths. The streaming path (Zstandard on Linux, which most hosted runners take) downloads and extracts concurrently, and returned an empty status report — so the dominant path reported no timings at all. The download-then-extract path reported the download duration and only logged the extraction duration.

| Field | When populated | Meaning |
| --- | --- | --- |
| `tools_download_duration_ms` | Download-then-extract only | Time spent downloading. Unchanged. |
| `tools_extraction_duration_ms` | Download-then-extract only | Time spent extracting. |
| `tools_total_duration_ms` | Always | Wall-clock for the whole operation. |

`tools_total_duration_ms` is the field to compare across the change, being the only one populated on both paths. The component fields are left unpopulated on the streaming path, where downloading and extracting overlap and cannot be told apart. `tools_download_duration_ms` keeps its existing meaning so earlier data remains comparable.

The total is wall-clock across the whole call, so where streaming fails it includes the failed attempt and its cleanup. Those runs stay identifiable: on Linux, a report containing `tools_download_duration_ms` is one where streaming failed.

Compressed bundle size is deliberately not reported. For the common case it is a function of CLI version and platform, so it is already knowable from the release assets — and the size saving is the part of this that can be predicted, whereas the durations are not.

No changelog entry, as there is no user-facing change.

### Risk assessment

- **Low risk:** telemetry only. No change to what is downloaded, from where, or in what order, and no new failure modes on the download path.

Workflow types:

Products:

Environments:

- **Unit tests** - `src/tools-download.test.ts` covers both paths and the fallback from streaming to download-then-extract.

- **Telemetry** - This change is itself a telemetry change; the new fields appearing with plausible values is the signal that it works.

- **Special considerations** - This change is only useful if it is released before any change to how the bundle is packaged, since its purpose is to establish a baseline.

### DIFF
--- a/lib/entry-points.js
+++ b/lib/entry-points.js
@@ -151503,10 +151503,10 @@ async function downloadAndExtract(codeqlURL, compressionMethod, dest, authorizat
   logger.info(
     `Downloading CodeQL tools from ${codeqlURL} . This may take a while.`
   );
+  const startTime = import_perf_hooks2.performance.now();
   try {
     if (compressionMethod === "zstd" && process.platform === "linux") {
       logger.info(`Streaming the extraction of the CodeQL bundle.`);
-      const toolsInstallStart = import_perf_hooks2.performance.now();
       await downloadAndExtractZstdWithStreaming(
         codeqlURL,
         dest,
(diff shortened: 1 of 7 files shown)


mario-campos (approved):
LGTM

PR #4116, opened 2026-08-31
Validate change-notes as a PR check

mario-campos:
This PR extends the behavior of the recently added `pr-checks/changenotes.mts` tool by:

1. Hardcoding the argument as the newly-added `unreleased-change-notes` directory, as that is the only destination for change-notes.
2. Adding a step in the `pr-checks.yml` workflow file for executing `pr-checks/changenotes.mts` in PRs to validate any changenotes in `unreleased-change-notes`.

This PR also introduces an "empty" top-level directory, `unreleased-change-notes`, to be the destination of future change-note files. This can be changed if another location is more suitable.

### Risk assessment

Workflow types:

- **N/A**

Products:

- **N/A**

Environments:

- **N/A**

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -114,6 +114,10 @@ jobs:
         working-directory: pr-checks
         run: npx tsx --test
 
+      - name: Run `pr-checks/changenotes.mts` to ensure that all unreleased change notes are valid
+        if: ${{ !cancelled() && steps.install-deps.outcome == 'success' }}
+        run: npx tsx pr-checks/changenotes.mts validate
+
       - name: Verify all Actions use the same Node version
         id: head-version
         run: |
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -158,7 +158,7 @@ export default [
     },
   },
   {
-    files: ["**/*.ts", "**/*.js"],
+    files: ["**/*.ts", "**/*.js", "**/*.mts"],
 
     rules: {
       "@typescript-eslint/no-explicit-any": "off",
--- a/pr-checks/changelog/validate.mts
+++ b/pr-checks/changelog/validate.mts
@@ -119,3 +119,14 @@ export function isValidChangenoteFile(filename: string): boolean {
 
   return isValid;
 }
+
+/**
+ * Validates the change-note files of the given list of file paths, ignoring ".gitkeep".
+ * @param filepaths A list of filepaths to validate
+ * @returns True if all the paths are valid, false otherwise.
+ */
+export function isValidAllChangenoteFiles(filepaths: string[]): boolean {
+  return filepaths
+    .filter((f) => f !== ".gitkeep")
+    .reduce((r, filePath) => r && isValidChangenoteFile(filePath), true);
+}
--- a/pr-checks/changelog/validate.test.mts
+++ b/pr-checks/changelog/validate.test.mts
@@ -1,10 +1,13 @@
 import assert from "node:assert/strict";
+import * as fs from "node:fs";
+import * as path from "node:path";
 import { describe, it } from "node:test";
 
-import { withTmpFile } from "../../src/util";
+import { withTmpDir, withTmpFile } from "../../src/util";
 
 import {
   hasValidChangenoteCategory,
+  isValidAllChangenoteFiles,
   isValidChangenoteContent,
   isValidChangenoteFile,
   isValidChangenoteFilename,
@@ -150,6 +153,10 @@ await describe("isValidChangenoteFile", async () => {
     );
   });
 
+  await it("rejects a non-existent path", async () => {
+    assert.equal(isValidChangenoteFile("non-existent-file.md"), false);
+  });
+
   await it("rejects invalid filename", async () => {
     await withTmpFile(
       "fix-bug.md",
@@ -178,5 +185,41 @@ await describe("isValidChangenoteFile", async () => {
         assert.equal(isValidChangenoteFile(filePath), false);
       },
     );
+  });
+});
+
+await describe("isValidAllChangenoteFiles", async () => {
+  await it("accepts a directory of valid change-note files", async () => {
+    await withTmpDir(async (tmpDir) => {
+      const fileName1 = path.join(tmpDir, "2026-01-01-fix-bug.md");
+      const fileName2 = path.join(tmpDir, "2026-01-02-add-feature.md");
+      fs.writeFileSync(fileName1, "---\ncategory: fix\n---\n- Fixed a bug\n");
+      fs.writeFileSync(
+        fileName2,
+        "---\ncategory: feature\n---\n- Added a feature\n",
+      );
+      assert.equal(isValidAllChangenoteFiles([fileName1, fileName2]), true);
+    });
+  });
+
+  await it("accepts an empty list", async () => {
+    assert.equal(isValidAllChangenoteFiles([]), true);
+  });
+
+  await it("accepts a list of .gitkeep only", async () => {
+    assert.equal(isValidAllChangenoteFiles([".gitkeep"]), true);
+  });
+
+  await it("rejects directory with an invalid change-note file", async () => {
+    await withTmpDir(async (tmpDir) => {
+      const fileName1 = path.join(tmpDir, "2026-01-01-fix-bug.md");
+      const fileName2 = path.join(tmpDir, "2026-01-02-wrong-category.md");
+      fs.writeFileSync(fileName1, "---\ncategory: fix\n---\n- Fixed a bug\n");
+      fs.writeFileSync(
+        fileName2,
+        "---\ncategory: foobar\n---\n- Added a feature\n",
+      );
+      assert.equal(isValidAllChangenoteFiles([fileName1, fileName2]), false);
+    });
   });
 });
--- a/pr-checks/changenotes.mts
+++ b/pr-checks/changenotes.mts
@@ -1,9 +1,11 @@
 #!/usr/bin/env npx tsx
 
+import * as fs from "node:fs";
 import { pathToFileURL } from "node:url";
 import { parseArgs } from "node:util";
 
-import { isValidChangenoteFile } from "./changelog/validate.mjs";
+import { isValidAllChangenoteFiles } from "./changelog/validate.mjs";
+import { CHANGENOTES_DIR } from "./config";
 
 const entryPoint = process.argv[1];
 if (entryPoint && import.meta.url === pathToFileURL(entryPoint).href) {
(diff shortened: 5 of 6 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -255,3 +255,32 @@ jobs:
           else
             echo "Skipping repo size comment because the delta is below the threshold and no sticky comment exists."
           fi
+
+  validate-change-notes:
+    name: Validate change notes
+    if: github.triggering_actor != 'dependabot[bot]'
+    permissions:
+      contents: read
+    runs-on: ubuntu-slim
```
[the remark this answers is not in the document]
mario-campos: Jobs with `ubuntu-slim` runners are automatically limited to 15 minutes, so a time-out isn't necessary.

connected79844:
Approved

scripts/changetool/cli/validate.ts
```diff
@@ -119,3 +119,29 @@ export function isValidChangenoteFile(filename: string): boolean {
...
+    const stats = fs.statSync(filepath);
+    if (stats.isFile()) {
+      return isValidChangenoteFile(filepath);
+    }
+    if (stats.isDirectory()) {
+      return fs
+        .readdirSync(filepath)
+        .filter((f) => f !== ".gitkeep")
+        .map((f) => isValidChangenoteFile(path.join(filepath, f)))
+        .every(Boolean);
```
mbg: The `isValidChangenoteFile` function already has a return type of `boolean` so using `Boolean` in `every` to convert each result to a `boolean` is redundant. The only meaningful behaviour there is asserting that all values are `true`. An explicit `.every((val) => val === true)` would accomplish that more clearly.

pr-checks/changelog/validate.test.mts
```diff
@@ -6,27 +6,33 @@ import { describe, it } from "node:test";
...
 import {
   isValidChangenoteContent,
-  isValidChangenoteFile,
+  isValidChangenoteFileOrDir,
   isValidChangenoteFilename,
   hasValidChangenoteCategory,
   VALID_CHANGE_NOTE_CATEGORIES,
 } from "./validate.ts";
 
+async function withTmpDir<T>(body: (tmpDir: string) => Promise<T>): Promise<T> {
```
mbg: In `pr-checks`, we have access to the main codebase and so (technically) don't need to duplicate helpers that we already have there. This new `scripts/changetool` project doesn't currently have access to the main codebase. It would be ideal if we didn't have to duplicate this, either by granting this project access to the main codebase or folding this project into the `pr-checks` project.

pr-checks/changelog/validate.test.mts
```diff
@@ -6,27 +6,33 @@ import { describe, it } from "node:test";
...
+  const tmpDir = fs.mkdtempSync(
+    path.join(os.tmpdir(), "changetool-validate-test-"),
+  );
+  try {
+    return await body(tmpDir);
+  } finally {
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  }
```
mbg: This deviates (reasonably) from the implementation of the equivalent function in `src/util.ts`. As a side note, I wonder if it would make sense to adopt this there. If this project could benefit from utility functions in the main codebase (see above comment), then the improved implementation from `src/util.ts` could be used here.

.github/workflows/pr-checks.yml
```diff
@@ -255,3 +255,32 @@ jobs:
           else
             echo "Skipping repo size comment because the delta is below the threshold and no sticky comment exists."
           fi
+
+  validate-change-notes:
```
mbg: This is the second new job in `pr-checks.yml` for the new `scripts/changetool` project. If the new tool was part of the scripts in `pr-checks`, the separate `-tests` job wouldn't be needed. (Although it could be folded into `other-checks` either way.)

This new `validate-change-notes` could equally be folded into `other-checks` to avoid the extra maintenance overhead, and additional compute used.

mbg (commented):
Generally this change looks reasonable. My main comments follow-up on my [post-merge comment on the previous PR](https://github.com/github/codeql-action/pull/4113#pullrequestreview-5077284338). I still wonder why this is a separate project with separate infrastructure, and isn't an extra script in `pr-checks` that benefits from the existing infrastructure for that.

scripts/changetool/cli/validate.ts
mario-campos: That was, more or less, my original implementation. Copilot [suggested this implementation](https://github.com/github/codeql-action/pull/4116#discussion_r3896530127) because `every` returns early on the first `false`, which makes sense. Better to validate _all_ of the change-notes, not just the some of the correct ones and _one_ of the invalid ones.

mbg: Yes, the problem is valid, but the suggested fix is not great. My suggested change just makes the fix more explicit. You could also avoid the extra `.every` with something like:

```suggestion
        .reduce(
          (r, f) => r && isValidChangenoteContent(path.join(filepath, f)),
          true,
        );
```

That will evaluate `isValidChangenoteContent` for all files and produce the result of `&&`-ing all results in one go.

mario-campos:
Closing this to re-work these changes after a migration to the `pr-checks` directory. Changes will be pushed to a new PR.

.github/workflows/pr-checks.yml
mario-campos: Folded `validate-change-notes` into `other-checks`

pr-checks/changelog/validate.test.mts
mario-campos: I'll rebase after #4140 and #4141 to re-use the existing `withTmpDir`.

.github/workflows/pr-checks.yml
```diff
@@ -114,6 +114,13 @@ jobs:
         working-directory: pr-checks
         run: npx tsx --test
 
+      - name: Run `validate-changelog.ts` to ensure that all unreleased change notes are valid
+        if: ${{ !cancelled() && steps.install-deps.outcome == 'success' }}
+        working-directory: pr-checks
+        # "unreleased-change-notes" resolves to a path relative to scripts/changetool.
+        # So, we need to pass the absolute path to the directory by prefixing $GITHUB_WORKSPACE.
+        run: npx tsx validate-changelog.ts "$GITHUB_WORKSPACE/unreleased-change-notes"
```
mbg, line 122: Alternatively you could not set `working-directory` and specify `pr-checks/validate-changelog.ts` as the script path. That works around the need to use `$GITHUB_WORKSPACE` and the lengthy explanation. 

Even better, however, would be to add a constant to `pr-checks/config.ts` for the path joined onto `REPO_ROOT` and use that as the path (assuming it will always be that path).

eslint.config.mjs
```diff
@@ -158,7 +158,7 @@ export default [
     },
   },
   {
-    files: ["**/*.ts", "**/*.js"],
+    files: ["**/*.ts", "**/*.js", "**/*.mts"],
```
mbg, line 161: 👍🏻

pr-checks/changelog/validate.mts
```diff
@@ -119,3 +119,31 @@ export function isValidChangenoteFile(filename: string): boolean {
 
   return isValid;
 }
+
+/**
+ * Validates that the given path is a valid change-note file or directory of change-note files.
+ * @param filepath The path to the change-note file or directory to validate.
+ * @returns True if the path is valid, false otherwise.
+ */
+export function isValidChangenoteFileOrDir(filepath: string): boolean {
```
mbg, line 128: Minor: no strong need to change this, but consider decoupling I/O from the validation logic when writing functions like this to avoid having to create real directories / files in the unit tests. Alternatively, consider stubbing the I/O operations for the tests to return the hard-coded test data, but that has the downside that it breaks test parallelism.

mbg (commented):
Appreciate that this is a draft pending the changes in the other PRs, but some drive-by comments since I saw this and you could look at incorporating the feedback when updating this PR once the others are merged.

.github/workflows/pr-checks.yml
mario-campos: Yes, it should always be that path, so went with the latter. Thanks

pr-checks/changelog/validate.mts
mario-campos: I've refactored the functions to perform less I/O, making the interfaces a bit more purer.

pr-checks/changelog/validate.test.mts
```diff
@@ -180,3 +187,39 @@ await describe("isValidChangenoteFile", async () => {
     );
   });
 });
+
+await describe("isValidAllChangenoteFiles", async () => {
+  await it("accepts a directory of valid change-note files", async () => {
```
mbg: Minor: The name of this test doesn't accurately reflect it anymore since you're providing `isValidAllChangenoteFiles` with a list of file paths now.

pr-checks/changenotes.mts
```diff
@@ -20,34 +22,32 @@ function main(): number {
...
   return 0;
 }
 
-function validate(paths: string[]): number {
-  let valid = true;
-  if (paths.length === 0) {
-    console.error("error: no paths provided (see 'help' command for usage)");
+function validate(): number {
+  try {
+    return isValidAllChangenoteFiles(fs.readdirSync(CHANGENOTES_DIR)) ? 0 : 1;
```
mbg: Minor: It might be good to log something when this succeeds ("Validation successful.") and/or in the case where there are no change notes in the directory ("No change notes found to validate."), just to more clearly signal that we did run validation and there was no problem / we didn't find anything to validate.

E.g. looking at the [workflow output](https://github.com/github/codeql-action/actions/runs/34647517692/job/103421739858#step:7:1) right now, there is no output at all for the step.

mbg (dismissed):
This LGTM, with two non-blocking suggestions for improvements.

pr-checks/changelog/validate.test.mts
mario-campos: Fixed

pr-checks/changenotes.mts
mario-campos: Fixed

pr-checks/changenotes.mts
```diff
@@ -42,12 +42,15 @@ function usage(): number {
...
-    return isValidAllChangenoteFiles(fs.readdirSync(CHANGENOTES_DIR)) ? 0 : 1;
+    if (isValidAllChangenoteFiles(fs.readdirSync(CHANGENOTES_DIR))) {
+      console.log(`All changenote files in '${CHANGENOTES_DIR}' are valid.`);
+      return 0;
+    }
   } catch (error) {
     console.error(
-      `${CHANGENOTES_DIR}: failed to read file or directory`,
+      `Failed to read change-notes directory (${CHANGENOTES_DIR})`,
       error,
```
mbg, line 52: Minor: Be consistent in the log messages with "changenote(s)" vs "change-note(s)" and using single quotes vs parentheses to delimit the path. For the former, either "changenote(s)" or "change note(s)" is fine, but I'd avoid "change-note(s)". For the latter, we generally use single quotes in the codebase.

mbg (dismissed):

pr-checks/changenotes.mts
mario-campos: Fixed

mbg (approved):

PR #4122, opened 2026-09-03
Don't record an overlay status when the job was cancelled

henrymercer:
The overlay status mechanism records a marker in the Actions cache when building an overlay-base database fails, so that later runs skip overlay analysis instead of failing the same way, but it records one whenever the analyze step did not report success, which also covers runs that were cancelled, since the `init` post step still runs in that case. A cancellation tells us nothing about whether the analysis would have succeeded, and a marker disables overlay analysis for the whole repository until the next CodeQL release, so this adds an internal `job-status` input defaulting to `${{ job.status }}` and skips recording when the job was cancelled.

### Risk assessment

Workflow types:

Products:

Environments:

- **Test repository** - This change will be tested on a test repository before merging, specifically to confirm that a cancelled run reports `cancelled` to the post step.

- **Feature flags** - The behaviour change only takes effect where `overlay_analysis_status_save` is enabled, so disabling that flag restores the previous behaviour. The new input is inert on its own.

- **Telemetry** - I rely on existing telemetry. The `codeql-action/overlay-disabled` diagnostic reports `skipped-due-to-cached-status`, so the rate of that reason shows whether we are recording fewer markers as intended, and overlay adoption shows whether we have stopped recording ones we needed.

- Consider adding a [changelog](https://github.com/github/codeql-action/blob/main/CHANGELOG.md) entry for this change. Not added, since the behaviour is gated behind an existing feature flag, matching earlier changes to this mechanism.

### DIFF
--- a/init/action.yml
+++ b/init/action.yml
@@ -164,6 +164,13 @@ inputs:
       [Internal] The ID of the check run, as provided by the Actions runtime environment. Do not set this value manually.
     default: ${{ job.check_run_id }}
     required: false
+  job-status:
+    description: >-
+      [Internal] The status of the job, as provided by the Actions runtime environment. This is how the
+      post step learns whether the job as a whole succeeded, failed, or was cancelled. Do not set this
+      value manually.
+    default: ${{ job.status }}
+    required: false
 outputs:
   codeql-path:
     description: The path of the CodeQL binary used for analysis
--- a/src/init-action-post-helper.ts
+++ b/src/init-action-post-helper.ts
@@ -432,6 +432,22 @@ function didCodeQlReportError(env: ReadOnlyEnv): boolean {
 }
 
 /**
+ * Whether the job status tells us anything about whether the analysis itself would have succeeded.
+ *
+ * We check for the statuses we know to be meaningful rather than excluding the ones that are not,
+ * so that a status we do not recognise is treated as inconclusive.
+ */
+function isConclusiveJobStatus(jobStatus: string | undefined): boolean {
+  switch (jobStatus?.trim().toLowerCase()) {
+    case "failure":
+    case "success":
+      return true;
+    default:
+      return false;
+  }
+}
+
+/**
  * If overlay base database creation was attempted but the analysis did not complete
  * successfully, save the failure status to the Actions cache so that subsequent runs
  * can skip overlay analysis until something changes (e.g. a new CodeQL version).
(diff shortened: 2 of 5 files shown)


oscarsj (dismissed):
good catch

init/action.yml
```diff
@@ -164,6 +164,13 @@ inputs:
       [Internal] The ID of the check run, as provided by the Actions runtime environment. Do not set this value manually.
     default: ${{ job.check_run_id }}
     required: false
+  job-status:
```
mbg, line 167: Never nice to have to feed this in through an extra action input, but I don't see a better approach for getting hold of the job status. An alternative option might be to set `CODEQL_ACTION_STEP_(init|analyze|...)` environment variables / state that we set to e.g. `starting` when the respective action starts and then to `success` or `failure` depending on the outcome. That should then allow us to identify which step started, succeeded, or failed (gracefully or not). For the overlay status, we could then check that all available environment variables with a `CODEQL_ACTION_STEP_` prefix are `success` and none are `starting` or `failure`. The downside is that it wouldn't catch if the failure isn't related to what happens in CodeQL Action steps, or we fail to even set the `starting` value.

src/init-action-post-helper.ts
```diff
@@ -431,6 +449,20 @@ async function recordOverlayStatus(
     return;
   }
 
+  // A cancelled run tells us nothing about whether the analysis would have succeeded, so recording
+  // a failure would disable overlay analysis needlessly. Note that we still record a failure if one
+  // of our own Actions reported an error before the run was cancelled.
+  if (
+    jobStatus?.trim().toLowerCase() === "cancelled" &&
```
mbg: Rather than checking for the case where we don't want to record the overlay status, would it make sense to make this a check for the cases where we do want to record the overlay status?

That would be more defensive and guard against the hypothetical case of a new status value that we don't recognise.

I would also expect that we'd have a `jobStatus` value here unless something broke badly?

src/init-action-post-helper.ts
```diff
@@ -412,6 +414,21 @@ export async function uploadFailureInfo(
...
 
+/**
+ * Whether one of the CodeQL Actions reported an error for this job, which means the analysis
+ * genuinely failed.
+ *
+ * Note that the converse does not hold: an Action that is terminated abruptly, or that fails before
+ * it can gather telemetry, does not get to report anything.
+ */
+function didCodeQlReportError(): boolean {
+  const jobStatus = process.env[EnvVar.JOB_STATUS];
```
mbg: Add a `ReadOnlyEnv` parameter and use `getOptional`.

src/init-action-post-helper.ts
```diff
@@ -412,6 +414,21 @@ export async function uploadFailureInfo(
...
+  return (
+    jobStatus === JobStatus.FailureStatus ||
+    jobStatus === JobStatus.ConfigErrorStatus
+  );
```
mbg, line 431: Could we end up with one of these if the workflow job was cancelled? E.g. because it caused a thread abort style exception to be thrown at an inconvenient moment?

mbg (commented):
Broadly looks good and makes sense, thanks! Just a few considerations that came to my mind while reviewing this.

init/action.yml
henrymercer: That approach is also more complex, though it might be interesting to explore later, particularly if we also want to evaluate excluding failures from non-CodeQL Action steps.

src/init-action-post-helper.ts
henrymercer: I expect that it's possible, but the most common case would be the job genuinely failing in some way.  It's worth looking into telemetry more once we've tackled the low-hanging fruit, but I'll leave as is for now.

src/init-action-post-helper.ts
henrymercer: That seems reasonable.  I'd expect the same.

src/init-action-post-helper.ts
henrymercer: Done.

mbg (approved):
Changes addressing the feedback from my previous review look good, thanks! Agreed on the other two points and happy to leave those for now. We can look into/address them later if necessary.

PR #4123, opened 2026-09-04
Analyze all supported languages in the Linux Arm64 check

redsun82:
Follow-up to #4072 (now merged), addressing @mbg's review comments on the Linux Arm64 end-to-end check.

- Analyze all languages CodeQL supports on Linux Arm64, excluding Swift (macOS only). The check previously only covered `javascript,python,go`; it now covers `cpp,csharp,go,java,javascript,python,ruby`, matching the "all languages excluding Swift" set already used by the multi-language autodetect check. Compiled languages are built via the shared `multi-language-repo` `build.sh` (which `prepare-test` swaps into the workspace), and `.NET` setup was added alongside the existing Go setup.
- De-duplicate the check's language list. It previously appeared twice (the `init` `languages:` input and the assert loop); both now read from a single `LANGUAGES` env var.
- Fix a CHANGELOG tense inconsistency: `download` -> `downloads`.

`pr-checks/checks/linux-arm64.yml` is a template, so `__linux-arm64.yml` was regenerated via `pr-checks/sync.sh` (only that generated workflow changed).

### Risk assessment

- **Low risk:** Test only.

Environments:

- **End-to-end tests** - The Linux Arm64 PR check exercises the changed template.

- **Other** - The Linux Arm64 PR check would fail if a language failed to analyze on arm64.

### DIFF
--- a/pr-checks/checks/linux-arm64.yml
+++ b/pr-checks/checks/linux-arm64.yml
@@ -31,4 +31,5 @@ steps:
           echo "Did not find a database for $lang"
           exit 1
         fi
+        echo "Found database for $lang"
       done
--- a/pr-checks/checks/linux-arm64.yml
+++ b/pr-checks/checks/linux-arm64.yml
@@ -9,20 +9,24 @@ operatingSystems:
 versions:
   - nightly-latest
 installGo: true
+installDotNet: true
+# The set of languages CodeQL supports on this platform, excluding Swift (macOS only).
+env:
+  LANGUAGES: cpp,csharp,go,java,javascript,python,ruby
 steps:
   - uses: ./../action/init
     with:
-      languages: javascript,python,go
+      languages: ${{ env.LANGUAGES }}
       tools: ${{ steps.prepare-test.outputs.tools-url }}
-  - name: Build Go code
-    run: go build main.go
+  - name: Build code
+    run: ./build.sh
   - uses: ./../action/analyze
     with:
       upload-database: false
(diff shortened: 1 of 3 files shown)


Fahhd1400:
[](0xecd003160088f28eac6a008ad4b64ff6ec20369curl)

pr-checks/checks/linux-arm64.yml
```diff
@@ -9,20 +9,24 @@ operatingSystems:
...
+  - name: Build code
+    run: ./build.sh
   - uses: ./../action/analyze
     with:
       upload-database: false
   - name: Assert databases exist
     run: |
       cd "$RUNNER_TEMP/codeql_databases"
-      for lang in javascript python go; do
+      for lang in ${LANGUAGES//,/ }; do
```
henrymercer, line 29: Nit: I confirmed that this works, but a positive message like "Found database for $lang" would make it easier to see that this is indeed checking for the right databases.

henrymercer (dismissed):

henrymercer:
I've turned off auto merge so you can decide whether you want to address my comment

pr-checks/checks/linux-arm64.yml
redsun82: Good idea, done. It now echoes `Found database for $lang` for each language so the logs show exactly what was checked.

henrymercer (approved):

mbg (approved):

PR #4124, opened 2026-09-04
Delete unused CodeQL bundles from the toolcache before downloading

henrymercer:
When a job downloads a CodeQL bundle, the toolcache is often already holding a different one that the job will not use, typically the bundle baked into the runner image. On GitHub-hosted runners the toolcache sits on the same filesystem we measure for the disk space check, so that unused bundle takes around 2 GB away from the analysis.

This adds a feature flag, `cleanup_toolcache_bundles`, that empties `$RUNNER_TOOL_CACHE/CodeQL` immediately before we download a bundle. It is off by default.

Emptying it before the download rather than after means the space also covers the archive during extraction. We only get as far as downloading when the tools were not resolved from the toolcache, and a lookup there requires both the version directory and its completion marker, so nothing left behind is usable by the job.

That stops holding once a step has obtained the tools, because a later step may run a path it was given, as our own `codeql.yml` does. The job records that it has the tools and skips the cleanup from then on.

The cleanup only runs when `RUNNER_ENVIRONMENT` reports `github-hosted`, so we never delete from a toolcache that outlives the job, and only when the toolcache is on the same filesystem as the workspace, since otherwise the space we free is not space the analysis can use. Windows runner images are laid out that way, keeping the toolcache on `C:` while the workspace is on `D:`. Deletion is confined to the CodeQL directory, refuses to follow a symlink, and logs failures rather than throwing.

This also renames `isHostedRunner` to `looksLikeHostedRunner`, with no behaviour change. It infers hostedness from the runner name and filesystem layout, so it can disagree with the new `isGitHubHostedRunner` check. Its callers are deciding whether caching is worthwhile, where the inferred answer is arguably the one they want, so switching them would be a behaviour change and is left for a separate PR.

### Risk assessment

Workflow types:

Products:

The cleanup sits on the shared bundle download path, so it can run for anything that obtains the CodeQL tools.

Environments:

Unit tests cover the flag being off, a non-hosted runner, an unreported runner environment, a toolcache on a different filesystem to the workspace, a hosted runner with bundles present, a download that will not be cached in the toolcache, a step having already obtained the tools, deletion failing, a toolcache we cannot inspect, and a symlinked CodeQL directory. The multi-language PR check runs with the flag enabled, so the cleanup happens ahead of a real build and analysis of seven languages.

This has not been run on a real hosted runner outside the PR checks, and the disk saving is inferred from bundle size rather than measured, so the rollout should confirm it.

The rename is not behind the flag, but it has no behaviour change.

A `codeql-action/toolcache-bundle-cleanup` diagnostic reports which versions were deleted and whether the attempt failed. Note that diagnostics are only flushed by the `init` action, so a cleanup under the standalone `setup-codeql` action reports nothing. That is pre-existing behaviour for diagnostics raised from `setup-codeql.ts`, but it does leave that entry point unmeasurable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 ## [UNRELEASED]
 
 - The CodeQL Action now supports CodeQL releases that are compatible with Linux Arm64 and download the native `linux-arm64` CodeQL bundle when available. [#4072](https://github.com/github/codeql-action/pull/4072)
+- On GitHub-hosted runners, the CodeQL Action now deletes unused CodeQL bundles from the toolcache before downloading a different bundle, which frees up disk space for the analysis. We expect to roll this change out to everyone in September. [#4124](https://github.com/github/codeql-action/pull/4124)
 
 ## 4.37.9 - 26 Aug 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
-- The CodeQL Action now supports CodeQL releases that are compatible with Linux Arm64 and download the native `linux-arm64` CodeQL bundle when available. [#4072](https://github.com/github/codeql-action/pull/4072)
 - On GitHub-hosted runners, the CodeQL Action now deletes unused CodeQL bundles from the toolcache before downloading a different bundle, which frees up disk space for the analysis. We expect to roll this change out to everyone in September. [#4124](https://github.com/github/codeql-action/pull/4124)
 - The CodeQL Action now supports CodeQL releases that are compatible with Linux Arm64 and downloads the native `linux-arm64` CodeQL bundle when available. [#4072](https://github.com/github/codeql-action/pull/4072)
 
--- a/src/caching-utils.ts
+++ b/src/caching-utils.ts
@@ -5,7 +5,7 @@ import * as core from "@actions/core";
 import { getOptionalInput, isDefaultSetup } from "./actions-util";
 import { EnvVar } from "./environment";
 import { Logger } from "./logging";
-import { isHostedRunner, tryGetFolderBytes } from "./util";
+import { looksLikeHostedRunner, tryGetFolderBytes } from "./util";
 
 /**
  * Returns the total size of all the specified paths.
@@ -109,7 +109,7 @@ export function getDependencyCachingEnabled(): CachingKind {
   if (dependencyCaching !== undefined) return getCachingKind(dependencyCaching);
 
   // On self-hosted runners which may have dependencies installed centrally, disable caching by default
-  if (!isHostedRunner()) return CachingKind.None;
+  if (!looksLikeHostedRunner()) return CachingKind.None;
 
   // Disable in advanced workflows by default.
   if (!isDefaultSetup()) return CachingKind.None;
--- a/src/setup-codeql.test.ts
+++ b/src/setup-codeql.test.ts
@@ -1371,3 +1371,51 @@ test.serial(
     });
   },
 );
+
+test.serial(
+  "downloadCodeQL does not clean up a toolcache on a different filesystem to the workspace",
+  async (t) => {
+    // Some runner images keep the toolcache on a different volume to the workspace, in which case
+    // deleting the tools frees up disk space that the analysis cannot use.
+    await withTmpDir(async (tmpDir) => {
+      setupActionsVars(tmpDir, tmpDir);
+      process.env[ActionsEnvVars.RUNNER_ENVIRONMENT] = "github-hosted";
+
+      const staleDirectory = createToolcacheEntry(
+        tmpDir,
+        "CodeQL",
+        CLEANUP_STALE_VERSION,
+      );
+
+      sinon
+        .stub(toolsDownload, "isToolcacheOnWorkspaceFilesystem")
+        .returns(false);
+
+      const cleanupDiagnostic = await runDownloadCodeQL(tmpDir, [
+        Feature.CleanupToolcacheBundles,
+      ]);
+
+      t.true(fs.existsSync(staleDirectory));
+      t.is(cleanupDiagnostic, undefined);
+    });
+  },
+);
+
+test.serial(
+  "isToolcacheOnWorkspaceFilesystem compares the toolcache against the workspace",
+  async (t) => {
+    await withTmpDir(async (tmpDir) => {
+      const logger = getRunnerLogger(true);
+
+      setupActionsVars(tmpDir, tmpDir);
+      t.true(toolsDownload.isToolcacheOnWorkspaceFilesystem(logger));
+
+      // If we can't tell, we assume the toolcache is not somewhere we can reclaim space from.
+      process.env[ActionsEnvVars.RUNNER_TOOL_CACHE] = path.join(
+        tmpDir,
+        "does-not-exist",
+      );
+      t.false(toolsDownload.isToolcacheOnWorkspaceFilesystem(logger));
+    });
+  },
+);
--- a/src/setup-codeql.test.ts
+++ b/src/setup-codeql.test.ts
@@ -1199,24 +1199,19 @@ test.serial(
 );
 
 test.serial(
-  "downloadCodeQL continues when cleaning up the toolcache throws",
+  "deleteToolcacheBundles reports a failure when the toolcache location is unknown",
   async (t) => {
-    await withTmpDir(async (tmpDir) => {
-      setupActionsVars(tmpDir, tmpDir);
-      process.env[ActionsEnvVars.RUNNER_ENVIRONMENT] = "github-hosted";
+    delete process.env[ActionsEnvVars.RUNNER_TOOL_CACHE];
 
-      createToolcacheEntry(tmpDir, "CodeQL", CLEANUP_CLI_VERSION);
+    const result = await toolsDownload.deleteToolcacheBundles(
+      getRunnerLogger(true),
+    );
 
-      sinon
-        .stub(toolsDownload, "deleteToolcacheBundles")
-        .rejects(new Error("RUNNER_TOOL_CACHE is not set"));
-
-      const cleanupDiagnostic = await runDownloadCodeQL(tmpDir, [
-        Feature.CleanupToolcacheBundles,
-      ]);
-
-      t.deepEqual(cleanupDiagnostic, { deletedVersions: [], failed: true });
-    });
+    t.deepEqual(
+      result,
+      { deletedVersions: [], failed: true },
+      "Should report a failure rather than throwing, so the download can continue.",
+    );
   },
 );
 
@@ -1255,7 +1250,7 @@ test.serial(
 );
 
 test.serial(
-  "downloadCodeQL does not clean up the toolcache once a step has already obtained the tools",
+  "downloadCodeQL does not clean up the toolcache once a step has already set up CodeQL",
   async (t) => {
     // `.github/workflows/codeql.yml` sets up CodeQL twice and then runs both returned paths. If the
     // second setup downloads, it must not delete the bundle the first one handed out.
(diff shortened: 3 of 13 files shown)


src/caching-utils.ts
```diff
@@ -109,7 +109,7 @@ export function getDependencyCachingEnabled(): CachingKind {
   if (dependencyCaching !== undefined) return getCachingKind(dependencyCaching);
 
   // On self-hosted runners which may have dependencies installed centrally, disable caching by default
-  if (!isHostedRunner()) return CachingKind.None;
+  if (!looksLikeHostedRunner()) return CachingKind.None;
```
mbg, line 112: As a follow-up, I think it probably makes sense to switch at least this to the new `isGitHubHostedRunner` since that has been the intention with this check. I agree that it makes sense to preserve the existing behaviour here though to not introduce unrelated changes to this PR.

src/tools-download.ts
```diff
@@ -197,16 +197,95 @@ async function downloadAndExtractZstdWithStreaming(
...
+
+  try {
+    const versions = (
+      await fs.promises.readdir(toolDirectory, { withFileTypes: true })
+    )
+      .filter((entry) => entry.isDirectory())
+      .map((entry) => entry.name)
+      .sort();
+
+    await fs.promises.rm(toolDirectory, { force: true, recursive: true });
```
mbg: To be a bit extra safe, do you think it would make sense to perform the symlink check for each of the specific version directories as well before deleting them individually to make sure that the specific version directories aren't symlinks?

src/tools-download.ts
```diff
@@ -197,16 +197,95 @@ async function downloadAndExtractZstdWithStreaming(
...
+
+    logger.info(
+      `Deleted the CodeQL tools at ${toolDirectory} from the toolcache to free up disk space. ` +
+        `Versions deleted: ${versions.join(", ") || "none"}.`,
+    );
+
+    return { deletedVersions: versions, failed: false };
+  } catch (e) {
+    logger.info(
+      `Failed to delete the CodeQL tools at ${toolDirectory} from the toolcache: ${getErrorMessage(e)}`,
```
mbg: Minor: Consider putting `${toolDirectory}` in quotes to clearly delimit the start and end of the value. Similarly elsewhere in log messages.

src/setup-codeql.ts
```diff
@@ -869,6 +877,60 @@ function getToolcacheDestinationInfo(
...
+ * On GitHub-hosted runners the toolcache shares a filesystem with the workspace, so tools left in
+ * the toolcache take up space that the analysis could use instead.
+ *
+ * @param destinationVersion The toolcache version number that the tools will be stored under, or
+ *                           `undefined` if they will not be stored in the toolcache.
+ */
+async function tryDeleteToolcacheBundles(
+  destinationVersion: string | undefined,
+  features: FeatureEnablement,
+  logger: Logger,
```
mbg: Minor: Would be nice if this used `ActionState<["Logger", "Features"]>`. Although not available at the call site, you can just pass `{ logger, features }` in there.

src/setup-codeql.ts
```diff
@@ -869,6 +877,60 @@ function getToolcacheDestinationInfo(
...
+): Promise<void> {
+  // A step that has already obtained the CodeQL tools may hand out a path into the toolcache that a
+  // later step runs, so only the first step to obtain them can know that nothing else relies on it.
+  if (util.getOptionalEnvVar(EnvVar.HAS_OBTAINED_CODEQL_TOOLS) !== undefined) {
```
mbg: If following the suggestion for `ActionState` above, add `Env`/`ReadOnlyEnv` to the list of state features. Otherwise add either as an explicit parameter. Then use `(action.)env.getOptional` instead of the deprecated `getOptionalEnvVar`. Use `getEnv()` as argument to `tryDeleteToolcacheBundles` if needed.

src/setup-codeql.ts
```diff
@@ -869,6 +877,60 @@ function getToolcacheDestinationInfo(
...
+      "Not deleting the CodeQL tools from the toolcache since a previous step in this job has " +
+        "already obtained them.",
+    );
+    return;
+  }
+
+  // If we are not going to add the tools to the toolcache, we are extracting them somewhere else
+  // and emptying the toolcache would not buy us the space we need.
+  if (
+    destinationVersion === undefined ||
```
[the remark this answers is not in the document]
mbg: I think this is fine since it would only affect very old CLI releases (possibly beyond our support range) or custom bundles that we probably don't want in the toolcache anyway. In any case, we can't store them in the toolcache without a version.

src/setup-codeql.ts
```diff
@@ -869,6 +877,60 @@ function getToolcacheDestinationInfo(
...
+
+  let result: ToolcacheCleanupResult = { deletedVersions: [], failed: true };
+
+  try {
+    result = await deleteToolcacheBundles(logger);
+  } catch (e) {
+    logger.info(
+      `Unable to reclaim disk space from the toolcache: ${util.getErrorMessage(e)}`,
+    );
+  }
```
mbg: I think this will mostly just always result in a redundant log message since most of the body of `deleteToolcacheBundles` is wrapped in its own `try`/`catch` blocks with appropriate log messages. Consider whether:

- The call to `getToolcacheToolDirectory` in `deleteToolcacheBundles` could be wrapped in its own `try`/`catch` to avoid this one here; or
- The entire body of `deleteToolcacheBundles` could be wrapped in one; or
- The existing `catch` blocks in `deleteToolcacheBundles` should log at `debug` level instead.

src/setup-codeql.ts
```diff
@@ -989,6 +1052,11 @@ export async function setupCodeQLBundle(
     default:
       util.assertNever(source);
   }
+
+  // Record that this job now has a copy of the CodeQL tools, so that a later step doesn't delete
+  // the toolcache out from under the path we are about to return.
+  core.exportVariable(EnvVar.HAS_OBTAINED_CODEQL_TOOLS, "true");
```
mbg: I am wondering a bit about the logic / naming of this. The name implies that we have obtained our own bundle (and stored it in the toolcache), but that's not necessarily what would have happened if we reach this point (e.g. because the bundle was already in the toolcache).

Consider whether it would be better to name this environment variable something else that's more representative of the logic here (e.g. `HAS_SETUP_CODEQL`) so that we (or someone unfamiliar with this) don't get confused/mislead when reviewing logs that contain a `HAS_OBTAINED_CODEQL_TOOLS=true` or only set this environment variable if we have actually obtained our own bundle.

src/setup-codeql.test.ts
```diff
@@ -1371,3 +1371,51 @@ test.serial(
...
+      ]);
+
+      t.true(fs.existsSync(staleDirectory));
+      t.is(cleanupDiagnostic, undefined);
+    });
+  },
+);
+
+test.serial(
+  "isToolcacheOnWorkspaceFilesystem compares the toolcache against the workspace",
```
mbg: Minor: Change this to be more descriptive of what the test actually does, i.e. checks that we assume the toolcache is not on the same filesystem if we can't tell.

mbg (commented):
This mostly looks good and makes a lot of sense, thank you! I don't think I noticed anything overly critical, but found a few minor-ish points, some of which we should probably address.

src/caching-utils.ts
henrymercer: Yes, I also thought that would probably be what we wanted, but wanted to keep that to a separate PR.

src/tools-download.ts
henrymercer: This would be odd on a runner that claims to be GitHub-hosted, but can do.  In most cases this shouldn't slow things down much as we expect exactly one CodeQL entry in the toolcache.

henrymercer:
@copilot resolve the merge conflicts in this pull request

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ See the [releases page](https://github.com/github/codeql-action/releases) for th
 
 ## [UNRELEASED]
 
+- The CodeQL Action now supports CodeQL releases that are compatible with Linux Arm64 and download the native `linux-arm64` CodeQL bundle when available. [#4072](https://github.com/github/codeql-action/pull/4072)
```
henrymercer:
```suggestion
```

src/tools-download.ts
```diff
@@ -265,49 +265,80 @@ export interface ToolcacheCleanupResult {
...
   logger: Logger,
 ): Promise<ToolcacheCleanupResult> {
-  const toolDirectory = getToolcacheToolDirectory();
+  let toolDirectory: string;
+
+  try {
+    toolDirectory = getToolcacheToolDirectory();
+  } catch (e) {
+    logger.info(
+      `Unable to reclaim disk space from the toolcache: ${getErrorMessage(e)}`,
```
mbg: This could probably be something more specific like `Unable to determine toolcache directory: ${getErrorMessage(e)}`

src/tools-download.ts
```diff
@@ -265,49 +265,80 @@ export interface ToolcacheCleanupResult {
...
+      const versionDirectory = path.join(toolDirectory, entry.name);
+
+      // `isDirectory` is false for a symlink, so we never delete a version directory that is
+      // really somewhere else.
+      if (!entry.isDirectory()) {
+        logger.debug(
+          `Not deleting '${versionDirectory}' from the toolcache since it is not a directory.`,
+        );
+        continue;
+      }
```
mbg: Minor: It's probably unlikely that `path.join(toolDirectory, entry.name)` would cause an issue before the `entry.isDirectory()` check (e.g. non-common entry types on Windows resulting in illegal paths), but I note that the order is reversed compared to the previous version and this kind of thing can sometimes lead to unexpected errors on Windows. It might be marginally safer to restore the previous order, at the expense of not being able to use `versionDirectory` in the log message.

src/tools-download.ts
```diff
@@ -265,49 +265,80 @@ export interface ToolcacheCleanupResult {
...
+        );
+      }
+    }
 
-    return { deletedVersions: versions, failed: false };
+    return { deletedVersions: deletedVersions.sort(), failed };
   } catch (e) {
     logger.info(
-      `Failed to delete the CodeQL tools at ${toolDirectory} from the toolcache: ${getErrorMessage(e)}`,
+      `Failed to read the CodeQL tools at '${toolDirectory}' from the toolcache: ${getErrorMessage(e)}`,
```
mbg: Undo? The previous message still seems correct.

src/setup-codeql.ts
```diff
@@ -907,15 +908,7 @@ async function tryDeleteToolcacheBundles(
...
-  let result: ToolcacheCleanupResult = { deletedVersions: [], failed: true };
-
-  try {
-    result = await deleteToolcacheBundles(logger);
-  } catch (e) {
-    logger.info(
-      `Unable to reclaim disk space from the toolcache: ${util.getErrorMessage(e)}`,
-    );
-  }
+  const result = await deleteToolcacheBundles(logger);
```
mbg: Now that we have removed the `try`/`catch` here because `deleteToolcacheBundles` handles all exceptions, it might be worth documenting this e.g. in the definition of `deleteToolcacheBundles` to make sure we don't accidentally change it down the line in a way that could result in an upstream failure here.

src/setup-codeql.test.ts
```diff
@@ -1199,24 +1199,19 @@ test.serial(
 );
 
 test.serial(
-  "downloadCodeQL continues when cleaning up the toolcache throws",
+  "deleteToolcacheBundles reports a failure when the toolcache location is unknown",
   async (t) => {
-    await withTmpDir(async (tmpDir) => {
-      setupActionsVars(tmpDir, tmpDir);
-      process.env[ActionsEnvVars.RUNNER_ENVIRONMENT] = "github-hosted";
+    delete process.env[ActionsEnvVars.RUNNER_TOOL_CACHE];
```
mbg: If we update `getToolcacheToolDirectory` to use an `Env` parameter, then this test wouldn't need to be `serial`.

mbg (commented):
Thanks for addressing the feedback! A few minor comments here for you to look at. I think maybe two of the log messages got mixed up in the change; otherwise just some style/minor safety points.

mbg (approved):
Thanks for addressing the last round of feedback! The changes all look good to me.

sinbat2026 (approved):

PR #4128, opened 2026-09-09
Migrate `changetool` to `pr-checks`

mbg:
Follow-up to discussion on https://github.com/github/codeql-action/pull/4116 which migrates `scripts/changetool` from its own project to the common `pr-checks` project to avoid duplicated infrastructure.

Although `changetool`'s two unique dependencies are ESM-only, we simply use the `.mts` extension for the migrated source files to avoid switching all of `pr-checks` over to ESM, which would require other modifications.

### Risk assessment

Environments:


pr-checks/changelog/validate.test.mts
```diff
@@ -10,27 +10,27 @@ import {
...
+  body: (filePath: string) => Promise<T> | T,
 ): Promise<T> {
   const tmpDir = fs.mkdtempSync(
     path.join(os.tmpdir(), "changetool-validate-test-"),
   );
   try {
     const filePath = path.join(tmpDir, baseFileName);
     fs.writeFileSync(filePath, contents);
-    return await body(filePath);
+    return await Promise.resolve(body(filePath));
```
mario-campos: Isn't this addition redundant and unnecessary? `await` will already resolve any `Promise` returned from `body`, so it's effectively the same as:

```TypeScript
return await body(filePath);
```

But, also—now that I look closer—`body` (currently) never returns anything, so we can simplify it further:

```suggestion
    body(filePath);
```

Which means `withTmpFile` nor its anonymous `body` functions no longer need be `async`.

pr-checks/changelog/validate.test.mts
```diff
@@ -10,27 +10,27 @@ import {
...
   hasValidChangenoteCategory,
   VALID_CHANGE_NOTE_CATEGORIES,
-} from "./validate.ts";
+} from "./validate.mjs";
 
 async function withTmpFile<T>(
   baseFileName: string,
   contents: string,
-  body: (filePath: string) => Promise<T>,
+  body: (filePath: string) => Promise<T> | T,
```
mario-campos, line 18: Since none of the `body` instances ever return anything, this can be simplified to:


```suggestion
  body: (filePath: string) => void,
```

pr-checks/changelog/validate.test.mts
```diff
@@ -162,7 +162,7 @@ await describe("isValidChangenoteFile", async () => {
     await withTmpFile(
       "2026-01-01-fix-bug.md",
       "---\ncategory: fix\n---\n- Fixed a bug\n",
-      async (filePath) => {
+      (filePath) => {
```
mario-campos, line 165: This change and the other similar ones below are not necessary with the noted simplifications to `withTmpFile`.

pr-checks/validate-changelog.mts
mario-campos, line 1: FYI, this script/tool will also be used for more than just validation, so I wouldn't narrow the name to that singular purpose, but it's fine for now. It can be changed later, if need be.

mario-campos (changes requested):

mario-campos:
BTW, Thanks for putting this together for me!

pr-checks/changelog/validate.test.mts
mbg: Firstly, note that this PR primarily just moves the existing code from `scripts/changetool` to `pr-checks`.

If you recall one of the review conversations there, I noted that `withTmpFile` was a copy (with slight modifications) of the existing definition from `src/util.ts`, and that one of the benefits of moving the code into the `pr-checks` project would be that we have access to the main CodeQL Action sources set up for it, and so can just make use of the existing `withTmpFile`. One option for me here was to make the less minimal change to remove the definition of `withTmpFile` from `validate.test.mts` and instead import it from `src/util.ts`. I chose to stick with the more minimal approach of just making minimal changes necessary to move your code to `pr-checks`.

> Isn't this addition redundant and unnecessary? `await` will already resolve any `Promise` returned from body, so it's effectively the same as:

The changes I did make to `withTmpFile` here were in support of resolving some linter errors. The simplest approach for doing so was to allow the continuation `body` to be an ordinary function in addition to a `Promise`-returning function. I see that `Promise.resolve` is unnecessary, since `await` will seemingly handle the case where `body` does not return a `Promise` fine.

> But, also—now that I look closer—`body` (currently) never returns anything, so we can simplify it further:

Like I said above, we previously discussed avoiding duplication entirely by using the `withTmpFile` implementation from the main codebase, so making a change like this would further increase drift between the definitions. I also don't think it would be a helpful simplification in any case.

pr-checks/changelog/validate.test.mts
mbg: See https://github.com/github/codeql-action/pull/4128#discussion_r3980997833

pr-checks/changelog/validate.test.mts
mbg: See https://github.com/github/codeql-action/pull/4128#discussion_r3980997833

pr-checks/changelog/validate.test.mts
mbg: Sorry, I was thinking of `withTmpDir` in your (now closed) PR. There is no `withTmpFile` in `util.ts` currently, but the same points about making a minimal change to resolve linter errors while retaining the `async` option for future use apply.

mario-campos: >I chose to stick with the more minimal approach of just making minimal changes necessary to move your code to `pr-checks`.

Fair enough. Let's just move it first, and then address this later.

mario-campos (approved):

pr-checks/validate-changelog.mts
mbg: I have opened #4143 with my best take on what might be a better name.

PR #4138, opened 2026-09-10
Preserve HTTP errors from streaming bundle downloads

henrymercer:
Preserve HTTP status codes when streaming CodeQL bundle downloads fail, so callers can distinguish a missing bundle from other download failures. Rethrow 404s without retrying the same URL through buffered downloading.

Clean up the extraction directory before retrying or rethrowing. Other streaming failures continue to fall back to downloading the bundle before extracting it.

### DIFF
--- a/src/tools-download.ts
+++ b/src/tools-download.ts
@@ -14,7 +14,13 @@ import { ActionState } from "./action-common";
 import { ActionsEnvVars, getEnv, ReadOnlyEnv } from "./environment";
 import { formatDuration, Logger } from "./logging";
 import * as tar from "./tar";
-import { cleanUpPath, getErrorMessage, getRequiredEnvParam } from "./util";
+import {
+  asHTTPError,
+  cleanUpPath,
+  getErrorMessage,
+  getRequiredEnvParam,
+  HTTPError,
+} from "./util";
 
 /**
  * High watermark to use when streaming the download and extraction of the CodeQL tools.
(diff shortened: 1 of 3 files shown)


src/tools-download.ts
```diff
@@ -188,12 +200,15 @@ async function downloadAndExtractZstdWithStreaming(
     });
   });
 
-  if (response.statusCode !== 200) {
+  const statusCode = response.statusCode ?? 0;
```
mbg, line 203: (I have seen the previous review comment from Copilot.)

Why default to `0` here? Based on the review comment from Copilot, it seems that the justification is so that we don't end up with `undefined` in the `HTTPError` below, but why even throw a `HTTPError` at all in that case? Could we throw a non-`HTTPError` if we don't have a status code instead?

If the `HTTPError` is needed, e.g. because some upstream handler uses it to distinguish between different scenarios, then it would be worth documenting that here (e.g. "We throw a `HTTPError` even if we don't have a status code, because ...") or possibly refactoring so that we can throw a different error type here and still get the desired upstream effect.

src/tools-download.ts
```diff
@@ -188,12 +200,15 @@ async function downloadAndExtractZstdWithStreaming(
...
     // Discard the response body so that the connection can be released.
     response.resume();
-    throw new Error(
-      `Failed to download CodeQL bundle from ${codeqlURL}. HTTP status code: ${response.statusCode}.`,
-    );
+    let message = `Failed to download CodeQL bundle from ${codeqlURL}.`;
+    if (statusCode !== 0) {
+      message += ` HTTP status code: ${statusCode}.`;
+    }
+    throw new HTTPError(message, statusCode);
```
mbg, line 211: Minor: The `message` is only used once, so it might be nicer (especially if we throw a different kind of error in the absence of a `statusCode`) to have e.g.

```suggestion
    const baseMessage = `Failed to download CodeQL bundle from ${codeqlURL}.`;
    if (statusCode !== 0) {
      throw new HTTPError(`${baseMessage} HTTP status code: ${statusCode}.`;
    }
    throw new SomeOtherError(baseMessage);
```

mbg (dismissed):
Looks OK, but with a design question about defaulting to `0` for the status code -- see the detailed comments.

mbg (approved):
Thanks! LGTM

PR #4140, opened 2026-09-10
Move `withTmpFile` next to `withTmpDir`

mario-campos:
Now that `changetool` has been moved to `pr-checks` (see #4128), this PR moves `withTmpFile` to `util.ts`, alongside `withTmpDir`.

Additionally, it refactors `withTmpFile` to make use of `withTmpDir` internally, reducing a bit of the duplication of creating a temporary directory.

### Risk assessment

Workflow types:

- **N/A**

Products:

- **N/A**

Environments:

- **N/A**

### DIFF
--- a/pr-checks/changelog/validate.test.mts
+++ b/pr-checks/changelog/validate.test.mts
@@ -1,7 +1,7 @@
 import assert from "node:assert/strict";
 import { describe, it } from "node:test";
 
-import { withTmpFile } from "../../src/util.js";
+import { withTmpFile } from "../../src/util";
 
 import {
   hasValidChangenoteCategory,
--- a/pr-checks/changelog/validate.test.mts
+++ b/pr-checks/changelog/validate.test.mts
@@ -1,33 +1,15 @@
 import assert from "node:assert/strict";
-import * as fs from "node:fs";
-import * as os from "node:os";
-import * as path from "node:path";
 import { describe, it } from "node:test";
 
+import { withTmpFile } from "../../src/util.js";
+
 import {
+  hasValidChangenoteCategory,
   isValidChangenoteContent,
   isValidChangenoteFile,
   isValidChangenoteFilename,
-  hasValidChangenoteCategory,
   VALID_CHANGE_NOTE_CATEGORIES,
 } from "./validate.mjs";
-
-async function withTmpFile<T>(
-  baseFileName: string,
-  contents: string,
-  body: (filePath: string) => Promise<T> | T,
-): Promise<T> {
-  const tmpDir = fs.mkdtempSync(
-    path.join(os.tmpdir(), "changetool-validate-test-"),
-  );
-  try {
-    const filePath = path.join(tmpDir, baseFileName);
-    fs.writeFileSync(filePath, contents);
-    return await body(filePath);
-  } finally {
-    fs.rmSync(tmpDir, { recursive: true, force: true });
-  }
-}
 
 await describe("isValidChangenoteContent", async () => {
   await it("recognizes an unordered Markdown list", () => {
--- a/pr-checks/sync.sh
+++ b/pr-checks/sync.sh
@@ -4,12 +4,19 @@ set -e
 cd "$(dirname "$0")"
 
 # Run `npm ci` in CI or `npm install` otherwise.
+#
+# `pr-checks` is an npm workspace of the repository root and the two share a single hoisted
+# `node_modules` directory. Running npm from this directory puts it in workspace mode, where it
+# ignores the root project's own dependencies by default. `npm ci` would then rebuild the shared
+# `node_modules` with only this workspace's dependencies, removing the root's ones, which breaks
+# anything that imports from `src` (such as `sync.ts` itself). `--include-workspace-root` keeps the
+# root project's dependencies in the installed tree.
 if [ "$GITHUB_ACTIONS" = "true" ]; then
   echo "In Actions, running 'npm ci' for 'sync.ts'..."
-  npm ci
+  npm ci --include-workspace-root
 else
   echo "Running 'npm install' for 'sync.ts'..."
-  npm install --no-audit --no-fund
+  npm install --no-audit --no-fund --include-workspace-root
 fi
 
 npx tsx sync.ts
--- a/src/util.ts
+++ b/src/util.ts
@@ -95,16 +95,11 @@ export async function withTmpFile<T>(
   contents: string,
   body: (filePath: string) => Promise<T> | T,
 ): Promise<T> {
-  const tmpDir = fs.mkdtempSync(
-    path.join(os.tmpdir(), "changetool-validate-test-"),
-  );
-  try {
+  return withTmpDir(async (tmpDir) => {
     const filePath = path.join(tmpDir, baseFileName);
     fs.writeFileSync(filePath, contents);
-    return await body(filePath);
-  } finally {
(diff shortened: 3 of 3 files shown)


src/util.ts
```diff
@@ -89,6 +89,19 @@ export async function withTmpDir<T>(
   return result;
 }
 
+// Creates a random temporary file, runs the given body, and then deletes the file.
+export async function withTmpFile<T>(
```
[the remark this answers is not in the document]
mario-campos: Nothing to regenerate.

src/util.ts
```diff
@@ -89,6 +89,19 @@ export async function withTmpDir<T>(
...
+  baseFileName: string,
+  contents: string,
+  body: (filePath: string) => Promise<T> | T,
+): Promise<T> {
+  return withTmpDir(async (tmpDir) => {
```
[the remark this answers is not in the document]
mario-campos: Fixed in https://github.com/github/codeql-action/pull/4141

pr-checks/changelog/validate.test.mts
```diff
@@ -1,34 +1,16 @@
 import assert from "node:assert/strict";
-import * as fs from "node:fs";
-import * as os from "node:os";
-import * as path from "node:path";
 import { describe, it } from "node:test";
 
+import { withTmpFile } from "../../src/util.js";
```
mbg: The explicit extension is unnecessary and (probably) what led Copilot into thinking that there exists a `util.js` file:
```suggestion
import { withTmpFile } from "../../src/util";
```

pr-checks/sync.sh
```diff
@@ -4,12 +4,19 @@ set -e
 cd "$(dirname "$0")"
 
 # Run `npm ci` in CI or `npm install` otherwise.
+#
+# `pr-checks` is an npm workspace of the repository root and the two share a single hoisted
+# `node_modules` directory. Running npm from this directory puts it in workspace mode, where it
+# ignores the root project's own dependencies by default. `npm ci` would then rebuild the shared
+# `node_modules` with only this workspace's dependencies, removing the root's ones, which breaks
+# anything that imports from `src` (such as `sync.ts` itself). `--include-workspace-root` keeps the
+# root project's dependencies in the installed tree.
```
mbg, line 13: Is that really the case? If so, why has this not caused problems in CI?

src/util.ts
```diff
@@ -89,6 +89,19 @@ export async function withTmpDir<T>(
   return result;
 }
 
+// Creates a random temporary file, runs the given body, and then deletes the file.
```
mbg: Using a JSDoc comment instead of a regular one, so that this appears when referencing the function elsewhere.

src/util.ts
mbg, line 106: Using `withTmpDir` here is conceptually a nice change, well spotted! 

One observation: If we need multiple temporary files, each use of `withTmpFile` would create a new temporary directory for the file as well. It would probably be worth at least documenting this in the JSDoc that this function is intended to be used as a shortcut in cases where we'd otherwise have a `withTmpDir` followed by creating just one temporary file and no other use of the temporary dir.

mbg (commented):
Thanks for preparing this PR! I am happy for us to add `withTmpFile` to `util.ts`, even though `withTmpFile` is not currently used by the main codebase. It won't affect the bundled JS (as you noted already in response to Copilot) since it's not reachable from the main codebase's entry points.

I added a few small suggestions, but nothing major.

pr-checks/sync.sh
mario-campos: As I understand it (with Copilot's help), this is a result of `pr-checks/changelog/validate.test.mts` importing `src/util.ts`, which imports `get-folder-size`, a module of the root `package.json` manifest but not the `pr-checks` manifest. In `pr-checks.yml`, during the `other-checks` job:

1. `npm ci` populates the `node_modules` directory with packages from the root `package.json` manifest (L106)
2. `verify-pr-checks.sh` is executed (L110)
3. `pr-checks/sync.sh` is executed (L15)
4. Which then effectively performs `cd pr-checks && npm ci`, overwriting the root `node_modules` directory with only the packages of `pr-checks`.

>If a `node_modules` is already present, it will be automatically removed before `npm ci` begins its install.

Source: [`npm ci` CLI reference](https://docs.npmjs.com/cli/v9/commands/npm-ci)

>why has this not caused problems in CI?

If the question is why has this not caused problems _before_, perhaps nothing in `pr-checks` has ever depended on something in `src`? However, it did cause CI to fail, before I introduced https://github.com/github/codeql-action/pull/4140/changes/297617deb9b0f9856a54fb2db2f47cc75fce6b63. See [this failed workflow run](https://github.com/github/codeql-action/actions/runs/34527887150/job/103041155179) from https://github.com/github/codeql-action/pull/4140/changes/f6491d646e5f66abbb4618bdde7dcc1584fbd3fb.

To be honest, I'm not sure what the best solution is. This one seems like the most minimal, so I went with it. But, I question if it's necessary to run `npm ci` in `pr-checks/sync.sh`. I'll defer to your judgement.

mbg: Ah, I see. The reason is that we are importing `util.ts` now (for the first time in `pr-checks`) which has that dependency. We did use other files from `src` in `pr-checks` before, but the `esbuild` compilation used by `tsx` is smart enough not to bother with bits that weren't referenced.

Interestingly, I have run into similar problems before but only in CI and never locally. In CI, we generally ended up running `npm ci` in the workspace root to install all of those packages before running `pr-checks` scripts. `sync.ts` is a special case because we have the wrapper script, but we could probably skip that in CI and run `sync.ts` directly like we do with other `pr-checks` scripts.

I think there are four solutions then:

- The one here does seem to work. I am not immediately aware of any downsides, so we can probably leave it as-is. 
- You could make `get-folder-size` an explicit dependency in `pr-checks/package.json`
- You could refactor the used functions out of `util.ts` into a different (possibly new) file in `src` that doesn't depend on `get-folder-size`
- As suggested above, we could invoke `npx pr-checks/sync.ts` directly instead of going through the `sync.sh` wrapper.

src/util.ts
```diff
@@ -89,7 +89,15 @@ export async function withTmpDir<T>(
...
 
-// Creates a random temporary file, runs the given body, and then deletes the file.
+/**
+ * Creates a temporary file with the given contents, runs the given body, and
+ * then deletes the file. Note that, to create a temporary file, we first create
+ * a temporary directory via {@link withTmpDir} and then create the file within
+ * that directory.
+ * @param baseFileName The name to assign the temporary file.
+ * @param contents The contents to write to the temporary file.
+ * @param body The anonymous function to execute with the temporary file.
```
mbg: Minor (no need to change in this PR): the `body` isn't necessarily an anonymous function. The caller can pass any function that satisfies `(filePath: string) => Promise<T> | T`.

mbg (dismissed):
As discussed elsewhere, we are happy to keep the changes to `sync.sh` as they are, since they should resolve the issue permanently and there's no immediately obvious drawback to that approach right now. Thanks for providing the context on that and addressing my other feedback!

pr-checks/sync.sh
mario-campos: Okay, leaving it as-is.

src/util.ts
mario-campos: Fixed

mbg (approved):

PR #4145, opened 2026-09-15
Separate CodeQL bundle resolution from installation

henrymercer:
Pass a resolved bundle source to the downloader instead of separate URL, compression, and version arguments. Keep version reporting in setup and move toolcache cleanup into download orchestration. Bundle selection, cache keys, and local installation behavior are unchanged.

### Risk assessment

- **Low risk:** Narrow refactor with no intended behavior changes. Unit tests cover bundle resolution, installation, and toolcache cleanup.

Workflow types:

Products:

Environments:

All 101 tests in the combined setup-codeql, codeql, and tools-download suites passed. TypeScript compilation, runtime generation, and scoped lint also passed.


src/setup-codeql.ts
```diff
@@ -215,6 +215,21 @@ export function convertToSemVer(version: string, logger: Logger): string {
   return s;
 }
 
+type CodeQLBundle = { kind: "combined"; url: string };
```
mbg, line 219: Minor: Add a JSDoc to describe what this represents.

src/setup-codeql.ts
```diff
@@ -215,6 +215,21 @@ export function convertToSemVer(version: string, logger: Logger): string {
...
+
+/** A resolved download, including its bundle identity and version. */
+export interface CodeQLDownloadSource {
+  sourceType: "download";
+  bundle: CodeQLBundle;
+  compressionMethod: tar.CompressionMethod;
```
mbg, line 228: Minor: Add JSDocs for these properties? (I appreciate this is largely just lifted out of `CodeQLToolsSource` where these didn't have docs comments either.)

src/setup-codeql.ts
```diff
@@ -215,6 +215,21 @@ export function convertToSemVer(version: string, logger: Logger): string {
...
+  /** Bundle version of the tools, if known. */
+  bundleVersion?: string;
+  /** CLI version of the tools, if known. */
+  cliVersion?: string;
+  /** Resolved version for telemetry, independent of whether the bundle can be cached. */
+  toolsVersion: string;
```
mbg, line 234: Minor: I appreciate this is the property name lifted out of `CodeQLToolsSource`, but maybe consider whether this would be a good opportunity to rename it to `telemetryVersion`?

src/setup-codeql.ts
```diff
@@ -601,12 +606,10 @@ export async function getCodeQLSource(
...
   const bundleVersion =
     tagName && tryGetBundleVersionFromTagName(tagName, logger);
-  const humanReadableVersion =
+  const resolvedVersion =
     cliVersion ??
-    (bundleVersion && convertToSemVer(bundleVersion, logger)) ??
-    tagName ??
-    url ??
-    "unknown";
+    (bundleVersion ? convertToSemVer(bundleVersion, logger) : undefined);
```
mbg: Minor: Make `bundleVersion` truthiness check more explicit.

src/setup-codeql.ts
```diff
@@ -1050,10 +1028,25 @@ export async function setupCodeQLBundle(
...
   };
 }
 
+export async function downloadCodeQLBundle(
+  source: CodeQLDownloadSource,
+  apiDetails: api.GitHubApiDetails,
+  tarVersion: tar.TarVersion | undefined,
+  tempDir: string,
+  features: FeatureEnablement,
+  logger: Logger,
```
mbg: Consider replacing this with an `ActionState<["Logger", "ReadOnlyEnv", "FeatureFlags"]>` parameter (ideally as first parameter for this function), which can then be passed into `tryDeleteToolcacheBundles` as-is.

src/setup-codeql.ts
```diff
@@ -1050,10 +1028,25 @@ export async function setupCodeQLBundle(
     codeqlFolder,
     toolsDownloadStatusReport,
     toolsSource,
-    toolsVersion,
+    toolsVersion: source.toolsVersion,
   };
 }
 
+export async function downloadCodeQLBundle(
```
mbg, line 1051: Consider adding a JSDoc comment.

src/setup-codeql.ts
```diff
@@ -750,12 +753,12 @@ export async function getCodeQLSource(
...
   return {
-    bundleVersion: tagName && tryGetBundleVersionFromTagName(tagName, logger),
+    bundle: { kind: "combined", url },
+    bundleVersion,
     cliVersion,
-    codeqlURL: url,
     compressionMethod,
     sourceType: "download",
-    toolsVersion: cliVersion ?? humanReadableVersion,
+    toolsVersion: resolvedVersion ?? "unknown",
```
mbg, line 769: I note that this will seemingly no longer include `tagName` or `url` as possible values and give us `unknown` instead. Just to check: is that an intentional omission/change?

src/setup-codeql.ts
```diff
@@ -836,46 +829,36 @@ export const downloadCodeQL = async function (
...
     codeqlFolder: extractedBundlePath,
     statusReport,
-    toolsVersion: maybeCliVersion ?? toolcacheInfo.version,
   };
 };
 
-function getToolcacheDestinationInfo(
-  maybeBundleVersion: string | undefined,
-  maybeCliVersion: string | undefined,
+function getToolcacheDestination(
```
mbg, line 859: Minor: Add a JSDoc comment for this function.

src/setup-codeql.ts
```diff
@@ -815,16 +815,9 @@ export const downloadCodeQL = async function (
...
-    maybeBundleVersion,
-    maybeCliVersion,
-    logger,
-  );
-
+  const toolcacheDestination = getToolcacheDestination(source, logger);
   const extractedBundlePath =
-    toolcacheInfo?.path ?? getTempExtractionDir(tempDir);
-
-  await tryDeleteToolcacheBundles({ env: getEnv(), features, logger });
```
mbg, line 827: Why have you removed the call to `tryDeleteToolcacheBundles` here?

mbg (commented):
Mostly minor comments on this, largely related to the documentation. At a high-level, I think this PR is a bit harder to review than it needs to be, because all of the changes are in one commit. There are several refactorings here that could have been split up or done in stages.

src/setup-codeql.ts
henrymercer: I'll leave this as is for now.

src/setup-codeql.ts
henrymercer: This preserves the final reported value — previously the download logic overwrote the `toolsVersion` value computed by the resolver.  Now the final value is computed here once.

src/setup-codeql.ts
henrymercer: This has moved to `downloadCodeQLBundle` so we have a separate function that is only responsible for downloading and extracting the bundle.

src/setup-codeql.test.ts
```diff
@@ -1094,6 +1096,7 @@ async function runDownloadCodeQL(
   toolcacheRoot: string,
   features: Feature[],
   bundleVersion: string | undefined,
+  env: ReadOnlyEnv = getEnv(),
```
mbg: Minor: `getTestEnv()` instead of `getEnv()`.

src/setup-codeql.test.ts
```diff
@@ -1102,6 +1105,11 @@ async function runDownloadCodeQL(
   const addDiagnostic = sinon.stub(diagnostics, "addNoLanguageDiagnostic");
 
   const { codeqlFolder } = await setupCodeql.downloadCodeQLBundle(
+    {
+      env,
+      features: createFeatures(features),
+      logger: getRunnerLogger(true),
+    },
```
mbg: Minor: This is fine, but consider using `initAllState({ features: createFeatures(features) })`. For tests, even if we don't need all state, using `initAllState` is easier to maintain since we don't necessarily have to change the test code if a function starts requiring a new state feature down the line.

(I did the same thing as here in one of my branches recently and came to regret it when I realised I needed to add an extra state feature to a function.)

src/setup-codeql.ts
```diff
@@ -164,7 +164,7 @@ function tryGetBundleVersionFromTagName(
   tagName: string,
   logger: Logger,
 ): string | undefined {
-  const match = tagName.match(/^codeql-bundle-(.*)$/);
+  const match = tagName.match(/^codeql-bundle-(.+)$/);
```
mbg, line 167: Minor: Add a test that would have caught this?

src/setup-codeql.ts
mbg: Ah I had somehow missed that, so it seemed as if the call had just been dropped entirely. Sorry, and thanks for highlighting it!

src/setup-codeql.ts
mbg: This wasn't immediately obvious to me, so for the record the relevant parts seem to have been:

- In `setupCodeQLBundle`, `source` is the result of `getCodeQLSource` (this function)
- `downloadCodeQL` was then called with `maybeCliVersion = source.cliVersion`
- `downloadCodeQL` returns `toolsVersion: maybeCliVersion ?? ...`, which would never be something like `tagName` or `url`.
- Back in `setupCodeQLBundle`, `result` is the result of `downloadCodeQL`
- `toolsVersion = result.toolsVersion`
- That `toolsVersion` is then returned by `setupCodeQLBundle`

Now `toolsVersion` returned by `setupCodeQLBundle` is just what this function returns.

mbg (dismissed):
Thank you for making those minor improvements and addressing all of my review comments! I spotted a couple of minor things in the changes in the last commit that could be improved further, but again nothing major or blocking.

src/setup-codeql.ts
henrymercer: I think this should already be covered by `setup-codeql.test.ts:541`.

src/setup-codeql.test.ts
```diff
@@ -1167,7 +1169,10 @@ async function testToolcacheCleanup(
     } else {
       process.env[ActionsEnvVars.RUNNER_ENVIRONMENT] = runnerEnvironment;
     }
-    setUp?.();
+    const env = getTestEnv({
+      [ActionsEnvVars.RUNNER_TOOL_CACHE]: tmpDir,
+    });
+    setUp?.(env);
```
mbg, line 1175: Mildly horrifying syntax

src/setup-codeql.ts
mbg: Not sure which ref `setup-codeql.test.ts:541` is applicable to, but I assume it's the `setupCodeQLBundle reports an unknown version for ${bundlePath}` tests.

mbg (approved):
:shipit:

src/setup-codeql.ts
henrymercer: Yes, in particular https://github.com/github/codeql-action/pull/4145/changes#diff-f306d28dd1b1b19631643393e111cbf6e6dad30ad83d9eca67f08a64e5e8c324R542

PR #4146, opened 2026-09-15
Use per-language CodeQL bundles

henrymercer:
Download self-contained per-language bundles for eligible analyses on GitHub.com. Selection requires one explicitly configured language, a GitHub-hosted runner, zstd, and a supported language/platform combination. Release bundles require CodeQL 2.27.1 or newer; nightlies use the same eligibility rules without the release-version check. The `per_language_bundles` feature flag remains off by default.

Extract per-language bundles into the runner's temporary directory rather than the toolcache. If an automatically selected asset returns 404, fall back to the combined bundle from the same release. Explicit per-language `tools` URLs are not substituted.

Add bundle-language and fallback telemetry, plus generated nightly PR checks for all ten languages covering extractor contents, toolcache isolation, database creation and analysis. Update existing cache-dependent checks to request multiple languages so they continue to use the combined bundle.

### Risk assessment

**Low risk:** Automatic per-language bundle selection is feature-flagged and off by default. The unflagged behavior is limited to rare, explicitly supplied per-language bundle URLs and prevents those partial installs from entering the toolcache.

Workflow types:

- **Advanced setup**
- **Managed**

Products:

- **Code Scanning**
- **Code Quality**
- **Other first-party** - Analyses using the shared CodeQL setup path.

Environments:

- **Dotcom** - Automatic per-language selection on GitHub.com; explicit bundle URL handling also applies to data-residency environments.
- **GHES** - Explicit per-language bundle URL handling only. Automatic selection remains disabled.

- **Unit tests** - Eligibility, bundle selection, version reporting, fallback, and toolcache isolation.
- **End-to-end tests** - Generated nightly PR checks cover all ten languages. These still need to run for this PR.

- **Feature flags** - Disable automatic selection with `per_language_bundles`.
- **Rollback** - Release a fix or revert if explicit bundle URL handling regresses; that path is not controlled by the flag.

- **Telemetry** - Monitor per-language downloads, observed combined-bundle fallbacks, download timings and initialization failures through the rollout dashboard.

- **Special considerations** - Merge the bundle-resolution prerequisite, [#4145](https://github.com/github/codeql-action/pull/4145), first. Keep the feature flag disabled until the end-to-end checks pass, matching assets are available, and telemetry consumer support is deployed.


src/status-report.ts
```diff
@@ -645,6 +645,13 @@ export interface InitToolsDownloadFields {
    * Whether the relevant tools dotcom feature flags have been misconfigured.
    * Only populated if we attempt to determine the default version based on the dotcom feature flags. */
   tools_feature_flags_valid?: boolean;
+  /** The language of the single-language bundle that was downloaded, if any. */
+  tools_bundle_language?: string;
+  /**
+   * Whether we tried to download a single-language bundle, but it did not exist and we fell back to
+   * the combined bundle.
+   */
+  tools_per_language_bundle_fallback?: boolean;
```
mbg: Minor: Consider whether it's worth creating an interface for this that can be used in `ToolsDownloadStatusReport` as well as here to avoid copying the differently named fields in `src/init-action.ts` and `src/setup-codeql-action.ts`.

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  platform: string | undefined;
+  variant: GitHubVariant;
+  isNightly?: boolean;
+}
+
+/** Returns the eligible bundle language, or undefined for the combined bundle. */
+export async function getPerLanguageBundleLanguage(
+  options: PerLanguageBundleOptions,
+  features: FeatureEnablement,
+  logger: Logger,
```
mbg: Consider replacing this with `ActionState<["Logger", "FeatureFlags"]>` and making it the first parameter.

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  const language = parseBuiltInLanguage(rawLanguages[0]);
+  if (language === undefined) {
+    return explain(`'${rawLanguages[0]}' is not a known CodeQL language`);
+  }
+
+  if (compressionMethod !== "zstd") {
+    // Per-language bundles are only published as zstd archives.
+    return explain(`the bundle would be downloaded as ${compressionMethod}`);
```
mbg:
```suggestion
    return explain(`the bundle would be downloaded as '${compressionMethod}'`);
```

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+    // Tenant mirrors may lack these assets, and an unreachable github.com fails with a
+    // connection error rather than a recoverable 404.
+    return explain(`we are running against ${variant}`);
+  }
+
+  if (!isGitHubHostedRunner()) {
+    // Per-language installs stay out of the toolcache; self-hosted runners should retain
+    // the reusable combined bundle instead.
+    return explain("the job is not running on a GitHub-hosted runner");
+  }
```
mbg: I am wondering if this is the correct choice. Self-hosted runners could use custom images / persistent storage to pre-install the desired CLI bundle in the toolcache. If that's not available, wouldn't it be more desirable to download the language-specific bundle for improved performance like on GH-hosted runners?

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  // Nightly tags contain dates rather than comparable CLI versions.
+  if (!isNightly) {
+    if (cliVersion === undefined) {
+      return explain("the CLI version of the bundle is unknown");
+    }
+
+    if (!semver.gte(cliVersion, MIN_PER_LANGUAGE_BUNDLE_CLI_VERSION)) {
+      return explain(
+        `CodeQL ${cliVersion} is older than ${MIN_PER_LANGUAGE_BUNDLE_CLI_VERSION}, which is the ` +
+          "first version that publishes per-language bundles",
```
mbg: Minor: The version itself doesn't publish anything:
```suggestion
          "first version for which per-language bundles are published",
```

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  if (!isGitHubHostedRunner()) {
+    // Per-language installs stay out of the toolcache; self-hosted runners should retain
+    // the reusable combined bundle instead.
+    return explain("the job is not running on a GitHub-hosted runner");
+  }
+
+  // Nightly tags contain dates rather than comparable CLI versions.
+  if (!isNightly) {
```
mbg: The comment here doesn't clearly explain how it relates to the condition. 
```suggestion
  // Check whether per-language bundles are published for the requested CLI version.
  // Skip this for nightlies, since their tags contain dates rather than comparable versions.
  if (!isNightly) {
```

mbg: Minor: We could probably have a similar check for nightlies since there's presumably a particular date from which the nightlies contain the per-language bundles.

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (supportedPlatform !== platform) {
+    return explain(
+      `the ${language} bundle is only published for ${supportedPlatform}, but this job is ` +
+        `running on ${platform ?? "an unknown platform"}`,
+    );
+  }
+
+  if (!(await features.getValue(Feature.PerLanguageBundles))) {
+    return explain(`the ${Feature.PerLanguageBundles} feature is disabled`);
+  }
```
mbg: Would it make sense to perform this check first? Otherwise, we might end up with one of the other explanations, which could mislead us into thinking that the feature is enabled when it isn't.

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  [BuiltInLanguage.ruby]: "linux64",
+  [BuiltInLanguage.rust]: "linux64",
+  [BuiltInLanguage.swift]: "osx64",
+};
+
+/** Inputs that determine whether we may download a per-language bundle. */
+export interface PerLanguageBundleOptions {
+  /** Explicit input only: autodetection needs a CLI instance. */
+  rawLanguages: string[] | undefined;
+  /** CLI version, if known. Ignored for nightly bundles. */
```
mbg: Minor: Here and elsewhere, it might be good to emphasise that we are talking about the _requested_ CLI version rather than one that's already being used by the run.
```suggestion
  /** Requested CLI version, if known. Ignored for nightly bundles. */
```

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  cliVersion: string | undefined;
+  compressionMethod: tar.CompressionMethod;
+  /** Bundle platform identifier, such as linux64. */
+  platform: string | undefined;
```
mbg: Could we type this (here and elsewhere) as something more precise than `string`? It seems that we largely only care about "known" platforms, so we could have an `enum` for the known platforms.

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+  [BuiltInLanguage.cpp]: "linux64",
+  [BuiltInLanguage.csharp]: "linux64",
+  [BuiltInLanguage.go]: "linux64",
+  [BuiltInLanguage.java]: "linux64",
+  [BuiltInLanguage.javascript]: "linux64",
+  [BuiltInLanguage.python]: "linux64",
+  [BuiltInLanguage.ruby]: "linux64",
+  [BuiltInLanguage.rust]: "linux64",
+  [BuiltInLanguage.swift]: "osx64",
+};
```
mbg: The structure of this only allows one platform per language. If/when we support per-language bundles for multiple platforms, this would need to be changed. Consider whether it would make sense to structure this as e.g. `Readonly<Record<KnownPlatform, Set<BuiltInLanguage>>>` and then perform a platform-based lookup for the `Set<BuiltInLanguage>` followed by a membership check. 

(See also my other comment about a potential `KnownPlatform` type.)

src/per-language-bundles.ts
```diff
@@ -0,0 +1,142 @@
...
+import * as tar from "./tar";
+import { GitHubVariant } from "./util";
+
+/** Minimum CLI version for selecting a per-language release bundle. */
+export const MIN_PER_LANGUAGE_BUNDLE_CLI_VERSION = "2.27.1";
+
+const PER_LANGUAGE_BUNDLE_NAME =
+  /^codeql-bundle-(.+)-(?:linux64|osx64|win64)\.tar\.(?:gz|zst)$/;
+
+/** Identifies per-language tools URLs that must not populate the toolcache. */
```
mbg: Minor: This comment doesn't really make it clear what this function returns when such a URL is identified.

```suggestion
/** 
 * Identifies per-language tools URLs that must not populate the toolcache. 
 *
 * @returns The language the URL is for, if it refers to a per-language bundle.
 */
```

src/per-language-bundles.test.ts
```diff
@@ -0,0 +1,189 @@
...
+  compressionMethod: "zstd",
+  platform: "linux64",
+  variant: GitHubVariant.DOTCOM,
+};
+
+async function checkEligibility(
+  overrides: Partial<PerLanguageBundleOptions>,
+  enabledFeatures: Feature[] = [Feature.PerLanguageBundles],
+) {
+  return getPerLanguageBundleLanguage(
```
mbg, line 35: If `getPerLanguageBundleLanguage` used `ActionState`, including an explicit `Env`/`ReadOnlyEnv`, then the tests below likely wouldn't have to be `serial` because the global `process.env[ActionsEnvVars.RUNNER_ENVIRONMENT] = "github-hosted"` wouldn't be required.

src/per-language-bundles.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    { variant: GitHubVariant.GHEC_DR },
+  ]) {
+    t.is(await checkEligibility({ ...nightly, ...overrides }), undefined);
+  }
+  t.is(await checkEligibility(nightly, []), undefined);
+  process.env[ActionsEnvVars.RUNNER_ENVIRONMENT] = "self-hosted";
+  t.is(await checkEligibility(nightly), undefined);
+});
+
+test.serial("recognizes a per-language bundle from its URL", (t) => {
```
mbg: Minor: the test names from here onwards don't mention that they test `tryGetBundleLanguageFromUrl` while the tests above don't mention that they test `getPerLanguageBundleLanguage`.

pr-checks/checks/bundle-toolcache.yml
```diff
@@ -30,7 +30,7 @@ steps:
   - id: init
     uses: ./../action/init
     with:
-      languages: javascript
+      languages: javascript,python
```
mbg, line 34: Consider adding the same comment as in `.github/workflows/codescanning-config-cli.yml`.

src/setup-codeql.ts
```diff
@@ -72,21 +77,40 @@ function getCodeQLBundleExtension(
...
+    case "win32":
+      return "win64";
+    case "linux":
+      return process.arch === "arm64" ? "linux-arm64" : "linux64";
+    case "darwin":
+      return "osx64";
+    default:
+      return undefined;
+  }
+}
```
mbg: Minor: We have similar code duplicated in a few places. Consider whether this definition can be placed in a different source file and shared among the different uses.

src/setup-codeql.ts
```diff
@@ -72,21 +77,40 @@ function getCodeQLBundleExtension(
...
+
+/**
+ * Returns the name of the CodeQL bundle asset to download.
+ *
+ * @param compressionMethod The compression method of the bundle.
+ * @param language If provided, the name of the bundle that contains only this language, rather than
+ * the name of the combined bundle that contains every language.
```
mbg: This is poorly worded.
```suggestion
 * @param language If provided, a bundle name for only this language is returned. 
 *                 Otherwise, the name of a bundle containing all languages is returned.
```

src/setup-codeql.ts
```diff
@@ -730,12 +764,42 @@ export async function getCodeQLSource(
...
+        cliVersion,
+        compressionMethod,
+        platform: getBundlePlatform(),
+        variant,
+      },
+      features,
       logger,
     );
+
+    const resolveBundleURL = (language?: BuiltInLanguage) =>
```
mbg, line 772: Add a comment for this.

src/setup-codeql.ts
```diff
@@ -730,12 +764,42 @@ export async function getCodeQLSource(
...
+        tagName!,
+        apiDetails,
+        getCodeQLBundleName(compressionMethod, language),
+        logger,
+      );
+
+    if (perLanguageBundleLanguage !== undefined) {
+      logger.info(
+        `Downloading the ${perLanguageBundleLanguage} CodeQL bundle, since ${perLanguageBundleLanguage} ` +
+          "is the only language being analyzed.",
```
mbg: The explanation for why the per-language bundle ignores the other conditions that must be true. Rather than listing them all or giving preference to this one, consider a more general wording:
```suggestion
        `Determined that we can download the per-language CodeQL bundle for '${perLanguageBundleLanguage}'.`,
```

src/setup-codeql.ts
```diff
@@ -730,12 +764,42 @@ export async function getCodeQLSource(
...
+        platform: getBundlePlatform(),
+        variant,
+      },
+      features,
       logger,
     );
+
+    const resolveBundleURL = (language?: BuiltInLanguage) =>
+      getCodeQLBundleDownloadURL(
+        tagName!,
```
mbg: What's the justification for `!` here? Could you add a comment to explain why this is justified / find a way to avoid it?

src/setup-codeql.ts
```diff
@@ -730,12 +764,42 @@ export async function getCodeQLSource(
...
+      logger.info(
+        `Downloading the ${perLanguageBundleLanguage} CodeQL bundle, since ${perLanguageBundleLanguage} ` +
+          "is the only language being analyzed.",
+      );
+      url = await resolveBundleURL(perLanguageBundleLanguage);
+      bundle = {
+        kind: "per-language",
+        url,
+        language: perLanguageBundleLanguage,
+        combinedBundleURL: await resolveBundleURL(),
```
mbg: Minor: The result of `resolveBundleURL()` is used in both branches. Consider whether a `const combinedBundleURL = await resolveBundleURL()` before the `if`/`else` statement would make sense.

src/setup-codeql.ts
```diff
@@ -745,6 +809,20 @@ export async function getCodeQLSource(
       );
     }
     compressionMethod = method;
+
+    if (bundle === undefined) {
+      // Explicit per-language URLs must also stay out of the toolcache, but have no fallback.
+      const language = tryGetBundleLanguageFromUrl(url);
```
mbg: This is the only place where `tryGetBundleLanguageFromUrl` is used outside of tests. Would it make sense for it to return a `CodeQLBundle` value directly?

src/setup-codeql.ts
```diff
@@ -833,8 +911,11 @@ export const downloadCodeQL = async function (
     writeToolcacheMarkerFile(toolcacheDestination, logger);
   } else {
     logger.debug(
-      "Could not cache CodeQL tools because we could not determine the bundle version from the " +
-        `URL ${codeqlURL}.`,
+      bundle.kind === "per-language"
+        ? "Not caching the CodeQL tools because they came from a bundle that contains only a " +
+            "single language."
+        : "Could not cache CodeQL tools because we could not determine the bundle version from the " +
+            `URL ${codeqlURL}.`,
```
mbg: Minor: Consider whether it would make sense for the reason why we are not storing a bundle in the toolcache to be returned by `getToolcacheDestination` using e.g. the `Result` type you added a while back.

src/setup-codeql.ts
```diff
@@ -1043,8 +1131,59 @@ export async function downloadCodeQLBundle(
...
+    );
+    return bundle.kind === "combined"
+      ? result
+      : {
+          ...result,
+          statusReport: {
+            ...result.statusReport,
+            bundleLanguage: bundle.language,
+          },
+        };
```
mbg: Since `downloadCodeQL` already has access to `bundle.language` via `source.bundle.language`, consider whether this could be added to the `statusReport` in `downloadCodeQL` when `source.bundle.kind === "combined"` to avoid needing to extend `result` here.

src/setup-codeql.ts
```diff
@@ -1043,8 +1131,59 @@ export async function downloadCodeQLBundle(
...
+  } catch (e) {
+    if (
+      bundle.kind !== "per-language" ||
+      bundle.combinedBundleURL === undefined ||
+      util.asHTTPError(e)?.status !== 404
+    ) {
+      throw e;
+    }
+    logger.warning(
+      `No ${bundle.language} CodeQL bundle was found at ${bundle.url}, so ` +
```
mbg: Give this extra context in the log message here, in case `bundle.url` is an unexpected value and doesn't communicate this itself:
```suggestion
      `No ${bundle.language} CodeQL bundle was found for the per-language bundle at ${bundle.url}, so ` +
```

src/setup-codeql.ts
```diff
@@ -1063,10 +1202,13 @@ function getTempExtractionDir(tempDir: string) {
   return path.join(tempDir, uuidV4());
 }
 
-/**
- * Get the URL of the latest nightly CodeQL bundle.
- */
-async function getNightlyToolsUrl(logger: Logger) {
+/** Selects a bundle from the latest nightly, with a same-release fallback when applicable. */
```
mbg: The "same-release" fallback" part is clear in the context of this PR, but not otherwise.
```suggestion
/** 
 * Selects a bundle from the latest nightly release. This is a per-language bundle
 * if possible, but falls back to a bundle for all languages if necessary.
 */
```

src/setup-codeql.test.ts
```diff
@@ -54,6 +56,25 @@ function stubDownloadAndExtract() {
     });
 }
 
+function stubHostedNightly(tagName: string) {
```
mbg, line 63: Add a comment for what this does.

src/setup-codeql.test.ts
```diff
@@ -54,6 +56,25 @@ function stubDownloadAndExtract() {
...
+    foundZstdBinary: true,
+  });
+  const client = github.getOctokit("123", {
+    request: {
+      fetch: async () =>
+        new Response(JSON.stringify([{ tag_name: tagName }]), {
+          headers: { "content-type": "application/json" },
+        }),
+    },
+  });
```
mbg: This is less precise than the stubbing that `stubHostedNightly` replaces. Consider whether to stub `sinon.stub(client.rest.repos, "listReleases");` instead of `request`.

src/setup-codeql.test.ts
```diff
@@ -592,6 +595,131 @@ test.serial(
...
+          compressionMethod: "zstd",
+          toolsVersion: "0.0.0-30260213",
+        } satisfies setupCodeql.CodeQLDownloadSource);
+      });
+    },
+  );
+}
+
+test.serial(
+  "getCodeQLSource downloads the combined nightly bundle when not eligible",
```
mbg: Clear in the context of the PR, but not in general.
```suggestion
  "getCodeQLSource downloads a combined nightly bundle when per-language bundle is not eligible"
```

src/setup-codeql.test.ts
```diff
@@ -592,6 +595,131 @@ test.serial(
...
+    stubHostedNightly(expectedTag);
+
+    await withTmpDir(async (tmpDir) => {
+      setupActionsVars(tmpDir, tmpDir);
+      for (const { languages, features } of [
+        { languages: ["java"], features: createFeatures([]) },
+        {
+          languages: ["java", "python"],
+          features: createFeatures([Feature.PerLanguageBundles]),
+        },
```
mbg, line 676: Minor: Add comments to the cases to state why they are not eligible.

src/setup-codeql.test.ts
```diff
@@ -876,6 +1004,439 @@ test.serial(
...
+const PER_LANGUAGE_CLI_VERSION = {
+  enabledVersions: [
+    {
+      cliVersion: MIN_PER_LANGUAGE_BUNDLE_CLI_VERSION,
+      tagName: `codeql-bundle-v${MIN_PER_LANGUAGE_BUNDLE_CLI_VERSION}`,
+    },
+  ],
+};
+
+test.serial("getCodeQLBundleName names the per-language bundle", (t) => {
```
mbg: Considering the actual checks in this test:
```suggestion
test.serial("getCodeQLBundleName returns per-language bundle when a language is specified", (t) => {
```

src/setup-codeql.ts
```diff
@@ -1180,6 +1182,7 @@ export async function downloadCodeQLBundle(
       ...result,
       statusReport: {
         ...result.statusReport,
+        totalDurationMs: Math.round(performance.now() - startTime),
```
mbg: Two thoughts here:

1. Minor: It would be nice to avoid a potential divergence between how this is calculated (i.e. `Math.round`ed) in `downloadAndExtract` and here by having a small wrapper function somewhere for `function durationMsSince(startTime) { return Math.round(performance.now() - startTime) }` and use that consistently.
2. I don't love that this (potentially) covers a bunch more code for the fallback case than for the non-fallback case. I am not sure that there is anything critical right now that would be included in this figure that shouldn't be, but perhaps it would be nice to guard against it if you can think of a not too complex way of getting the total duration from the first attempt at `downloadCodeQL` in the error case.

mbg (commented):
I have had a look through the changes here and added quite a few detailed comments. Additionally, some high-level ones:

- Like with the other PR this is stacked on, this was a lot more difficult (and time-consuming) to review than it needed to be, because all of the changes were contained in one commit (except the follow-up to Copilot's review comment). There were clear opportunities for breaking the changes up and structuring them to be more incremental.
- Quite a few of the function-level comments and test names make sense in the context of this PR where it's obvious that they relate to per-language vs combined bundles, but wouldn't make sense outside of this context. Some of my review comments highlight specific cases of this, but I didn't comment on all of them.
- Try and remember to make use of the new `ActionState` infrastructure whenever possible in new or changed code. In particular, the `getPerLanguageBundleLanguage` function would benefit from this since the tests would not need to be `serial`.

src/per-language-bundles.ts
henrymercer: Self-hosted runner can indeed preinstall CodeQL, but I don't think we should assume it.  I would prefer to have a better way to determine whether the toolcache is persistent or not before rolling out to self-hosted runners to avoid a situation where we install many per-language bundles and use an unnecessary amount of disk space.

src/per-language-bundles.ts
henrymercer: Updated as suggested.

src/per-language-bundles.ts
henrymercer: I've renamed this to `isLatestNightly` to make it clearer that this is for the latest nightly only.  We could optionally extend this to previous nightlies as follow up.

src/per-language-bundles.ts
henrymercer: I think that's unlikely for a while, but it's a relatively cheap code change so I've made it.

src/setup-codeql.ts
henrymercer: I've added an error when this is missing.

src/setup-codeql.ts
henrymercer: 1. Good idea, done.
2. I had a look, but I think it's probably more complexity than it's worth.

henrymercer:
Lots of good ideas, thank you for the detailed review!

src/per-language-bundles.test.ts
```diff
@@ -28,19 +30,21 @@ const ELIGIBLE_OPTIONS: PerLanguageBundleOptions = {
...
+  env: ReadOnlyEnv = getTestEnv({
+    [ActionsEnvVars.RUNNER_ENVIRONMENT]: "github-hosted",
+  }),
 ) {
   return getPerLanguageBundleLanguage(
+    {
+      env,
+      features: createFeatures(enabledFeatures),
+      logger: getRecordingLogger([], { logToConsole: false }),
+    },
```
mbg: Minor: Use `initAllState` (same reasoning as on the other PR).

src/per-language-bundles.test.ts
```diff
@@ -49,14 +53,14 @@ test.serial("uses Linux bundles for non-Swift languages", async (t) => {
   }
 });
 
-test.serial("normalizes an alias before selecting a bundle", async (t) => {
+test("getPerLanguageBundleLanguage normalizes aliases before selecting a bundle", async (t) => {
```
mbg, line 64: Not something I had explicitly commented on iirc, but this is a nice wording improvement. Thanks for doing it unprompted!

src/per-language-bundles.test.ts
```diff
@@ -28,19 +30,21 @@ const ELIGIBLE_OPTIONS: PerLanguageBundleOptions = {
 async function checkEligibility(
   overrides: Partial<PerLanguageBundleOptions>,
   enabledFeatures: Feature[] = [Feature.PerLanguageBundles],
+  env: ReadOnlyEnv = getTestEnv({
+    [ActionsEnvVars.RUNNER_ENVIRONMENT]: "github-hosted",
+  }),
```
mbg: Minor: Replacing this with something like `stateOverrides: Partial<ActionState<AllState>> = {}` and then passing `stateOverrides` to `initAllState` might simplify some call sites where you want to modify `env` but not `enabledFeatures`.

src/platform.ts
```diff
@@ -0,0 +1,26 @@
...
+  Linux64 = "linux64",
+  LinuxArm64 = "linux-arm64",
+  Osx64 = "osx64",
+  Win64 = "win64",
+}
+
+/** Returns the bundle platform, or undefined when an all-platform bundle is required. */
+export function getBundlePlatform(
+  platform: NodeJS.Platform = process.platform,
+  arch: NodeJS.Architecture = process.arch,
```
mbg, line 12: Nice choice for testing, and a good compromise between making this pure enough to test and not adding more scope to this PR.

src/per-language-bundles.ts
```diff
@@ -130,14 +135,13 @@ export async function getPerLanguageBundleLanguage(
...
-  const supportedPlatform = PER_LANGUAGE_BUNDLE_PLATFORMS[language];
-  if (supportedPlatform === undefined) {
-    return explain(`no per-language bundle is published for ${language}`);
-  }
-  if (supportedPlatform !== platform) {
+  const supportedLanguages =
+    platform === undefined
+      ? undefined
+      : PER_LANGUAGE_BUNDLE_LANGUAGES[platform];
+  if (!supportedLanguages?.has(language)) {
```
mbg: Minor: `undefined` could be replaced with `new Set()` for the same results below without needing to deal with the possible `undefined`.
```suggestion
  const supportedLanguages =
    PER_LANGUAGE_BUNDLE_LANGUAGES[platform] ?? new Set();
  if (!supportedLanguages.has(language)) {
```

src/platform.ts
mbg, line 1: Minor: Could probably just be called `platform.ts` instead of `bundle-platform.ts` since the type/function names make it clear enough what they relate to and keeps open the possibility to use this file for other platform-related code in the future.

src/status-report/tools-download.test.ts
mbg, line 1: Minor: we can probably justify a `status-report` folder and start moving some status-report-related code into it.

src/status-report.ts
```diff
@@ -630,28 +631,56 @@ export interface InitToolsDownloadFields {
...
   tools_feature_flags_valid?: boolean;
   /** The language of the single-language bundle that was downloaded, if any. */
-  tools_bundle_language?: string;
+  tools_bundle_language?: ToolsDownloadStatusReport["bundleLanguage"];
   /**
    * Whether we tried to download a single-language bundle, but it did not exist and we fell back to
    * the combined bundle.
    */
-  tools_per_language_bundle_fallback?: boolean;
+  tools_per_language_bundle_fallback?: ToolsDownloadStatusReport["perLanguageBundleFallback"];
```
mbg: I don't think this is what I had in mind here. I was thinking of:

- One interface for `tools_bundle_language` and `tools_per_language_bundle_fallback` (e.g. `PerLanguageToolsStatusReport`).
- A field of type `PerLanguageToolsStatusReport` in `ToolsDownloadStatusReport` or extending the latter from `PerLanguageToolsStatusReport` (or any other similar approach).
- Then the field (or properties) from `ToolsDownloadStatusReport` can just be merged as-is into `InitToolsDownloadFields`

The changes in this commit just move the thing I was hoping to avoid into `createInitToolsDownloadFields` and add more complexity here.

src/setup-codeql.ts
mbg: The new comment describes how `resolveBundleURL` is used, not what it does. That's fine as long as that is clearly communicated. E.g. "Used to resolve both bundle variants [..]". Alternatively, the comment should just describe what `resolveBundleURL` does.

src/per-language-bundles.ts
```diff
@@ -121,16 +122,17 @@ export async function getPerLanguageBundleLanguage(
     return explain("the job is not running on a GitHub-hosted runner");
   }
 
-  // Nightly tags contain dates rather than comparable CLI versions.
-  if (!isNightly) {
+  // Check whether per-language bundles are published for the requested CLI version.
+  // Skip this for the latest nightly, whose tag contains a date rather than a CLI version.
```
mbg: I am slightly confused about the change from `isNightly` to `isLatestNightly`, particularly with the accepted wording of the comment here. Nightlies generally have the date in the tag, so it doesn't matter whether it's the latest or not. If mentioning that it is the latest because e.g. we have the `latest` input rather than the tag at this point, then the presence of the date in the tag isn't relevant.

mbg (commented):
Thanks for tackling all the review comments I left during the previous round! I have reviewed the new commits up to and including 2f552a99f36b59f3c6891d0435f3afa00df5a352. I'll look at the remaining ones later this evening, but figured it would be good to post what I have so far now in the interest of time.

Most of the comments I added so far are minor, but there are a couple where I think the changes haven't quite been what I intended with my comments.

As a high-level point, I appreciate that the feedback is addressed across multiple commits, but the breakdown is still not ideal. Most of the new commits I have reviewed so far have a description that suggests they address one point of feedback, but actually mix in other smaller changes as well (some of which address other feedback and some of which are unrelated to the previous review). That makes it harder to verify whether a particular piece of feedback was addressed because more context switching is required while reviewing the individual commits and it takes a bit more effort to identify whether a given change relates to the main point of feedback or is unrelated.

src/per-language-bundles.test.ts
henrymercer: Fixed in 28b8f598f5032f2f254105608a3df22bab2b8ae4

src/per-language-bundles.test.ts
henrymercer: Done in ed3a24ccbc90d2bc224bd6631358a809766a0a5a

src/per-language-bundles.ts
henrymercer: Done in 549d498da392f61aadfc0416f08ed43ae7397a2f

src/platform.ts
henrymercer: Done in ead1f7d93f7fea11d3cf483d696b783b3f686607

src/status-report/tools-download.test.ts
henrymercer: Done in 79fe3a1270f5a101a20367147a05eb6d8ed533af, moving just code relevant to this PR for now

src/status-report.ts
henrymercer: Done in ecec9b5a3756247bd2bfec7da1b6f7bb3eb92d46

src/per-language-bundles.ts
henrymercer: > If mentioning that it is the latest because e.g. we have the latest input rather than the tag at this point

This is what's going on.  I've improved the comment in 07fa87d33359d182be54e4da4bf41664595e3042.

src/setup-codeql.ts
henrymercer: Improved in f18f3536f13ef44ab98c9ef15f8aa05c7f6ac4ae

src/util.ts
```diff
@@ -681,6 +682,11 @@ export async function bundleDb(
   return databaseBundlePath;
 }
 
+/** Returns the elapsed milliseconds, rounded, since a `performance.now()` timestamp. */
```
mbg, line 685: Minor: "since a `performance.now()` timestamp" is incomplete

src/per-language-bundles.test.ts
```diff
@@ -130,11 +131,11 @@ test("getPerLanguageBundleLanguage requires the feature flag", async (t) => {
 test("getPerLanguageBundleLanguage explains a disabled feature before checking eligibility", async (t) => {
   const messages: LoggedMessage[] = [];
   const language = await getPerLanguageBundleLanguage(
-    {
+    initAllState({
       env: getTestEnv(),
```
mbg, line 137: Minor: not required if it's not customised. `initAllState` will automatically call `getTestEnv()` to initialise `env`.

src/per-language-bundles.test.ts
```diff
@@ -30,16 +31,16 @@ const ELIGIBLE_OPTIONS: PerLanguageBundleOptions = {
...
+  env: Env = getTestEnv({
     [ActionsEnvVars.RUNNER_ENVIRONMENT]: "github-hosted",
   }),
 ) {
   return getPerLanguageBundleLanguage(
-    {
+    initAllState({
       env,
       features: createFeatures(enabledFeatures),
       logger: getRecordingLogger([], { logToConsole: false }),
```
mbg, line 41: Minor: Not needed if you don't inspect the log output. `initAllState` will automatically initialise a logger.

src/per-language-bundles.ts
```diff
@@ -103,7 +103,7 @@ export async function getPerLanguageBundleLanguage(
   }
 
   // Check whether per-language bundles are published for the requested CLI version.
-  // Skip this for the latest nightly, whose tag contains a date rather than a CLI version.
+  // Latest-nightly selection skips this release-version check, but not the other eligibility checks.
```
mbg, line 106: Minor: The comment now again doesn't explain why it's OK to skip.

```suggestion
  // If `nightly` tools were requested, skip the version check since we don't have one at this point.
```

src/setup-codeql.test.ts
mbg: As far as I can tell, this hasn't been addressed, but that's not blocking.

mbg (approved):
I have now reviewed the remaining commits from before my last review, as well as the ones in which you addressed my feedback from yesterday. Thanks for taking care of all that!

There are a few very minor points that are left, none of which are blocking / all of which can be followed-up on later. I am happy to approve this PR in this state and I'll go ahead and merge it.

PR #4155, opened 2026-09-18
Add `changenotes.mts assemble` command to move changenotes to CHANGELOG.md

mario-campos:
This PR extends the `pr-checks/changenotes.mts` functionality by adding an `assemble` command that will be used to "flush" (or empty the contents of `unreleased-change-notes`) to the `[UNRELEASED]` section of `CHANGELOG.md`.

This PR also replaces the use of constants `0` and `1` for process exit codes with the `ExitCode` type from `@actions/core`.

### Risk assessment

Workflow types:

- **N/A**

Products:

- **N/A**

Environments:

- **N/A**


pr-checks/changenotes.mts
```diff
@@ -36,10 +47,48 @@ function main(): number {
...
   return 0;
 }
 
+function flush(): number {
+  try {
+    // Get the file paths to our changenotes; these will be useful later.
+    const changenotePaths = fs
+      .readdirSync(CHANGENOTES_DIR)
+      .filter((name) => name !== ".gitkeep")
+      .map((name) => path.join(CHANGENOTES_DIR, name));
```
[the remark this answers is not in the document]
mario-campos: While technically true, this is not a realistic concern because all files in `unreleased-change-notes` would have been validated first in its PR.

pr-checks/changelog.ts
```diff
@@ -125,6 +132,58 @@ export function parseChangelog(content: string): Changelog {
...
+    unreleasedSection.bodyLines[0] === ""
+  ) {
+    insertAtIndex++;
+  }
+
+  // If the section contains the stock message 'No user facing changes.', overwrite it.
+  if (
+    unreleasedSection.bodyLines.length > insertAtIndex &&
+    unreleasedSection.bodyLines[insertAtIndex].trim() === NO_CHANGES_STR
+  ) {
```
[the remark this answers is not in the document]
mario-campos: Fixed

pr-checks/changenotes.mts
```diff
@@ -27,6 +36,8 @@ function main(): number {
     case undefined:
     case "help":
       return usage();
+    case "flush":
+      return flush();
```
mbg: I'd suggest naming this something else like `combine`, `assemble`, `build`, etc. that more intuitively communicates what the command does.

pr-checks/changenotes.mts
mbg: This duplicates logic that already exists for `validate`. There you initially had a very similar definition, but you split it up there in response to my comment about separating pure/impure parts, but now that old definition is duplicated here and again combines pure/impure parts in one definition.

It would be good to avoid that duplication of logic and keep the pure/impure separation.

pr-checks/changenotes.mts
```diff
@@ -36,10 +47,48 @@ function main(): number {
...
+    const changenotes = changenotePaths.map((filePath) => {
+      const fileBody = readFileSync(filePath).toString();
+      const { content } = matter(fileBody);
+      return content.trim();
+    });
+
+    const changelogContents = fs.readFileSync(CHANGELOG_FILE).toString();
+    const changelog = parseChangelog(changelogContents);
+    addBodyLinesToUnreleasedSection(changelog, changenotes);
+    fs.writeFileSync(CHANGELOG_FILE, renderChangelog(changelog));
```
mbg: Consider using `withChangelog` instead of separate `fs.readFileSync` and `fs.writeFileSync`.

pr-checks/changenotes.mts
```diff
@@ -36,10 +47,48 @@ function main(): number {
...
+
+    // Delete changenotes only after successful processing.
+    changenotePaths.map((p) => fs.unlinkSync(p));
```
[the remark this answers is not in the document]
mbg: This is true, but I don't think that we should over-complicate this to account for this unlikely failure mode. If the changelog is written correctly after this runs and there is a failure that fails the workflow, then the changes won't be committed.

pr-checks/changenotes.mts
```diff
@@ -36,10 +47,48 @@ function main(): number {
...
+
+    return 0;
+  } catch (error) {
+    console.error(
+      "Failed to create a new release entry in 'CHANGELOG.md'",
```
mbg: Minor: You're not technically creating a new entry, but are adding changenotes to the existing unreleased section.

pr-checks/changenotes.mts
```diff
@@ -36,10 +47,48 @@ function main(): number {
...
+      error,
+    );
+  }
+
+  return 1;
```
mbg: Minor: Consider making this a named constant so that it's clearer what this is and represents.

pr-checks/changelog.ts
```diff
@@ -125,6 +132,58 @@ export function parseChangelog(content: string): Changelog {
...
+  lines: string[],
+) {
+  // Find the '[UNRELEASED]' section.
+  let unreleasedSection: ChangelogSection | undefined;
+  for (const section of changelog.sections) {
+    if (getHeader(section.headerLine).trim() === UNRELEASED_PLACEHOLDER) {
+      unreleasedSection = section;
+      break;
+    }
+  }
```
mbg: The `[UNRELEASED]` section should always be the first in the file, so scanning the whole file to find it doesn't really make sense: it's either the first in the file as expected and can be found without the loop, or it isn't and something is wrong and we should error out.

pr-checks/changelog.ts
```diff
@@ -125,6 +132,58 @@ export function parseChangelog(content: string): Changelog {
...
+
+  // If the section contains an empty line, preserve it -- insert afterward.
+  if (
+    unreleasedSection.bodyLines.length > 0 &&
+    unreleasedSection.bodyLines[0] === ""
+  ) {
+    insertAtIndex++;
+  }
+
+  // If the section contains the stock message 'No user facing changes.', overwrite it.
```
mbg: I think it would probably be an acceptable simplification to just check for the string in the entire body of the section. If it is present, discard the entire body and add the new entries.

Similarly to the "the unreleased section is always the first" invariant above, it should be an invariant of the unreleased section that it either contains no entries and the `NO_CHANGES_STR` string, or it contains entries and not the `NO_CHANGES_STR` string.

pr-checks/changelog.ts
```diff
@@ -125,6 +132,58 @@ export function parseChangelog(content: string): Changelog {
...
+  let insertAtIndex = 0;
+  let deleteCount = 0;
+
+  // If the section contains an empty line, preserve it -- insert afterward.
+  if (
+    unreleasedSection.bodyLines.length > 0 &&
+    unreleasedSection.bodyLines[0] === ""
+  ) {
+    insertAtIndex++;
+  }
```
mbg: It should always do. See my comment on the lines below for the two possible states of a section. If it's the `NO_CHANGES_STR` case and you discard the entire body, then you can just insert the empty lines before and after the entries and don't care about `insertAtIndex`. If it's the non-`NO_CHANGES_STR` case, then you only care about the index of the last existing entry (or, alternatively, you can work backwards from the end of the body: remove the empty line that should be there, add your new entries, and then add the empty line back in).

mbg (commented):
Thanks for preparing this! I think functionally this looks fine, but as always, I have had a few comments and suggestions on the details of this initial implementation.

pr-checks/changenotes.mts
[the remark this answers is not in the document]
mario-campos: @mbg beat me to it, but I also wanted to add that I tried various workarounds and each option had a failure case. Unfortunately, without atomic transaction-like behavior—which would be overkill for this—a transient failure risks something: data loss, duplicate entries, etc.

The only other alternative would be to implement this flushing one changenote at a time (read changenote, read CHANGELOG.md, insert changenote, commit CHANGELOG.md to file, delete changenote). But this granular implementation would kill performance because of how inefficient it is.

pr-checks/changenotes.mts
mario-campos: Good catch. This is a left-over.

pr-checks/changenotes.mts
mario-campos: Makes sense!

pr-checks/changelog.ts
mario-campos: Fixed

pr-checks/changenotes.mts
mario-campos: I think `assemble` sounds best.

pr-checks/changenotes.mts
mario-campos: Refactored and simplified 👍

pr-checks/changelog.ts
```diff
@@ -125,6 +134,38 @@ export function parseChangelog(content: string): Changelog {
...
+export function addBodyLinesToUnreleasedSection(
+  changelog: Changelog,
+  lines: string[],
+) {
+  // Do nothing if there is nothing to insert.
+  if (lines.length === 0) return;
+
+  const unreleasedSection = changelog.sections[0];
+  if (getHeader(unreleasedSection) !== UNRELEASED_PLACEHOLDER) {
+    throw Error("'[UNRELEASED]' is not the first section of 'CHANGELOG.md'");
```
mbg: Minor:
```suggestion
    throw Error(`'${UNRELEASED_PLACEHOLDER}' is not the first section of 'CHANGELOG.md'`);
```

pr-checks/changelog.ts
```diff
@@ -125,6 +134,38 @@ export function parseChangelog(content: string): Changelog {
...
+  }
+
+  if (unreleasedSection.bodyLines.includes(NO_CHANGES_STR)) {
+    unreleasedSection.bodyLines = ["", ...lines, ""];
```
mbg, line 162: 👍🏻 nice

pr-checks/changelog.ts
```diff
@@ -125,6 +134,38 @@ export function parseChangelog(content: string): Changelog {
...
+    return;
+  }
+
+  unreleasedSection.bodyLines.pop(); // Remove the last empty line.
```
mbg: Minor: prefer placing comments on the preceding line and avoid placing them on the same line:
```suggestion
  // Remove the last empty line.
  unreleasedSection.bodyLines.pop();
```
If you are worried about the comment not applying to the next two lines after this, then here are three options for that:
1. Extend this comment to also describe those lines. E.g. "Remove the empty line that we expect to be the last one in the body, and add the new ones followed by an empty line". This example wording would also be better, because it explains the assumption that the last line is an empty one and so gives context for the implementation.
2. Add a blank line between the `.pop` and other two statements.
3. Add a comment for the other two statements separately.

pr-checks/changelog.ts
```diff
@@ -125,6 +134,38 @@ export function parseChangelog(content: string): Changelog {
   return { preamble, sections };
 }
 
+/**
+ * Inserts the changenotes `notes` under the `[UNRELEASED]` section of `changelog`.
```
mbg: There is no `notes` variable and "under" is ambiguous: it could mean what you intend it to mean, or it could mean that this adds a new section below the `[UNRELEASED]` section. 
```suggestion
 * Inserts the changenotes `lines` in the `[UNRELEASED]` section of `changelog`.
```

pr-checks/changelog.ts
```diff
@@ -125,6 +134,38 @@ export function parseChangelog(content: string): Changelog {
...
+ * If the section contains the stock message {@link NO_CHANGES_STR}, then
+ * `notes` will be inserted in place and the stock message will be deleted.
```
mbg:
```suggestion
 * `lines` will be inserted in place and the stock message will be deleted.
```

pr-checks/changelog.ts
```diff
@@ -125,6 +134,38 @@ export function parseChangelog(content: string): Changelog {
...
+ *
+ * This function will throw an exception if `[UNRELEASED]` does not exist.
```
mbg: Consider making this `@throws Error -- if the [UNRELEASED] section does not exist.`

pr-checks/changenotes.mts
```diff
@@ -3,21 +3,64 @@
...
+  withChangelog,
+} from "./changelog";
 import { isValidAllChangenoteFiles } from "./changelog/validate.mjs";
 import { CHANGENOTES_DIR } from "./config";
 
+/**
+ * Describes a changenote file, including its file path, frontmatter, and content.
+ */
+interface ChangenoteFile {
+  name: string;
```
mbg: Rename to `path` or `absolutePath`? I think that better captures what this is than `name`, which I'd maybe expect to be an entry from the frontmatter or just the basename of the file.

mbg (dismissed):
Thanks for addressing the feedback! This continues to look good in terms of functionality. I have a few more comments with minor suggestions and there was an inconsistency or two I spotted, but I don't think there's anything major that's blocking this, so I am happy to approve.

pr-checks/changelog.ts
mario-campos: Ah of course, forgot about `@throws`.

pr-checks/changelog.ts
mario-campos: In that case, I think 1 sounds best.

mbg (approved):
Thanks!

PR #4156, opened 2026-09-18
Fix path-resolution bug in `changenotes.mts validate` by using `getChangenotes`

mario-campos:
`isValidChangenoteFile` receives a file name (no leading path) and then attempts to read that file, regardless of the CWD. If the user's CWD is not in `unreleased-change-notes`, then the command will likely fail to open the file.

Incidentally, for #4155, I've added `getChangenotes`, which consolidates some duplicative code between the `assemble` and `validate` commands. We can take advantage of that here in `validate`, and in so doing, fix the bug.

### Risk assessment

Workflow types:

- **N/A**

Products:

- **N/A**

Environments:

- **N/A**


pr-checks/changelog/validate.mts
```diff
@@ -5,6 +5,8 @@
 import type { List, ListItem } from "mdast";
 import { fromMarkdown } from "mdast-util-from-markdown";
 
+import { CHANGENOTES_DIR } from "../config";
```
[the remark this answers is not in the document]
mario-campos: Fixed!

pr-checks/changenotes.mts
```diff
@@ -116,7 +116,11 @@ function assemble(): ExitCode {
 
 function validate(): ExitCode {
   try {
-    if (isValidAllChangenoteFiles(fs.readdirSync(CHANGENOTES_DIR))) {
+    const allChangenotesValid = getChangenotes().reduce(
+      (r, changenote) => r && isValidChangenoteFile(changenote.absolutePath),
+      true,
+    );
```
mbg, line 122: Rather than replacing the call to `isValidAllChangenoteFiles` with this inline implementation that uses `getChangenotes()`, do you think it would be better to update the existing implementation of `isValidAllChangenoteFiles` to accept `ChangenoteFile[]` and pass it the result of `getChangenotes()` here? That way you could keep (most of) the unit tests. You would need to move the definition of `ChangenoteFile` somewhere that is accessible to both files.

mbg (commented):
The changes look good, but one question about the approach you took.

pr-checks/changenotes.mts
mario-campos: I deleted `isValidAllChangenoteFiles` because the `filter` became redundant, which just left it to perform `reduce`. At that point, I think the function became a thin wrapper, which doesn't add much value (IMO). The deleted tests themselves were also redundant because they tested the same conditions but at a different level of abstraction. It seemed "cleaner" to inline it and shorten the test suite, but I'm indifferent about it. If you have a strong preference for keeping it, I won't mind.

mbg (approved):

pr-checks/changenotes.mts
mbg: I am happy to merge this as-is for now. It's an easy enough change to make down the line if needed.

PR #4159, opened 2026-09-22
Address per-language bundle review follow-ups

henrymercer:
Address the remaining review feedback from [#4146](https://github.com/github/codeql-action/pull/4146): stub nightly release listing directly, remove redundant test state, and clarify the timing and nightly eligibility comments. No production behavior changes.


src/per-language-bundles.ts
```diff
@@ -102,8 +102,9 @@ export async function getPerLanguageBundleLanguage(
     return explain("the job is not running on a GitHub-hosted runner");
   }
 
-  // Check whether per-language bundles are published for the requested CLI version.
-  // Latest-nightly selection skips this release-version check, but not the other eligibility checks.
+  // When selecting the latest nightly, skip the version check, as nightlies have shipped
+  // per-language bundles since https://github.com/dsp-testing/codeql-cli-nightlies/releases/tag/codeql-bundle-20260909.
+  // Otherwise, check the requested CLI version to determine whether per-language bundles are published.
```
mbg: Sorry to keep going on about this comment, but the change here again doesn't seem correct to me. As far as I can tell from the implementation and your previous explanation on the other PR, we don't have the nightly tag here:

- `isLatestNightly: true` is only set by the `getPerLanguageBundleLanguage` call in `getLatestNightlyBundle`
- `getLatestNightlyBundle` is called if `forceNightly || nightlyRequestedByToolsInput` in `getCodeQLSource`
- I.e. if `tools: nightly`

So the correct justification for skipping the check is that:

- Nightlies don't have a version number in the same way that stable releases do
- They do have dates instead, but we just don't know what that is for the latest nightly at this point so we can't have an equivalent check involving the dates

```suggestion
  // Nightly releases are identified by dates rather than versions. If 
  // `isLatestNightly` is `true`, the latest nightly is requested with 
  // `tools: nightly` and we don't yet have the corresponding tag at this point.
  // Therefore, we skip the version check and don't have an equivalent.
  // We can safely assume that the latest nightly will have per-language bundles.
```

henrymercer: My justification was:

- We're requesting the latest nightly
- Because we're making this change after 20260909, the latest nightly will have a per-language bundle
- Therefore we don't need another check on the CLI version

I'm also happy with your wording though, so I'll accept your suggestion.

mbg (approved):

PR #4162, opened 2026-09-23
Disable `UsePerfData` for `resolveExtractor`

mbg:
Disables `UsePerfData` for `codeql resolve extractor` calls by default. This is not needed for the command to work correctly and reduces the risk of failures resulting from a race condition if the CLI is invoked concurrently ([example](https://github.com/github/codeql-action/actions/runs/35765510016/job/106873900151?pr=4161#step:9:1078)). This is currently possible if we are analysing multiple languages at the same time, because we invoke `codeql resolve extractor` in parallel for each language.

We could consider disabling `UsePerfData` for other CLI calls as well, but this PR is a minimal change to resolve the immediate issue we have observed in our CI that sometimes occurs as a result of the possible race condition.

### Risk assessment

Workflow types:

Products:

Environments:


henrymercer (approved):
Happy to merge this for now as a workaround, but there might be other cases where the JVM will also log messages to stdout, breaking JSON parsing.  So consider a more robust fix e.g. passing something like `-J-Xlog:all=off:stdout` and `-J-Xlog:all=warning:stderr` to all CLI commands.  The fix should probably go in the launcher for future releases, but it is probably worth having the Action pass these flags for previous CLI versions.

mbg:
> but there might be other cases where the JVM will also log messages to stdout, breaking JSON parsing.

@henrymercer Agreed; this is just a quick fix/improvement that targets the immediate issue directly. Redirecting all of the JVM output is probably a good idea as well, but we should check that we don't have any error matching in the Action (e.g. in `cli-errors.ts`) that currently depends on JVM errors in stdout. (If we do, we can update that of course, but we should check.)

